PR #15, opened 2017-09-20
Cross Platform Performance Monitoring

brianrob:
This document covers a proposed design for consuming performance and tracing data and exposing it to performance monitoring systems.

### DIFF
--- a/accepted/cross-platform-performance-monitoring.md
+++ b/accepted/cross-platform-performance-monitoring.md
@@ -38,12 +38,17 @@ The concept of performance counters will be covered in this document, but counte
 	- Have a top-level bitmask that knows whether an event has been requested by anyone and use that to decide whether an event is on.
 	- At a minimum, once an event has been requested by anyone, use an observer pattern to dispatch the event to each consumer, and allow them to deliver it if they want (or ignore it).  We can also be smarter and only dispatch to the systems that requested the event (think of each of these logging systems as sessions).
 - Update native runtime events to use the EventData serialization format.
-	- This is fairly straight forward and just requires updating the script and serialization functions.
+	- This is fairly straight forward and just requires updating the script and serialization functions.  The purpose of this is to ensure that all events have a consistent serialization format and that the format easily allows for access to individual payload items.  Once a particular logging system has taken over it can choose to convert the payload into a different format. 
 - Enable TraceLogging support in EventPipe.
 	- This is required for use of EventCounters.  EventCounters log counter update events using TraceLogging.
 	- OPEN ISSUE: What work is required here?
 - Enable the EventPipe to target EventListeners as dispatch targets.
 	- Allow the EventListener to pinvoke into the runtime and register itself as a target.
 	- Implement a GC safe buffering mechanism to allow native events to be dispatched to managed code.
-- 
-
+- Plumb payload names for all events through the EventPipe.
+	- This is required to fill in the PayloadNames field of EventWrrittenEventArgs (EventListener support).
+	- Right now, EventSource specifies a metadata blob.
+	- CONSIDER: Make EventSource pass a strongly typed object representing the payload fields.
+- Extra Credit: Move the LTTng implementation behind the EventPipe
+	- Right now, calls to LTTng sit right next to the EventPipe.  They should be moved behind the EventPipe just like ETW (above).
+	- This isn't specifically required for performance counters, but can/should be done especially if it cleans up the code or makes things easier.
--- a/accepted/cross-platform-performance-monitoring.md
+++ b/accepted/cross-platform-performance-monitoring.md
@@ -2,17 +2,29 @@
 
 ### Introduction ###
 
-This document describes both the end-to-end scenario, and the platform support required to enable cross-platform performance monitoring.
+This document describes the end-to-end scenarios and platform support required to enable cross-platform performance monitoring of the .NET runtime, framework, and .NET applications.
 
-The concept of performance counters will be covered in this document, but counters are not the only monitoring mechanism included in this design.
+Important Points:
+
+ - This proposal is designed to provide a mechanism for managed code to consume both managed and native tracing events (GC, JIT, ThreadPool, Tasks, EventSource, EventCounters, etc.).
+ - This proposal leaves the existing EventSource/EventListener model in place and simply adds a mechanism to allow EventListeners to subscribe to events that are emitted from the .NET runtime instead of from managed code via an EventSource (GC, Jit ThreadPool, etc.)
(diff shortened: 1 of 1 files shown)


brianrob:
@vancem, @davidfowl, @anurse, @glennc , @SergeyKanzhelev

brianrob:
@goldshtn

brianrob:
@cwe1ss

anurse:
Looking good! Definitely glad we'll be able to get native runtime events in-proc via EventListener.

brianrob:
Thanks for the review @anurse!

brianrob:
@nategraf, FYI

davidfowl:
@brianrob what is EventPipe?

brianrob:
EventPipe is the underlying native mechanism in CoreCLR that is responsible for sending events to their destinations.  Think about it as the CoreCLR telemetry pipeline.  All events go into it, and then we can send them wherever we want (ETW, LTTng, EventListeners, our own cross-platform file format, etc.)

Not all of these destinations are implemented yet, but that's the idea.

SergeyKanzhelev:
I want to learn more about EventPipe. 
1. When even created - how much of execution context will be preserved? If any - can we correlate those events with the `ID` of DiagnosticSources's activity? Maybe only for heavier events? Great scenario for that is associating wait times with http request.
2. Am I correct that aggregation will happen in managed code (in EventCounter). Or EventPipe will also aggregate?
3. Do you have a registry of events in EventPipe? Maybe profiler API's callbacks like exception unwinding?
4. Is EventPipe calls EventListener asynchronously? How fast?

I like the approach!

vancem:
> I want to learn more about EventPipe.

Don't get too hung up on EventPipe.   It is arguably an implementation detail internal to the native code of the .NET Runtime.   It is not really trying to introduce a new model, but simply allow an 'os independent path' for events generated in the runtime (e.g. GC and JIT events), to flow to subscribers (which might be ETW on Windows, LttNG on Linux and EventListeners on either. 

It is this last capability (EventListeners can subscribe to GC and JIT events), that is new, but from a users point of view, it just looks like we added a new EventSource that happens to be implemented in the native code of the runtime (and happens to use EventPipe as its implementation).  

> When even created - how much of execution context will be preserved? If any - can we correlate those events with the ID of DiagnosticSources's activity? Maybe only for heavier events? Great scenario for that is associating wait times with http request.

The events will look like any other EventSource event.   In particular most events will simply be on the thread which caused the event (which means all normal correlation mechanisms should work).   

> Am I correct that aggregation will happen in managed code (in EventCounter). Or EventPipe will also aggregate?

Whether aggregation happens is on a event-by-event basis.   Initially we would probably just expose the existing events but we definitely could add EventCounters for anything people need (those would be a separate work item).  

> Do you have a registry of events in EventPipe? Maybe profiler API's callbacks like exception unwinding?
Is EventPipe calls EventListener asynchronously? How fast?

The model is exactly EventListeners.   Thus you learn about new EventSources as they come on line or when you create your EventListener (which ever is last).  This gives you an opportunity to subscribe.  

We have toyed with the idea of having a profiler API that fires when EventSources fire.   That is a separate work item, and it is not explicitly part of this plan but certainly on the table if we need it.  

For the most part calls to EventListeners are synchronous (so you can carry context on the thread).   There may be exceptions to this, and if there is enough abuse (people doing things that are too complex in EventListener callbacks and creating very fragile systems), we may have to 'protect' ourselves, but simplicity and functionality argue for a simple synchronous model and clients don't abuse the trust that the system places in them by calling them in 'deep' places in the runtime/framework.

SergeyKanzhelev:
@vancem thanks for details. How one will ensure not to fall into infinite recursion if EventListener for GC event causes the same event by allocating something or causing GC to fire?

Today managed code cannot be called from inside the `JITCompilationStarted` and other callbacks of Profiler API. When you saying it's a separate work item - is it because of this problem?

vancem:
> How one will ensure not to fall into infinite recursion if EventListener for GC event causes the same event by allocating something or causing GC to fire?

Ultimately this will be handled in the 'standard' way of having some (thread) state that indicates whether you are currently in the process of logging and do something 'lean' (e.g. nothing) when that happens.  However exactly whose responsibility this is at this point an open question.   My current thinking is that it is better if this is a problem for the EventListener to solve (the instrumentation simply unconditionally logs).   The reasons are 

1.  Depending on what the Listener does and what events it subscribes to  it may or may not be a problem.  Only the Listener has the complete picture.  
2.  Sending in all events gives the EventListener maximum flexibility (e.g. they could have pre-declared buffers etc.
3. Listeners will probably have to have such logic for other things (e.g. if they subscribe to network events and ever send messages over the network).  

Thus my inclination is that this work item ignores the problem to the degree we can.   Clearly if JUST CALLING CODE, will cause an event, then the instrumentation code itself has to protect itself not to emit those events while in the callback.  I am open about this.   If we find it is just too error prone to let EventListeners do their own protection, having the runtime do it for them is reasonable.   

> Today managed code cannot be called from inside the JITCompilationStarted and other callbacks of Profiler API. When you saying it's a separate work item - is it because of this problem?

Certainly events associated with JIT compilation will need some sort of recursion avoidance (as mentioned above).

brianrob:
GC events will also require special handling because we can't call managed code during foreground GCs.  I expect the runtime will buffer the events until they can be delivered.  One option for the JIT compilation case is to use this strategy as well, and I suspect it's what we want to do.

SergeyKanzhelev:
Thanks! Just make sure to put a big note in comments about this risk. I'd expect that some events may cause recursion in only in production under high load and work just fine during testing.

brianrob:
Thanks @SergeyKanzhelev.  Have you seen instances of this before?  If so, can you share an example of what you've seen?  I'd like to proactively plan for this in as many cases as we know of.

SergeyKanzhelev:
Most situations with recursion we saw were related to additional settings set on production. Like extra EventListener enabled and such.

For system events behavior of GC, tasks and locks is different. async/await may be very fast and not even go into different task in test and be in resource starvation mode in production. Locks will rarely be taken in test and in production it is typical to wait for resource. We saw situations like that quite often so listener should be ready for the changed behavior pattern. 

Another thing we found once is UI programming and thread affinity. When the background task started incorrectly from UI thread it may cause and application to hang. So listener should be super careful in how it instantiates tasks.

Very vague notes. Hope it helps. I'll let you know if I'll remember more specific cases.

brianrob:
Thanks @SergeyKanzhelev.  I read that as that you've seen stress behavior in production.  Makes sense.  Locks are another good example - it's possible that a contention event could cause infinite recursion if an EventListener takes a lock and happens to contend on that lock (thus producing a contention event).  We probably need to consider what kinds of generic recursion protection we can put in the runtime for this.

brianrob:
Based on comments on and off this thread, we will consider this design accepted.

PR #17, opened 2017-10-26
Extension of IL verification rules to address ref returns

VSadov:
A proposal for extension of IL verification rules that would allow byref returns in verifiable code and validate that they are used in type-safe manner.

---
Motivation -

There seems to be enough interest in making typesafe code emitted by current compilers formally verifiable, so we need a central location where we discuss and agree on the rules. 

The other areas that would need to be addressed are: [readonly references](https://github.com/dotnet/designs/pull/21) (type-safe code may not be verifiable according to current rules) and ref-structs (opposite: current rules are insufficient to detect type-safety violations).

However, we need to start with something, so let’s start with ref returns. 

Immediately actionable pieces:
-	Who should be on board when we take decision on this?
-	Do we agree on the overall proposal?

### DIFF
--- a/proposed/verifiable-ref-returns.md
+++ b/proposed/verifiable-ref-returns.md
@@ -1,0 +1,64 @@
+# Ref returns in C# and IL verification. #
+
+The goal of this document is to specify additional IL verification rules that would allow C# code that utilizes ref returns be formally verifiable.  
+
+Even in a post-CAS world verifiability is an important property. It is a subset of type-safe code that can be validated by a tool. As such it is often a soundness prerequisite for other kind of analysis. Besides, verifier is a great testing tool that keeps compilers honest about their type-safe claims.  
+
+While we believe that the current rules of the language result in a type-safe code, the formal verification rules for the emitted IL and PEVerify (or any tool of similar purpose) have not yet been updated to accept such code.  
+
+## C# rules for reference: ##
+
+1)	refs to variables on the heap are safe to return
+2)	ref parameters are safe to return
+3)	out parameters are safe to return (but must be definitely assigned, as is already the case today)
+4)	instance struct fields are safe to return as long as the receiver is safe to return
+5)	“this” is not safe to return from struct members
+6)	a ref, returned from another method is safe to return if all refs/outs passed to that method as formal parameters were safe to return.
+*Specifically it is irrelevant if receiver is safe to return, regardless whether receiver is a struct, class or typed as a generic type parameter.*  
+
+NOTE: C# rules operate with high level language entities such as scopes, expressions, local variables, etc., that do not exist at the level of IL or do not map one-to-one.
+The following is the attempt to introduce/adjust verification rules to allow validation of the safety of the ref returns at IL level.
+
+# Verification of ref returns. #
+
+## Returnable managed pointer. ##
+
+CLI needs to add a notion of another transient type - returnable managed pointer. 
+*(need to adjust “III.1.8.1.1 Verification algorithm” and in “I.8.7 Assignment compatibility” accordingly)*
+
+NOTE: there are similarities with already existing concept of controlled mutability managed references.
+
+Returnable managed pointer is a transient type used for verification purposes. When certain conditions are met, managed pointers can be classified as returnable. Unlike ordinary managed pointer types, returnable managed pointers can be used as operands of `RET` instruction in verifiable code. 
+
+Returnable managed pointer may be a result of one of the following:
+-	`LDSFLDA`, `LDELEMA`, `UNBOX` always returnable   (see: C# rule #1)
+-	`LDFLDA` when the receiver is a reference type     (see: C# rule #1).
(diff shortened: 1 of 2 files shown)


proposed/verifiable-ref-returns.md
```diff
@@ -0,0 +1,75 @@
...
+-   a local byref slot may be marked as returnable. 
+-	`STLOC` into returnable slots is verifiable only when the value is a returnable managed pointer. 
+-	`LDLOC` from a returnable slot produces a returnable managed pointer.
+
+The rationale for introducing returnable locals is to allow for a single pass verification analysis. 
+While, in theory, it may be possible that the �returnable� property could be inferred via some form of fixed point flow analysis, it would make verification substantially more complex and expensive and would not retain the single-pass property.
+
+## Metadata encoding of returnable local slots ## 
+
+Local slots can be marked as returnable by applying `modopt[System.Runtime.CompilerServices.IsReturnableAttribute]` in the local signature.
```
MichalStrehovsky: Is there a particular reason to use an `Attribute` suffix here? This will serve similar purpose as `System.Runtime.CompilerServices.IsVolatile` or `IsConst`. Those are not attributes either.

proposed/verifiable-ref-returns.md
```diff
@@ -0,0 +1,75 @@
...
+
+Returnable managed pointer is a transient type used for verification purposes. When certain conditions are met, managed pointers can be classified as returnable. Unlike ordinary managed pointer types, returnable managed pointers can be used as operands of `RET` instruction in verifiable code. 
+
+Returnable managed pointer may be a result of one of the following:
+-	`LDSFLDA`, `LDELEMA`, `UNBOX` always returnable   (see: C# rule #1)
+-	`LDFLDA` when the receiver is a reference type     (see: C# rule #1).
+-	`LDFLDA` when the receiver is a returnable ref 	(see: C# rule #4)
+-	`LDARG` of a byref parameter 			(see: C# rule #2, #3), 
+except for arg0 in a struct method 		(see: C# rule #5).
+-	`CALL`, `CALLVIRT` when all reference arguments are returnable. (see: C# rule 6)
```
MichalStrehovsky: Should this also apply to `CALLI`?

VSadov: The rule could include `CALLI` for completeness. However `CALLI` cannot be present in verifiable code so generalization would not make much difference.

MichalStrehovsky: > However `CALLI` cannot be present in verifiable code

Why not? ECMA-335 defines verification rules for when `CALLI` is verifiable. It's not like `JMP` (whose verification rules are simply "The jmp instruction is never verifiable").

proposed/verifiable-ref-returns.md
VSadov: No, no reason to be an attribute. 

In the parallel case with readonly references, it seems appropriate to use `IsReadOnlyAttribute` since that is used in other places to express readonliness.
I could not think of a similar existing type for "returnable", so I suggest `IsReturnableAttribute` just for similarity, but you are right - it does not need to be an attribute or even a new type. 

Any alternative ideas here would be appreciated, but we need to pick some type.

proposed/verifiable-ref-returns.md
VSadov: Is it? Perhaps I think of it as unverifiable because it invariably involves pointers.
Anyways, no problem with including it.

VSadov: I think, I should mention varargs/arglist case as well. 

C# compiler does consider them for the purpose of "any ref argument can potentially come back as a ref return"

VSadov:
Merging this, since this has been opened for a long time, and informally approved by interested parties. 

NOTE: The proposal does not have direct impact on the language, but has effect on the compiler.
It codifies the metadata shape which has direct impact on interop/verification scenarios. 
(that is why it fits in csharplang repo)

MichalStrehovsky:
@VSadov Based on your own comments *"I think, I should mention varargs/arglist case as well."*, the question around verification of CALLI, and not giving this `Attribute` suffix (that can lead someone to believe they can apply this as an attribute on somethings), I thought this is not fully fleshed out yet...

VSadov:
@MichalStrehovsky - will make the update regarding `CALLI`, I thought I did...

As for the type to mark returnable local slots - any good ideas?

Ideally that would be a preexisting type that exists everywhere - like `IsConst` or even `System.Object`, but there is a danger that it is already used by somebody to `modopt` locals.

MichalStrehovsky:
> As for the type to mark returnable local slots - any good ideas?

I think we need a new type (`IsConst` is only available starting NetStandard 1.3 and I can't think of anything existing that wouldn't be confusing). I would be fine with `System.Runtime.CompilerServices.IsReturnable`. If we would like to eventually make it part of the framework so that the compiler doesn't need to always emit it, we might want to run it through a CoreFX API review. (This is the part where making this a custom attribute would be weird because users might think they can apply it to things and something will happen as a result.)

PR #24, opened 2017-11-30
Initial draft of file enumeration design doc.

JeremyKuhne:
This is a proposal to provide a performant extensibility point for file enumeration in System.IO.

This is based on the existing internal Windows implementation, see the CoreFX master branch for some portion of this in action. A branch implemented to spec (for Windows only) can be found here: https://github.com/JeremyKuhne/corefx/tree/findpublic

@terrajobst, @pjanotti, @danmosemsft

### DIFF
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -6,7 +6,7 @@ Enumerating files in .NET provides limited configurability. You can specify a si
 
 Recursive enumeration is also problematic in that there is no way to handle error states such as access issues or cycles created by links.
 
-These restrictions have a significant impact on file system intensive applications, a key example being MSBuild. We can address these restrictions and provide performant, configurable enumeration.
+These restrictions have a significant impact on file system intensive applications, a key example being MSBuild. This document proposes a new set of primitive file and directory traversal APIs that are optimized for providing more flexibility while keeping the overhead to a minimum so that enumeration becomes both more powerful as well as more performant.
 
 ## Scenarios and User Experience
 
@@ -15,20 +15,31 @@ These restrictions have a significant impact on file system intensive applicatio
 
 ## Requirements
 
-1. Filtering based on common file system data is possible
-	a. Name
-	b. Attributes
-	c. Time stamps
-    d. File size
-2. Result transforms can be of any type
-3. We provide common filters and transforms
-4. Recursive behavior is configurable
-5. Error handling is configurable
 
 ### Goals
 
-1. API is cross platform generic
-2. API minimizes allocations while meeting #1
+
+1. Custom filtering based on common file system data
+	- Name
+	- Attributes
+	- Time stamps
+    - File size
+2. Result transforms can be of any desired output type
+	- Like Linq Select(), but keeps FileData on the stack
+3. API minimizes allocations
+4. API is cross platform abstract
+3. We provide common filters and transforms
+	- To file/directory name
+	- To full path
+	- To File/Directory/FileSystemInfo
+	- DOS style filters (Legacy- `*/?` with DOS semantics, e.g. `*.` is all files without an extension)   
+	- Simple Regex filter
+	- Simpler globbing (`*/?` without DOS style variants)
+	- Set of extensions (`*.c`, `*.cpp`, `*.cc`, `*.cxx`, etc.)
+4. Recursive behavior is configurable
+	- On/Off
+	- Predicate based on FileData
+5. Can avoid throwing access denied exceptions
 
 ### Non-Goals
 
@@ -45,12 +56,12 @@ namespace System.IO
     /// <summary>
     /// Delegate for filtering out find results.
     /// </summary>
-    internal delegate bool FindPredicate<TState>(ref RawFindData findData, TState state);
+    public delegate bool FindPredicate<TState>(ref RawFindData findData, TState state);
 
     /// <summary>
     /// Delegate for transforming raw find data into a result.
     /// </summary>
-    internal delegate TResult FindTransform<TResult, TState>(ref RawFindData findData, TState state);
+    public delegate TResult FindTransform<TResult, TState>(ref RawFindData findData, TState state);
 
     [Flags]
     public enum FindOptions
@@ -89,7 +100,7 @@ namespace System.IO
             FindOptions options = FindOptions.None);
     }
 
-    public static class DirectoryInfo
+    public class DirectoryInfo
     {
         public static IEnumerable<TResult> Enumerate<TResult, TState>(
             FindTransform<TResult, TState> transform,
@@ -191,8 +202,7 @@ public static FindEnumerable<string, string> GetFiles(string directory,
         (ref RawFindData findData, string expr) => FindTransforms.AsFullPath(ref findData),
         (ref RawFindData findData, string expr) =>
         {
-            return FindPredicates.NotDotOrDotDot(ref findData)
-                && !FindPredicates.IsDirectory(ref findData)
+            return !FindPredicates.IsDirectory(ref findData)
                 && DosMatcher.MatchPattern(expr, findData.FileName, ignoreCase: true);
         },
         state: DosMatcher.TranslateExpression(expression),
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -292,6 +292,49 @@ namespace System.IO
 }
 ```
 
+### Potential testability variant
+
+Having `FindData` be a struct is critical to meet the performance goals for the feature. To allow testing of Filters and Predicates we could also expose an interface of `IFindData`:
+
+``` C#
+namespace System.IO
+{
+    public interface IFindData<TState>
+    {
+        string Directory { get; }
+        string OriginalDirectory{ get; }
+        string OriginalUserDirectory { get; }
+        TState State { get; }
+
+        ReadOnlySpan<char> FileName { get; }
+        FileAttributes Attributes { get; }
+        long Length { get; }
+
+        DateTime CreationTimeUtc { get; }
+        DateTime LastAccessTimeUtc { get; }
+        DateTime LastWriteTimeUtc { get; }
+    }
+
+    public unsafe struct FindData<TState> : IFindData<TState>
+    // ... same as defined earlier
+}
+```
+
+Predicates and transforms could then be modified as in the following example:
+
+```C#
+namespace System.IO
+{
+    public static partial class FindTransforms
+    {
+        public static DirectoryInfo AsDirectoryInfo<TFindData, TState>(ref TFindData findData)
+            where TFindData : struct, IFindData<TState>
+    }
+}
+```
+
+This adds a bit more syntactical complexity, but should allow testing transforms and predicates against mocked IFindData structs.
+
 ### Existing API summary
 
 ``` C#
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -215,7 +215,7 @@ namespace System.IO
     /// <summary>
     /// Used for processing and filtering find results.
     /// </summary>
-    public ref struct FindData
+    public ref struct FindData<TState>
     {
         // This will have private members that hold the native data and
         // will lazily fill in data for properties where such data is not
@@ -238,6 +238,8 @@ namespace System.IO
         public DateTime CreationTimeUtc { get; }
         public DateTime LastAccessTimeUtc { get; }
         public DateTime LastWriteTimeUtc { get; }
+
+        public TState State;
     }
 }
 ```
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -49,6 +49,8 @@ public static IEnumerable<string> GetFileFullPaths(string directory,
 
 While using anonymous delegates isn't strictly necessary, they are cached, so all examples are written with them. The second argument above, for example, could simply be `FindTransforms.AsUserFullPath`, but it wouldn't take advantage of delegate caching.
 
+> Note that while you can write a solution that doesn't allocate a `FileInfo` by using the `string[]` APIs and `GetFullPath()` it would still allocate unneeded strings and introduce costly normalization overhead.
+
 
 ### Getting full paths of a given extension
 
@@ -328,6 +330,19 @@ namespace System.IO
 
 ## Q & A
 
+#### Why is this so complicated?
+
+This isn't intended to be a common-use API. The existing APIs will be kept, maintained, and extended based on customer demand. We don't want to:
+
+1. Have scenarios blocked waiting for new APIs to work their way through the system
+2. Have to write "normal" APIs to address more corner cases
+
+In order to make this a usable extension point we have to sacrifice some usability to get the neccessary characteristics. Note that what people build on this will directly impact our future designs of the standard, usability focused, APIs.
+
+#### Why are you using Linq in your examples?
+
+For example clarity. Clearly if we were trying to fully optimize for perf/allocations we wouldn't.
+
 #### Why aren't you providing a filter that does _`X`_?
 
 We want to only provide pre-made filters that have broad applicability. Based on feedback we can and will consider adding new filters in the future.
@@ -351,7 +366,7 @@ This is something we're investigating for the future. See discussions below.
 Right now we provide a single cross platform view. Some scenario might benefit from viewing OS specific data, such as the `inode` or the underlying UTF-8 data. A few options:
 
 1. Provide interfaces for platform specific data on `FindData`, such as `UnixFindData`.
-2. Make raw OS data structs public and provide a method to get the data: `bool TryGetNativeData<T>(ref T nativeData)`
+2. Make raw OS data structs public and provide a method to get the data: `bool TryGetNativeData<T>(ref T nativeData, ref FindData findData)`
 
 Related to this, we could potentially provide `FindOptions` flags to specify exactly what raw data we get. On Windows, for example, there are a number of different data structs that can be returned from [NtQueryDirectoryFile](https://msdn.microsoft.com/en-us/library/windows/hardware/ff567047.aspx). We could abstract this away. If we went this route we'd need to carefully measure performance impact of having conditions on our `FindData` properties.
 
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -123,11 +123,14 @@ The key goal is to provide an advanced API that keeps allocations to a minimum b
 		- (Pri1) Simple Regex filter (e.g. IsMatch())
 		- (Pri2) Simpler globbing (`*/?` without DOS style variants)
 		- (Pri2) Set of extensions (`*.c`, `*.cpp`, `*.cc`, `*.cxx`, etc.)
-4. Recursive behavior is configurable
+6. Recursive behavior is configurable
 	- On/Off via flag
 	- Predicate based on FileData
-5. Can avoid throwing access denied / security exceptions via Options flags
-6. Can hint buffer allocation size via flag
+7. Behavior flags
+	1. Recurse
+	2. Ignore inaccessible files (e.g. no rights)
+	3. Hint to use larger buffers (for remote access)
+	4. Optimization to skip locking for single-thread enumeration 
 
 ### Non-Goals
 
@@ -165,7 +168,13 @@ namespace System.IO
         IgnoreInaccessable = 0x2,
 
 		// Hint to use larger buffers for getting data (notably to help address remote enumeration perf)
-		UseLargeBuffer = 0x4
+		UseLargeBuffer = 0x4,
+
+		// Allow .NET to skip locking if you know the enumerator won't be used on multiple threads
+		// (Enumerating is inherently not thread-safe, but .NET needs to still lock by default to
+		//  avoid access violations with native access should someone actually try to use the
+		//  the same enumerator on multiple threads.)
+		NoLocking = 0x8
 
         // Future: Add flags for tracking cycles, etc. 
 	}
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -325,3 +325,16 @@ namespace System.IO
 
 
 ## Q & A
+
+#### Why aren't you providing a filter that does _`X`_?
+
+We want to only provide pre-made filters that have broad applicability. Based on feedback we can and will consider adding new filters in the future.
+
+#### Why did you put data in the struct that is expensive to get on Unix?
+
+While Windows gives all of the data you see in `FindData` in a single enumeration call, this isn't true for Unix. We're trying to match the current `System.IO.*Info` class data, not the intersection of OS call data. We will lazily get the extra data (time stamps, for example) on Unix to avoid unnecessarily taxing solutions that don't need it.
+
+#### Why doesn’t the filename data type match the OS format?
+
+We believe that it is easier and more predictable to write cross-plat solutions based on `char` rather than have to deal directly with encoding. The current plan is that we'll optimize here by not converting from UTF-8 until/if needed and we'll also minimize/eliminate GC impact by keeping the converted data on the stack or in a cached array.
+
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -72,8 +72,8 @@ public static IEnumerable<string> GetFileFullPathsWithExtension(string directory
 {
     return new FindEnumerable<string, string[]>(
         directory,
-        (ref FindData<string> findData) => FindTransforms.AsUserFullPath(ref findData),
-        (ref FindData<string> findData) =>
+        (ref FindData<string[]> findData) => FindTransforms.AsUserFullPath(ref findData),
+        (ref FindData<string[]> findData) =>
         {
             return !FindPredicates.IsDirectory(ref findData)
                 && findData.State.Any(s => findData.FileName.EndsWith(s, StringComparison.OrdinalIgnoreCase));
@@ -105,7 +105,7 @@ The key goal is to provide an advanced API that keeps allocations to a minimum b
 	- Name
 	- Attributes
 	- Time stamps
-    - File size
+	- File size
 4. Transforms can provide results in any type
 	- Like Linq Select(), but keeps FileData on the stack
 5. Filters and transforms will be provided
@@ -167,14 +167,14 @@ namespace System.IO
         // Skip files/directories when access is denied (e.g. AccessDeniedException/SecurityException)
         IgnoreInaccessable = 0x2,
 
-		// Hint to use larger buffers for getting data (notably to help address remote enumeration perf)
-		UseLargeBuffer = 0x4,
+        // Hint to use larger buffers for getting data (notably to help address remote enumeration perf)
+        UseLargeBuffer = 0x4,
 
-		// Allow .NET to skip locking if you know the enumerator won't be used on multiple threads
-		// (Enumerating is inherently not thread-safe, but .NET needs to still lock by default to
-		//  avoid access violations with native access should someone actually try to use the
-		//  the same enumerator on multiple threads.)
-		NoLocking = 0x8
+        // Allow .NET to skip locking if you know the enumerator won't be used on multiple threads
+        // (Enumerating is inherently not thread-safe, but .NET needs to still lock by default to
+        //  avoid access violations with native access should someone actually try to use the
+        //  the same enumerator on multiple threads.)
+        NoLocking = 0x8
 
         // Future: Add flags for tracking cycles, etc. 
 	}
@@ -197,7 +197,7 @@ namespace System.IO
             string path,
             FindTransform<TResult, TState> transform,
             FindPredicate<TState> predicate,
-			FindPredicate<TState> recursePredicate = null,
+            FindPredicate<TState> recursePredicate = null,
             TState state = default,
             FindOptions options = FindOptions.None);
     }
@@ -207,7 +207,7 @@ namespace System.IO
         public static IEnumerable<TResult> Enumerate<TResult, TState>(
             FindTransform<TResult, TState> transform,
             FindPredicate<TState> predicate,
-			FindPredicate<TState> recursePredicate = null,
+            FindPredicate<TState> recursePredicate = null,
             TState state = default,
             FindOptions options = FindOptions.None);
     }
@@ -221,7 +221,7 @@ namespace System.IO
         // will lazily fill in data for properties where such data is not
         // immediately available in the current platform's native results.
 
-		// The full path to the directory the current result is in
+        // The full path to the directory the current result is in
         public string Directory { get; }
 
         // The full path to the starting directory for enumeration
--- a/accepted/file-enumeration.md
+++ b/accepted/file-enumeration.md
@@ -14,7 +14,7 @@ These restrictions have a significant impact on file system intensive applicatio
 
 To get full paths for files, one would currently have to do the following:
 
-```
+``` C#
 public static IEnumerable<string> GetFileFullPaths(string directory,
     string expression = "*",
     bool recursive = false)
@@ -292,49 +292,6 @@ namespace System.IO
 }
 ```
 
-### Potential testability variant
-
-Having `FindData` be a struct is critical to meet the performance goals for the feature. To allow testing of Filters and Predicates we could also expose an interface of `IFindData`:
-
-``` C#
-namespace System.IO
-{
-    public interface IFindData<TState>
-    {
-        string Directory { get; }
-        string OriginalDirectory{ get; }
-        string OriginalUserDirectory { get; }
-        TState State { get; }
-
-        ReadOnlySpan<char> FileName { get; }
-        FileAttributes Attributes { get; }
-        long Length { get; }
-
-        DateTime CreationTimeUtc { get; }
-        DateTime LastAccessTimeUtc { get; }
-        DateTime LastWriteTimeUtc { get; }
-    }
-
-    public unsafe struct FindData<TState> : IFindData<TState>
-    // ... same as defined earlier
-}
-```
-
-Predicates and transforms could then be modified as in the following example:
-
-```C#
-namespace System.IO
-{
-    public static partial class FindTransforms
-    {
-        public static DirectoryInfo AsDirectoryInfo<TFindData, TState>(ref TFindData findData)
-            where TFindData : struct, IFindData<TState>
-    }
-}
-```
-
-This adds a bit more syntactical complexity, but should allow testing transforms and predicates against mocked IFindData structs.
-
 ### Existing API summary
 
 ``` C#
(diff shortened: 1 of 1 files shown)


poke:
I have to say that the API looks quite complicated. The fact that your sample is basically building a helper method *on top* of the proposed API kind of shows that consuming it might be too difficult and it is maybe not meant as a high-level way to enumerate the file system after all. I fear that people will just end up using the existing methods which they can then filter using comfortable LINQ instead of dealing with this new API. And I certainly don’t believe this new API should be limited to those that *care* about its benefits—ideally, everybody should be able to use it without having to invest time to learn completely new API styles.

When I saw the `FindPredicates.NotDotOrDotDot(ref findData) && !FindPredicates.IsDirectory(ref findData)` part, I was wondering whether it wasn’t possible to at least make those extensions methods, so one could do `findData.NotDotOrDotDot().IsNotDirectory()` which would make it fluent and a lot easier to use. But I don’t know if extension methods support refs (yet)?

Wouldn’t we be able to convert this to a fluid LINQ-like API, so we could do something like `FindFiles(directory, recursive: true).Where(f => f.NotDotOrDotDot() && !f.IsDirectory()).AsFullPath()`? Or will that not allow us to avoid allocations?

svick:
My comments:

1. The API description lists several types as `internal`, I think those were supposed to be `public`.
2. What's the point of having a `FindEnumerable` constructor, and also `Directory.Enumerate`? If they're different, that's confusing. If they're the same, then I think the constructor is unnecessary.
3. Would it make sense to include `state` inside `RawFindData`? That would make the signatures and lambdas simpler, especially if `state` was not used.
4. Why is `RawFindData` named like that? Wouldn't just `FindData` suffice? 
5. If common predicates and transforms are going to be provided, I think their signatures should be the same as what the methods accept, so that you can use them as method groups, e.g.: `Directory.Enumerate<string, object>(path, FindTransforms.AsFullPath<object>, FindPredicates.IsDirectory<object>)`.
6. If `state` is considered optional, then it could make sense to provide overloads where you don't have to specify its type (so the example above would become just `Directory.Enumerate<string>(path, FindTransforms.AsFullPath, FindPredicates.IsDirectory)`).
7. `NotDotOrDotDot` is fun to say out loud but I think it's unclear what it's supposed to do. If it's meant to ignore files whose names start with a dot, which are considered hidden on UNIX, maybe it would make sense to introduce some platform-independent notion of "hidden files"?

@poke An important part of the proposal is limiting recursion. How would you do that using a LINQ-like API?

JeremyKuhne:
> I have to say that the API looks quite complicated.

@poke It is a bit complicated due to the need to keep allocations to a minimum. It is meant to be an advanced extension point that others can build solutions on for high-impact scenarios that aren't broadly applicable. Think MSBuild trying to find all files with a set of extensions under a certain folder tree as one example.

@svick 

> The API description lists several types as internal, I think those were supposed to be public.

Thanks, I'll fix them. This started from our current internal Windows implementation. :)

> What's the point of having a FindEnumerable constructor, and also Directory.Enumerate?

We don't need to have both if we seal the class. I'm not settled on what to do there.

> Would it make sense to include state inside RawFindData? That would make the signatures and lambdas simpler, especially if state was not used.

https://github.com/dotnet/corefx/pull/25691 explores doing this.

> Why is RawFindData named like that? Wouldn't just FindData suffice?

I don't think we would introduce a higher level abstraction, so FindData is probably ok.

> If common predicates and transforms are going to be provided, I think their signatures should be the same as what the methods accept

Was trying to encourage leveraging that anonymous method delegates are cached. I'll definitely give that more thought.

> If state is considered optional, then it could make sense to provide overloads where you don't have to specify its type

Seems rational at first- I'll try it out

> NotDotOrDotDot is fun to say out loud but I think it's unclear what it's supposed to do.

This one is a struggle. Could go with `NotJustDotOrDotDot`?

CZEMacLeod:
@svick @JeremyKuhne 
Assume `NotDotOrDotDot` is a predicate that excludes the current directory and parent directory entries that are returning in directory enumerations e.g. Not . and .. I think the name is fine as is, although it might make more sense in fluent form as @poke suggests? In which case it might be `findData.ExceptDotAndDotDot()`?
If this is referring to Linux type hidden files/folders starting with a . then the name is confusing.
Looking at your example case, should the predicate include the Not at all?
```
public static FindEnumerable<string, string> GetFiles(string directory,
    string expression = "*",
    bool recursive = false)
{
    return new FindEnumerable<string, string>(
        directory,
        (ref RawFindData findData, string expr) => FindTransforms.AsFullPath(ref findData),
        (ref RawFindData findData, string expr) =>
        {
            return !FindPredicates.DotOrDotDot(ref findData)
                && !FindPredicates.IsDirectory(ref findData)
                && DosMatcher.MatchPattern(expr, findData.FileName, ignoreCase: true);
        },
        state: DosMatcher.TranslateExpression(expression),
        options: recursive ? FindOptions.Recurse : FindOptions.None);
}
```
Also, aren't . and .. directories and would be excluded by the IsDirectory predicate anyway?

One other thing is that if `RawFindData` can contain directory entries for enumerating and filtering directories as well as files, then perhaps `FileName` is not the right propery name?

JeremyKuhne:
> it might make more sense in fluent form as @poke suggests?

It would make more sense, but would defeat key design goals, notably to minimize allocations. The struct has to be a ref struct to make this possible. If we had https://github.com/dotnet/csharplang/issues/186 it would be worth exploring.

> Also, aren't . and .. directories and would be excluded by the IsDirectory predicate anyway?

They are, good catch.

> perhaps FileName is not the right propery name?

Perhaps `Name` would be ok.

CZEMacLeod:
@JeremyKuhne Just finished reading dotnet/csharplang#186 so I now understand why this doesn't (currently) have fluent / extension methods.
I must admit I mainly write in vb.net and didn't realise that C# _couldn't_ do ByRef extension methods.

JeremyKuhne:
@CZEMacLeod It is one of my top wishlist items. That and https://github.com/dotnet/csharplang/issues/187. :)

CZEMacLeod:
@JeremyKuhne I notice that some of this comes off the back of MSBuild 'Globbing' (as an aside - is that actually a word?)
Not wanting to take this off-topic at all but I was involved in a discussion on a very similar issue on an asp.net core equivalent of the old asp.net optimization. The feature request was to enable recursively searching for files to bundle using patterns. 
It was noted there that this should be able to work over `IFileProvider` rather that the raw filesystem to allow for the same functionality when using 'virtual' files of one sort or another - e.g. embedded resources, zips, dynamically generated files, etc.
Do you think it might be possible to ensure that this 'concrete' filesystem API can also work this way so we don't need to learn 2 (or more) APIs?
For reference the conversation is Shazwazza/Smidge#47
I believe Nuget also uses (used?) an implementation of IFileProvider in its pack and restore commands to determine files to include - and again includes a lot of pattern matching.
It might be worth considering the scenarios for NuGet and ASP.Net in this spec.

JeremyKuhne:
> 'Globbing' (as an aside - is that actually a word?)

It is. https://en.wikipedia.org/wiki/Glob_%28programming%29

> Do you think it might be possible to ensure that this 'concrete' filesystem API can also work this way so we don't need to learn 2 (or more) APIs?

I'm open to suggestions on how we could make FindData more flexible without creating too much overhead. Biggest thing is keeping allocations down, but speed is important too. As System.IO currently has a version of this proposal, we can experiment and see real-world impact of design options. I've been measuring impact on >400K file sets on multiple types of drives (SSD, HDD, DriveSpace arrays...).

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
+# Extensible File Enumeration
+
+**PM** [Immo Landwerth](https://github.com/terrajobst) | **Dev** [Jeremy Kuhne](https://github.com/jeremykuhne)
+
+Enumerating files in .NET provides limited configurability. You can specify a simple DOS style pattern and whether or not to look recursively. More complicated filtering requires post filtering all results which can introduce a significant performance drain.
+
+Recursive enumeration is also problematic in that there is no way to handle error states such as access issues or cycles created by links.
+
+These restrictions have a significant impact on file system intensive applications, a key example being MSBuild. We can address these restrictions and provide performant, configurable enumeration.
```
terrajobst: >We can address these restrictions and provide performant, configurable enumeration. [](start = 113, length = 83)

You want to make sure that your intro has a concrete proposal. Maybe something like:

> This proposes a new set of primitive file and directory traversal APIs that are optimized for minimal overhead in order to make enumeration both more flexible as well as more performant.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+
+## Scenarios and User Experience
```
terrajobst: >Scenarios and User Experience [](start = 3, length = 29)

Your scenarios way too short. You want to make those headings and show some sample code consuming the APIs you're proposing. Scenarios are meant to illustrate the value your APIs are adding. They shouldn't be longer than a few paragraphs but they also shouldn't be that abstract.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+
+1. MSBuild can custom filter filesystem entries with limited allocations and form the results in any desired format.
+2. Users can build custom enumerations utilizing completely custom or provided commonly used filters and transforms.
+
+## Requirements
+
+1. Filtering based on common file system data is possible
+	a. Name
```
terrajobst: >a. [](start = 1, length = 2)

Doesn't render on GitHub as an enumeration. Use numbers or bullet points instead.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+	b. Attributes
+	c. Time stamps
+    d. File size
```
terrajobst: > d. File size [](start = 3, length = 13)

What doesn't pop here is how the filter an be provided. I think you want to call out explicitly that the filter is a customer function and thus supports all sorts of filtering criteria, including calling custom functions and arbitrary combinators (AND, OR, NOT etc).

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+2. Result transforms can be of any type
```
terrajobst: >Result transforms can be of any type [](start = 3, length = 36)

I think this needs more details. Basically, you need to call out how this would differ, from, say, regular Linq.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+3. We provide common filters and transforms
```
terrajobst: >We provide common filters and transforms [](start = 3, length = 40)

I'd explicity list the ones you think are common, such as recursively searching by file name prefix, suffix, and exensions (or whatever the ones are you have in mind)

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+4. Recursive behavior is configurable
```
terrajobst: >ecursive behavior is configurable [](start = 4, length = 33)

I think you want more detail here. I assume you want ON, OFF, and delegate so that user code can use the criteria listed in (1) decide whether to walk a sub tree.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+5. Error handling is configurable
```
terrajobst: >Error handling is configurable [](start = 3, length = 30)

Presumably you want to call out that this should be able to avoid throwing & catching exeptions.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+Recursive enumeration is also problematic in that there is no way to handle error states such as access issues or cycles created by links.
+
+These restrictions have a significant impact on file system intensive applications, a key example being MSBuild. We can address these restrictions and provide performant, configurable enumeration.
+
+## Scenarios and User Experience
+
+1. MSBuild can custom filter filesystem entries with limited allocations and form the results in any desired format.
+2. Users can build custom enumerations utilizing completely custom or provided commonly used filters and transforms.
+
+## Requirements
```
terrajobst: This section is intended to be empty; the requirements you want to have should go under goals and the ones you want to scope out under non-goals.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+
+### Goals
+
+1. API is cross platform generic
+2. API minimizes allocations while meeting #1
+
+### Non-Goals
+
+1. API will not expose platform specific data
+3. Error handling configuration is fully customizable
```
terrajobst: >Error handling configuration is fully customizable [](start = 3, length = 50)

That needs more detail as a goal says *Error handling is configurable* while a non-goal says *Error handling configuration is fully customizable*. You need to draw that line.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+        public static IEnumerable<TResult> Enumerate<TResult, TState>(
+            string path,
+            FindTransform<TResult, TState> transform,
+            FindPredicate<TState> predicate,
+			FindPredicate<TState> recursePredicate = null,
+            TState state = default,
+            FindOptions options = FindOptions.None);
+    }
+
+    public static class DirectoryInfo
```
danmosemsft: DirectoryInfo is not static of course

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+### Proposed API surface
+
+``` C#
+namespace System.IO
+{
+    /// <summary>
+    /// Delegate for filtering out find results.
+    /// </summary>
+    internal delegate bool FindPredicate<TState>(ref RawFindData findData, TState state);
+
```
terrajobst: Presumably you mean `public`, not `internal`, right?

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+### Proposed API surface
+
+``` C#
+namespace System.IO
+{
+    /// <summary>
+    /// Delegate for filtering out find results.
+    /// </summary>
+    internal delegate bool FindPredicate<TState>(ref RawFindData findData, TState state);
```
terrajobst: >internal [](start = 4, length = 8)

Presumably you mean `public`, not `internal`, right?

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+
+    /// <summary>
+    /// Delegate for transforming raw find data into a result.
+    /// </summary>
+    internal delegate TResult FindTransform<TResult, TState>(ref RawFindData findData, TState state);
```
terrajobst: >internal [](start = 4, length = 8)

Presumably you mean `public`, not `internal`, right?

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+        public static FileInfo AsFileInfo(ref RawFindData findData)
+        public static FileSystemInfo AsFileSystemInfo(ref RawFindData findData)
+        public static string AsFileName(ref RawFindData findData)
+        public static string AsFullPath(ref RawFindData findData)
+    }
+}
+
+```
+
+### DosMatcher
```
danmosemsft: I assume you would have a RegexMatcher as well.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+
+        /// <summary>
+        /// Return true if the given expression matches the given name.
+        /// </summary>
+        public unsafe static bool MatchPattern(string expression, ReadOnlySpan<char> name, bool ignoreCase = true)
+    }
+}
+```
+
+### Samples
```
terrajobst: >Samples [](start = 4, length = 7)

I'd take both the `DosMatcher` as well as this sample and make it part of a scenario.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,238 @@
...
+
+Getting full path of all files matching a given name pattern (close to what FindFiles does, but returning the full path):
+
+``` C#
+public static FindEnumerable<string, string> GetFiles(string directory,
```
danmosemsft: This sample has a heavy cognitive load:

- Tuples
- ref
- Two delegates, likely wrapping other BCL API, which intellisense won't help find
- concept of passing in a state at the start of an enumeration

It's the ultimate API, but it is likely intimidating for someone who simply wants look for files matching a regex, without caring about every drop of performance

danmosemsft:
@KrzysztofCwalina I'd be interested in your feedback on the API.

/cc @stephentoub who's out.

svick:
Regarding `NotJustDotOrDotDot`: I get that `.` and `..` are commonly included in directory listings. But are there actually cases when they are useful when enumerating a directory? If not, `NotJustDotOrDotDot` could be implicit, making any issues with naming it moot.

accepted/file-enumeration.md
JeremyKuhne: For sure. Doing this doesn't preclude us adding simpler overloads, and I would actually expect to eventually.

---
In reply to: [154808558](https://github.com/dotnet/designs/pull/24#discussion_r154808558) [](ancestors = 154808558)

accepted/file-enumeration.md
JeremyKuhne: Ok

---
In reply to: [154807983](https://github.com/dotnet/designs/pull/24#discussion_r154807983) [](ancestors = 154807983)

accepted/file-enumeration.md
JeremyKuhne: Ideally. We might have to have less than ideal perf to start with if we don't get the Span overloads on Regex at first.

---
In reply to: [154807693](https://github.com/dotnet/designs/pull/24#discussion_r154807693) [](ancestors = 154807693)

accepted/file-enumeration.md
JeremyKuhne: Yup, misstype

---
In reply to: [154807425](https://github.com/dotnet/designs/pull/24#discussion_r154807425) [](ancestors = 154807425)

accepted/file-enumeration.md
JeremyKuhne: Yep, copy paste fail

---
In reply to: [154807630](https://github.com/dotnet/designs/pull/24#discussion_r154807630) [](ancestors = 154807630)

accepted/file-enumeration.md
JeremyKuhne: Got it

---
In reply to: [154807326](https://github.com/dotnet/designs/pull/24#discussion_r154807326) [](ancestors = 154807326)

JeremyKuhne:
> I get that . and .. are commonly included in directory listings. But are there actually cases when they are useful when enumerating a directory?

In theory they can just be assumed with the right context. I'm hesitant to not make all filtering explicit, but I'm definitely giving this one serious thought. I'd hate to fully block something valid unwittingly. Perhaps we can make it implicit in the filters we provide?

MichalStrehovsky:
This feels kind of like a very thin wrapper over `FindFirstFile` Windows API (e.g. the `public ReadOnlySpan<char> FileName { get; }` doesn't sound like something that will translate well to anything that is not Windows).

Will the API surface allow for an implementation that lets the underlying filesystem run pieces of the query? E.g. some filesystems have options to query just for directories. This API seems to be geared towards "give me everything and I'll filter out what I want", not giving much chance to optimize the query. Might be important for network filesystems.

Kind of similar question about FindRawData - would there be any way to indicate what metadata I would like to access when doing the query? E.g. SMB defines multiple information levels to return with their FindData equivalent - will this API allow for asking for "just the right amount" (do not transfer data I won't need, but also avoid lazily fetching data that could have been provided in one go).

JeremyKuhne:
> (e.g. the public ReadOnlySpan<char> FileName { get; } doesn't sound like something that will translate well to anything that is not Windows)

We could potentially provide a ReadOnlySpan<byte> with an encoding in the future, say by putting an interface on the type (along with other things such as the inode). I thought about doing using byte initially, but it makes writing cross plat extensions much more complicated. We can optimize internally for the common ones we provide in any number of ways.

> Will the API surface allow for an implementation that lets the underlying filesystem run pieces of the query?

Only in what we provide if you're talking about just files or just directories. The goal of this is to provide the best performance we can in a cross-platform abstracted way. NT has no way to filter other than by name. That said, I'd like to see examples of Unix APIs that let you filter or concrete suggestions for our surface area here. 

We could provide a special PlatformFilter or something like that, but it reintroduces the problem we have now with recursion- you'd have to iterate twice to find subdirectories. 

> Kind of similar question about FindRawData - would there be any way to indicate what metadata I would like to access when doing the query?

 The FindData struct exposes the data through properties specifically to allow lazily filling in the data on Unix (names included- we won't convert from UTF-8 until needed, for example). The surface area is designed around what Windows provides "normally". Anything beyond this set would need to be grabbed by querying further on each hit.

accepted/file-enumeration.md
danmosemsft: I think it would be helpful to propose those here.

poke:
@svick 
> @poke An important part of the proposal is limiting recursion. How would you do that using a LINQ-like API?

Not sure what exactly you mean here, but I’m thinking of a `IQueryable` way of constructing a query first before executing it against the file system. So you would just create a query configuration using a fluent syntax which could then be executed in a very similar way to this API.

---

@JeremyKuhne 
> @poke It is a bit complicated due to the need to keep allocations to a minimum.

Is this about allocations while enumerating the files, or would you also want to reduce allocations for the query itself?

> It is meant to be an advanced extension point that others can build solutions on for high-impact scenarios that aren't broadly applicable.

I understand that but I still believe that we should try to design an API that will allow others to use it just as well for less critical applications. At least as long as we’re still in an early design phase, we could at least *try* to evaluate other options here.

Since  dotnet/csharplang#186 came up a few times now; what is the time plan for this API? Would waiting for / building upon ref extension methods be an acceptable thing here? That might allow us to make everything a bit simpler.

---

Anyway, I’m thinking of an implementation like this (rough and unflexible sketch, just to show the direction):

```
public class FindEnumeratorQuery
{
    private string _directory;
    private string _state;
    private FindOptions _options = FindOptions.None;
    private IList<FindPredicate<string>> _predicates = new List<FindPredicate<string>>();
    private FindTransform<TResult, TState> _transform;

    public FindEnumeratorQuery(string directory, bool recursive = false)
    {
        _directory = directory;
        if (recursive)
            _options |= FindOptions.Recurse;
    }

    public FindEnumeratorQuery MatchPattern(string expression, bool ignoreCase = false)
    {
        _state = DosMatcher.TranslateExpression(expression);
        _predicates.Add((ref RawFindData findData, string expr) => DosMatcher.MatchPattern(expr, findData.FileName, ignoreCase: ignoreCase));
    }
    public FindEnumeratorQuery IsDirectory() { … }
    public FindEnumeratorQuery IsNotDirectory() { … }

    public FindEnumeratorQuery AsFullPath() { … }

    public FindEnumerable<string, string> GetFiles<TResult>()
    {
        return new FindEnumerable<string, string>(_directory, _transform,
            CombinePredicates(_predicates),
            state: _state, options: _options);
    }

    private static FindPredicate<string> CombinePredicates(IEnumerable<FindPredicate<string>> predicates) {…}
}
```

I believe that, if we accept the allocations in the query construction, we could actually design this in a way that the API becomes more friendly without hurting the performance during the actual enumeration.

MichalStrehovsky:
> The goal of this is to provide the best performance we can in a cross-platform abstracted way

I don't see the abstraction here; this API basically assumes that the best way to query the filesystem on the underlying platform is with a FindFirstFile-like API.

The approach that @poke is going for (with `IQueryable`-like fluent syntax) would let the underlying implementation choose the best possible way to get the information user asks for.

The design as proposed here is focusing too much on making *just the managed parts* fast, forgoing the fact that the underlying platform also needs to do allocations (and I/O, plus potentially networking operations) to satisfy the query. We might be able to get significant savings from limiting that work, but I don't see how the provider could do that, given an API surface that doesn't let the provider know what the user is going for. The time spent doing that extra I/O is orders of magnitude bigger than time spent doing a Gen0 collection of a couple short lived objects (that are per query - not per result).

Even just looking at Windows - with the proposed API, do we know whether we should call FindFirstFile with [FIND_FIRST_EX_LARGE_FETCH](https://blogs.msdn.microsoft.com/oldnewthing/20131024-00/?p=2843/)?

jnm2:
> Even just looking at Windows - with the proposed API, do we know whether we should call FindFirstFile with FIND_FIRST_EX_LARGE_FETCH?

I have business code calling `FindFirstFile` with `FIND_FIRST_EX_LARGE_FETCH` because the alternatives I tried was measured to be thousands of times slower for network shares with thousands of files.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Scenarios and User Experience
+
+1. MSBuild can custom filter filesystem entries with limited allocations and form the results in any desired format.
+2. Users can build custom enumerations utilizing completely custom or provided commonly used filters and transforms.
+
+## Requirements
+
+
+### Goals
```
terrajobst: >## Goals [](start = 1, length = 8)

The primary goal seems to be:

* Provide an API shape that is as expressive as the current API shape while *also* reducing the number of allocations to a minimum. We're willing to compromise some usability, but the complexity should be about the same as many of the other `Span<T>`-based APIs.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,248 @@
...
+	- DOS style filters (Legacy- `*/?` with DOS semantics, e.g. `*.` is all files without an extension)   
+	- Simple Regex filter
+	- Simpler globbing (`*/?` without DOS style variants)
+	- Set of extensions (`*.c`, `*.cpp`, `*.cc`, `*.cxx`, etc.)
+4. Recursive behavior is configurable
+	- On/Off
+	- Predicate based on FileData
+5. Can avoid throwing access denied exceptions
+
+### Non-Goals
```
terrajobst: One non-goal to call out is that  we don't intend to replace the existing IO APIs -- these are meant to be advanced APIs for folks that really have to care about performance.

accepted/file-enumeration.md
```diff
@@ -0,0 +1,248 @@
...
+            FindPredicate<TState> predicate,
+			FindPredicate<TState> recursePredicate = null,
+            TState state = default,
+            FindOptions options = FindOptions.None);
+    }
+
+    /// <summary>
+    /// Used for processing and filtering find results.
+    /// </summary>
+    public ref struct RawFindData
```
MichalStrehovsky: Btw: why is this a `ref struct`? Is it expected that the struct will hold byreflike fields in the future? Looking at the [current implementation](https://github.com/dotnet/corefx/blob/eb0d4388c339288a589bdd7f5e0c9fcd61da4362/src/System.IO.FileSystem/src/System/IO/RawFindData.cs) there aren't any.

JeremyKuhne:
@poke

> Is this about allocations while enumerating the files, or would you also want to reduce allocations for the query itself?

>I believe that, if we accept the allocations in the query construction

It is primarily about reducing string allocations for results we don't care about- which is the primary measured perf problem we're trying to solve. We also, however, want to keep the allocations constrained on construction where possible. _Additionally_ we want to keep performance high. Using anonymous methods gets you caching of the delegates, which is part of why I settled on the current pattern after trying a number of others. `String.Create()` also lets you at internals, which helped shape this.

@MichalStrehovsky 

> I don't see the abstraction here; this API basically assumes that the best way to query the filesystem on the underlying platform is with a FindFirstFile-like API.

This API is trying to provide the set of data that goes into making a `FileSystemInfo`, letting you provide logic that filters out results while pushing string allocations as late as possible. That is the abstraction, and the problem it is meant to solve. The managed allocations are the biggest impact, by far, on measured directory iteration performance in key workloads like MSBuild and VS. Making the change as shown (which is currently in Core) improved our own perf two-fold on large filesets (checking thousands or hundreds of thousands of files). We can and will provide common predicates for both platforms and optimize further on those where possible/warranted by providing platform specific implementations that get at more internals. 

> The design as proposed here is focusing too much on making just the managed parts fast

That is what this proposal is. It isn't to provide direct access to the OS APIs- it's to provide access to the internals of our iterator. I don't see a way to extend the providing of `RawFindData` on the iterator without making this way more complicated or defeating the allocation reductions goal.

> The time spent doing that extra I/O 

What extra I/O? There is no way to filter out the disk access for a directory enumeration. The file system has to get all of the data to make a decision. How many extra allocations you have up the stack is what makes all of the difference. I removed a fundamental block on Windows by simply _not_ calling FindFirstFile. It has to copy the NT structure into a Win32 structure and doesn't have context for a "session". Additionally, it allocates a new 4K buffer for _every_ single directory you access.

@jnm2 

> I have business code calling FindFirstFile with FIND_FIRST_EX_LARGE_FETCH because each of the alternatives I tried was measured to be thousands of times slower for network shares with thousands of files.

I've skipped FindFirstFile entirely and use our own 4K buffer. I saw no impact on local searches when I played with various sizes here. I'm happy to make the buffer size configurable and would love to get some real world data on the impact of buffer sizes on shares. If you have the opportunity to play with tweaking what we have please do. I'll try and experiment some here as well.

https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/src/System/IO/FindEnumerable.Windows.cs#L114

We could provide a single flag option for `UseLargeBuffer` or something similar. Again, data to inform that would be fantastic.

JeremyKuhne:
@davidwrighton If you have a chance to look at this I'd appreciate it. I'm also interested in your thoughts on the impact of trying to put an interface on the data struct and passing the interface through the delegates rather than the struct itself (for testability, etc.).

accepted/file-enumeration.md
```diff
@@ -0,0 +1,342 @@
...
+
+These restrictions have a significant impact on file system intensive applications, a key example being MSBuild. This document proposes a new set of primitive file and directory traversal APIs that are optimized for providing more flexibility while keeping the overhead to a minimum so that enumeration becomes both more powerful as well as more performant.
+
+## Scenarios and User Experience
+
+### Getting full paths
+
+To get full paths for files, one would currently have to do the following:
+
+```
```
terrajobst: Add `charp` to get syntax highlighting

accepted/file-enumeration.md
```diff
@@ -0,0 +1,385 @@
...
+        /// Return true if the given expression matches the given name.
+        /// </summary>
+        public unsafe static bool MatchPattern(string expression, ReadOnlySpan<char> name, bool ignoreCase = true)
+    }
+}
+```
+
+### Potential testability variant
+
+Having `FindData` be a struct is critical to meet the performance goals for the feature. To allow testing of Filters and Predicates we could also expose an interface of `IFindData`:
```
terrajobst: We should probably discuss in the API review in more detail, but I feel  I'm not sure this minor tweak is worth the complexity. It seems in order to support testing, we'd need to allow passing something around that can serve to *produce* `FindData<T>` instances. In other words, for testing it seems just being able to to query a single find data is very limited, compared to an `IFileSystem` (whatever that would look like).

migueldeicaza:
My comments:

I find this new API difficult to read, I am hoping that this can live in a parallel universe and not in the core of the .NET frameworks.
 
* Just how much memory is being allocated today in these scenarios that you are looking for, and what are the savings with this design?
* Has someone looked into rewriting the alloc-happy code in the framework to avoid this, like Mono used to?

The examples shown also rely on APIs that require FileInfos to be created, and are themselves not a pinnacle of low-allocation work, starting with the IEnumerable right there, and the use of LINQ ☺
 
The first one, can be rewritten without allocating FileInfos, like this:
 
```
foreach (var t in Directory.GetFiles ("/tmp")) Console.WriteLine (Path.GetFullPath (t))
 ```

The second example, filtering for extensions that is given is a poor implementation, if you want to match on extensions, you can let Directory do the work:
 
```
foreach (var t in Directory.GetFiles (“*.ext”)) Console.WriteLine (Path.GetFullPath (t))
```

The last one addresses all three items from the list of pros on that document.

I checked both of those, like this, and not a single FileInfo was created:
 
``` 
mac$ MONO_ENV_OPTIONS='--trace=T:System.IO.FileInfo' csharp -e 'foreach (var t in System.IO.Directory.GetFiles ("/etc", "*.conf")) Console.WriteLine (System.IO.Path.GetFullPath (t))' | grep FileInfo
mac$ MONO_ENV_OPTIONS='--trace=T:System.IO.FileInfo' csharp -e 'foreach (var t in System.IO.Directory.GetFiles ("/etc", "*")) Console.WriteLine (System.IO.Path.GetFullPath (t))' | grep FileInfo
```

As the doc points out, recursive directory scanning will be more expensive on Unix, no matter what you do, as you need to probe each file listed that you get back from the OS.   But if we move the scanning to unmanaged code, no FileInfos will ever be produced, nor stats surfaced on Unix.
 
I just looked at the implementation in CoreFX (which we just took for Mono) and it looks allocation happy by default, something that the original Mono version did not do.   Mono used to do the work in unmanaged code, and surfaced a properly-sized array, while the new implementation creates a `List<T>` to scan, and then does a `ToArray()` at the end.
 
Some questions:

danmosemsft:
@migueldeicaza did the questions you mention get cut off?  Great feedback BTW

JeremyKuhne:
> I find this new API difficult to read, I am hoping that this can live in a parallel universe and not in the core of the .NET frameworks.

We won't be getting rid of the existing APIs. This is not intended for general day-to-day use- we'll keep the "simple" APIs as the primary entry points. It's meant to allow advanced users such as MSBuild to do custom filtering, caching, and transforms without writing their own native implementations.

> The first one, can be rewritten without allocating FileInfos, like this

Yes, but you trade in extra string allocations and an expensive normalization call for every entry. Same for the second example. I kept these initial two relatively simple to illustrate usage. I also did not fully optimize the third example to keep it conceptually simpler (e.g. I wouldn’t use Linq in reality). I’ll add some comments.

> I just looked at the implementation in CoreFX (which we just took for Mono) and it looks allocation happy by default

For the CoreFX implementation are you looking at what is in master? Master was changed recently to use an internal implementation that looks like this proposal (for Windows). Even with the changes, our APIs that return arrays currently use IEnumerable internally. I've opened an issue about potentially optimizing those scenarios. (https://github.com/dotnet/corefx/issues/25863)

JeremyKuhne:
Note: This is based on the existing internal Windows implementation, see the CoreFX master branch for some portion of this in action. A branch implemented to spec  (for Windows only) can be found here: https://github.com/JeremyKuhne/corefx/tree/findpublic.

terrajobst:
@migueldeicaza 

By the way, merging doesn't mean we're ignoring your feedback. It just means we've accepted the business goal and are committed to ship the feature. The design will likely evolve, especially after an API review. So the discussion is hardly over!

powercode:
PowerShell, specifically the FileSystemProvider, would be a customer for this to.

PowerShell today has perf issues with enumerating files recursively and getting the `Mode`  and `LinkType` synthetic properties.

It would be good to make sure that these new APIs will work well with that core scenario. @SteveL-MSFT.

PR #25, opened 2017-12-01
Package information in the API reference

terrajobst:
This document captures what @mairaw, @dend, @weshaggard and myself discussed earlier this week.


accepted/package-doc/package-doc.md
```diff
@@ -0,0 +1,170 @@
...
+This file should be an input to the process that tags APIs with specific
+framework monikers (e.g. that tags `System.String` to be contained in `net45`,
+`net461`, `netcoreapp1.0`, `netcoreapp1.1` etc).
+
+Please note that the process as outlined above will ensure that if a package
+only provides an API for .NET Standard it will show up on all frameworks that
+implement that version of the standard, which is the desired behavior.
+Furthermore, if an API is made available for a specific framework, it will also
+show up as being available on all later versions of the same framework, which is
+also desired behavior.
```
terrajobst: Hey @weshaggard. We need to think about what this means for cases where an API is moving in-box. In that case, the package needs to provide an empty folder for that framework with the special marker file `_._` -- or -- a facade that type forwards. If it's the former, we get the behavior for free. If its the latter, the indexer probably just needs to ignore type forwards, but a more correct behavior would be to follow type forwards but ignore assemblies it cannot resolve, i.e. are outside the package.

Thoughts?

weshaggard: For indexing purposes I would suggest we just let whatever we define as "in the platform" override whatever might come from a nuget package for the given platform. That way we don't really have to worry about this case. There are going to be cases were we have a library in a package and one in the platform so we should just have the overriding characteristic.

weshaggard:
Doc looks reasonable to me.

svick:
How is this going to behave for types like the mentioned `SqlClient`, which require a package on some frameworks (.Net Core), but not on others (.Net Framework)?

Also, do I understand it correctly that this would fix https://github.com/dotnet/docs/issues/1849?

accepted/package-doc/package-doc.md
terrajobst: Makes sense. I've clarified this in 2e3398845d935a01fb5516bc9ecd8b6d7f92505c.

PR #26, opened 2017-12-05
Add Performance Profiling Controller Design

brianrob:
Design for out-of-proc exposure of EventPipe profiling and tracing functionality.

### DIFF
--- a/accepted/performance-profiling-controller.md
+++ b/accepted/performance-profiling-controller.md
@@ -1,0 +1,53 @@
+# Performance Profiling Controller Design #
+
+### Introduction ###
+
+This document describes the end-to-end scenarios and work required to build a performance profiling controller for .NET Core.  The controller is responsible for control of the profiling infrastructure and exposure of performance data produced by .NET performance diagnostics components in a simple and cross-platform way.
+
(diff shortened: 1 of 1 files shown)


brianrob:
@Petermarcu, @valenis, @vancem, review feedback would be very much appreciated.

SergeyKanzhelev:
CC: @nikmd23 @mmilirud @karolz-ms

@brianrob would it be more logical to offload configuration and data collection part to Application Insights SDK. ASP.NET Core 2.0 already uses Application Insights SDK as a Visual Studio debug session experience - see https://blogs.msdn.microsoft.com/webdev/2017/08/14/announcing-asp-net-core-2-0/ 

I like the idea of http-based output & configuration experience. So you can configure Application Insights SDK using this experience and get more than just http page. You'd also get channels to Elastic Search, stdout, ETW on windows.

brianrob:
Collected feedback offline and merging.

pharring:
I like the idea of a remote-control for profiling. But hosting an HTTP server seems (to me) like a big pill to swallow. Isn't there something lower level?

brianrob:
@pharring, an HTTP server just just one option - I expect that we'll architect things such that other transports can be used as desired.  I would also personally like to have a way to control tracing without HTTP (something on the command line), but I'm not clear on exactly what the mechanism will be.

PR #27, opened 2017-12-07
Add initial draft of "Automatic Binding Redirection By Default"

terrajobst:
This document describes the changes necessary for turning on automatic binding redirect generation by default.

### DIFF
--- a/accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
+++ b/accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
@@ -16,7 +16,7 @@ assembly in version `X+N`. The application author has a way to influence this by
 using a feature called [binding redirects]. Those are entries in the
 `app.config` that instruct the CLR to unify specific version ranges.
 
-These cases frequently occur in these cases:
+This frequently occurs in these cases:
 
 * **NuGet graphs**. The application author often references a graph of NuGet
   packages while simultaneously also referencing newer versions of some of the
@@ -44,8 +44,8 @@ again. The net effect is that the vast majority of our customers have this
 setting still turned off.
 
 The proposal is to turn this setting on by default but allow application authors
-to turn the setting off if necessary. In other words, we change it from an opt-
-in feature to an opt-out feature.
+to turn the setting off if necessary. In other words, we change it from an
+opt-in feature to an opt-out feature.
 
 [binding redirects]: https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/redirect-assembly-versions
 [abrg]: https://docs.microsoft.com/en-us/dotnet/framework/configure-apps/how-to-enable-and-disable-automatic-binding-redirection
@@ -68,14 +68,14 @@ with a `FileLoadException`.
 
 ### Upgrading a NuGet dependency
 
-Robert is using FizzBuzz, a documentation framework that produces help files out
-of Markdown documents. FizzBuzz depends on Markdig, a popular .NET markdown
+Robert is using DocBuzz, a documentation framework that produces help files out
+of Markdown documents. DocBuzz depends on Markdig, a popular .NET markdown
 library. Robert encounters an issue with some of his recent documentation
 changes which causes Markdig to crash. He notices that while there is no updated
-version of FizzBuzz, there is an updated version of Markdig that fixes his
-issue. Robert updates his (indirect) dependency on Markdig to the latest version
-that fixes the issue. He triggers a rebuild of the documentation and is happy to
-see that the crash no longer occurs.
+version of DocBuzz, there is an updated version of Markdig that fixes his issue.
+Robert updates his (indirect) dependency on Markdig to the latest version that
+fixes the issue. He triggers a rebuild of the documentation and is happy to see
+that the crash no longer occurs.
 
 Robert doesn't know that this is thanks to automatic binding redirect generation
 being on by default. Otherwise his application would have crashed with a
@@ -87,7 +87,8 @@ being on by default. Otherwise his application would have crashed with a
 
 * **Turning this setting on must not break working applications**. We expect
   this setting to make more applications work, but we must not break
-  applications where adding binding redirects breaks them.
+  applications where adding binding redirects breaks them. See Q&A section why
+  we believe this is achievable.
 
 * **It must be possible to opt-out**. The customer base on .NET Framework is
   fairly large and has an unfathomable amount of different configurations. We
--- a/accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
+++ b/accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
@@ -1,0 +1,137 @@
+# Automatic Binding Redirection By Default
+
+**PM** [Immo Landwerth](https://github.com/terrajobst) |
+**Dev** *TBD* |
+**Feature**: [#dotnet/sdk/1405](https://github.com/dotnet/sdk/issues/1405)
+
+Four years ago, .NET Framework 4.5.1 shipped an MSBuild feature that
+[automatically generates binding redirects][abrg]. Unfortunately, it's
+effectively off for most of our customers. This proposal is about turning the
+feature on by default by making it an opt-out.
+
+The .NET Framework runtime is the only runtime that uses an exact binding match
+policy. That means that if the runtime needs to resolve a reference to an
+assembly in version `X`, it will not succeed if the application deployed this
+assembly in version `X+N`. The application author has a way to influence this by
+using a feature called [binding redirects]. Those are entries in the
+`app.config` that instruct the CLR to unify specific version ranges.
+
+These cases frequently occur in these cases:
+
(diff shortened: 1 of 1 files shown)


accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -0,0 +1,137 @@
...
+Robert is using FizzBuzz, a documentation framework that produces help files out
+of Markdown documents. FizzBuzz depends on Markdig, a popular .NET markdown
+library. Robert encounters an issue with some of his recent documentation
+changes which causes Markdig to crash. He notices that while there is no updated
+version of FizzBuzz, there is an updated version of Markdig that fixes his
+issue. Robert updates his (indirect) dependency on Markdig to the latest version
+that fixes the issue. He triggers a rebuild of the documentation and is happy to
+see that the crash no longer occurs.
+
+Michelle doesn't know that this is thanks to automatic binding redirect
```
bording: This should be Robert, not Michelle!

terrajobst: LOL. Good catch.

terrajobst:
@dsplaisted 

> It's worth calling out that this still doesn't do anything to improve ASP.NET projects.

Good call. Done with 9944a8d.

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -0,0 +1,161 @@
...
+Michelle maintains a .NET Framework 4.5 application. She tries to install the
+new `System.IO.Pipelines` NuGet package that targets .NET Standard 2.0. She gets
+an error message from NuGet that this isn't compatible with .NET Framework 4.5.
+After some research, she learns that she must at least target .NET Framework
+4.6.1 in order to use .NET Standard 2.0 packages. After upgrading to .NET
+Framework 4.6.1, the package installs without errors. Michelle then starts to
+use pipelines from her app. When she F5s her application it just works fine.
+
+Michelle doesn't know that this is thanks to automatic binding redirect
+generation to be on by default. Otherwise her application would have crashed
```
jnm2: Should be: "generation _being_ on"

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -0,0 +1,161 @@
...
+of Markdown documents. FizzBuzz depends on Markdig, a popular .NET markdown
+library. Robert encounters an issue with some of his recent documentation
+changes which causes Markdig to crash. He notices that while there is no updated
+version of FizzBuzz, there is an updated version of Markdig that fixes his
+issue. Robert updates his (indirect) dependency on Markdig to the latest version
+that fixes the issue. He triggers a rebuild of the documentation and is happy to
+see that the crash no longer occurs.
+
+Robert doesn't know that this is thanks to automatic binding redirect generation
+to be on by default. Otherwise his application would have crashed with a
```
jnm2: Same [Should be: "generation being on"]

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -0,0 +1,161 @@
...
+
+We generally don't want to compute and emit binding redirects because for most
+class library projects it's just wasting time. We need to figure out how we can
+detect unit testing projects during build and turn the setting on.
+
+## Q & A
+
+### Why is the setting not conditioned on the target framework?
+
+Over the last years we have not encountered a single instance where binding
```
weshaggard: That is a bit bold. We just hit an issue with ValueTuple where having the binding redirect actually caused the application to fail when upgrading to 4.7.1 on the box. Granted it was a bug we fixed in the netstandard facade but non-the-less if there wasn't a binding redirect it would have worked just but with it it failed.

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -0,0 +1,161 @@
...
+  already existing feature in MSBuild which uses the also existing binding
+  redirect feature of the CLR.
+
+## Design
+
+Somewhere in the common props we need to add the following snippet:
+
+```xml
+<PropertyGroup>
+    <AutoGenerateBindingRedirects Condition="'$(AutoGenerateBindingRedirects)' == ''">True</AutoGenerateBindingRedirects>
```
weshaggard: So the opt-out is for the customer to set this to false in their project?

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
terrajobst: But the problem isn't with Automatic Generation of Binding Redirects (i.e. false or incorrect generation of them). In other words, turning of ABRG wouldn't have fixed ValueTuple either.

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
terrajobst: Correct

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
weshaggard: That's correct but my point is having binding redirects can cause some issues which is what I was point out. Mainly I think your statement may be a little strong and will likely get some feedback because of it.

terrajobst: That's fair, all I'll try to reword. I think it's important to point out that ABRG isn't the culprit thought and that turning it on by default is safe

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -0,0 +1,166 @@
...
+being on by default. Otherwise his application would have crashed with a
+`FileLoadException`.
+
+## Requirements
+
+### Goals
+
+* **Turning this setting on must not break working applications**. We expect
+  this setting to make more applications work, but we must not break
+  applications where adding binding redirects breaks them.
```
Andrew-MSFT: I think you need a little bit more on the risk assessment, e.g. when we talked in person, you explained why you didn't think it is likely to break many people

terrajobst: I believe I have that covered in the [Q&A section](https://github.com/terrajobst/designs/blob/f1133dbb244f26b65d4fb4f085854c3af4f32ccf/accepted/automatic-redirects-by-default/automatic-redirects-by-default.md#why-is-the-setting-not-conditioned-on-the-target-framework)

terrajobst:
I believe I've addressed all the feedback. Before merging, I'd like to close which team/person will own the implementation.

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -0,0 +1,167 @@
...
+* **NuGet graphs**. The application author often references a graph of NuGet
+  packages while simultaneously also referencing newer versions of some of the
+  indirectly referenced packages. Getting consistent graphs is often not
+  possible or even desirable.
+
+* **.NET Standard 2.0 support files**. Binding redirects are also required for
+  the .NET Standard 2.0 support files that .NET Framework 4.6.1 applications
+  need to deploy in order to load .NET Standard libraries.
+
+Correctly authoring binding redirects if non-trivial as it requires the author
```
AlexGhiondea: Typo — redirects *is*, not *if*

terrajobst: Eagle eye! Fixed.

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
nguerrera: Blindly turning this true everywhere is going to cause perf issues.

terrajobst:
I have incorporated the feedback from the feature crew review.

accepted/automatic-redirects-by-default/automatic-redirects-by-default.md
```diff
@@ -85,6 +85,37 @@ being on by default. Otherwise his application would have crashed with a
...
+  doesn't do harm but it does impact performance as the MSBuild has to walk more
+  assemblies.
+
+* **Turned off in web projects**. Web projects cannot use an `app.config` file
+  as they use `web.config`, which is usually taken from the source directory. At
+  best, dropping the extra file is just noise but at worse it might confuse the
+  build or the runtime.
+
+* **Automatically turned on when targeting .NET Framework 4.6.1 or higher and a
+  .NET Standard 2.0 library is consumed**. This addresses the primary issues
```
nguerrera: The extensions code injects facades/libs when referencing .NET Standard 1.5 or greater IIRC. I presume that we should be enabling it for the same cases.

terrajobst: Excellent point. Clarified.

PR #30, opened 2018-01-22
WIP: Add proposal for range

terrajobst:
@KrzysztofCwalina

### DIFF
--- a/accepted/system-range/system-range.md
+++ b/accepted/system-range/system-range.md
@@ -160,11 +160,12 @@ namespace System {
     public struct Range : IEnumerable, IEnumerable<int>, IEquatable<Range> {
         public static Range Empty { get; }
         public static Range All { get; }
-        public static Range Between(int start, int end);
-        public static Range From(int start);
-        public static Range To(int end);
+        public static Range Between(int startIndex, int endIndex);
+        public static Range From(int startIndex);
+        public static Range To(int endIndex);
         public Range(int index, int length);
         public int Index { get; }
+        public int EndIndex { get; }
         public int Length { get; }
         public bool IsUnbounded { get; }
         public Range Bind(int length);
@@ -258,10 +259,11 @@ namespace System {
 
 ### Open Design Points
 
-* Right now, we cannot retrieve `Index` or `Length` if the range is unbounded
-  (it throws `NotSupportedException`).
-    - We could expose `Start: int?` and `End: int?` and then add a factory
-      method `Between(int? start, int? end)`
+* Right now, we cannot retrieve `Index`, `EndIndex` or `Length` if the
+  range is unbounded (it throws `NotSupportedException`).
+    - We could expose `Index: int?` and `EndIndex: int?` and then add a
+      factory method `Between(int? start, int? end)` but that feels
+      overengineered
 
 * Should `Range` support containment checks?
     - Positions and/or ranges?
--- a/accepted/system-range/system-range.md
+++ b/accepted/system-range/system-range.md
@@ -11,19 +11,18 @@ language has to be bound to a .NET type, which we call `System.Range`.
 ### Construcing ranges
 
 ```csharp
-// Ranges are constructed with an index and a length
+// Ranges are constructed with an index and a length:
 
 var fiveToTen = new Range(index: 5, length: 6);
 
-// By using a factory method, they can also be constructed from two indices.
+// By using a factory method, they can also be constructed from two indices:
 
 var alsoFiveToTen = Range.Between(5, 11);
 
-// The upper bound is considered exclusive so this will print the 1, 2, 3
+// The upper bound is considered exclusive so this will print 1, 2, 3:
 
 foreach (var value in Range.Between(1, 4))
     Console.WriteLine(value);
-
 ```
 
 The C# syntax for ranges isn't finalized and is actively worked on. While this
@@ -32,25 +31,33 @@ so that we can look at source code as it's likely written:
 
 ```csharp
 var fiveToTen = 5..11; // Equivalent to Range.Between(5, 11)
+
+// You can also construct ranges with index and length:
+var fiveToTen = 5.:6; // Equivalent to new Range.Between(5, 6)
 ```
 
-### Open ended ranges
+### Unbounded ranges
 
 ```csharp
-// Ranges can be open ended:
+// Ranges can be unbounded:
 
 var fiveToEnd = 5..;  // Equivalent to Range.From(5) i.e. missing upper bound
 var startToTen = ..1; // Equivalent to Range.To(11) i.e. missing lower bound
 var everything = ..;  // Equivalent to Range.All i.e. missing upper and lower bound
 
-// Asking for Index, Length or enumerating an open range will throw
+// When the lower bound is omitted, it's interpreted to be zero.
+//
+// When the upper bound is omitted, it's interpreted to be the length of the
+// receiving collection.
+//
+// Asking for Length or enumerating an unbound range will throw
 // NotSupportedException. The consumer of the range is expected to 
-// close it over the particular data:
+// bind it for a particular length:
 
 void Print(int[] data, Range range)
 {
-    if (range.IsOpen)
-        range = range.Close(0..data.Length);
+    if (range.IsUnbound)
+        range = range.Bind(data.Length);
 
     foreach (var index in range)
         Console.WriteLine(data[index]);
@@ -60,6 +67,19 @@ var data = new [] { 'a', 'b', 'c' };
 Print(data, 2..); // Prints c
 Print(data, ..2); // Prints a, b
 Print(data, ..);  // Prints a, b, c
+```
+
+### Negative bounds
+
+```csharp
+// You can also use negative bounds. They are interpreted as being relative
+// to the length of the receiving collection, so -1 is the last element,
+// -2 is the second to last element and so on.
+
+var data = new [] { 'a', 'b', 'c' };
+Print(data, -2.:2);  // Prints b, c
+Print(data, -1..);   // Prints c
+Print(data, -3..-1); // Prints a, b
 ```
 
 ### Using ranges in APIs
@@ -80,7 +100,7 @@ This allows creating substrings simply by using the indexer:
 ```csharp
 var helloWorld = "Hello, World!";
 var hello = helloWorld[..5];
-var world = helloWorld[7..12];
+var world = helloWorld[7.:5];
 ```
 
 While this might appear to just be minor syntactic sugar it allows for creating
@@ -106,10 +126,12 @@ This might look like this:
 ```csharp
 Tensor<float> testDataTable = GetTestData();
 int columnCount = testDataTable.GetLength(1);
-int outputColumn = columnCount - 1;
 
-Tensor<float> inputTable = testDataTable[.., ..outputColumn];
-Tensor<float> outputTable = testDataTable[.., outputColumn..];
+// Take all rows and all but the last column
+Tensor<float> inputTable = testDataTable[.., ..-1];
+
+// Take all rows and only the last column
+Tensor<float> outputTable = testDataTable[.., -1..]; 
 ```
 
 ## Requirements
@@ -144,8 +166,8 @@ namespace System {
         public Range(int index, int length);
         public int Index { get; }
         public int Length { get; }
-        public bool IsClosed { get; }
-        public Range Close(Range range);
+        public bool IsUnbounded { get; }
+        public Range Bind(int length);
         public override string ToString();
         public override int GetHashCode();
         public bool Equals(Range other);
@@ -156,33 +178,16 @@ namespace System {
 }
 ```
 
-### Companion APIs
+### Negative indices
 
-```csharp
-namespace System {
-    public partial class String {
-        public string this[Range range] { get; }
-    }
-    public partial class T[] {
-        public T[] this[Range range] { get; }
-    }
-    public partial struct ArraySegment<T> {
-        public ArraySegment<T> this[Range range] { get; }
-    }      
-    public partial struct Span<T> {
-        public Span<T> this[Range range] { get; }
-    }
-}
-```
+The semantics of negative indices are as follows:
 
-### Semantics of negative indices
-
-The semantics of negative indices would be as follows:
-
-* Both `start` and `end` can be negative.
+* `length` can never be negative, but both `start` and `end` can be. In other
+  words, calling the constructor with a negative length will fail but you can
+  pass negative values to `From()`, `Between()`, and `To()`.
 * If the value is negative, the size of container is added. So `-1` is the last
   element.
-* Logically, negative bounds requires the range to be closed.
+* Logically, negative bounds requires the range to be bound.
 
 The internal encoding would be as follows:
 
@@ -191,7 +196,7 @@ The internal encoding would be as follows:
   - `length := (start < 0 || end < 0) ? end : end - start`
 
 * `..end`
-  - `index := int.MinValue`
+  - `index := 0`
   - `length := end`
 
 * `start..`
@@ -205,36 +210,69 @@ length of a range is `int.MaxValue` which negated is `int.MinValue + 1`.
 done if either both are positive or both are negative. As soon if only one side
 is negative one needs to know the container size in order to verify that `start`
 is indeed less than or equal to `end`. However, it seems bad form to throw from
-the `Close()` method. Instead, we should follow what Python does and simply
-produce the range `start..start`.
+the `Bind()` method. Instead, we should follow what Python does and clamp the
+produced ranges to the container boundaries if either side is negative.
 
-This encoding ensures that checking whether the range is open only requires
+This encoding ensures that checking whether the range is unbound only requires
 checking if either `index` or `length` is negative. This will cover both missing
 bounds as well as bounds that require adding the container size.
+
+### Companion APIs
+
+```csharp
+namespace System {
+    public partial class String {
+        public string this[Range range] { get; }
+    }
+    public partial class T[] {
+        public ArraySegment<T> this[Range range] { get; }
+    }
+    public partial struct ArraySegment<T> {
+        public ArraySegment<T> this[Range range] { get; }
+    }      
+    public partial struct Span<T> {
+        public Span<T> this[Range range] { get; }
+    }
+    // These aren't needed but they are provided for aesthetics.
+    //
+    // Instead of:
+    //
+    //      data.AsSpan()[start:.length]
+    //      data.AsSpan().Slice(start:.length)
+    //
+    // You can just say:
+    //
+    //      data.AsSpan(start:.length)
+    //
+    public static partial class MemoryExtensions {
+        public static Span<T> AsSpan<T>(this T[] array, Range range);
+        public static Span<T> AsSpan<T>(this ArraySegment<T> arraySegment, Range range);
+        public static Span<T> AsReadOnlySpan<T>(this T[] array, Range range);
+        public static ReadOnlySpan<T> AsReadOnlySpan<T>(this ArraySegment<T> arraySegment, Range range);
+        public static ReadOnlySpan<char> AsReadOnlySpan(this string text, Range range);
+    }
+}
+```
 
 ## Q & A
 
 ### Open Design Points
 
-* Right now, we cannot retrieve `Index` or `Length` if the range is open on any
-  side (it throws `NotSupportedException`).
+* Right now, we cannot retrieve `Index` or `Length` if the range is unbounded
+  (it throws `NotSupportedException`).
     - We could expose `Start: int?` and `End: int?` and then add a factory
       method `Between(int? start, int? end)`
 
-* We should check whether we can handle negative numbers for `Length` so that we
-  can support addressing relative to the end. Check Python to see what they did
-  there.
-
 * Should `Range` support containment checks?
     - Positions and/or ranges?
-    - Question is how to define semantics for open ended ranges.
+    - Question is how to define semantics for unbounded ranges.
 
-* Should `Range` support combining/intersecting open ended ranges?
-    - Be careful though. For open ended slices this might produce invalid
-      values. For instance, you might think combining `3..` and `..12` would
-      produce `3..12` but this would mean indexing into an array of size 5 would
-      fail, while `3..` would have worked.
-    - We could define the operation as `Close(Range)` though.
+* Should `Range` support combining/intersecting unbounded ranges?
+    - Be careful though. For unbounded ranges this might produce invalid values.
+      For instance, you might think combining `3..` and `..12` would produce
+      `3..12` but this would mean indexing into an array of size 5 would fail,
+      while `3..` would have worked.
+    - We could define the operation as `Bind(Range)` though.
 
 * Should we add an interface for range access?
     ```csharp
@@ -262,10 +300,11 @@ bounds as well as bounds that require adding the container size.
         }
         ```
 * Should we suppport steps in ranges? Syntax is unclear and so is whether this
-  means we'd need a separate type.
+  means we'd need a separate type, such as `SteppedRange`. Python supports this:
+  `lower:upper:step` with `:step` being optional and `:1` by default.
 * For the companion types, should we support setters for range-based indexers?
   `myarray[0..10] = Enumerable.Range(1, 10)`
 * Python doesn't create errors for cases where the indexes aren't valid. It
-  looks like when they close ranges the just clamp the start and end position to
+  looks like when they bind ranges the just clamp the start and end position to
   the container. And if the start:end are overlapping, they simply produce an
   empty set.
--- a/accepted/system-range/system-range.md
+++ b/accepted/system-range/system-range.md
@@ -33,7 +33,7 @@ so that we can look at source code as it's likely written:
 var fiveToTen = 5..11; // Equivalent to Range.Between(5, 11)
 
 // You can also construct ranges with index and length:
-var fiveToTen = 5.:6; // Equivalent to new Range.Between(5, 6)
+var fiveToTen = 5.:6; // Equivalent to new Range(5, 6)
 ```
 
 ### Unbounded ranges
@@ -42,7 +42,7 @@ var fiveToTen = 5.:6; // Equivalent to new Range.Between(5, 6)
 // Ranges can be unbounded:
 
 var fiveToEnd = 5..;  // Equivalent to Range.From(5) i.e. missing upper bound
-var startToTen = ..1; // Equivalent to Range.To(11) i.e. missing lower bound
+var startToTen = ..1; // Equivalent to Range.To(1) i.e. missing lower bound
 var everything = ..;  // Equivalent to Range.All i.e. missing upper and lower bound
 
 // When the lower bound is omitted, it's interpreted to be zero.
@@ -125,7 +125,6 @@ This might look like this:
 
 ```csharp
 Tensor<float> testDataTable = GetTestData();
-int columnCount = testDataTable.GetLength(1);
 
 // Take all rows and all but the last column
 Tensor<float> inputTable = testDataTable[.., ..-1];
@@ -141,8 +140,7 @@ Tensor<float> outputTable = testDataTable[.., -1..];
 * We need a type that both, plays well with existing concepts in the base class
   library, as well as the intended syntax for C#
 * We need to ensure we can extend the notion of ranges in the base class library
-  in the future (for example, by supporting negative ranges or element types
-  other than `Int32`).
+  in the future (for example, by supporting element types other than `Int32`).
 
 ### Non-Goals
 
--- a/accepted/system-range/system-range.md
+++ b/accepted/system-range/system-range.md
@@ -1,0 +1,271 @@
+# System.Range
+
+**PM** [Immo landwerth](https://github.com/terrajobst)
+
+The C# team is considering providing language syntax for specifying ranges to
+represent parts of a string, array, or span. In order to make this work, the
+language has to be bound to a .NET type, which we call `System.Range`.
+
+## Scenarios and User Experience
+
+### Construcing ranges
+
+```csharp
+// Ranges are constructed with an index and a length
+
+var fiveToTen = new Range(index: 5, length: 6);
+
+// By using a factory method, they can also be constructed from two indices.
+
+var alsoFiveToTen = Range.Between(5, 11);
+
+// The upper bound is considered exclusive so this will print the 1, 2, 3
+
+foreach (var value in Range.Between(1, 4))
+    Console.WriteLine(value);
+
+```
+
+The C# syntax for ranges isn't finalized and is actively worked on. While this
+is part of a separate design group it makes sense to introduce the syntax here
+so that we can look at source code as it's likely written:
+
+```csharp
+var fiveToTen = 5..11; // Equivalent to Range.Between(5, 11)
+```
+
+### Open ended ranges
+
+```csharp
+// Ranges can be open ended:
+
+var fiveToEnd = 5..;  // Equivalent to Range.From(5) i.e. missing upper bound
+var startToTen = ..1; // Equivalent to Range.To(11) i.e. missing lower bound
+var everything = ..;  // Equivalent to Range.All i.e. missing upper and lower bound
+
+// Asking for Index, Length or enumerating an open range will throw
+// NotSupportedException. The consumer of the range is expected to 
+// close it over the particular data:
+
+void Print(int[] data, Range range)
+{
+    if (range.IsOpen)
+        range = range.Close(0..data.Length);
+
+    foreach (var index in range)
+        Console.WriteLine(data[index]);
+}
+
+var data = new [] { 'a', 'b', 'c' };
+Print(data, 2..); // Prints c
+Print(data, ..2); // Prints a, b
+Print(data, ..);  // Prints a, b, c
+```
+
+### Using ranges in APIs
+
+Ranges provide a concise way to provide an `index` and a `length`. This allows
+for making common operations more compact and less confusing at the call side.
+For example, consider this version of the indexer for `String`:
+
+```csharp
+partial class String
+{
+    public string this[Range range] => Substring(range.Index, range.Length);
+}
+```
+
+This allows creating substrings simply by using the indexer:
+
+```csharp
+var helloWorld = "Hello, World!";
+var hello = helloWorld[..5];
+var world = helloWorld[7..12];
+```
+
+While this might appear to just be minor syntactic sugar it allows for creating
+powerful APIs, such as the indexer for `Tensor<T>` (which you can think of as
+multidimensional array where the dimensions aren't statically known).
+
+```csharp
+partial class Tensor<T>
+{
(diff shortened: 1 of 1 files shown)


terrajobst:
Marked as work-in-progress as this proposal is still under discussion. This PR is mostly meant to capture the current line of thinking.

g0t4:
Is it going to be confusing if the range syntax sometimes operates on start/end and other times operates on start/length? 

For example, 
`var fiveToTen = 5..10; // Equivalent to Range.Between(5, 10)` 
This would be [5,6,7,8,9,10]

`var world = helloWorld[5..10];`
This would be a substring of characters from 5 to 14? 
I could easily confuse this as characters 5 to 10. 

Obviously, this is less confusing: 
`var world = helloWorld[7..3];`

`helloWorld[..5]` 
If 5 is length, then this would be characters 0 to 4?
If 5 is index, then this would be characters 0 to 5?

I assume that indexer + range syntax is start/length, and non-indexer range syntax is start/end? If so at least that differentiates the two cases but still I wonder if it would be confusing.

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,170 @@
...
+language has to be bound to a .NET type, which we call `System.Range`.
+
+## Scenarios and User Experience
+
+### Constructing and enumerating ranges
+
+```csharp
+// Ranges are constructed with an index and a length
+
+var fiveToTen = new Range(index: 5, length: 5);
```
OliaG: If the length is 5, should it be fiveToNine?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,170 @@
...
+    public string this[Range range] => Substring(range.Index, range.Length);
+}
+```
+
+This allows creating substrings simply by using the indexer:
+
+```csharp
+var helloWorld = "Hello, World!";
+var hello = helloWorld[..5];
+var world = helloWorld[7..5];
```
OliaG: Should it be 5..7?

OliaG: Could you add as a comment what is expected result is for each line, it'll be easier to understand your thinking.

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,170 @@
...
+```csharp
+partial class Tenstor<T>
+{
+    public Tenstor<T> this[params Range[] ranges] { get; }
+}
+```
+
+The indexer takes a list of ranges that can be used to get slices of the tensor.
+Imagine you have a two dimensional tensor (in other words a table). Imagine it
+represents test data where each each row is a test case and the columns are
```
OliaG: Word "each" is repeated twice

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,170 @@
...
+This might look like this:
+
+```csharp
+Tensor<float> testDataTable = GetTestData();
+int columnCount = testDataTable.GetLength(1);
+int outputColumn = columnCount - 1;
+int lastDataColumn = outputColumn - 1;
+
+Tensor<float> inputTable = testDataTable[.., ..lastDataColumn];
+Tensor<float> outputTable = testDataTable[.., outputColumn..1];
```
OliaG: "outputColumn..1" - here you mean length? outputColumn:1

poke:
I would strongly suggest to make ranges have an exclusive upper bound. This will more likely match expectations programmers will have when looking at it, especially when they come from other languages.

For example, it makes a lot more sense to me that a `new Range(0, foo.length)` would produce all *valid* indexes to access items of `foo`. So it would produce indexes `[0, 1, 2, …, foo.length - 2, foo.length - 1]` making this have `foo.length` items. This also makes for the nice property that a `Range(0, n)` and a `Range(n, m)` would nicely combine into a `Range(0, m)`.

Having the upper bound non-exclusive would also fit use cases like this more easily:

```c#
foreach (var index in new Range(foo.length))
{
    DoSomething(foo[index]);
}
```

Compare this to a common for loop that people usually write:

```c#
for (var i = 0; i < foo.length; i++)
{
    DoSomething(foo[index]);
}
```

Furthermore, I would suggest the syntax to actually refer to indexes, not start and length. As you can see just from the comments so far, having `helloWorld[7..5]` is *really* confusing when it’s supposed to mean “start at index 7 and take 5 elements”. It is a lot clearer if this was `helloWorld[7..12]`, to mean “start at index 7 and take until index 12”. This will also open up the range object to some other functionalities, like the ability to *skip* elements: `foo[new Range(0, 10, step: 2)])` would yield indexes `[0, 2, 4, 6, 8]`. Or even the ability to have negative indexes in there to refer to indexes from the end: `foo[new Range(0, -2)]` would yield `[0, 1, 2, …, foo.length - 2, foo.length - 1]`.

People familiar with Python will recognize this properly since that’s exactly how both the built-in [`range`](https://docs.python.org/3/library/stdtypes.html#range) and [`slice`](https://docs.python.org/3/library/functions.html#slice) objects work there. There are a few Python related discussions on why the exclusive upper bound is a good idea: [Why are slice and range upper-bound exclusive?](https://stackoverflow.com/q/11364533/216074), [Why are Python ranges half-open (exclusive) instead of closed (inclusive)?](https://www.quora.com/Why-are-Python-ranges-half-open-exclusive-instead-of-closed-inclusive). Of course, there’s also the famous paper by Edsger W. Dijkstra on [“why numbering should start at zero” and why an exclusive upper bound “is to be preferred”](http://www.cs.utexas.edu/users/EWD/transcriptions/EWD08xx/EWD831.html).

---

Finally, I would also suggest the Range type to have some common interface that people could implement to make the indexing syntax `foo[rangeObject]` work for custom objects (I would be fine if that was just `IEnumerable<int>`). And I would also like for people to be able to override the item access with a range instead of an index, so people could implement their custom logic for `collection[range]` without having to rely on the compiler to make this individual `collection[start], collection[start+1], …` item accesses. One problem there though that I’m not sure how to handle the return type there: Is `str[range]` returning a substring, a `char[]`, or an `IEnumerable<char>`, or something else?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,170 @@
...
+        public static readonly Range All;
+        public static Range Between(int start, int end);
+        public static Range From(int start);
+        public static Range To(int end);
+        public Range(int index, int length);
+        public int Index { get; }
+        public int Length { get; }
+        public bool IsClosed { get; }
+        public Range Close(int index, int length);
+        public override string ToString();
```
MichalStrehovsky: Should this also override `GetHashCode` and `Equals`?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,170 @@
...
+  mean we shouldn't be doing it, but that it will be covered by a separate
+  design proposal).
+
+## Design
+
+### Proposed API
+
+```csharp
+namespace System {
+    public struct Range : IEnumerable, IEnumerable<int> {
```
MichalStrehovsky: Would `IEquatable<Range>` make sense too?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,170 @@
...
+        public static readonly Range Empty;
+        public static readonly Range All;
```
MichalStrehovsky: Static properties might be better. It's not like we need to make sure there's a single instance of this struct and as a property this would be nicely inlineable, with constant propagation likely kicking in and we would avoid type initializer checks as well.

olivier-spinelli:
@poke Couldn't agree more. 
Ranges should have an exclusive upper bound.
Dijkstra was definitely right!

sethjuarez:
There are a couple of things that I've been thinking about with `Range`.
- Under the provided `[start..end]` syntax if `start > end` should this represent the range enumerated in reverse: `start, start - 1, start -2, ..., end`?
- Additionally, _negative indices_ generally represent indexing but from the back of the `Enumerable` - if we support _negative indices_, how will these be represented?

poke:
@sethjuarez I wouldn’t expect range indexing (or “slicing”) to work on all enumerables. And I don’t think the current proposal even goes in that direction. Slicing only really makes sense for things that have a particular size and where indexing individual items is also possible. So you could slice an array or a string, but not just any enumerable.

Python makes that distinction too where you can slice *sequences* but not *iterables*. `ICollection` would be the equivalent type to sequences, and `IEnuerable` for iterables.

As for `[start..end]` with `start > end`, taking again Python as the model, this would result in an empty result. Slicing does not implicitly wrap around, neither does it reverse its direction (unless you would tell it to do so with a negative step).

accepted/system-range/system-range.md
terrajobst: Fixed

accepted/system-range/system-range.md
terrajobst: That's what the variable names are for ;-)

---
In reply to: [163118819](https://github.com/dotnet/designs/pull/30#discussion_r163118819) [](ancestors = 163118819)

accepted/system-range/system-range.md
terrajobst: Good point, will add.

---
In reply to: [163200458](https://github.com/dotnet/designs/pull/30#discussion_r163200458) [](ancestors = 163200458)

accepted/system-range/system-range.md
terrajobst: Fair enough. I don't have a preference and for primitive types I can go either way. It's also not like we'll ever need to run code there.

---
In reply to: [163201749](https://github.com/dotnet/designs/pull/30#discussion_r163201749) [](ancestors = 163201749)

accepted/system-range/system-range.md
terrajobst: Yep.

---
In reply to: [163200319](https://github.com/dotnet/designs/pull/30#discussion_r163200319) [](ancestors = 163200319)

terrajobst:
@g0t4 

>I assume that indexer + range syntax is start/length, and non-indexer range syntax is start/end? If so at least that differentiates the two cases but still I wonder if it would be confusing.

Not sure what you mean. The indexer is defined as taking a value of type `Range`. The `..` is C# syntax, the brackets are the C# syntax for indexers. There is no compound syntax of `[<lower> .. <upper>]`, if that's what you mean.

terrajobst:
> I would strongly suggest to make ranges have an exclusive upper bound. This will more likely match expectations programmers will have when looking at it, especially when they come from other languages.

We've discussed this recently in the C# language design meeting and concluded the same. The proposal now reflects that.

terrajobst:
@sethjuarez

> Under the provided [start..end] syntax if start > end should this represent the range enumerated in reverse: start, start - 1, start -2, ..., end?

Absolutely not. Same reason why no language will offer reverse for loops just by swapping upper and lower bound. Otherwise walking an empty list would yield a single value of `-1`.

> Additionally, negative indices generally represent indexing but from the back of the Enumerable - if we support negative indices, how will these be represented?

That's on my list to figure out. I'll talk to you about it :-)

poke:
@terrajobst 
> We've discussed this recently in the C# language design meeting and concluded the same. The proposal now reflects that.

Yay! :)

> Maybe that got lost in the proposal but `l..u` means `Range.Between(l, u)`, which constructs a range from two positions.

Okay, I did get confused there then. I was taking the original `world = helloWorld[7..5]` example and was thinking this was supposed to be `world = helloWorld[new Range(start: 7, length: 5)]`. Guess that was just a mistake after all then. If `l..u` means `Range.Between(l, u)` then that’s perfect with me.

> but currently most seem to think that negative values might be off the table to reduce complexity

That would be a real shame especially since it’s such a simple construct in Python. But I do get that this is a lot more complex here when `Range` is supposed to implement `IEnumerable<int>` to be able to iterate on all indexes. That instantly “destroys” negative indexes since you cannot iterate a range with negative indexes until you know what collection it is linked to. I guess that’s why Python differentiates between `range`, being basically an `IEnumerable<int>`, and `slice` which is the object that gets passed to objects when indexing it. Only the latter supports negative indexes, and the slice syntax is actually just allowed inside indexers.

But maybe having a second `Slice` type would be an acceptable solution here too? I might be thinking a bit too far ahead here, but what do you think about this:

```c#
interface ISlice
{
    public int Start { get; }
    public int End { get; }
    public ISlice Adjust(int length);
}

struct Slice : ISlice, IEquatable<ISlice> {
    …
    public ISlice Adjust(int length)
    {
        // based on https://github.com/python/cpython/blob/3.6/Objects/sliceobject.c#L239
        int start = Start;
        int end = End;
        if (start < 0)
        {
            start += length;
            if (start < 0)
                start = 0;
        }
        else if (start >= length)
        {
            start = length - 1;
        }

        if (end < 0)
        {
            end += length;
            if (end < 0)
                end = 0;
        }
        else if (end >= length)
        {
            end = length - 1;
        }

        return new Slice { Start = start, End = end };
    }
}

struct Range : ISlice, IEnumerable<int>, IEquatable<Range> {
    …
    public ISlice Adjust(int length) => Close(0, length);
}

// Example idea of an implementation:
// the indexer would take an `ISlice` and `Close` it first on the sequence length before using it
// then to actually slice the collection. That way, an `ISlice` that had 
public partial class String
{
    public string this[ISlice slice]
    {
        get
        {
            slice = slice.Adjust(this.length);
            return this.Substring(slice.Start, slice.End - slice.Start);
        }
    }
}
```

While this won’t make the syntax support negative indexes (considering that the `..` operator would still create a `Range` object), it would allow us to do `someString[new Slice(-4, -2)]`. Indexers could then take an `ISlice` and adjust it to fit the current collection size. For normal ranges, this would just close the range. For slices, this would adjust the negative indexes to be proper indexes fitting to the collection size.

And this would also provide that base interface that I was asking for, so people could come up with their own slice implementations.

> By default no type can be indexed unless the type provides an indexer that accepts `Range`, there is no compiler magic along the lines you describe.

Oh, yes, you’re right, I was thinking too complicated here. I was expecting this to be a compiler feature along the lines the `foreach` works for enumerating things. If this works as a simple indexer, then that’s good!

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,200 @@
...
+        public Span<T> this[Range range] { get; }
+    }
+}
+```
+
+## Q & A
+
+### Open Design Points
+
+* Right now, we cannot retreive `Index` or `Length` if the range is open on any
```
poke: *“retrieve”*

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,200 @@
...
+    - We could expose `Start: int?` and `End: int?` and then add a factory
+      method `Between(int? start, int? end)`
+
+* We should check whether we can handle negative numbers for `Length` so that we
+  can support addressing relative to the end. Check Python to see what they did
+  there.
+
+* Should `Range` support containment checks?
+    - Positions and/or ranges?
+    - Question is how to define semantics for open ended ranges.
```
poke: If we understand open ended ranges as not having an end on one side, then `Range.From(x).Contains(y)` for all `y >= x`, and `Range.To(x).Contains(y)` for all `y < x`. Similarly for ranges:
* `Range.From(x).Contains(Range.Between(y, z))` and `Range.From(x).Contains(Range.From(y))` iff `y >= x`
* `Range.To(x).Contains(Range.Between(y, z))` and `Range.From(x).Contains(Range.To(z))` iff `z < x`
* `Range.To(x).Contains(Range.From(y))` and `Range.From(x).Contains(Range.To(y))` would then never be true, since they might be overlapping but cannot be contained since they both go to infinity on different sides (at least assuming that a `Range.To()` does not implicitly stop at 0).

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,200 @@
...
+
+* Should `Range` support combining/intersecting open ended ranges?
+    - Be careful though. For open ended slices this might produce invalid
+      values. For instance, you might think combining `3..` and `..12` would
+      produce `3..12` but this would mean indexing into an array of size 5 would
+      fail, while `3..` would have worked.
+    - We could define the operation as `Close(Range)` though.
```
poke: I would personally always assume that we close the range over the collection first before attempting any range-based indexing. Just taking pagination as an example: `items[new Range(page * pageSize, length: pageSize)]`. I would totally expect this to return less than `pageSize` items on the last page when there aren’t enough items, instead of failing.

terrajobst:
@poke

> That would be a real shame especially since it’s such a simple construct in Python. But I do get that this is a lot more complex here when `Range` is supposed to implement `IEnumerable<int>` to be able to iterate on all indexes. That instantly “destroys” negative indexes since you cannot iterate a range with negative indexes until you know what collection it is linked to.

While this is true we already have complexity due to the fact that we plan on supporting open ended ranages (such as `4..`, which means from index `4` all the way to the end). Enumerating those ranges will result in an exception. The expectation is that consumers will have to close the range. In practice, I don't think this will be become a real problem as people will just apply to the range to the collection they want to enumerate and that receiver will take care of closing it. Take this example:

```
var myCollection = new List<int> { 1, 2, 3 };
foreach (var x in myCollection[2..])
   Console.WriteLine(x);
```

The indexer of `List<T>` will accept a range and take care of closing the range before the returning a value that repesents the slice of the list. How this is represented is still an open issue.

> I guess that’s why Python differentiates between range, being basically an IEnumerable<int>, and slice which is the object that gets passed to objects when indexing it. Only the latter supports negative indexes, and the slice syntax is actually just allowed inside indexers. But maybe having a second `Slice` type would be an acceptable solution here too?

I think your idea of `Slice` is the same as `Range` except with support for negative numbers. I'm currently exploring if and how we can support negative ranges with that type rather than creating a separate type for it.

> `interface ISlice`

What would be the point of this interface over the `Range` primitive? In othw words, who would implement it and how would consumers use it? Using an interface will force allocating an object which seems way to heavy for most scenarios.

accepted/system-range/system-range.md
terrajobst: Fixed

accepted/system-range/system-range.md
terrajobst: I think the heart of the question is whether we want to define it for partially closed ranges or whether we consistently fail with an exception. That latter might be easier to understand and debug. My knee-jerk reaction is that you shouldn't have partially closed ranges when you perform range checks.

accepted/system-range/system-range.md
terrajobst: > I would personally always assume that we close the range over the collection first before attempting any range-based indexing.

Sort of. I expect the indexer of array to take care of it. But the indexer will throw an argument out of range exception if the resulting range is invalid. And for `3..` it would close it to `3..5` and succeed while `3..12` would fail. Hence, thinking of intersecting is logically wrong.

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,226 @@
...
+var hello = helloWorld[..5];
+var world = helloWorld[7..12];
+```
+
+While this might appear to just be minor syntactic sugar it allows for creating
+powerful APIs, such as the indexer for `Tensor<T>` (which you can think of as
+multidimensional array where the dimensions aren't statically known).
+
+```csharp
+partial class Tenstor<T>
```
pharring: typo: `Tensor<T>` (and on line 93)

svick:
> Enumerating [open ended] ranges will result in an exception.

Why? I think it would be nice if the following code worked, because it reads much nicer than the equivalent `for` loop:

```c#
foreach (int x in 1..)
{
    …
    break;
    …
}
```

Though this leaves some questions open:

1. Question: What happens if iteration continues after `int.MaxValue`?

    Answer: An exception is thrown. You asked to iterate infinitely and `int` is not able to do that without overflowing, so you get an overflow exception.

2. Question: What would `foreach (long x in 1..)` do?

    Answer: If nothing special is done, that would behave exactly the same as the `int` version, which I think is not ideal. Maybe as an alternative (though fairly complicated) solution, `Range` could, in addition to implementing `IEnumerable<int>` implicitly, also implement `IEnumerable<long>` explicitly? That way, the `long` version of the `foreach` could iterate above `int.MaxValue`.

    Except that wouldn't actually work, because `foreach` is not target-typed. If you write `foreach (long x in e)` and `e` implements `IEnumerable<int>` implicitly and `IEnumerable<long>` explicitly, the compiler chooses `IEnumerable<int>`. And changing that would be a breaking change and also way outside the scope of this issue.

In the end, assuming making it throw leaves the door open for a future improvement, that's probably fine. If it does not leave the door open, then I think it should be considered now.

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,226 @@
...
+var everything = ..;  // Equivalent to Range.All i.e. missing upper and lower bound
+
+// Asking for Index, Length or enumerating an open range will throw
+// NotSupportedException. The consumer of the range is expected to 
+// close it over the particular data:
+
+void Print(int[] data, Range range)
+{
+    if (range.IsOpen)
+        range = range.Close(0..range.Length);
```
eerhardt: Personally I find this awkward.  What's the reasoning for not allowing users to query if Index or Length is set?

There is also a bug in your code.  From the comments `// Asking for Index, Length or enumerating an open range will throw`, however you are accessing `range.Length` inside of the `if (range.IsOpen)` check.

terrajobst:
@svick

> > Enumerating [open ended] ranges will result in an exception.
> Why? I think it would be nice if the following code worked

On top of the overflow issue you pointed out:

1. It's unlikely you want to enumerate all ints to `int.MaxValue`.
2. It's not defined for missing lower bounds

So I'd say an exception is the way to got because there is no sensible definition for the request.

> Question: What would `foreach (long x in 1..)` do?

There are currently no plans for a range with `long` values. However, we believe we can add it later if necessary. In that case the compiler would infer `LongRange` if either end of the range is of type `long`. We' also likely add an implicit case from `Range` to `LongRange` and an explicit cast from `LongRange` to `Range`. However, your code would simply extract ints from a `Range` which are converted to `long`. So in order to create an open ended `LongRange` you'd have to write:

```csharp
foreach (var x in (long)1..)
{
}
```

> Range could, in addition to implementing `IEnumerable<int>`` implicitly, also implement `IEnumerable<long>` explicitly? 

I don't see how this would be helpful as `Range` would still store the index and length as an `int`. In order to benefit from larger ranges, you'll have to use a range type that actually stores them as `long`, i.e. you'll need a `LongRange` type.

accepted/system-range/system-range.md
terrajobst: > There is also a bug in your code.

Oops, the code is meant to read as follows:

```csharp
void Print(int[] data, Range range)
{
    if (range.IsOpen)
        range = range.Close(0..data.Length);
```

> What's the reasoning for not allowing consumers to query if `Index` or `Length` is set? 

Originally that's what I did but then I got feedback that it makes the type more busy. How would you model it? The most natural representation would be `int?` but that makes writing checks more awkward. It also means that that callers would have to check for both ends when in practice I think you don't care which end is open, you just want to close it. Partially closing seems like a non-scenario to me (although you can if `Close` accepts a range).

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,230 @@
...
+
+### Proposed API
+
+```csharp
+namespace System {
+    public struct Range : IEnumerable, IEnumerable<int>, IEquatable<Range> {
+        public static Range Empty { get; }
+        public static Range All { get; }
+        public static Range Between(int start, int end);
+        public static Range From(int start);
```
KrzysztofCwalina: Should the parameter be called index?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,230 @@
...
+        public static Range To(int end);
+        public Range(int index, int length);
+        public int Index { get; }
+        public int Length { get; }
```
KrzysztofCwalina: Do we need End property?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,230 @@
...
+        public bool IsClosed { get; }
+        public Range Close(Range range);
```
KrzysztofCwalina: Do we want to call it Bind/IsBound?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,230 @@
...
+* Should `Range` support combining/intersecting open ended ranges?
+    - Be careful though. For open ended slices this might produce invalid
+      values. For instance, you might think combining `3..` and `..12` would
+      produce `3..12` but this would mean indexing into an array of size 5 would
+      fail, while `3..` would have worked.
+    - We could define the operation as `Close(Range)` though.
+
+* Should we add an interface for range access?
+    ```csharp
+    interface IRangeable<T>
```
KrzysztofCwalina: ISlicable

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,271 @@
...
+  mean we shouldn't be doing it, but that it will be covered by a separate
+  design proposal).
+
+## Design
+
+### Proposed API
+
+```csharp
+namespace System {
+    public struct Range : IEnumerable, IEnumerable<int>, IEquatable<Range> {
```
jaredpar: readonly

accepted/system-range/system-range.md
jaredpar: How can Close operate without further context? I expected to see an extra parameter here like `ICollection`, `IList`, etc ...

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,271 @@
...
+    }
+}
+```
+
+### Companion APIs
+
+```csharp
+namespace System {
+    public partial class String {
+        public string this[Range range] { get; }
```
jaredpar: Why does this return a string vs. a `Span<char>`?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,271 @@
...
+    }
+    public partial class T[] {
+        public T[] this[Range range] { get; }
```
jaredpar: Why does this return a `T[]` vs. a `Span<T>`?

accepted/system-range/system-range.md
KrzysztofCwalina: I think the method should take an int length

accepted/system-range/system-range.md
terrajobst: It's currently not because my prototype needed to write to the underlying fields in order to bypass the argument validation of the constructor: `Range(int index, int length)` rejects negative lengths, but unbound ranges need to set a negative number. I guess we could use unsafe code to set the fields directly but that seems weird..

accepted/system-range/system-range.md
terrajobst: Because the vast majority of our customers will not care for getting spans out of `string` but would really enjoy the shorthand for `Substring`. For the folks that want spans, they use `str.AsSpan(start:end)`.

accepted/system-range/system-range.md
```diff
@@ -32,25 +31,33 @@ so that we can look at source code as it's likely written:
 
 ```csharp
 var fiveToTen = 5..11; // Equivalent to Range.Between(5, 11)
+
+// You can also construct ranges with index and length:
+var fiveToTen = 5.:6; // Equivalent to new Range.Between(5, 6)
 ```
 
-### Open ended ranges
+### Unbounded ranges
```
KrzysztofCwalina: I think it should be "Unbound"

accepted/system-range/system-range.md
terrajobst: Fixed

accepted/system-range/system-range.md
terrajobst: Yep, added.

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,312 @@
...
+
+The C# syntax for ranges isn't finalized and is actively worked on. While this
+is part of a separate design group it makes sense to introduce the syntax here
+so that we can look at source code as it's likely written:
+
+```csharp
+var fiveToTen = 5..11; // Equivalent to Range.Between(5, 11)
+
+// You can also construct ranges with index and length:
+var fiveToTen = 5.:6; // Equivalent to new Range.Between(5, 6)
```
eerhardt: ‘new Range.Between’?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+### Unbounded ranges
+
+```csharp
+// Ranges can be unbounded:
+
+var fiveToEnd = 5..;  // Equivalent to Range.From(5) i.e. missing upper bound
+var startToTen = ..1; // Equivalent to Range.To(11) i.e. missing lower bound
```
eerhardt: Code says ‘1’ comment says ‘11’

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,312 @@
...
+represents test data where each row is a test case and the columns are
+representing the inputs with the last column containing the expected result. Now
+let's say you want to split this data into two tensors, one representing the
+inputs and one the outputs.
+
+This might look like this:
+
+```csharp
+Tensor<float> testDataTable = GetTestData();
+int columnCount = testDataTable.GetLength(1);
```
eerhardt: Is this line (columnCount) necessary?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,312 @@
...
+```
+
+## Requirements
+
+### Goals
+
+* We need a type that both, plays well with existing concepts in the base class
+  library, as well as the intended syntax for C#
+* We need to ensure we can extend the notion of ranges in the base class library
+  in the future (for example, by supporting negative ranges or element types
```
eerhardt: Should be updated since we are proposing negative ranges above.

accepted/system-range/system-range.md
terrajobst: No, `Between` is a static factory method, hence no `new`.

accepted/system-range/system-range.md
terrajobst: By design. It's the Spinal Tap fix.

accepted/system-range/system-range.md
terrajobst: Yes, it's testing whether you're paying attention! Deleted.

accepted/system-range/system-range.md
terrajobst: Good call

accepted/system-range/system-range.md
terrajobst: Ah, never mind. I'm an idiotic :-)

accepted/system-range/system-range.md
```diff
@@ -11,19 +11,18 @@ language has to be bound to a .NET type, which we call `System.Range`.
 ### Construcing ranges
 
 ```csharp
-// Ranges are constructed with an index and a length
+// Ranges are constructed with an index and a length:
 
 var fiveToTen = new Range(index: 5, length: 6);
```
jaredpar: Why are we using a ctor for length but a factory method for start / end? Instinct is to be consistent and have factory methods in both places.

accepted/system-range/system-range.md
```diff
@@ -11,19 +11,18 @@ language has to be bound to a .NET type, which we call `System.Range`.
...
 
-// By using a factory method, they can also be constructed from two indices.
+// By using a factory method, they can also be constructed from two indices:
 
 var alsoFiveToTen = Range.Between(5, 11);
 
-// The upper bound is considered exclusive so this will print the 1, 2, 3
+// The upper bound is considered exclusive so this will print 1, 2, 3:
```
jaredpar: Consider removing considered. 

Both funny and serious.

accepted/system-range/system-range.md
jaredpar: The `.:` syntax is equivalent to calling the .ctor, not the Between method.

accepted/system-range/system-range.md
```diff
@@ -32,25 +31,33 @@ so that we can look at source code as it's likely written:
...
 
 var fiveToEnd = 5..;  // Equivalent to Range.From(5) i.e. missing upper bound
 var startToTen = ..1; // Equivalent to Range.To(11) i.e. missing lower bound
 var everything = ..;  // Equivalent to Range.All i.e. missing upper and lower bound
 
-// Asking for Index, Length or enumerating an open range will throw
+// When the lower bound is omitted, it's interpreted to be zero.
+//
+// When the upper bound is omitted, it's interpreted to be the length of the
+// receiving collection.
```
jaredpar: I think we should specify this is only true when methods, as an implementation detail, call the Bind method. It's possible for methods to interpret unbounded ends any way they choose. Although most will simply use Bind.

accepted/system-range/system-range.md
```diff
@@ -32,25 +31,33 @@ so that we can look at source code as it's likely written:
...
+//
+// Asking for Length or enumerating an unbound range will throw
```
jaredpar: Unbounded on end or unbounded on either? The comment earlier says omitting the lower bound is interpreted as 0. Hence it should follow that `..3` is a range equivalent to `0..3`

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,310 @@
...
+      `ListRange<T>`.
+        ```csharp
+        public struct ListRange<T> : IList<T>
+        {
+            public ListRange(IList<T> list, Range range);
+            public IList<T> List { get; }
+            public Range Range { get; }
+        }
+        ```
+* Should we suppport steps in ranges? Syntax is unclear and so is whether this
```
jnm2: 'suppport' is typo

accepted/system-range/system-range.md
justinvp: > I guess we could use unsafe code to set the fields directly but that seems weird...

Use a private .ctor in that case?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,310 @@
...
+  mean we shouldn't be doing it, but that it will be covered by a separate
+  design proposal).
+
+## Design
+
+### Proposed API
+
+```csharp
+namespace System {
+    public struct Range : IEnumerable, IEnumerable<int>, IEquatable<Range> {
```
justinvp: Is `IEnumerable<int>` implemented explicitly? Will there be a public `GetEnumerator()` that returns a struct enumerator?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,310 @@
...
+        public static Range Empty { get; }
+        public static Range All { get; }
+        public static Range Between(int startIndex, int endIndex);
```
justinvp: Why does `Range` use `startIndex` but `Span<T>` uses `start`?

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,310 @@
...
+        public Range(int index, int length);
+        public int Index { get; }
+        public int EndIndex { get; }
+        public int Length { get; }
+        public bool IsUnbounded { get; }
+        public Range Bind(int length);
+        public override string ToString();
+        public override int GetHashCode();
+        public bool Equals(Range other);
+        public override bool Equals(object other);
```
justinvp: Nit: The parameter name for `override Equals(object)` should be `obj` to match the parameter name on [`object.Equals`](https://github.com/dotnet/corefx/blob/90027686dc6ee078ae190bfa471962efeab7a123/src/System.Runtime/ref/System.Runtime.cs#L1871). (Unlike `Equals(Range)` where it should be `other`).

terrajobst:
Superseded by #48

PR #31, opened 2018-01-24
PR feedback on the "verifiable ref returns" proposal

VSadov:


VSadov:
@MichalStrehovsky - please take a look.

After thinking a bit about whether IsReturnable should be an attribute, I am getting convinced that it should be just a regular class like IsConst. Its only foreseeable target is IL local slots and that is not even in AttributeTargets enum. - Making this type an attribute seems unnecessary and perhaps confusing.

VSadov:
Thanks!!

PR #32, opened 2018-03-07
Add debugging specs

tmat:


tmat:
@richlander

akoeplinger:
@tmat this seems to add the CI environment variables file from https://github.com/dotnet/designs/pull/19, was that intentional?

tmat:
@akoeplinger Yes. We moved it to the accepted specs. I'll close #19.

PR #35, opened 2018-03-23
Adds STANDARD_CI_REPOSITORY_TYPE

tmat:
This is the name of the SCM the repo is using.

Although there might be better names than `STANDARD_CI_REPOSITORY_TYPE` (like `STANDARD_CI_SCM`) I chose it for pragmatic reasons - the term "repository type" is already used by NuGet in nuspec metadata. We don't strictly need to be consistent but for the sake of less complexity it's better to be.

### DIFF
--- a/accepted/build/standard-ci-env-variables.md
+++ b/accepted/build/standard-ci-env-variables.md
@@ -14,6 +14,7 @@ Source control information needed:
 
 * Commit ID/hash
 * Repository URL
+* Source Control Manager name (`git`, `TFVC`, `mercurial`, `svn`, etc.)
 
 Places to store this information:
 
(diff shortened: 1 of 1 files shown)


tmat:
@richlander

tmat:
@akoeplinger @loic-sharma @tmds @dtretyakov @normj @hypersw @FeodorFitsner Adding one more variable that would be useful to standardize on. Let me know what you think.

FeodorFitsner:
What are allowed values for `STANDARD_CI_REPOSITORY_TYPE` variable?

tmat:
@FeodorFitsner Updated the PR to include well-known values

PR #37, opened 2018-04-10
Update the Source Link spec for VS15.7 auth support

chuckries:
This updates the spec to provide an example of generating VSTS source
link. It also adds that VS 2017 15.7 supports authentication.


chuckries:
@tmat

PR #38, opened 2018-04-11
Fixed typos in source-link.md

jnm2:


accepted/diagnostics/source-link.md
```diff
@@ -103,7 +103,7 @@ Using this URL scheme, it is possible to generate Source Link for VSTS git repos
 {
     "$schema": "http://json-schema.org/draft-04/schema#",
     "title": "SourceLink",
-    "description": "A mapping of source file paths to URL's",
+    "description": "A mapping of source file paths to URLs",
```
jnm2: (Apostrophes are only used to pluralize acronyms when each letter of the acronym is followed by a dot. For acronyms without dots, `'s` is used for possessive.)

dfederm:
While you're here, I also noticed that in both the GitHub and VSTS examples the `WriteLinesToFile` uses `File="$(IntermediateOutputPath)source_link.json"` when it could just use the property it defined earlier, `File="$(SourceLink)"`

jnm2:
@dfederm No problem! Done.

PR #39, opened 2018-04-25
STANDARD_CI_REPOSITORY_URL_MAP proposal

tmat:


accepted/build/standard-ci-env-variables.md
```diff
@@ -47,8 +47,25 @@ It is important that these environment variables do not conflict with other vari
 
 * **STANDARD\_CI\_SOURCE\_REVISION\_ID** -- Commit hash / ID; Example: 2ba93796dcf132de447886d4d634414ee8cb069d
 * **STANDARD\_CI\_REPOSITORY\_URL** -- URL for repository; Example: https://github.com/dotnet/corefx
+* **STANDARD\_CI\_REPOSITORY\_URL\_MAP{n}** -- optional mapping of URLs 
 * **STANDARD\_CI\_REPOSITORY\_TYPE** -- Source control manager name; Example: `git`, `TFVC`, `mercurial`, `svn`
 
+### STANDARD_CI_REPOSITORY_URL
+
+The URL to be used to refer to the repository from artifacts produced by the build. If the source control provider supports permanent URLs (i.e. URLs resilient to renames) this should be such permanent URL.
```
akoeplinger: I'm confused by this, is `https://github.com/dotnet/corefx` still preferred?
Or should it be `https://raw.githubusercontent.com/dotnet/corefx/master/` (doesn't make sense here imho).

If it's the former then I'm not sure why we need this, renames of a repository should be handled by the hosting provider with proper redirects.

accepted/build/standard-ci-env-variables.md
```diff
@@ -47,8 +47,25 @@ It is important that these environment variables do not conflict with other vari
...
 * **STANDARD\_CI\_SOURCE\_REVISION\_ID** -- Commit hash / ID; Example: 2ba93796dcf132de447886d4d634414ee8cb069d
 * **STANDARD\_CI\_REPOSITORY\_URL** -- URL for repository; Example: https://github.com/dotnet/corefx
+* **STANDARD\_CI\_REPOSITORY\_URL\_MAP{n}** -- optional mapping of URLs 
 * **STANDARD\_CI\_REPOSITORY\_TYPE** -- Source control manager name; Example: `git`, `TFVC`, `mercurial`, `svn`
 
+### STANDARD_CI_REPOSITORY_URL
+
+The URL to be used to refer to the repository from artifacts produced by the build. If the source control provider supports permanent URLs (i.e. URLs resilient to renames) this should be such permanent URL. 
+
+### STANDARD_CI_REPOSITORY_URL_MAP{n}
```
akoeplinger: I feel like this shouldn't be the responsibility of the CI service. The CI service in many cases doesn't host the repository so it can't know the mappings to permanent URLs, short of hardcoding popular hosting services.

The CI service would also need to parse/know about the submodules and they could in turn be stored on different hosting services.

I think it'd make more sense if this value would be set by the _user_ who controls the repo.

akoeplinger: This also needs examples of how the mapping should look like, there are many ways to refer to a repository in submodules:

- `https://github.com/dotnet/corefx`
- `git://github.com/dotnet/corefx`
- `git://github.com/dotnet/corefx.git`
- `git@github.com:dotnet/corefx`
- ...

I guess you'd only support http(s) ?

And the mapping would look like `STANDARD_CI_REPOSITORY_URL_MAP=https://github.com/dotnet/corefx;https://raw.githubusercontent.com/dotnet/corefx/master/` ?

accepted/build/standard-ci-env-variables.md
tmat: It's the former. Redirects work to some extent but not completely. If you rename repo A to B and then repo C to A. Some hosts (VSTS) could handle it by using GUIDs. Something like `https://github.com/{organization-guid}/{repository-guid}`.

accepted/build/standard-ci-env-variables.md
tmat: Thanks for feedback. These are good points. One thing that we were considering was a custom build step that is invoked by the CI that could set this environment variable. But yeah, I agree it's not ideal. Still trying to figure out a better approach.

Re multiple ways to refer to a repo in submodule - are there actually more cases than the four you list above?

PR #42, opened 2018-06-19
Add Link-time framework feature removal spec

MichalStrehovsky:
A sample of how this would work is here: https://github.com/MichalStrehovsky/corefx/commit/2a7fa763d177d2a075fa40272ad113c0da4eaef6

If we had a switch like the one in the linked sample usage commit, the People app that ships with Windows would get 1.2 MB smaller on AMD64 with post-NetStandard2.0 .NET Native because we would no longer need to carry the full HTTP/HTTPS/FTP stack just because the People app uses `DataContractSerializer`.

Other samples:
* Making the [reflection fallback](https://github.com/MichalStrehovsky/corefx/commit/372c8d7b1dfb7586be023049bec6f243b4ea31a1) in DataContractSerializers optional saves 335 kB.
* Making the [reflection fallback](https://github.com/MichalStrehovsky/corefx/commit/eb76499e287fa45f4054bfe214682f6142232b58) in XmlSerializers optional saves 410 kB.

### DIFF
--- a/proposed/link-time-feature-removal.md
+++ b/proposed/link-time-feature-removal.md
@@ -1,0 +1,93 @@
+# Link-time framework feature removal
+
+[Self-contained deployment](https://docs.microsoft.com/en-us/dotnet/core/deploying/) of .NET Core applications (a deployment that includes both the runtime/framework, and app code) provides a unique challenge and opportunity to tailor the deployed bits to the needs of the application. A whole program analysis step that runs as part of publishing can identify parts of the application and libraries (framework or otherwise) that are not used – and remove them from the deployment package. This has a positive impact on the size of the deployment and saves considerable amounts of storage space and bandwidth for the end user.
+
+[IL Linker](https://github.com/dotnet/announcements/issues/30) (for CoreCLR-based .NET Core applications) and Dependency Reducer (for .NET Native-based .NET Core applications) have demonstrated that performing static reachability analysis of methods and types can lead to significant savings. Static reachability analysis assumes that there are certain roots in the application (places that are always reachable, such as the Main method of the app); the analysis then builds a transitive closure of all the methods and types reachable from roots. Methods and types that are not reachable from roots are considered unused and can be discarded.
+
+## Beyond static reachability analysis
+
+Static reachability analysis has its limits though: the fact that something is statically reachable doesn’t mean it’s going to be used at runtime. While sometimes it’s possible to change the library code in a way that enables removal of unused code through static reachability analysis alone ([example](https://github.com/dotnet/corefx/pull/23867)), this doesn’t always work, and often leads to clunky and fragile code.
+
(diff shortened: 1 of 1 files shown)


MichalStrehovsky:
Cc @jkotas @marek-safar @morganbr @davidwrighton @stephentoub

sergiy-k:
Thanks @MichalStrehovsky! This looks great. The only question I have is what everyone thinks about the names of attributes? After reviewing multiple samples, it seems to me that the 'Removable' name is quite ambiguous and it does not speak clearly for its purpose. Maybe something more concrete like RemovableFeatureRoot attribute or just FeatureRoot attribute would be better? I'm not super good at coming up with good names so please feel to ignore. Thank you.

MichalStrehovsky:
> The only question I have is what everyone thinks about the names of attributes?

I could file an issue over in the CoreFX repo and run the attribute through the API review process. They tend to also comment on names. It's not critical, but maybe having something approved saves us from having to process two attributes (unofficial and official one) in the future.

marek-safar:
@MichalStrehovsky please run the names via API review. We have more attributes we want to add/import and it might even make a sense to have them in a new namespace.

MichalStrehovsky:
Filed dotnet/corefx#30643.

terrajobst:
My primary concern with the feature as proposed is that it's hard to diagnose issues introduced by incorrect attribution if the default behavior is returning zero/null.

I'd rather we FailFast by default. Looking at [your example](https://github.com/MichalStrehovsky/corefx/commit/eb76499e287fa45f4054bfe214682f6142232b58) I do however buy the scenario that you want to define query methods for features, which is as easy as doing this:

```csharp
private static bool SupportsReflectionBasedSerializationFeature
{
    [Removable(ReflectionBasedSerializationFeatureName)]
    get
    {
        return true;
    }
}
```

My gut feel is -- and that might wrong -- that the vast majority of attribute applications will be implementations of regular methods. So maybe we should be using two different attributes or just change the attribute to this:

```csharp
private static bool SupportsReflectionBasedSerializationFeature
{
    [Removable(ReflectionBasedSerializationFeatureName, DoNotFail=true)]
    get
    {
        return true;
    }
}
```

terrajobst:
Do we provide an accordance where the developer can manually specify the list of features the linker has to keep regardless of reachability? This would allow developers to work around problems with the linker in case a method fail fast by just keeping the feature entirely.

MichalStrehovsky:
> Do we provide an accordance where the developer can manually specify the list of features the linker has to keep regardless of reachability? This would allow developers to work around problems with the linker in case a method fails fast by just keeping the feature entirely.

Feature removal is opt in - the developer has to specify the name of the feature to remove and only then the body replacement happens. If we want the linker not to do that, we would simply not ask it to strip the feature.

Or are you asking for this attribute to serve a double purpose (acting like a conditional `[Preserve]` attribute?). Do you have a scenario where this would be useful? The motivation behind the proposed feature is to allow treeshaking in places that are *reachable* in the static call graph (the fact that we're keeping the methods is the problem; the problem is not their unintended removal) - I don't see a scenario where this would be beneficial.

PR #44, opened 2018-06-23
lowercase tfvc to match rest of spec

ctaggart:


terrajobst:
If we go with matching the command line, it should be `tf`.

tmat:
@terrajobst This is not command line tool name, this is name of the source control provider. TF is not a source control manager.

PR #45, opened 2018-06-23
hg for mercurial

ctaggart:
`hg` feels a lot more consistent with the other repository types. It matches the command line names used.


ctaggart:
Or may be spell out subversion. I don’t know. Just didn’t feel consistent

tmat:
I don't think the goal here is to match the command line tool name. 
How would you name the corresponding SoruceLink package? Microsoft.SourceLink.Hg?

tmat:
I think we need more data on how people usually refer to Mercurial. Like a list of tools for Mercurial and how they are named. The same for Subversion. We can decide based on that.

ctaggart:
I’d spell out both Mercurial and Subversion if you are going for how people usually refer to them and want to match package names

PR #48, opened 2018-12-03
Add proposal for System.Index and System.Range

terrajobst:


accepted/system-range/system-range.md
```diff
@@ -0,0 +1,246 @@
...
+SZ-arrays (i.e. `T[]` in C#) do not have an API that represents the indexer.
+Instead, they are specialized as IL opcodes to load and store elements in such
+an array. The indexer is part of the language semantics and thus doesn't have an
+API representation.
+
+For indexing using `Index` the compiler doesn't have to do much, thus the
+handling is done directly by the compiler, i.e. this code:
+
+```C#
+T[] array = GetArray(;)
```
poke: `GetArray();` – same below in line 148

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,246 @@
...
+var fiveToTen = 5..11; // Equivalent to Range.Create(5, 11)
+```
+
+### Unbounded ranges
+
+```C#
+// Ranges can be unbounded:
+
+var fiveToEnd = 5..;  // Equivalent to Range.FromStart(5) i.e. missing upper bound
+var startToTen = ..1; // Equivalent to Range.ToEnd(1) i.e. missing lower bound
```
poke: This range should be `..11` and `Range.ToEnd(11)` going by the variable name.

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,246 @@
...
+    public sealed class RuntimeHelpers
+    {
+        public static T[] GetArrayRange<T>(T[] array, Range range);
+    }
+}
+```
+
+### Companion APIs
+
+To decide which indexers we offer we followed the following principles:
```
poke: nit: Comma before *“we followed”*

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,246 @@
...
+
+* Scalar indexers should return the element type
+* Range indexers should return the type they are on
+
+If developers whish to optimize for allocations they can convert the type to an
```
poke: nit: Comma before *“they can convert”*

poke:
I kind of missed the discussion where the syntax for from-the-end indexes was changed from negative indexes to `^` indexes, so I don’t know what has been said there. I do have a bit of a problem with the wording though.

An index *“from end”* suggests to me that I basically index the collection like normal but starting at the end. Since indexing starts at zero, `^0` should be the first element from the end, i.e. the last element. I totally get that `^1` should be `length - 1`, just like `-1` in the old spec does (I actually still prefer that but I guess that has already been decided now), but I think the wording should be adjusted in some way.

DustinCampbell:
@poke: The C# LDM decided on `^` rather than `-` because it would be significant breaking change to use negative numbers. The problem is that `a[-1]` compiles and means something today -- it throws an exception, which the application might catch. It would be a major breaking change to stop throwing an exception and return a result that the application isn't expecting.

FWIW, this is the design that @MadsTorgersen and I showed demoed at Build back in May. The discussion thread for this is on the C# language design repo here: https://github.com/dotnet/csharplang/issues/185. Here's a pointer to my comment describing the design decision: https://github.com/dotnet/csharplang/issues/185#issuecomment-389419970.

poke:
@DustinCampbell Ah, thanks a lot for the explanation and the pointers. That does make sense! :) I’ll read that other discussion later to catch up.

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,246 @@
...
+instance of `Span<T>`, `ReadOnlySpan<T>`, `Memory<T>` or `ReadOnlyMemory<T>` by
+calling the appropriate API (e.g. `AsSpan()` or `AsMemory()`).
+
+We'll also provide overloads to common operations that take indexes and ranges,
+such as `String.Substring(int)`, `String.Substring(int, int)` and
+`AsSpan(int, int)`.
```
bartonjs: Should the "such as" here be the existing overloads, or the new ones?  (I expected to see `String.Substring(Index)`, `String.Substring(Range)`, and `String.AsSpan(Range)`)

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,246 @@
...
+
+```C#
+namespace System
+{
+    public partial class String
+    {
+        public char this[Index index] { get; }
+        public String this[Range range] { get; }
+        public String Substring(Index index);
```
bartonjs: `Substring(Index startIndex)` (same with Slice).

accepted/system-range/system-range.md
terrajobst: It's a new language feature. It makes sure that the calling method doesn't allocate. In this case though, the method has to allocate so I changed it back to `GetArray();`.

DustinCampbell: :trollface:

accepted/system-range/system-range.md
```diff
@@ -0,0 +1,246 @@
...
+```
+
+We propose the following method:
+
+```C#
+namespace System.Runtime.CompilerServices
+{
+    public sealed class RuntimeHelpers
+    {
+        public static T[] GetArrayRange<T>(T[] array, Range range);
```
eerhardt: Does this return a _new_ array with the data _copied_ out of `array`? If so, that seems wrong.

Instead, I would assume this should return `Span<T>`.

stephentoub: > Does this return a new array with the data copied out of array? If so, that seems wrong.

If it's being used by the compiler to generate the code for:
```
T[] elements = array[range];
```
as above, then returning a span doesn't help.

stephentoub: ps This is based on the idea that a) we should have range-based indexers on array and string, and b) that all range-based indexers should return the same type as the one they're on.  I argued that indexers should be non-allocating/non-copying, and so if we want to stick with (b), then we shouldn't have these indexers on array and string.  I lost.

eerhardt: It's frustrating that we spend so much effort on non-allocating scenarios, only to bring them back front and center in new scenarios.

stephentoub: Yes, yes it is.  I'll assume your thumbs down on my comment isn't about my opinion but about the fact that I lost.

eerhardt: In my mind, `string` is a different scenario than `array` and they shouldn't be grouped together.  (strings have semantics - like equality - that arrays don't, so you can't say "string" is just an array of characters. If it was, we would have just used `char[]`. Also, arrays are mutable and strings aren't.)

With string, you can call `Substring`, and that is well known to be allocating a new string. Also, why do we have _both_ the indexer `[Range]` and `Substring` on `string`? They do the same thing right? Why would I choose one over the other?

So instead, `string[Range]` could return `ReadOnlyMemory|Span<char>` and `string.Substring(Range)` return `string`. The difference is clear.

Then `array[Range]` could return `Span<T>` (or `ArraySegment<T>`).

Honestly, I don't see a need for the `array.Subarray(Range)` scenario. If you really want that - allocate a new array, and call `Array.Copy`.

> about the fact that I lost.

Correct.

jkotas: >  I argued that indexers should be non-allocating/non-copying

:+1:

The allocating array indexers look like a  major performance trap. I do not think we should have them.

stephentoub: > The allocating array indexers look like a major performance trap.

Yes, especially since if you write code like:
```C#
string s = "hello, world";
ReadOnlySpan<char> span = s[1..3];
```
it's far from obvious that allocates and is different from using AsSpan(…).

terrajobst: This was discussed in the C# LDM last week. I presented my opinion which is as follows:

1. Arrays and strings are vastly more useful to the majority of our customers than span-based APIs.
2. The rules for the range indexer are much simpler if all types follow the same pattern, namely that they return the same type as the container.
3. Performance, while certainly a major focus for us, isn't the most important goal for the majority of code that is written. C# has plenty of features that implicitly allocate. While we often see those as problematic for performance critical code, they are also a major reason why C# is loved as a productive language.

It seemed the overwhelming majority of people in the room shared that opinion. That being said, I don't believe the issue that Stephen pointed out was considered at the time. But this strikes me as yet another instance of an analyzer that we should offer. The offered fix should be `ReadOnlySpan<char> span = s.AsSpan(1..3);`

terrajobst: The other argument that Mads brought is that we should think of indexers with ranges as a different thing, if we believe this deserves compact syntax. In the end, one of the biggest customers of range will be `Tensor<T>`. Slicing that guy won't be an O(1) allocation-free operation either.

If we truly believe that index-syntax must mean allocation-free and O(1) access, just removing the range-based indexer from array alone isn't the right fix, we'd also need to remove it from string. However, I suspect that this means the compacter syntax benefits much fewer users. The difference between `s.Substring(start, length)` and `s.Substring(start..end)` isn't nearly as significant as `s[start..end]`.

jkotas: > C# has plenty of features that implicitly allocate.

Yes, that's correct. However, I believe that it is important that the rought performance characteristics are predictable by looking at the code without thinking about it too hard. It is part of the balance that makes C# good productive language.

For example, if you see Linq, you know that you are on the slow plan. If you see `async`, you know that you are paying some overhead, but it is not terrible. The allocating/copying indexers fail on this.

eerhardt: > In the end, one of the biggest customers of range will be Tensor<T>. Slicing that guy won't be an O(1) allocation-free operation either.

I implemented `Tensor<T>.Slice` to not copy data. You can just change the strides you are using, and now you can project into the same data without making a copy.

See https://github.com/dotnet/corefxlab/blob/38fb993f61467f24fb845323f25e807c4a618827/src/System.Numerics.Tensors/System/Numerics/Tensors/Tensor.cs#L1272

[numpy does the same thing](https://scipy-cookbook.readthedocs.io/items/ViewsVsCopies.html).

From the numpy link above:

> Technically, that means that the data of both objects is shared. You can create views by selecting a slice of the original array

So it is definitely not the case that slicing a Tensor will copy the data out.
__________

> The rules for the range indexer are much simpler if all types follow the same pattern, namely that they return the same type as the container.

That does not hold for all cases. It definitely doesn't hold for the `DenseTensor` case, where you project into the same data.

Imagine you are processing a video (or 3D image) using Tensor, and you wanted to slice out a single image in order to do some processing on it. Forcing a copy of the data for every frame in the video is not going to produce great performance.

terrajobst: > I implemented `Tensor<T>.Slice` to not copy data. You can just change the strides you are using, and now you can project into the same data without making a copy.

I didn't say copy, I said not cheap (compared to returning span or an element). Returning views that can mutate the original data have their own problems and can surprise people too. But that's a different problem we should discuss when reviewing tensors.

> > The rules for the range indexer are much simpler if all types follow the same pattern, namely that they return the same type as the container.
> 
> That does not hold for all cases. It definitely doesn't hold for the `DenseTensor` case, where you project into the same data.

That depends on how it's exposed. Looking at the thing you linked, it looks like the `Tensor<T>` class has a a virtual `Slice` method that is typed to return a `Tensor<T>`. I'd say that's close enough to what I said.

eerhardt: > I didn't say copy, I said not cheap 

You said `won't be an O(1) allocation-free operation`. The current implementation of `Tensor<T>.Slice` is `O(1)`, it is not allocation-free however because Tensor is a class.

> That depends on how it's exposed. Looking at the thing you linked, it looks like the Tensor<T> class has a a virtual Slice method that is typed to return a Tensor<T>. I'd say that's close enough to what I said

When a user explicitly has a `DenseTensor<float>` and they slice it, they won't be getting back a `DenseTensor<float>`. Instead they will get a view into the current `DenseTensor`, which is the base type `Tensor<float>`. So here it isn't the same type as the container.

terrajobst: Alright, for now I'm going to resolve this as the current spec lists the POR as agreed between FXDC/LDM. We can re-evaluate this based on customer feedback in Preview 2.

jkotas: Would it make more sense to not add questionable indexers first, and only add them later if the customer feedback tells us that they are really needed?

PR #51, opened 2019-02-02
Runtime binding proposal

richlander:
Proposal for .NET Core 3.0

Related content:

* [Roll Forward On No Candidate Fx](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md) doc defines existing behavior. 
* [Roll forward improvements for COM -- dotnet/core-setup #5062](https://github.com/dotnet/core-setup/issues/5062) specifies a proposal (largely) specific to COM. 
* [--fx-version argument to dotnet exec should apply roll-forward policy -- dotnet/core-setup #4146](https://github.com/dotnet/core-setup/issues/4146) specifies a proposal to enable roll-forward for `--fx-version`.


proposed/runtime-binding.md
```diff
@@ -0,0 +1,265 @@
...
+
+### RollForward
+
+We will introduce a new knob, called `RollForward`, with the following values:
+
+* `patch` -- Roll forward to the highest patch verison. This disables minor version roll forward.
+* `minor` -- Roll forward to the lowest higher minor verison, if requested minor version is missing.
+* `major` -- Roll forward to lowest higher major version, and lowest minor version, if requested major version is missing
+* `UseLatestMinor` -- Roll forward to latest minor version, even if requested minor version is on machine. Intended for COM hosting scenario.
+* `UseLatestMajor` -- Roll forward to latest major and highest minor version, even if requested major is on machine. Intended for COM hosting scenario.
```
nguerrera: Nit 1: I think the values should all have the same canonical casing, so Patch, Minor, Major

Nit 2: I don't this Use is adding much value, how about LatestMinor, LatestMajor

proposed/runtime-binding.md
```diff
@@ -0,0 +1,265 @@
...
+
+[Framework-dependent](https://docs.microsoft.com/en-us/dotnet/core/deploying/#framework-dependent-deployments-fdd)  applications require a .NET Core host to find a compatible runtime from a central location. The behavior of runtime binding is critical for both application compatibility and runtime deployment convenience.
+
+This document proposes a model that works for both apps and managed COM components/hosting.
+
+Related content:
+
+* [Roll Forward On No Candidate Fx](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md) doc defines existing behavior. 
+* [Roll forward improvements for COM -- dotnet/core-setup #5062](https://github.com/dotnet/core-setup/issues/5062) specifies a proposal specific to COM. 
+* [--fx-version argument to dotnet exec should apply roll-forward policy -- dotnet/core-setup #4146](https://github.com/dotnet/core-setup/issues/4146) specifies a proposal to enable roll-forward for `--fx-version`
```
nguerrera: That is not what dotnet/core-setup#4146 was meant to propose. The issue there is about when a specific fx-version is forced via fx-version, that version should behave the same as if the app had gotten on to it by roll forward.

nguerrera: I do like --fx-version composing with --roll-forward in this proposal, it's just orthogonal to that issue. I commented on the issue and changed it's title to be clearer about this.

nguerrera:
LGTM. I really like the single option, and the elimination of magic [0-2] numeric values. I also never like the "onNoCandidateFx" name.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,265 @@
...
+The example above will attempt to load the highest `3.0` patch version, which may be `3.0.0` or `3.0.27`, for example, whichever is the highest path version on the machine.
+
+### Configuration Precedence
+
+The host will consult these settings in order (later scopes take precendence over earlier ones):
+
+* `.runtimeconfig.json` properties (AKA "json")
+* CLI arguments
+* Environment variables (AKA "ENV")
+
```
KathleenDollard: Earlier a project (.csproj/.vbproj) setting was proposed. It's not mentioned here in precedence. Is that because it rolls into the runtimeconfig? Cool, but we should mention it I think.

richlander: >  Is that because it rolls into the runtimeconfig? 

Yes. Project values have no meaning at runtime. I will add a note for that.

proposed/runtime-binding.md
richlander: Good feedback. Let me share some of my thinking and see what you say. I original went with exactly what you have (ignoring the casing topic) `Minor` and `LatestMinor`. The problem is that the two settings are not symmetrical:

* `Minor` -- Do something only as a fallback case if a preferred runtime version is not found.
* `LatestMinor` -- Do something always.

That's why I added `Use` to `LatestMinor` to make it more obvious that this setting is more of a command than a request/hint.

I also thought of splitting these two forms of configuration into two different settings. That seemed like the worst idea, both because it would be confusing and there is no scenario where both settings would be set.

Thoughts?

vitek-karas: Maybe:
`AlwaysLatestMinor`
`AlwaysLatestMajor`

proposed/runtime-binding.md
richlander: I re-read that issue. I now see that it is unrelated.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,265 @@
...
+We will introduce a new knob, called `RollForward`, with the following values:
+
+* `patch` -- Roll forward to the highest patch verison. This disables minor version roll forward.
+* `minor` -- Roll forward to the lowest higher minor verison, if requested minor version is missing.
+* `major` -- Roll forward to lowest higher major version, and lowest minor version, if requested major version is missing
+* `UseLatestMinor` -- Roll forward to latest minor version, even if requested minor version is on machine. Intended for COM hosting scenario.
+* `UseLatestMajor` -- Roll forward to latest major and highest minor version, even if requested major is on machine. Intended for COM hosting scenario.
+
+There is no default setting. See **Configuration Precedence** for more information.
```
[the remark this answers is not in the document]
richlander: This is what I see at runtime:

```console
C:\testapps\twooneapp>dotnet bin\Debug\netcoreapp2.1\twooneapp.dll
Hello World!
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.1.7\System.Private.CoreLib.dll

C:\testapps\twooneapp>dotnet --fx-version 2.1.0 bin\Debug\netcoreapp2.1\twooneapp.dll
Hello World!
C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.1.0\System.Private.CoreLib.dll
```

for this code:

```csharp
static void Main(string[] args)
{
    Console.WriteLine("Hello World!");
    Console.WriteLine(typeof(Object).Assembly.Location);
}
```

This is the basis of the logic I wrote.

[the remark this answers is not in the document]
vitek-karas: I think that is expected. By default (1st case) the app rolls forward to the latest patch only. Note that "Minor" in the new setting is the same 1 in `rollForwardOnNoCandidateFx`, which only kicks in if the requested `Major.Minor` doesn't exist. In this case it does exist (the machine apparently has at least 2.1.0 and 2.1.7). So the roll forward logic doesn't really kick in.
Rolling forward on patches is a separate piece of functionality, and that always rolls forward to latest (unless completely disabled).

The second case uses the `--fx-version` which currently means exactly that version and no other. Absolutely no roll forward will happen (Major/Minor or Patch). So since you requested 2.1.0, you got 2.1.0.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Configuration Knobs
+
+The host exposes the following configuration knobs that you can use to control binding behavior:
+
+### RollForward
+
+We will introduce a new knob, called `RollForward`, with the following values:
+
+* `None` -- Do not roll forward. Only bind to specified version.
```
nguerrera: Does this have a runtimeconfig equivalent today?

proposed/runtime-binding.md
```diff
@@ -0,0 +1,304 @@
...
+* `Patch` -- Roll forward to the highest patch verison. This disables minor version roll forward.
+* `Minor` -- Roll forward to the lowest higher minor verison, if requested minor version is missing. If the requested minor version is present, then the `Patch` policy is used.
+* `Major` -- Roll forward to lowest higher major version, and lowest minor version, if requested major version is missing. If the requested major version is present, then the `Minor` and `Patch` policies are used, as appropriate.
+* `AlwaysLatestMinor` -- Roll forward to latest minor version, even if requested minor version is present. Intended for COM hosting scenario.
+* `AlwaysLatestMajor` -- Roll forward to latest major and highest minor version, even if requested major is present. Intended for COM hosting scenario.
+
+`Minor` is the default setting. See **Configuration Precedence** for more information.
```
nguerrera: Last time we spoke in a big group about this topic, there was a lot of openness to Major becoming the default. Is that now off the table? cc @DamianEdwards 

I think that this proposal makes configuration easier and the whole process easier to understand, I was really hoping that global tools would start to just workin mismatched major case. cc @wli3

proposed/runtime-binding.md
richlander: I was thinking [`applypatches`](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md#applypatches). On that note, is the third example wrong on that page?

@vitek-karas was advocating for not adding `None` to `RollForward`. I find it hard to reason about `--fx-version` behavior in absence of a `None` option.

proposed/runtime-binding.md
richlander: There was more openness to it at the time. I'm not proposing that, however, as the way to move forward. When we flip that lever, the journey to re-inventing the .NET Framework will be complete. Yeahh!

Another way of putting is asking which UX experience we want (since both will exist to some degree):

- Why didn't think app start? Oh, I have to install something from Microsoft. Ain't got time for that. Screw this app!
- Why did this app crash? I don't know. Screw this app!

I'm pushing choice 1 as the good default. The second one is too hard for everyone involved.

proposed/runtime-binding.md
vitek-karas: I guess `None` would mean `rollForwardOnNoCandidateFx=0` and `applyPatches=0` in current `runtimeconfig`.
I'm not sure we should "advocate" it by including it in the setting here. Basically we don't want anybody to use that setting unless there's no other way as it disables security patching.

I think that if we decide to have an option for this behavior in `RollForward` it should not be called `None` - mostly because people tend to think about `None` as the safe default - which this is not (neither default nor safe).
If we include it I would call it `DisableAll` or something like that - to make it more explicit and more "scary".

proposed/runtime-binding.md
vitek-karas: There's also the argument about hosting environments (think ASP.NET hosting). That is running my app on a container/VM which is managed by somebody else. If in such case I try to run a 2.* app on a host which only has 3.0 and it silently works... everybody thinks (including the hosting provider), that it will work just fine. In reality there might be some slight behavioral difference which can't be tested for upfront before deployment. It makes it harder for both sides to have a solid experience.

proposed/runtime-binding.md
```diff
@@ -229,37 +225,80 @@ Can be specified via the following ways:
...
+The following example enables a .NET Core 2.1 application to run on .NET Core 2.2 without having to select the correct 2.2 patch version. In this case `--fx-version` specifies the floor version. In this scenario, 2.2.0 is not on the machine, but a later patch version is.
 
 ```console
-C:\testapps\threezeroapp>bin\Debug\netcoreapp3.0\threezero.exe --fx-version 3.0.0 --roll-forward patch
+C:\testapps\twooneapp>dotnet --fx-version 2.2.0 bin\Debug\netcoreapp2.1\twooneapp.dll
+It was not possible to find any compatible framework version
+The specified framework 'Microsoft.NETCore.App', version '2.2.0' was not found.
+C:\testapps\twooneapp>dotnet --fx-version 2.2.0 --roll-forward Patch bin\Debug\netcoreapp2.1\twooneapp.dll
+Hello World!
+C:\Program Files\dotnet\shared\Microsoft.NETCore.App\2.2.3\System.Private.CoreLib.dll
```
vitek-karas: I like the overall outcome of this. The only confusion might arise when all the different ways to specify these settings is combined. If specified on command line together like this, it makes perfect sense. If the `RollFoward` is specified through an environment variable which is effectively invisible, it gets a bit more complicated (as compared to existing behavior of `--fx-version`).
Basically if I specify `--fx-version` today I get some relatively predictable behavior. After this change that behavior becomes less predictable since the ambient env. variable may modify it. Now I have to know about the env. variable (and the `runtimeconfig` setting to be able to decide the exact behavior.

I'm not sure if it's OK to leave it like that, or if there's something we should do... and I don't know what... just raising a concern.

proposed/runtime-binding.md
```diff
@@ -229,37 +225,80 @@ Can be specified via the following ways:
...
-The example above will attempt to load the highest `3.0` patch version, which may be `3.0.0` or `3.0.27`, for example, whichever is the highest path version on the machine.
-
 ### Configuration Precedence
 
-The host will consult these settings in order (later scopes take precendence over earlier ones):
+The host will consult the various ways of setting `RollForward`, in order (later scopes take precendence over earlier ones):
+
+1. `.runtimeconfig.json` properties (AKA "json")
+2. CLI arguments
+3. Environment variables (AKA "ENV")
```
vitek-karas: Can we add some reasoning why this order?
I understand why we want `runtimeconfig` to be the least important. I'm not sure about ENV winning over CLI. ENV may be "global" while CLI is always very specific to the app.

proposed/runtime-binding.md
```diff
@@ -229,37 +225,80 @@ Can be specified via the following ways:
...
 
-* `.runtimeconfig.json` properties (AKA "json")
-* CLI arguments
-* Environment variables (AKA "ENV")
+Note: The project file property is not listed because the project file is a build-time not runtime artifact. Runtime-oriented values in project files need to be written to `.runtimeconfig.json` as part of the build in order to influence runtime behavior.
 
 The `version` and `roll-forward` settings compose in the following way:
```
vitek-karas: It would be good to list the various ways to specify `version`. Is it the version of framework in `runtimeconfig` and `--fx-version` CLI, or is there something else as well?

proposed/runtime-binding.md
```diff
@@ -229,37 +225,80 @@ Can be specified via the following ways:
...
 ```
 
-The example above will attempt to load the highest `3.0` patch version, which may be `3.0.0` or `3.0.27`, for example, whichever is the highest path version on the machine.
-
 ### Configuration Precedence
 
-The host will consult these settings in order (later scopes take precendence over earlier ones):
+The host will consult the various ways of setting `RollForward`, in order (later scopes take precendence over earlier ones):
+
+1. `.runtimeconfig.json` properties (AKA "json")
```
vitek-karas: Note that `runtimeconfig` has two levels. For example `rollForwardOnNoCandidateFx` can be specified globally (for all frameworks in the app) or locally per-framework.
I assume we would allow the same for `rollForward`, in which case there's a precedence to define.

proposed/runtime-binding.md
```diff
@@ -229,37 +225,80 @@ Can be specified via the following ways:
 * Environment variable: `DOTNET_ROLL_FORWARD`
```
vitek-karas: As mentioned below, `runtimeconfig` should probably support both global and per-framework setting.

proposed/runtime-binding.md
richlander: Ya. I think you are right. I will switch.

proposed/runtime-binding.md
richlander: I am only aware of those two and cannot think of a useful other option. We could have an ENV to specify version, but I suspect you'd largely want that to test latest and we now have an ENV that does that.

proposed/runtime-binding.md
richlander: What would that look like? I don't think I'm familiar with this scenario.

proposed/runtime-binding.md
vitek-karas: OK - those two make sense - maybe we should list them here, as they are not explicitly mentioned together anywhere in this doc I think...

proposed/runtime-binding.md
vitek-karas: `.runtimeconfig.json` can look like this for example:
``` json
{
  "runtimeOptions": {
    "tfm": "netcoreapp3.0",
    "applyPatches": false,
    "rollForwardOnNoCandidateFx": 0,
    "framework": {
      "name": "Microsoft.NETCore.App",
      "version": "3.0.0",
      "applyPatches": true,
      "rollForwardOnNoCandidateFx": 2
    }
  }
}
```

The settings per-framework win over those on the global level.

proposed/runtime-binding.md
```diff
@@ -205,20 +205,37 @@ We will offer a configuration knobs for hosts that enable artificial roll-forwar
...
+
+
+If `RuntimeFrameworkVersion` is not set, then `version` defaults to the `0` patch for the specified target framework. For example, `version` would be set to `3.0.0` by default for the `netcoreapp3.0` target framework.
+
+MSBuild project settings are only consulted at build-time and are then written as properties to `.runtimeconfig.json`, which can be consulted at runtime.
+
 ### RollForward
 
-We will introduce a new knob, called `RollForward`, with the following values:
+`RollForward` specifies the roll-forward policy for an application, either as a fallback in case a specific version
```
vitek-karas: Probably a copy paste error - the sentence seem to end too early.

proposed/runtime-binding.md
```diff
@@ -205,20 +205,37 @@ We will offer a configuration knobs for hosts that enable artificial roll-forwar
...
-* `Major` -- Roll forward to lowest higher major version, and lowest minor version, if requested major version is missing. If the requested major version is present, then the `Minor` and `Patch` policies are used, as appropriate.
-* `AlwaysLatestMinor` -- Roll forward to latest minor version, even if requested minor version is present. Intended for COM hosting scenario.
-* `AlwaysLatestMajor` -- Roll forward to latest major and highest minor version, even if requested major is present. Intended for COM hosting scenario.
+* `LatestPatch` -- Roll forward to the highest patch verison. This disables minor version roll forward.
+* `Minor` -- Roll forward to the lowest higher minor verison, if requested minor version is missing. If the requested minor version is present, then the `LatestPatch` policy is used.
+* `Major` -- Roll forward to lowest higher major version, and lowest minor version, if requested major version is missing. If the requested major version is present, then the `Minor` policy is used.
+* `LatestMinor` -- Roll forward to latest minor version, even if requested minor version is present. Intended for COM hosting scenario.
```
vitek-karas: Not sure we should explicitly state that this is for COM... it is useful in other cases as well. But if you think it makes sense, I'm fine with it.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,349 @@
...
+
+**User Impact:** When a user runs an application on a machine that does not have `x` installed, but has a later runtime installed, then the app will transparently run on that version. The user does not need to configure their applications to use that later minor version. There is some risk in this scenario that the app will crash due to incompatibilities in the runtime relative to the runtime that the app targets and that it will be difficult to diagnose the cause (incompatibility)and the solution (installing a older runtime). If that app runs successfully, then the user will be happy. The tension between compatibility and deployment convenience is why this scenario is opt-in.
```
normj: I think you meant "applications to use that later **major** version."

normj: Will there be any info or warning text when the app starts to warn the user for the major version roll forward? 

I see a developer just assuming there app will work for all future versions without knowing that version 4.0 changed his assumptions and the app crashes. If the app **transparently** rolls forward and crashes the user never realizes they can solve the problem them selves by installing the appropriate version of .NET Core and instead blame the original developer.

proposed/runtime-binding.md
nguerrera: OK. Next question: do you consider it OK if we recommend to global tool authors to put RollForward=Major in their csproj. We have a lot of feedback that folks don't like that their global tools do not work with only 3.0 installed.

proposed/runtime-binding.md
normj: I would have thought the order precedence would have been:

1. CLI arguments // This is the app user who knows the running the environment the best and is choosing to opt in newer behavior.
2. `.runtimeconfig.json` properties (AKA "json") // The developer believes he is using low level functionality that won't change much and is willing to take the risk it will work in the future but doesn't really not for sure.
3. Environment variables (AKA "ENV") // Useful when you don't control the how the process starts.

At AWS I can't imagine us ever setting the Environment variables on behalf of a customer because we can't guarantee the app would work. But we would want the developer running on AWS the option to override the default behavior by setting the environment variable.

For example a developer has deployed a .NET Core 2.1 Lambda Function (Serverless). Then we add .NET Core 3.X support. Currently the user would have to recompile the app to move it to the 3.X runtime. If the environment variable exists they could just modify the environment variable and change the runtime to be 3.X with no recompilation required. They just need to retest to make sure it works in 3.X.

proposed/runtime-binding.md
richlander: Our philosophy is that the developer owns stdout and stderr. In part, one app's stdout in another app's stdin. stderr is more an option, but is also not perfect.

My belief is that it should be uncommon for apps to run in environments where major version rollforward is enabled. If apps crash in such an environment, it should be possible to connect the opt-in rollforward setting with the crash and consider that the setting is the cause of the crash if no other more obvious causes are identified.

Fair?

proposed/runtime-binding.md
jkotas: > OK if we recommend to global tool authors to put `RollForward=Major`

It is ok as long as the global tool authors understand the consequences. This setting increases probability that the tool won't work.

I think it is manageable for small tools that are unlikely to break. I would not recommend it for large complex tools.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,351 @@
...
+
+Given a request for `x.y.z` version, the host will look for an `x` runtime version, as described for **Minor Version Selection** above. If it cannot find an `x` runtime, then the host will look for all runtimes that are higher than `x`. It will select the lowest major version within that set and then select the lowest minor and highest patch version within that set. This behavior is oriented on a best effort attempt at selecting a compatible installed product version. That's why the lowest minor version is selected, not the highest one.
+
+**User Impact:** When a user runs an application on a machine that does not have `x` installed, but has a later runtime installed, then the app will transparently run on that later major version.
+
+There is some risk in this scenario that the app will crash due to incompatibilities in the runtime relative to the runtime that the app targets and that it will be difficult to diagnose the cause (incompatibility)and the solution (installing a older runtime). If that app runs successfully, then the user will be happy. The tension between compatibility and deployment convenience is why this scenario is opt-in.
```
jkotas:
```suggestion
There is some risk in this scenario that the app will crash due to incompatibilities in the runtime relative to the runtime that the app targets and that it will be difficult to diagnose the cause (incompatibility) and the solution (installing a older runtime). If that app runs successfully, then the user will be happy. The tension between compatibility and deployment convenience is why this scenario is opt-in.
```

proposed/runtime-binding.md
vitek-karas: What other mechanism do you think should be used to notify about the roll-forward happening.
On Windows we were discussing that maybe we should write an information event into the event log.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,349 @@
...
+* If `ROLL_FORWARD_ON_NO_CANDIDATE_FX` is set, it will be honored.
+* If `ROLL_FORWARD` is set, it will be honored.
+* If both settings are set, it is an error.
+* If neither settings are set, then the existing default behavior is used, as described in the **Patch Version Selection** and **Patch Version Selection** sections above.
+
+Note: The ENV syntax is used above, but the same rules apply for any way that the two settings are set.
+
+### Configuration Precedence
+
+The host will consult the various ways of setting `RollForward`, in order (later scopes take precendence over earlier ones):
```
normj: typo precendence -> precedence

proposed/runtime-binding.md
normj: I see that I missed the key detail above `(later scopes take precedence over earlier ones)` and cli arguments will override anything in `.runtime.config` or environment variables.

For environment variables I think they should have the least amount of power to change the behavior given their potential large scope. If somebody explicitly made a setting in `.runtime.config` or command line arguments, those should be chosen over environment variables since the variable might not have been set for that particular application.

This logic is based on the comment earliar that tooling would not automatically set any the project settings so that a developer always has to explicitly make the choice to opt into their behavior.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,351 @@
...
```
richlander:
```suggestion
The host will consult the various ways of setting `RollForward`, in order (later scopes take precedence over earlier ones):
```

proposed/runtime-binding.md
normj: I like the idea of it getting put out as a runtime event.

normj:
LGTM

It requires a high bar of compatibility between minor versions and protects .NET Core's ability to evolve with breaking changes when they are important for major versions.

proposed/runtime-binding.md
```diff
@@ -0,0 +1,401 @@
...
+
+As of 3.0 Preview 2, previews have special binding behavior. In this proposal, previews are proposed to not have special behavior. Apps that depend on preview runtimes can bind to non-preview runtimes and vice-versa. Previews are not considered to be special going forward, with this proposal. This behavior is oriented on enabling applications built with previews to get a chance to run (likely successfully) on a stable version and to make testing stable applications on previews easy.
+
+Assumption: Stable Visual Studio versions only install stable .NET Core versions.
+
+## Handling Frameworks
+
+In .NET Core 3.0, components like ASP.NET Core, WPF and Windows Forms are modeled as *frameworks*. Frameworks follow the same binding rules as the runtime. The runtime is modeled as the lowest-level framework in the system.
+
+The following example, the application depends on the `Microsoft.NETCore.App` framework, version `3.0.0-preview-27324-5`. This framework is defined within the .NET Core installation and expresses a dependency on `Microsoft.NETCore.App`, version `3.0.0-preview-27324-5`. The host binds the frameworks according to these declarations.
```
vitek-karas:
```suggestion
The following example, the application depends on the `Microsoft.AspNetCore.App` framework, version `3.0.0-preview-19075-0444`. This framework is defined within the .NET Core installation and expresses a dependency on `Microsoft.NETCore.App`, version `3.0.0-preview-27324-5`. The host binds the frameworks according to these declarations.
```

PR #52, opened 2019-02-07
Publishing to Single-file in .Net Core

swaroop-sridhar:
The is a design proposal for supporting single file distribution in .Net Core.

To run apps published as a single-file, [CoreCLR#20287](https://github.com/dotnet/coreclr/issues/20287) proposed extracting out all the embedded dependencies to files at startup.

However, based on customer feedback through public and private fourms, the current design uses a hybrid of in-memory execution and extraction to files.

Further stages of improvements are outlined in `staging.md`.

### DIFF
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -118,7 +118,7 @@ dotnet publish --single-file
 By default, `PublishSingleFile` bundles all files in the publish directory into the single executable. This behavior can be altered by using the following content meta-data:
 
 ```xml
-<IncludeInSingleFilePublish>{Always, PreserveNewest, Never}</IncludeInSingleFilePublish>
+<IncludeInSingleFile>{Always, PreserveNewest, Never}</IncludeInSingleFile>
 ```
 
 For example, to place some files in the publish directory but not bundle them in the single-file:
@@ -127,7 +127,7 @@ For example, to place some files in the publish directory but not bundle them in
 <ItemGroup>
     <Content Update="*.xml">
       <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
-      <IncludeInSingleFilePublish>PreserveNewest</CopyToPublishDirectory>
+      <IncludeInSingleFile>PreserveNewest</IncludeInSingleFile>
     </Content>
   </ItemGroup>
 ```
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -78,7 +78,7 @@ To any host (native binary) specified, the bundler tool will add:
 
 ##### Implementation
 
-The bundler should ideally be located close to the dotnet host, since their implementation is closely related. Therefore, the bundler will be implemented in the `core-setup` repo. The `corehost` code is expected to move to the CoreCLR repo, at which point, the bundler should move with it. The bundler will be implemented in managed code.
+The bundler should ideally be located close to the core-host, since their implementation is closely related. Therefore, the bundler will be implemented in the `core-setup` repo. The bundler will be implemented in managed code.
 
 ##### Command Line Interface
 
@@ -237,7 +237,18 @@ The semantics of assembly location should be tuned further based on user feedbac
 * A subset of CoreCLR Tests
 * Tests to ensure that MSIL files with embedded PDBs are handled correctly
 * End-to-End testing on real world apps such as Roslyn, MusicStore
-* *Measurements*: Publish size and run-time (first run, subsequent runs) for HelloWorld, Roslyn and MusicStore.
+
+#### Measurements
+
+Measure publish size and run-time (first run, subsequent runs) for HelloWorld (framework-dependent and self-contained), Roslyn and MusicStore.
+
+#### Telemetry
+
+Collect telemetry for single-file published apps with respect to parameters such as:
+
+* Framework-dependent vs self-contained apps.
+* Whether the apps are Pure managed apps, ready-to run compiled apps, or have native dependencies.
+* Embedding of additional/data files, and use of file access API.
 
 ## Further Work
 
@@ -259,8 +270,6 @@ The above design should be extended to seamlessly support single-file publish fo
 - For host/runtime support, the options are:
   - Implement plugins using existing infrastructure. For example: Take control of assembly/native binary loads via existing `AssemblyLoadContext` callbacks and events. Extract the files embedded within the single-file plugin using the `GetFileStream()` API and load them on demand.
   - Have new API to load a single-file plugin, for example: `AssemblyLoadContext.LoadWithEmbeddedDependencies()`.
-
-
 
 #### VS Integration
 
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -78,21 +78,9 @@ To any host (native binary) specified, the bundler tool will add:
 
 ##### Implementation
 
-The bundler should ideally be located close to the core-host, since their implementation is closely related. Therefore, the bundler will be implemented in the `core-setup` repo. The bundler will be implemented in managed code.
+The bundler should ideally be located close to the core-host, since their implementation is closely related. Therefore, the bundler will be implemented in the `core-setup` repo. 
 
-##### Command Line Interface
-
-The bundler will be a standalone tool with the following command line:
-
-```bash
-bundle -a <App>      The name of the managed app
-       -h <Host>     The path to the native host
-       -c <contents> The path to a directory containing the files to bundle
-       [-o <output>] The path to output bundle
-       [-v]          Generate verbose output
-```
-
-Most users are only expected to interact with the bundler via the build system.
+The bundler will be implemented in managed code as a standalone tool. Application developers are expected to interact with the bundler only through the build system. The `dotnet` SDK will implement msbuild-tasks that use this tool to publish single-file apps.  
 
 ##### Build System Interface
 
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -108,10 +108,28 @@ Publishing to a single file can be triggered by adding the following property to
 * The `PublishSingleFile` property applies to platform-specific builds with respect to a given runtime-identifier. The output of the build is a native binary for the specified platform.
 * Setting the `PublishSingleFile`property causes the managed app, managed dependencies, platform-specific native dependencies, configurations, etc. (basically the contents of the publish directory when `dotnet publish` is run without setting the property) to be embedded within the native `apphost`. The publish directory will only contain the single bundled executable (and the PDB file).
 
-Optionally, we can add the following switch to the CLI, as a shortcut to setting the `PublishSingleFile` property.
+Optionally, we can add a switch to the CLI, as a shortcut to setting the `PublishSingleFile` property.
 
 ```bash
+# Alias for dotnet publish /p:PublishSingleFile=true
 dotnet publish --single-file
+```
+
+By default, `PublishSingleFile` bundles all files in the publish directory into the single executable. This behavior can be altered by using the following content meta-data:
+
+```xml
+<IncludeInSingleFilePublish>{Always, PreserveNewest, Never}</IncludeInSingleFilePublish>
+```
+
+For example, to place some files in the publish directory but not bundle them in the single-file:
+
+```xml
+<ItemGroup>
+    <Content Update="*.xml">
+      <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+      <IncludeInSingleFilePublish>PreserveNewest</CopyToPublishDirectory>
+    </Content>
+  </ItemGroup>
 ```
 
 #### Interaction with other tools
@@ -164,6 +182,9 @@ the TPA with single-exe publish will be (assuming `?` is the bundle marker):
 Since all the files of an app published as a single-file live together, we can perform the following optimizations
 
 - R2R compile the app and all of its dependent assemblies in a single version-bubble
+
+  Single-file apps compiled cross-platform may have this optimization disabled until the ready-to-run compiler  (`crossgen`) supports cross-compilation. 
+
 - Investigate whether collectively signing the files in an assembly saves space for certificates.
 
 #### APIs to access Bundled files
@@ -198,33 +219,7 @@ The above design should be extended to seamlessly support single-file publish fo
   - Implement plugins using existing infrastructure. For example: Take control of assembly/native binary loads via existing `AssemblyLoadContext` callbacks and events. Extract the files embedded within the single-file plugin using the `GetFileStream()` API and load them on demand.
   - Have new API to load a single-file plugin, for example: `AssemblyLoadContext.LoadWithEmbeddedDependencies()`.
 
-#### Project File Options
 
-By default, `PublishSingleFile` bundles all files in the publish directory into the single executable. This behavior can be altered via the following configurations in the application's project file.
-
-* Exclude certain files in the publish directory from being bundled (ex: because of licensing issues or for testing purposes).
-
-```xml
-<ItemGroup>
-    <SingleFileExclude Include="List of patterns to exclude"/>
-</ItemGroup>
-```
-
-- Add additional files to the bundle from other locations (ex: add data files to the app). The developer will be responsible for handling these additional files.
-
-```xml
-<ItemGroup>
-    <SingleFileAdd Include="List of file paths"/>
-</ItemGroup>
-```
-
-An app may choose to have all of its dependencies extracted to disk at runtime (instead of loading certain files directly from the bundle) by setting the following property.
-
-```xml
-<PropertyGroup>
-    <SingleFileExtractAll>true</SingleFileExtractAll>
-</PropertyGroup>
-```
 
 #### VS Integration
 
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -310,5 +310,6 @@ Most applications are expected to work without any changes. However, apps with a
 * [Mono/MkBundle](https://github.com/mono/mono/blob/master/mcs/tools/mkbundle) 
 * [Fody.Costura](https://github.com/Fody/Costura)
 * [Warp](https://github.com/dgiagio/warp)
+* [dotnet-warp](https://github.com/Hubert-Rybak/dotnet-warp)
 * [BoxedApp](https://docs.boxedapp.com/index.html)
 
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -166,26 +166,23 @@ Where
 the TPA with single-exe publish will be (assuming `?` is the bundle marker):
 `?/il.dll`; `/tmp/.net/e53xf3/r2r.dll`; `/usr/local/nuget/dep.dll`
 
-## Testing
+### New API
 
-* Unit Tests to achieve good coverage on apps using managed, ready-to-run, native code 
-* Tests to verify that framework-dependent and self-contained apps can be published as single file.
-* Tests for ensure that every app model template supported by .NET Core (console, wpf, winforms, web, etc.) can be published as a single file.
-* A subset of CoreCLR Tests
-* End-to-End testing on real world apps such as Roslyn, MusicStore
-* *Measurements*: Publish size and run-time (first run, subsequent runs) for HelloWorld, Roslyn and MusicStore.
+We propose adding the following API to the .net core framework to facilitate operation of apps published as single-files 
 
-## Further Work
+#### Single-file? 
 
-#### Bundler Optimizations
+Check whether an app is running from a single-file bundle:
 
-Since all the files of an app published as a single-file live together, we can perform the following optimizations
-
-- R2R compile the app and all of its dependent assemblies in a single version-bubble
-
-  Single-file apps compiled cross-platform may have this optimization disabled until the ready-to-run compiler  (`crossgen`) supports cross-compilation. 
-
-- Investigate whether collectively signing the files in an assembly saves space for certificates.
+```cs
+namespace System.Runtime.Loader
+{
+    public partial class Bundle
+    {
+        public static bool IsBundle(Assembly assembly);
+    }
+}
+```
 
 #### APIs to access Bundled files
 
@@ -209,6 +206,29 @@ namespace System.Runtime.Loader
 ```
 
 We can also provide an abstraction that abstracts away the physical location of a file (bundle or disk). For example, add a variant of  `GetFileStream` API that looks for a file in the bundle, and if not found, falls back to disk-lookup.
+
+## Testing
+
+* Unit Tests to achieve good coverage on apps using managed, ready-to-run, native code 
+* Tests for the newly added API 
+* Tests to verify that framework-dependent and self-contained apps can be published as single file.
+* Tests for ensure that every app model template supported by .NET Core (console, wpf, winforms, web, etc.) can be published as a single file.
+* A subset of CoreCLR Tests
+* Tests to ensure that MSIL files with embedded PDBs are handled correctly
+* End-to-End testing on real world apps such as Roslyn, MusicStore
+* *Measurements*: Publish size and run-time (first run, subsequent runs) for HelloWorld, Roslyn and MusicStore.
+
+## Further Work
+
+#### Bundler Optimizations
+
+Since all the files of an app published as a single-file live together, we can perform the following optimizations
+
+- R2R compile the app and all of its dependent assemblies in a single version-bubble
+
+  Single-file apps compiled cross-platform may have this optimization disabled until the ready-to-run compiler  (`crossgen`) supports cross-compilation. 
+
+- Investigate whether collectively signing the files in an assembly saves space for certificates.
 
 #### Single-file Plugins
 
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -168,7 +168,8 @@ the TPA with single-exe publish will be (assuming `?` is the bundle marker):
 
 ### New API
 
-We propose adding the following API to the .net core framework to facilitate operation of apps published as single-files 
+In this section, we propose adding a few APIs to facilitate common operations on bundled-apps.
+This is only a draft of the proposed APIs. The actual shape of the APIs will be decided via API review process.
 
 #### Single-file? 
 
@@ -192,7 +193,7 @@ The binaries that are published in the project are expected to be handled transp
 - Open an assembly for reflection/inspection
 - Load plugins built as single-file class-libs using existing Loader APIs.
 
-In facilitate this usage, we can add an API similar to [GetManifestResourceStream](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getmanifestresourcestream?view=netframework-4.7.2#System_Reflection_Assembly_GetManifestResourceStream_System_String_) to obtain a stream corresponding to an embedded file. 
+Therefore, we propose adding an API similar to [GetManifestResourceStream](https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getmanifestresourcestream?view=netframework-4.7.2#System_Reflection_Assembly_GetManifestResourceStream_System_String_) to obtain a stream corresponding to an embedded file. 
 
 ```C#
 // Open a file embedded in the bundle built for the specified assembly 
@@ -205,7 +206,27 @@ namespace System.Runtime.Loader
 }
 ```
 
-We can also provide an abstraction that abstracts away the physical location of a file (bundle or disk). For example, add a variant of  `GetFileStream` API that looks for a file in the bundle, and if not found, falls back to disk-lookup.
+We can also provide an abstraction that abstracts away the physical location of a file (bundle or disk). For example, add a variant of  `GetFileStream` API that looks for a file in the bundle, and if not found, falls back to disk-lookup. However such abstractions are also easy to build outside of the .Net Framework.
+
+### Existing API
+
+We need to determine the semantics of current APIs such as `Assembly.Location` that return the information about an assembly's location on disk. For example, `Assembly.Location`  for a bundled assembly could return:
+
+* A fixed literal (ex: `null`) indicating that no actual location is available.
+* The simple name of the assembly (with no path).
+* A special UNC notation such as `//?/asm.dll` to denote files that come from the bundle. If multiple single-file bundles are used in an app (ex: plugins), this notation could be extended as `//?bundle-assembly/name.dll` -- if it is important to make a distinction with respect to which bundle an assembly came from.
+* The path of the assembly as if it were not to be packaged into the single-file.
+* In the case of files spilled to disk (ex: ready-to-run assemblies) -- the actual location of the extracted file.
+* A configurable selection of the above, etc.
+
+Most of the app development can be agnostic to whether the app is published as single-file or not. However, the parts of the app that deal with physical locations of files need to be aware of the single-file packaging. 
+
+In the first implementation, we propose keeping the default behavior of `Assembly.Location`. That is, 
+
+* If the assembly is loaded directly from the bundle, return empty-string
+* If the assembly is extracted to disk, return the location of the extracted file.
+
+The semantics of assembly location should be tuned further based on user feedback, and such that it aligns with other bundling technologies in .Net Core ecosystem.
 
 ## Testing
 
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -266,6 +266,7 @@ To summarize, here's the overall experience for creating a HelloWorld single-fil
 
   * The bundled app and configuration files are processed directly from the bundle.
   * Remaining 216 files will be extracted to disk at startup. 
+  * If reuse of extracted files is enabled, subsequent runs of the app may skip the extraction step.
 
 
 Most applications are expected to work without any changes. However, apps with a strong expectation about absolute location of dependent files may need to be made aware of bundling and extraction aspects of single-file publishing. No difference is expected with respect to debugging and analysis of apps.
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -1,0 +1,279 @@
+# Single-file Publish
+
+Design for publishing apps as a single-file in .Net Core 3.0
+
+## Introduction
+
+The goal of this effort is enable .Net-Core apps to be published and distributed as a single executable.
+
+There are several strategies to implement this feature -- ranging from bundling the published files into zip file (ex: [Warp](https://github.com/dgiagio/warp)), to native compiling and linking all the binaries together (ex: [CoreRT](https://github.com/dotnet/corert)). These options, along with their cost/benefit analysis is explored in this [staging document](staging.md).
+
+#### Goals
+
+In .Net Core 3.0, we plan to implement a solution that 
+
+* Is widely compatible: Apps containing MSIL assemblies, ready-to-run assemblies, native binaries, configuration files, etc. can be packaged into one executable.
+* Can run framework dependent pure managed apps directly from bundle:
+  * Executes IL assemblies, and processes configuration files directly from the bundled executable.
+  * Extracts ready-to-run and native binaries to disk before loading them.
+* Usable with debuggers and tools:  The single-file should be debuggable using the generated symbol file. It should also be usable with profilers and tools similar to a non-bundled app.
+
+This feature-set is described as Stage 2 in the [staging document](staging.md), and can be improvised in further releases.
+
+#### Non Goals
+
+* Optimizing for development: The single-file publishing is typically not a part of the development cycle. It is typically used in release builds as a packaging step. Therefore, the single-file feature will be designed with focus on consumption rather than production.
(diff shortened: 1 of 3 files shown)


ericsampson:
For self-contained, what happens the second (and subsequent) runs of the .exe? I'd hope that it would be smart enough to only extract the files the first time.

swaroop-sridhar:
@ericsampson the proposal to reuse extracted files on subsequent runs is discussed here: https://github.com/dotnet/designs/pull/52/files#diff-5e93215a94e626495c10448a40fa4ccbR40

ericsampson:
Thanks @swaroop-sridhar I did end up finding that, it just wasn't mentioned in the design.md user User Experience which makes it seem like the extraction was always going to happen each time: Cheers!

Run: HelloWorld.exe

The bundled app and configuration files are processed directly from the bundle.
Remaining 216 files will be extracted to disk at startup.

swaroop-sridhar:
Thanks for pointing that out @ericsampson. I added a note about it [here](https://github.com/dotnet/designs/pull/52/files#diff-28dba553f5f12520afc2d90d1587ef56R269)

accepted/single-file/extract.md
```diff
@@ -0,0 +1,81 @@
...
+* Be short, to reduce the risk of running over path length limit.
+* Be randomized, in order to support concurrent launches of the same app.
+
+The proposed extraction path is `%TEMP\.net\<random-code>` / `$TMPDIR/.net/<random-code>` . 
+
+#### Pros
+
+* Simpler implementation: fewer complexities with respect to app-updates and concurrency.
+* User need not worry about explicit cleanup.
+* Some customers don't prefer a click-and-run model, rather than an app-install model. This option better suits them.
```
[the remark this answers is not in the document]
swaroop-sridhar: Click-and-run model: App runs with necessary extraction on startup and cleanup on exit. There are no side-effects after the app is run.
Install model: App installs the extracted files to user-visible non-temporary storage, for reuse across multiple runs.

The point here is that the temporary-extraction solution suits situations that don't prefer apps leaving behind state on disk.

accepted/single-file/staging.md
```diff
@@ -0,0 +1,139 @@
...
+
+* Unsuitable for environments that have native dependencies that must be bundled and cannot be extracted.
+
+## 4. Run from Bundle: .Net Core libraries 
+
+### 4.1 Description
+
+In this stage
+
+* .Net Core native libraries are statically linked to the single-file host executable.
```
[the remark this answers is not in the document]
swaroop-sridhar: Yes

accepted/single-file/design.md
```diff
@@ -0,0 +1,279 @@
...
+We can also provide an abstraction that abstracts away the physical location of a file (bundle or disk). For example, add a variant of  `GetFileStream` API that looks for a file in the bundle, and if not found, falls back to disk-lookup.
+
+#### Single-file Plugins
+
+The above design should be extended to seamlessly support single-file publish for plugins.
+
+- The bundler tool will mostly work as-is, regardless of whether we publish an application or class-lib. The binary blob with dependencies can be appended to both native and managed binaries.
+- For host/runtime support, the options are:
+  - Implement plugins using existing infrastructure. For example: Take control of assembly/native binary loads via existing `AssemblyLoadContext` callbacks and events. Extract the files embedded within the single-file plugin using the `GetFileStream()` API and load them on demand.
+  - Have new API to load a single-file plugin, for example: `AssemblyLoadContext.LoadWithEmbeddedDependencies()`.
```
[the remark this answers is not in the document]
jeffschwMSFT: @ayende this is a good suggestion.  Can you describe a potential scenario?

accepted/single-file/extract.md
```diff
@@ -0,0 +1,81 @@
...
+
+Mono extracts native libraries at startup. Costura extracts and explicitly loads the native-libraries at startup.
+
+## Temporary Extraction
+
+The first option is to extract out the necessary files to a temporary location, and clean them out on exit. The requirements for this path are:
+
+* Extract to a location within the temp-directory, so that if some files are not removed (ex: because an app crashed before cleanup), they are removed when the temp-directory is purged.
+* Be short, to reduce the risk of running over path length limit.
+* Be randomized, in order to support concurrent launches of the same app.
```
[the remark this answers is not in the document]
jeffschwMSFT: We have discussed having reuse as an option.  Both 'extract everytime' and 'reuse' have different pros and cons.  We plan on leaving it up to the developer to choose.

swaroop-sridhar:
>> how do I select which native deps I'll bundle with each?
Yes @ayende  Single-file apps are platform-specific. So, they must be generated from a platform specific publish (wrt a particular RID), whether framework-dependent or self-contained. 

Embedding the correct native dependencies into a single-file is no different from copying the native dependencies to the RID-specific publish directory. The single-exe part of the build only embeds the contents of the RID-specific publish directory into the host. 

Are you asking about adding additional native binaries that are not in the publish directory?

accepted/single-file/design.md
```diff
@@ -0,0 +1,279 @@
...
+
+```xml
+<PropertyGroup>
+    <PublishSingleFile>true</PublishSingleFile>
+</PropertyGroup>    
+```
+
+* The `PublishSingleFile` property applies to both framework dependent and self-contained publish operations.
+* Setting the `PublishSingleFile`property causes the managed app, dependencies, configurations, etc. (basically the contents of the publish directory when `dotnet publish` is run without setting the property) to be embedded within the native `apphost`. The publish directory will only contain the single bundled executable (and the symbol file).
+
```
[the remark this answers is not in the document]
ayende: That is actually a concern for me. I want to be able to have a single package that include the PDB. We currently do that with embedded pdbs, and I assume this will continue to work?

accepted/single-file/extract.md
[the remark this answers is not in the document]
ayende: The issue is that Windows has to load the native library from a file, right?
In the click and run scenario, can you write the files with `FILE_FLAG_DELETE_ON_CLOSE`, call `LoadLibrary` on the file and then close it?
That would mean that the OS will take care of cleanup automatically.

accepted/single-file/design.md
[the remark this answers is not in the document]
ayende: Consider the case of resources that are bundled into the app. In our case, we have a web app that has a lot of files (js, png, css, etc).
We actually have three separate options here right now. During dev, we might be working on the raw files, the compiled files or during release, loading them from a zip file.

Being able to tell how we are running would be helpful because if `IsBundled == false`, I need to probe a whole different location to see where the file I need to serve it located.

MichalStrehovsky:
We might need to define what APIs like `Assembly.Location` and `Assembly.CodeBase` return for bundled assemblies. This is a problem that has been plaguing CoreRT/ProjectN for a long time (see e.g. https://github.com/dotnet/corefx/pull/35148#issuecomment-461803725, dotnet/corert#5647, https://github.com/dotnet/corert/issues/6679#issuecomment-446919279, dotnet/corert#6947, dotnet/corert#6946).

I haven't been able to find a solution that would work for everyone and the best I have is for us to offer a new `AppContext` switch to select one of these behaviors (one of them should be the default):

* `Location` returns an empty string. This is a documented behavior of Location if the assembly got loaded from byte array. Theoretically, user code should be able to handle that. In reality, this is a very good way to break everyone using this API.
* `Location` returns the name of the assembly, without a file path. This should work for people who want to roundtrip the location through e.g. `Assembly.Load`.
* `Location` returns the name of the assembly, qualified by `AppContext.BaseDirectory`. This makes existing code that does `Path.GetFilePath(Assembly.GetEntryAssembly().Location)` to get to the app's directory. This pattern is extremely common.

ViktorHofer:
I don't have much context here but naively speaking, couldn't option 3 (AppContext.BaseDirectory + assembly name) also allow round-tripping with Assembly.Load. Means if the full path passed into Assembly.Load is the current executing assembly it won't attempt to load that assembly and just returns.

jkotas:
> This is a problem that has been plaguing CoreRT/ProjectN for a long time

I think the large part of the problem has been that CoreRT/ProjectN were niches and not many folks cared to make their projects compatible with the "single file" model used by them. The good news is that making "single file" model more first class citizen will make more folks to care about it.

We are going to need new APIs to allow apps and libraries to do what they need to do in the (bundled) single file mode. It is unlikely that we would be able to make things "just work" well by quirking existing APIs. I would worry about making sure that we have all the right API first, and only worry about quirking existing APIs last based on the feedback.

accepted/single-file/design.md
```diff
@@ -0,0 +1,281 @@
...
+* End-to-End testing on real world apps such as Roslyn, MusicStore
+* *Measurements*: Publish size and run-time (first run, subsequent runs) for HelloWorld, Roslyn and MusicStore.
+
+## Further Work
+
+#### Bundler Optimizations
+
+Since all the files of an app published as a single-file live together, we can perform the following optimizations
+
+- R2R compile the app and all of its dependent assemblies in a single version-bubble
```
nguerrera: Is this at odds with the other goal of "Able to generate cross-platform bundles (ex: publish a single-file for Linux target from Windows)"? My understanding is that at present, we do not support "cross-crossgen", correct?

accepted/single-file/design.md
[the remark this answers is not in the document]
jeffschwMSFT: Thanks for the feedback, we could make this configurable.  I think by default (to save size on disk) this would be off.

accepted/single-file/design.md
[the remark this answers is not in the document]
jeffschwMSFT: Good feedback, and thanks for the scenario details.

accepted/single-file/design.md
```diff
@@ -0,0 +1,281 @@
...
+<ItemGroup>
+    <SingleFileExclude Include="List of patterns to exclude"/>
+</ItemGroup>
+```
+
+- Add additional files to the bundle from other locations (ex: add data files to the app). The developer will be responsible for handling these additional files.
+
+```xml
+<ItemGroup>
+    <SingleFileAdd Include="List of file paths"/>
```
nguerrera: There are IDE considerations for adding new item types. This may not give the best experience for maintaining such files in the IDE.

We should consider if this can be metadata on Content. Today there is CopyToOutputDirectory, CopyToPublishDirectory. There can be IncludeInSingleFilePublish or something.

Today, CopyToOutputDirectory implies CopyToPublishDirectory by default and you can override.

accepted/single-file/design.md
```diff
@@ -0,0 +1,281 @@
...
+
+*  Create a new HelloWorld app: `HelloWorld$ dotnet new console`
+
+#### Framework Dependent HelloWorld
+
+* Normal publish: `dotnet publish` 
+
+  * Publish directory contains the host `HelloWorld.exe` ,  the app `HelloWorld.dll`, configuration files `HelloWorld.deps.json`, `HelloWorld.runtimeconfig.json`, and the PDB-file `HelloWorld.pdb`.
+
+* Single-file publish: `dotnet publish /p:PublishSingleFile=true`
```
nguerrera: Should PublishSingleFile change the default output directory of the publish? Incremental publish today already has some problems. I suspect publishing alternating between single-file and not will compound them if we're not careful. We might also take the opportunity to spec out the incrementality better, fix bugs, and handle this case carefully.

nguerrera: Or we could not advertise using a global swtich on publish and instead recommend setting PublishSingleFile in project file based on configuration. In general, switching global properties between msbuild invocations is tricky business with respect to incrementality.

nguerrera: We may want to note the option to add `dotnet publish --single-file` as sugar for `dotnet publish /p:PublishSingleFile=trure` . Whether this is a good idea depends on how we answer my other questions above.

accepted/single-file/design.md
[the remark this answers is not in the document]
swaroop-sridhar: OK, thanks for the details @ayende. For this situation, I think it would be better served by providing  an abstraction that handles bundles and files seamlessly, as proposed [here](https://github.com/dotnet/designs/pull/52/files#diff-28dba553f5f12520afc2d90d1587ef56R190).

But I think it is also reasonable to expose an API like the following, to support custom scenarios.
```C#
namespace System.Runtime.Loader
{
    public partial class Bundle
    {
        public static bool IsBundle(Assembly assembly);
    }
}
```

accepted/single-file/design.md
swaroop-sridhar: That's a good suggestion @nguerrera, I'll change `SingleFile Include/Exclude` to be meta-data.

accepted/single-file/design.md
swaroop-sridhar: The output directory issue is should be handled similarly for all post-processing tools (currently ILLinker, crossgen, and single-file). The publish process may temporarily use other directories, but ultimately, the publish directory should contain the single-file, I think.

I agree with the idea of marking the PublishSingleFile on the project file rather than the command line, in order to reduce confusion wrt incremental build.

accepted/single-file/design.md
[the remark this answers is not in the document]
swaroop-sridhar: @ayende you're asking about assemblies generated with debug info embedded  (ex: `csc -debug:embedded`) right? Those are fine, I'll add it as a scenario to test.
 If a separate PDB file is generated, we won't add it to the single-file. But as @jeffschwMSFT says, we can make it configurable if necessary.

accepted/single-file/extract.md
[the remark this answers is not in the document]
swaroop-sridhar: Thanks @ayende. We can consider that for implementation on Windows.
However, we will still need a clean up step on exit, because there may be files other files (ex: data files) that are bundled and extracted out to disk.

accepted/single-file/design.md
swaroop-sridhar: Thanks for pointing that out @nguerrera. Yes, we can generate cross-platform bundles without any new cross-compilation optimizations, until cross-crossgen is available.

MichalStrehovsky:
> We are going to need new APIs to allow apps and libraries to do what they need to do in the (bundled) single file mode. It is unlikely that we would be able to make things "just work" well by quirking existing APIs.

Not having tools available to help existing code bridge the gap until it can move to the new API is always a source of sadness for the customer, and for the person trying to help the customer. I'm not as enthusiastic about libraries adapting to this quickly. Single file deployments will be niche, unless we make them the default.

I had conversations like this many times, and some of them were literally about `Assembly.Location`:

\- Just don't call this API, call this other new API instead.
\- The call is in a NuGet package.
\- Oh.

accepted/single-file/design.md
ericsampson: I know it seems really minor, but IMHO having CLI sugar for this would be really nice in terms of lightweight usage, demos, discoverability, etc.

accepted/single-file/design.md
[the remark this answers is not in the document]
ericsampson: +1 on configurably embedded PDB.

jeffschwMSFT:
@MichalStrehovsky thanks for bringing this perspective.  The good thing is that in these cases these files could be spilled to disk.  In general we strive to enable the spill to disk with high fidelity to the original apps intention.

ericsampson:
Is someone from Azure App Service looped in this design discussion, because it could play in very nicely with Run From Package (which is also used by Azure Functions, so coldstart perf improvements if the 'extract once' functionality is leveraged automatically? Or eventually the Stage 4/5 'extract and run in-mem' could be huge?) @odhanson @jeffhollan

swaroop-sridhar:
Thanks @ericsampson. That's a great idea.
@KetanChawda-MSFT can you please tag the team that handles "Run from package" app service, to take a look at this design? Thanks.

accepted/single-file/extract.md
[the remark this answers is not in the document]
ayende: Can probably do that as well. Extract them, open with delete on close and "leak" the handle.
The OS will handle that for you.
On Linux, I think we can load a `memfd_create` and then `dlopen` , so we don't even need to touch the disk

MichalStrehovsky:
> The good thing is that in these cases these files could be spilled to disk

They'll be spilled to the temporary location so the code that assumes it can read e.g. a config file from `Path.GetFilePath(Assembly.GetEntryAssembly().Location)` is still going to get some path in %TEMP%. We might as well return the same string we would return for the embedded case so that the implementation detail of extracting to a temporary location won't be visible (especially if we're extracting due to a current implementation limitation and not because the user asked for that).

It's not an academic scenario - we see this pretty often in ProjectN. Here's one [in xunit](https://github.com/xunit/xunit/blob/cbf28f6d911747fc2bcd64b6f57663aecac91a4c/src/xunit.console/ConsoleRunner.cs#L131).

jeffschwMSFT:
@MichalStrehovsky who would we hide the paths?  In general being single-exe is a chosen path and spilling to disk is also a choice.  I would opt for defaults that enable the most things to 'just work'.  I would presume that if we spill to disk and not hide the path that most uses of Location, etc. would just work.  Are there cases that do not?

accepted/single-file/design.md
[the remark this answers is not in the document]
ayende: Yes, I am. If those work, I'm happy with that.

MichalStrehovsky:
> I would presume that if we spill to disk and not hide the path that most uses of Location, etc. would just work. Are there cases that do not?

The cases when the assembly is not interesting, only the directory where the assembly is. [Path.GetDirectoryName(Assembly.GetEntryAssembly().Location)](https://www.google.com/search?ei=0ZheXIyoBv2Ak74P_KGeoA8&q=%22Path.GetDirectoryName%28Assembly.GetEntryAssembly%28%29.Location%29%22&oq=%22Path.GetDirectoryName%28Assembly.GetEntryAssembly%28%29.Location%29%22&gs_l=psy-ab.3...0.0..28780...0.0..0.0.0.......0......gws-wiz.yragdAh-HDY) is a common way to get to the "app directory" to read config files or load plugins from. Doing that is roughly equivalent to calling `AppContext.BaseDirectory` right now. If we do nothing, it will either be an empty string (loaded from memory), or a path to the extracted files, depending on implementation details (that might change).

We either come up with a standardized way to approach this (e.g. single file deployments on Xamarin seem to prefer returning directory-unqualified assembly file name), or every implementation is going to have their own way to do this, breaking user code in a different way. There should be just a single way to do this in .NET Core.

accepted/single-file/design.md
```diff
@@ -0,0 +1,281 @@
...
+
+The goal of this effort is enable .Net-Core apps to be published and distributed as a single executable.
+
+There are several strategies to implement this feature -- ranging from bundling the published files into zip file (ex: [Warp](https://github.com/dgiagio/warp)), to native compiling and linking all the binaries together (ex: [CoreRT](https://github.com/dotnet/corert)). These options, along with their cost/benefit analysis is explored in this [staging document](staging.md).
+
+#### Goals
+
+In .Net Core 3.0, we plan to implement a solution that 
+
+* Is widely compatible: Apps containing MSIL assemblies, ready-to-run assemblies, native binaries, configuration files, etc. can be packaged into one executable.
```
poke: I feel like configuration shouldn’t be included in this list by default. If I think about a common use case for this single-file system, then I am thinking of easy-to-distribute utility applications that likely come with a default configuration setup and can be further configured by a file that’s next to the application. If the configuration file gets embedded by default, then there needs to be additional logic to support an external configuration file (outside of the bundle).
But if the configuration file does not get embedded, this can be simply handled as the file being optional (this would be very easy to implement with M.E.Configuration).

accepted/single-file/design.md
```diff
@@ -0,0 +1,281 @@
...
+
+To summarize, here's the overall experience for creating a HelloWorld single-file app 
+
+*  Create a new HelloWorld app: `HelloWorld$ dotnet new console`
+
+#### Framework Dependent HelloWorld
+
+* Normal publish: `dotnet publish` 
+
+  * Publish directory contains the host `HelloWorld.exe` ,  the app `HelloWorld.dll`, configuration files `HelloWorld.deps.json`, `HelloWorld.runtimeconfig.json`, and the PDB-file `HelloWorld.pdb`.
```
poke: Don’t framework-dependent publishes only include the app’s `.dll` and to run it, you have to run `dotnet HelloWorld.dll`? Where does this `.exe` come from?

accepted/single-file/design.md
```diff
@@ -0,0 +1,281 @@
...
+}
+```
+
+We can also provide an abstraction that abstracts away the physical location of a file (bundle or disk). For example, add a variant of  `GetFileStream` API that looks for a file in the bundle, and if not found, falls back to disk-lookup.
+
+#### Single-file Plugins
+
+The above design should be extended to seamlessly support single-file publish for plugins.
+
+- The bundler tool will mostly work as-is, regardless of whether we publish an application or class-lib. The binary blob with dependencies can be appended to both native and managed binaries.
```
poke: What happens when plugins require a common dependency that would already be there with the application? Would they have to bundle it again?
For example, let’s assume application `Foo.exe` uses a library `Foo.Common.dll`. Now, plugins can be built that use the `Foo.Common.dll`. The application then has some mechanism to pick these up.

With the classic approach, or even .NET Framework, the application might have exactly those two outputs: `Foo.exe` and `Foo.Common.dll`. And when you create the plugin, you might end up with `plugin.dll` and `Foo.Common.dll`. Now since the common library would already be bundled with the application, distributions of the plugin would only have to contain the `plugin.dll`. You could paste that into the application folder, enable the plugin, and it would work.

With .NET Core and single-file publishing, this would now mean that both the application and the plugin would bundle `Foo.Common.dll` and all the runtime dlls into their final output. So you have those twice for this setup. Of course, you could opt in to make the library _not_ bundle into a single file, but the plugin might also have some other dependencies, so it likely would _want_ to bundle into a single file. So in the end, this would require a lot of configuration for the plugin author to produce an efficient output.

Do you have any thoughts on this situation?

accepted/single-file/extract.md
```diff
@@ -0,0 +1,81 @@
...
+
+The proposed extraction path is `%TEMP\.net\<random-code>` / `$TMPDIR/.net/<random-code>` .
```
poke: Would this be configurable? I could imagine that some applications might rather have this be `%TEMP%\MyCompany\<random-code>`.

accepted/single-file/extract.md
```diff
@@ -0,0 +1,81 @@
...
+The requirements for persistent extraction are:
+
+* Reuse: Extract on first-run, reuse on subsequent runs.
+* Fault-tolerance: Recover from failure after partial extraction on first run.
+* Concurrency: Handle concurrent launches on first start.
+* Upgrade: Each version of the app extracts to a unique location, supporting side-by-side use of multiple versions.
+* Uninstall: Users can identify and delete extracted files when the app is no longer needed.
+* Access control: Processes running with elevated access can extract to admin-only-writable locations.
+
+At first startup, the host takes a disk lock (in order to handle concurrent launches), extracts files to an *install-location*, verifies them, and writes and *end-marker* before releasing the lock. If the host fails to extract all the resources at startup, it will attempt to clean up all contents of the install-location before termination.
```
poke: Typo: ”writes **an** *end-marker*”

accepted/single-file/extract.md
```diff
@@ -0,0 +1,81 @@
...
+By default, the host extracts dependencies to `BASE_PATH/APP_NAME/ID/PERM/`
+
+- `BASE_PATH` is  `%HOMEPATH%\.dotnet\AppDependencies ` /  `$HOME/.dotnet/AppDependencies`
+- `APP_NAME` is simple the name of the app (the host executable containing embedded data).
+- `ID` is a hash-code based on the host binary to distinguish app version and architecture specifications.
+- `PERM` is `user` for normal runs, and `admin` for elevated runs with appropriate write-permissions.
+
+A different extraction location may be specified via a `runtimeconfig` option.
+
+The cleanup of extracted files in the install-location will be manual in this version. We can consider adding `dotnet CLI` commands for cleanup in future. However, future versions are expected to spill fewer artifacts to disk, making the cleanup commands a lower priority feature.
```
poke: For self-contained apps, users are not expected to have a dotnet CLI, so there might be some other approach needed to clean it up. Would it be possible for the app to clean up itself, to implement some *“Exit and uninstall”* functionality into the application?

accepted/single-file/staging.md
```diff
@@ -0,0 +1,139 @@
...
+
+* A  packaging tool (bundler) that embeds the application and all of its dependencies (essentially the contents of the publish directory) into a host executable. 
+	* This version will not support compression of the bundled assemblies.
+* When the executable runs, it will extract all of those files into a temporary directory, and then run as though the app was published to that temporary directory.
+
+The mechanism will support:
+
+* Reuse: Extract on first-run, reuse on subsequent runs.
+* Upgrade: Each version of the app extracts to a unique location, supporting side-by-side use of multiple versions.
+* Uninstall: Users can identify and delete extracted files when the app is no longer needed.
```
poke: In extract.md, the temporary extraction was explained to clean up the temporary location on exit, so how does this work with this explanation where there the extracted files are reused?

And how exactly would users be able to *identify* no longer needed files? Especially when running an application through multiple versions, I would imagine that it would be quite difficult for users to identify which version they are running and which files they can delete (especially since this would still be a manual process).

accepted/single-file/staging.md
```diff
@@ -0,0 +1,139 @@
...
+## 2. Run from Bundle: MSIL
+
+### 2.1 Description
+
+This stage improves on Stage 1 in that 
+
+* MSIL files bundled into the single-file will load and execute directly from the executable.
+* Native libraries will still need to 
+	* Remain in the publish directory unmerged, or
+	* Extracted to the disk like the previous self-extractor stage
```
poke: What about an option that combines these two approaches? I.e. extract the native libraries to the current folder on first launch (and whenever they get deleted).

ayende:
WRT `Location`, what about `/home/me/app.bundle;app.dll` ? That would still maintain the get the base direction method, and it is also clear that this is something new.

ayende:
About config files, etc. 
I can tell you what we would do in this case. We currently have a `settings.default.json` file that we bundle with our distribution. At startup, we search first for `settings.json` in the base directory and copy the `settins.default.json` file.

We'll absolutely need that to be in the same directory as the bundled app for discovery to the user.

On the other hand, I can certainly see the need for throwing some data in such an bundled app somewhere that isn't clearly visible (`~/.myapp`) works great and there are already APIs for it.

accepted/single-file/design.md
jkotas: That was changed for .NET Core 3.0. You get `.exe` for framework dependent publishes by default in .NET Core 3.0.

accepted/single-file/design.md
jkotas: This problem of what dependencies to include with plugins is not specific to single .exe files. You are right that one needs to be aware what comes with the app vs. what needs be included with the plugin. 

Take a look at plugin sample https://github.com/dotnet/samples/tree/master/core/extensions/AppWithPlugin that shows how it can be done.

accepted/single-file/design.md
jkotas: The problem is how you define "configuration". What is configuration for you may not be a configuration for me. As mentioned elsewhere in the doc, you will be able to customize what to include in the bundle.

accepted/single-file/extract.md
swaroop-sridhar: The `.net` part just signifies that these are temporary files created by .net-core. This way, the entire `.net` folder can be easily cleaned up if necessary. We could, of course, have `%temp%\.net\app\...` etc. But we choose shorter file names -- so that we don;t hit the max-path limit.

accepted/single-file/design.md
swaroop-sridhar: I agree, the goal of this work is to just package the contents of the `publish` directory to a single file.  It shouldn't change the contents of the publish directory. The way to resolve missing dependencies for a plugin depends on it's configuration and isolation settings. 

For example, if we didn't package `Foo.Common.dll` with `plugin.dll`, and `plugin.dll` is loaded in it's own `AssemblyLoadContext`, then it could:
* Choose to not load `Foo.Common.dll` in its own ALC, in which case, the load-logic falls back to the default ALC (of `Foo.exe`) and loads `Foo.Common.dll` in that ALC.
* Choose to explicitly load `Foo.Common.dll` from the `Foo.exe` bundle, using `GetFileStream` API proposed [here](https://github.com/dotnet/designs/pull/52/files#diff-28dba553f5f12520afc2d90d1587ef56R185).

accepted/single-file/extract.md
swaroop-sridhar: You mean not cleanup on every exit, but on a specific uninstall run, right? Yes, the app-developers can implement this, because the install-location of extracted files is predictable.

accepted/single-file/design.md
swaroop-sridhar: By configuration, here I mean .net-core's configuration files which are recognized and handled by the .net-core host/runtime. All other files are data-files from the point of view of the runtime, and should be handled as such.

accepted/single-file/staging.md
swaroop-sridhar: Removal of persistent-extracted files will be manual in this version, as noted [here](https://github.com/dotnet/designs/pull/52/files#diff-5e93215a94e626495c10448a40fa4ccbR64). That is, user can look for the app in `$HOME/.dotnet/Apps/<AppName>` and delete the apps he doesn't want to use anymore.

Here, versioning is based on a hash-of the entire single-file (including assemblies, binary and data files, signatures etc). That is, if the app, or any of its dependencies changed when recompiled, we will extract to a new location. So, it is tricky to provide an accurate human-readable description generically for all apps.

However, we can provide some information to help this situation. For example, as part of the end-marker, we can write out the full name and assembly version of the app's main assembly (or version of all assemblies in the app).

accepted/single-file/design.md
jkotas: E.g. ASP.NET comes with bunch of its own configuration .json files that are just data-files from the point of the runtime.

I do not think the distinction between data files and configuration files matters much. All publishes stuff will get bundled (correct?) by default, it does not really matter what it is.

swaroop-sridhar: Yes that's correct, Whatever is in the publish directory will get bundled, other files can be included too, as you mentioned above. Thanks.

accepted/single-file/staging.md
swaroop-sridhar: This option can be achieved by using persistent-extraction, and setting the extraction-location as the current directory, using the configuration mentioned [here](https://github.com/dotnet/designs/pull/52/files#diff-5e93215a94e626495c10448a40fa4ccbR62).

accepted/single-file/design.md
poke: Oh cool, didn’t know that! So the `.exe` will then pick up the “correct” runtime if one is installed automatically? Do you happen to have a link for me where this process is described?

accepted/single-file/design.md
poke: > This problem of what dependencies to include with plugins is not specific to single .exe deployment.

The difference for non single file deployments is that you can simply diff what files would be added by your plugin and only distribute those.

> […] the load-logic falls back to the default ALC (of `Foo.exe`) and loads `Foo.Common.dll` in that ALC.

So it would work if you deploy the app as a single file and choose not to do that with the plugin (distributing only the `plugin.dll`), and the `plugin.dll` could then still (automatically) access the common library and (more importantly) the framework dependencies from the `Foo.exe`?

The only problem I would see if the plugin has additional dependencies and wants to do a single-file deployment of only those dependencies that are specific to the plugin (and aren’t already included in the app). But I guess this would work with customization of the bundling behavior, so all the framework assemblies are not bundled.

On that note, I could actually imagine just bundling the framework assemblies for every single self-contained application, so that there’s only the .exe, the concrete dependencies, and then maybe a `framework.dll`. – Would that work somehow or would that already be problematic because the executing assembly needs to be a host file that is able to unpack the files?

accepted/single-file/extract.md
poke: Yeah, I do understand the idea behind this but I could still see some situations where having a `.net` in the path would be unwanted. The fact that the application is a .NET application is an implementation detail after all.

accepted/single-file/extract.md
poke: Yeah, using a persistent location (so that the extracting only happens once) but at the same time offering an option _within_ the application to remove the extracted files at that location.

accepted/single-file/design.md
vitek-karas: You can find most of this process described [here](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/host-components.md)

accepted/single-file/extract.md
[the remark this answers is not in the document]
KathleenDollard: The sentence construction is confusing: "don't prefer...rather than". I do not know what you are saying. 

Can you restate this intent in different words.

accepted/single-file/staging.md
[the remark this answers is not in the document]
KathleenDollard: This is going to be a great state. The approach of this proposal allows a step-wise approach so we don't have to have everything ready before we provide anything.

accepted/single-file/design.md
KathleenDollard: @swaroop-sridhar is there a "not" missing from the second sentence? Is it that we cannot until we have cross-crossgen?

accepted/single-file/design.md
KathleenDollard: Is there just the single switch? 

Agree that we want some CLI love on this. Want to design it with the full richness we'll need later.

accepted/single-file/design.md
swaroop-sridhar: Yes that's a typo, sorry.

accepted/single-file/extract.md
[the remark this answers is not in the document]
swaroop-sridhar: Sorry I'll fix the typo.

accepted/single-file/design.md
swaroop-sridhar: @nguerrera I've updated the doc now to use `IncludeInSingleFilePublish` metadata.

swaroop-sridhar:
@MichalStrehovsky I added the discussion about `Assembly.Location` [here](https://github.com/dotnet/designs/pull/52/files#diff-28dba553f5f12520afc2d90d1587ef56R211).
I think `Assembly.Codebase` can remain unchanged for bundled or normal assemblies.

accepted/single-file/design.md
swaroop-sridhar: @poke : A few points to clarify:
* The plugin and app should be developed with a proper expectation of each other's dependencies (native binaries and assemblies). 
* If diffing published files is desirable, we still have the option to publish without the single-file option. The dependencies should not be different whether we publish single-file or not.
* Framework is typically never packaged with plugins.
* The runtime will provide call-backs extract file embedded within plugins. We may be able to refactor the host code such that the callbacks may reuse that code for handling single-file plugins.

accepted/single-file/design.md
```diff
@@ -0,0 +1,314 @@
...
+  * If reuse of extracted files is enabled, subsequent runs of the app may skip the extraction step.
+
+
+Most applications are expected to work without any changes. However, apps with a strong expectation about absolute location of dependent files may need to be made aware of bundling and extraction aspects of single-file publishing. No difference is expected with respect to debugging and analysis of apps.
+
+## Related Work
+
+* [Mono/MkBundle](https://github.com/mono/mono/blob/master/mcs/tools/mkbundle) 
+* [Fody.Costura](https://github.com/Fody/Costura)
+* [Warp](https://github.com/dgiagio/warp)
```
jkotas: Also: [dotnet-warp](https://github.com/Hubert-Rybak/dotnet-warp) global tool

(mentioned in https://www.hanselman.com/blog/BrainstormingCreatingASmallSingleSelfcontainedExecutableOutOfANETCoreApplication.aspx)

NSentinel:
It may make sense to look at the https://github.com/MiloszKrajewski/LibZ design with similar goals for the .NET Framework.

We have used it for a long time without any issues  in many projects (thanks to @MiloszKrajewski)

swaroop-sridhar:
Thanks for the reference @NSentinel.  
The core of LibZ is to embed assemblies as resources, and instrumenting the app to use resolvers to load the assemblies from resources. LibZ also adds a container abstraction to reduce assembly size. 
I'll add LibZ to the list of related work. Thanks.

cup:
https://github.com/dotnet/cli/issues/7737
https://github.com/dotnet/core/issues/1278
https://github.com/dotnet/core/issues/2647
https://github.com/dotnet/coreclr/issues/17015
https://github.com/dotnet/coreclr/issues/20287
https://github.com/dotnet/corefx/issues/13329
https://github.com/dotnet/corert/issues/6819
https://github.com/dotnet/core-setup/pull/5286

https://github.com/dotnet/designs/pull/52
https://github.com/dotnet/sdk/pull/3074
https://github.com/dotnet/sdk/pull/3132

jjxtra:
Looks like single file and razor is broken until this is resolved

PR #60, opened 2019-04-03
Add design proposal for better obsoletion

terrajobst:


terrajobst:
Not ready yet.

PathogenDavid:
I know you said it's not ready yet, but I read it anyway 😬

I feel like the majority of projects using this would have the same `UrlFormat` for every `Obsolete` attribute across the entire project. It'd be nice if the default value for `UrlFormat` could be applied at the module level. Something like

```csharp
[module: DefaultObsoleteUrl(UrlFormat="https://aka.ms/obsolete/{0}")]
```

As an added bonus, this attribute could be injected by the build system. This would allow libraries hosting their documentation out of source control to link to specific versions of the documentation. (IE: Use `https://github.com/dotnet/platform-compat/blob/532b760d5e2016ef84696dfd2f0132eb04693651/docs/DE0001.md` instead of `https://github.com/dotnet/platform-compat/blob/master/docs/DE0001.md`)

terrajobst:
If you're insisting, you can comment here: https://github.com/terrajobst/designs/pull/1 😄

PathogenDavid:
Done! I hope you don't mind I left some other comments that I thought of since that one.

terrajobst:
Not at all! As long as you don't mind me disagreeing 😉

PathogenDavid:
Disagree away!

PR #62, opened 2019-04-05
Better Obsoletion

terrajobst:


terrajobst:
> You should probably address browser visiblity for hiding the api from intellisense.

What do you mean? Presumably, you believe the existing mechanism isn't good enough?

terrajobst:
> would it be a goal or non-goal to mark APIs which were never shipped in stable packages as experimental? Or even wider - have an `Experimental` attribute which is similar to `Obsolete` but works in opposite way. You are marking something as non-committal. i.e. library haven't committed on yet and it may be gone when library will move from `-beta` to `-stable`.

We have only briefly talked about that. It's an interesting idea that might make it easier for us to ship previews and inform customers what's not stable yet. However, this warrants more thought and I'll consider this out-of-scope for the discussion of obsoletion.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+
+Abigail would like to use the brand new JSON APIs so she updates the project
+from .NET Core 2.1 to .NET Core 3.1. However, after upgrading the notices that
+we obsoleted non-generic collections and she gets a bunch of warnings against
+her code:
+
+| Code    | Description              | Project                        |
+|---------|--------------------------|--------------------------------|
+| BCL0006 | 'ArrayList' is obsolete  | Fabrikam.Billing.BusinessLogic |
+| BCL0006 | 'Hashtable' is obsolete  | Fabrikam.Billing.BusinessLogic |
```
jkotas: FWIW, Hashtable is a perfectly good type that does not have direct equivalent anywhere else in the framework when it comes to multi-threading. We should stop trying to obsolete it. It does not work. Replacing it with something else just creates bugs.

`BinaryFormatter` is a good example of a type to obsolete.

MeikTranel:
I agree to almost anything said. That said.. while agreeing with the idea that breaking the obsolete attribute is madness, i sure hope you are really sure about the naming of the DeprecatedAttribute.

The word obsolete and deprecated may be equivalently used in the english language... but to the best of my knowledge in other languages such as the german language "obsolete" implies that something becomes no longer required, while deprecation implies a plea not to use.

I am by no means an expert on languages so please feel free to chime in non-english speakers. I am also by no means asking you not to design english-first. It's just a consideration worth taking in imho

accepted/better-obsoletion/better-obsoletion.md
terrajobst: Sure, we can discuss that when we actually apply the attribute. I've no strong opinions on `Hashtable`. But duly noted 😄

terrajobst:
> I agree to almost anything said. That said.. while agreeing with the idea that breaking the obsolete attribute is madness, i sure hope you are really sure about the naming of the DeprecatedAttribute.

If we would be starting from scratch, I'd push for the name of `DeprecatedAttribute` because it does indeed sound stronger than deprecated. At the same time, obsolete is a well understood concept in .NET and is already recognized by the compiler. I don't think inventing a new concept here is worth the cost.

PathogenDavid:
> "obsolete" implies that something becomes no longer required, while deprecation implies a plea not to use.

FWIW, this is true in English as well. It's software developers that tend to use them interchangeably.

MeikTranel:
Maybe we could still have the best of both worlds by allowing both URL or Message in the constructor. We could tag the message attribute as obsolete or leave it (not every library author has the time or resources to host a URL for error codes) and defaulting the root URL to msdocs and the error code to the generic obsolete diagnosis. That way we don't have any massive breaking changes but still allow for a better system to take it's place.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+```
+
+Instead of taking the URL directly, the API takes a format string. This allows
+having a generic URL that includes the diagnostic ID. This avoids having to
+repeat the ID twice and thus making a copy & paste mistake.
+
+### Compiler Changes
+
```
KathleenDollard: Maybe you are talking in more detail elsewhere, but doesn't the compiler now have two jobs:

- Recognize to ignore (or set to error) by DiagnosticId
- Create a message for the warning

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+#### String Encoding
+
+Instead of introducing new APIs, we could invent a textual format that we put
+in the attribute's message property, such as:
+
+```C#
+namespace System.Collections
+{
+    [Obsolete("::BCL0006|http://aka.ms/obsolete/{0}")]
```
KathleenDollard: This would be great if we were just lining up help. But if we are also suppressing warnings on this, I think typos would be too serious to use a pattern based approach.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+Here is how the attribute would be applied:
+
+```C#
+internal static class Obsolete
+{
+    public const string Url = "http://aka.ms/obsolete/{0}";
+
+    // NOTE: Keep this sorted by the the ID's value,
+    //       not the name of the field!
+    public const string SecureStringId = "BCL0001";
```
richlander: I was surprised to see all BCL code being exposed (A) in one place, and (B) on the Obsolete type itself. I think a better model would be to have no codes exposed on this type whatsoever. Your goal is to establishing a naming scheme, which is good, but this design is missing an opportunity to communicate that. Instead, we could adopt the pattern of exposing types called [component]"Obsolete", like "BclObjsolete", "SystemObsolete" or "AspNetCoreObsolete".  We'd create such a type now with whatever name we liked.

I'm hoping we can come up with a better name than "BCL" since it doesn't show up anywhere in the platform. I'd prefer "SystemObsolete" but that is incredibly close to "System.Obsolete". Another option is "CoreFXObsolete", tying it to repo name. That may be a bad idea for other reasons, but it would be easy to communicate and for developers to grok / implement.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+    public const string NonGenericCollectionsId = "BCL0006";
+}
+
+namespace System.Collections
+{
+    [Obsolete(DiagnosticId = Obsolete.NonGenericCollectionsId, UrlFormat=Obsolete.Url)]
```
richlander: Any reason why `UrlFormat=Obsolete.Url` cannot be the default?

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+}
+```
+
+The benefit is that it doesn't require new API and that that the call sites
+are slightly more compact.
+
+The downside is more cryptic use and weak contract between the compiler and the
+API surface, which might also result in accidental recognition.
+
+#### Separate Attribute
```
KathleenDollard: No, just no.

Do we have to talk about this?

accepted/better-obsoletion/better-obsoletion.md
PathogenDavid: > on the Obsolete type itself.

Note that this is a different internal type, *not* `ObsoleteAttribute`.

The idea is to store the IDs and URL in a central location on a per-library basis to prevent typos and allow changing the URL from one place. (Basically this class addresses [my concerns on the earlier PR](https://github.com/dotnet/designs/pull/60#issuecomment-479606367).)

terrajobst: > I was surprised to see all BCL code being exposed (A) in one place

I only did this to showcase internal plumbing to help us centralize the diagnostic IDs. This wouldn't be exposed to the consumer.

accepted/better-obsoletion/better-obsoletion.md
terrajobst: Bad naming on my end. These are different types. Think of `UrlFormat=Obsolete.Url` as `UrlFormat=MyLibrariesObsolete.UrlConstance`

accepted/better-obsoletion/better-obsoletion.md
terrajobst: I agree. I wanted to call out why these are bad ideas because it was suggested before.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+the code is well-tested. Last year, the code was ported to .NET Core 2.1.
+
+The component is designed for a traditional two-tier environment and includes
+data access that handles retrieving and storing business objects from and to SQL
+Server. Abigail is now tasked to refactor that code to use REST-based services
+instead so that the application can eventually be converted to a cloud-based
+offering.
+
+Abigail would like to use the brand new JSON APIs so she updates the project
+from .NET Core 2.1 to .NET Core 3.1. However, after upgrading the notices that
```
ahsonkhan: `after upgrading the notices that` <= typo

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+  choice and require at least a page of explanation, rather than a one-liner. 
+
+This document proposes changes to how APIs can be obsoleted in a more structured
+way.
+
+## Scenarios and User Experience
+
+### Targeted Suppression
+
+Abigail is a developer and maintains the business logic layer component for
```
ahsonkhan: Presumably this is a fictitious example, correct? Add a preamble to clarify that.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+        public string DiagnosticId { get; set; }
+        public string UrlFormat { get; set; }
+        public string Url => UrlFormat == null ? null : string.Format(UrlFormat, DiagnosticId);
+    }
+}
+```
+
+The irony is that this would effectively mean that we're obsoleting the
+`ObsoleteAttribute` which now begs the question whether we should mark it
+`[Obsolete]` or `[Deprecated]`.
```
ahsonkhan: Lol

accepted/better-obsoletion/better-obsoletion.md
terrajobst: All specs follow that pattern. The fact that you’ve never seen this makes me sad because it means we no longer write proper specs 😬

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+{
+    public sealed class ObsoleteAttribute : Attribute
+    {
+        // Existing:
+        //
+        // public ObsoleteAttribute();
+        // public ObsoleteAttribute(string message);
+        // public ObsoleteAttribute(string message, bool error);
+        // public bool IsError { get; }
+        // public string Message { get; }
```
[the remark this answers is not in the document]
terrajobst: That behavior would be unchanged. The compiler emits a message. For C#:

CS0612: '\<API>`' is obsolete

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+having a generic URL that includes the diagnostic ID. This avoids having to
+repeat the ID twice and thus making a copy & paste mistake.
+
+### Compiler Changes
+
+The compilers would be enlightened about these new properties and use them as
+follows:
+
+* If `DiagnosticId` is `null`, use the existing diagnostic ID (e.g. `CS0618` in
+  C#). Otherwise, use the specified warning ID.
```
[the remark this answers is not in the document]
terrajobst: Yes. Fixed.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,345 @@
...
+* If `Url` is not `null`, use it as the diagnostic link when rendering them in
```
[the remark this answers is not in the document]
terrajobst: How would the compiler find that? Would you expect the compiler to search all APIs in that assembly for applications of `[Obsolete]` with the same warning ID? That feels wrong to me. @jaredpar?

accepted/better-obsoletion/better-obsoletion.md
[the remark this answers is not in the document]
terrajobst: Not sure what you mean. Are you asking whether we'd ever want to specify more than one URL/diagnostic ID? If so I'd say no.

accepted/better-obsoletion/better-obsoletion.md
terrajobst: Clarified.

accepted/better-obsoletion/better-obsoletion.md
terrajobst: No :-) I've only listed it because folks have asked. I wanted to record why they are bad ideas.

accepted/better-obsoletion/better-obsoletion.md
terrajobst: Agreed. I've renamed the section from `Possible Alternatives` to `Rejected Alternatives` to make it clear that I think these are bad ideas.

accepted/better-obsoletion/better-obsoletion.md
ahsonkhan: Or, I haven't read enough of them :)

Good to know.

terrajobst:
@KathleenDollard 

> Could you still suppress all Obsolete attributed things - new style and old style?

Good point. I've added an open issue with 6b146af.

> If we are touching this attribute, should we consider anything else - like a level or reason. Or is everything the same level of STOP, FIX THIS RIGHT NOW

Hmm, I don't think that's super useful at the level of the diagnostic. This should more more in the linked documentation page IMHO.

ericstj:
> @ericstj
> 
> > You should probably address browser visiblity for hiding the api from intellisense.
> 
> What do you mean? Presumably, you believe the existing mechanism isn't good enough?

Not necessarily, but I didn't notice any mention of it in the doc and it seemed relevant to what you were discussing.  EG: explain why just hiding the API from intellisense isn't good enough.

pascalberger:
> Why is providing coder fixers not a goal?

While I agree with the reasons listed to not provide code fixes, I still think it would be a nice opt-in feature for cases where there is a 1:1 replacement. Instead of not providing code fixes I would prefer to have the possibility to define a replacement member and let the author of the obsolete attribute decide if a specific case is an appropriate use case for a code fix or not.

poke:
@pascalberger
>  I would prefer to have the possibility to define a replacement member

As mentioned in the document, for most obsoletions, just finding a replacement that is mostly compatible will not work. It is a lot more likely that migrating to a newer functionality requires non-trivial changes throughout the code base. A fixer that only operates locally wouldn't really be helpful there.

terrajobst:
@ericstj 

> Not necessarily, but I didn't notice any mention of it in the doc and it seemed relevant to what you were discussing. EG: explain why just hiding the API from intellisense isn't good enough to obsolete an API.

Got it. I've clarified this in 74707d7.

@pascalberger 

> While I agree with the reasons listed to not provide code fixes, I still think it would be a nice opt-in feature for cases where there is a 1:1 replacement. Instead of not providing code fixes I would prefer to have the possibility to define a replacement member and let the author of the obsolete attribute decide if a specific case is an appropriate use case for a code fix or not.

Code fixers are an independent feature. On case by case basis, we can decide to add one if we want. It's just not in scope of this feature to enable that, i.e. I don't see a point in recording the alternative member as part of the obsolete API. The cases where a single one exists are rare and docs are better at explaining this. That being said, the fact that we're adding specific diagnostic IDs also make it easier for code fix providers to scan for violations without having to implement if the analysis from scratch.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,373 @@
...
+
+Obsoletion is a process that informs customers about problematic APIs. It's most
+useful for first time users of the API so that they don't take dependencies on
+less optimal technology. However, it's also a powerful tool for informing
+customers about potential issues they have in their code base. Neither of these
+goals require removing the API eventually.
+
+### Should obsoleted members be hidden from IntelliSense?
+
+No always. Using `[EditorBrowsable(EditorBrowsableState.Never)]` one can hide
```
PathogenDavid: "No always" typo.

rainersigwald:
One of the major concerns with the existing attribute (and the reason we haven't applied it in MSBuild) is that adding _any_ warning is a build-breaking change for users who have warnings-as-errors. What's the plan to resolve that problem in the core libraries?

terrajobst:
> One of the major concerns with the existing attribute (and the reason we haven't applied it in MSBuild) is that adding _any_ warning is a build-breaking change for users who have warnings-as-errors. What's the plan to resolve that problem in the core libraries?

.NET Core has a lower bar and we'd accept this. The entire proposal is to alow customers to apply supressions in a more targeted fashion. By definition, obsoletion is about introdcing new diagnostics. However, they only kick-in when people target a later version of the platform. My position is that this is an acceptable breaking change, especially for customers that turn on "warnings as errors" because they asked to be brokne.

accepted/better-obsoletion/better-obsoletion.md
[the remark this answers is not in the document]
eharshosh: If there is no good reason not to, why not make the properties virtual and un-seal the class?
This would allow users to simply inherit from the attribute and provide fixed properties.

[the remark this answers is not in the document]
PathogenDavid: The compiler needs to be able to figure out the values of all of the properties at compile-time. Allowing inheritance would mean that the compiler has to be able to instantiate the attribute at compile-time and read the values of the properties.

svick:
This might be only tangentially related, but it seems it's important that people will be able to reach the relevant documentation page. Will they?

First of all, if they use `dotnet build`, I don't think there is anything to indicate that such a page exists.

And even if they use VS, I think the UI is quite subtle ("light blue color in the Code column means I can click it?") and when they do figure that out, they might be discouraged from clicking by previous bad experiences (e.g. https://github.com/MicrosoftDocs/feedback/issues/1303).

I'm not sure if there's anything that can be done here, but I think it's worth thinking about.

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,373 @@
...
+entirely.*
+
+### Samples
+
+Here is how the attribute would be applied:
+
+```C#
+internal static class BclObsolete
+{
+    public const string Url = "http://aka.ms/obsolete/{0}";
```
KexyBiscuit: Should all the Urls be secured(`https://`)?

accepted/better-obsoletion/better-obsoletion.md
```diff
@@ -0,0 +1,373 @@
...
+
+* If `DiagnosticId` is `null`, use the existing diagnostic ID (e.g. `CS0618` in
+  C#). Otherwise, use `DiagnosticId`.
+    - The compiler should honor suppressions for the specified diagnostic id.
+* If `Url` is not `null`, use it as the diagnostic link when rendering them in
+  the IDE.
+* The compiler should assume that `Url` and `DiagnosticId` are independent
+  features, in other words both can be used, either, and neither.
+
+***OPEN ISSUE**: Should the compiler suppress all diagnostics from
```
mairaw: Your formatting since off here making the rest of the text all bold in preview. Perhaps:

```suggestion
**OPEN ISSUE:** *Should the compiler suppress all diagnostics from
```

accepted/better-obsoletion/better-obsoletion.md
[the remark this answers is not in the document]
terrajobst: Right.

accepted/better-obsoletion/better-obsoletion.md
terrajobst: Fixed

accepted/better-obsoletion/better-obsoletion.md
terrajobst: Fixed

terrajobst:
@svick 

> This might be only tangentially related, but it seems it's important that people will be able to reach the relevant documentation page. Will they?
> 
> First of all, if they use `dotnet build`, I don't think there is anything in the error message to indicate that such a page exists.
> 
> And even if they use VS, I think the UI is quite subtle ("light blue color in the Code column means I can click it?") and when they do figure that out, they might be discouraged from clicking by previous bad experiences (e.g. [MicrosoftDocs/feedback#1303](https://github.com/MicrosoftDocs/feedback/issues/1303) or when the link just sends you to Bing).
> 
> I'm not sure if there's anything that can be done here, but I think it's worth thinking about.

That's a fair point. I don't think including the URL when printing diagnostics on the command line would be bad, but I also don't think it will be super useful either. The CI build isn't the right place to find out that you're using something obsoleted, that should be when you're doing development in the inner loop. And while there are some people who do that entirely using a text editor and `dotnet build`, the 87.3412% of developers do that from inside an IDE (yes, I made that number up).

I agree with you that the IDE affordances for navigating to the provided URL seem too subtle. One thing I've been doing in my own analyzer is adding a code fixer like `About DE002...` which opens the browser. Having a generalized actions to promote this inside the the IDE would be useful. I've filed https://github.com/dotnet/roslyn/issues/42103 for this.

terrajobst:
While there are still some open design issues, I'll merge this because we agreed that we'll tackle this for .NET 5. Thanks everyone for the feedback:

Remaining issues (listed in the doc too):

* Should the compiler suppress all diagnostics from `ObsoleteAttribute` if the existing generic diagnostic ID is suppressed (e.g. `CS0618`)? If not, the developer has no way to turn this feature off entirely.
* We need to decide on guidance to avoid ID conflicts with the community. We should talk to the Roslyn analyzer team for best practices. Stake in the ground could be that we should use a prefix like `BCL` that is owned by a single party and responsible to avoid duplicated numbers. Other parties would need to choose a different prefix.
* Close the loop with XML Serializer team and lock down what we should do.

PR #65, opened 2019-04-20
Proposal for architecture specific apps

richlander:


proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+* With this build type, `--self-contained` is `false` by default, even when a RID is specified. `--self-contained` or `--self-contained true` must be specified to produce a self-contained build.
+
+WPF and Windows Forms templates should opt-in to this build type, with the following pattern:
+
+```xml
+<ArchitectureSpecific>true</ArchitectureSpecific>
+<RuntimeIdentifier>win-x86</RuntimeIdentifier>
+```
+
+The `win-x86` RID is proposed as the default to match the 32-bit default for .NET Framework projects. There is a perf benefit and it will be more compatible with existing .NET Framework projects that may have 32-bit only dependencies. It does mean that we need to install the Windows 32-bit .NET Core runtime by default (the 32-bit SDK is not needed). Visual Studio should make it straightforward to switch the default RID to `win-x64`.
```
tannergooding: There is also the twitter thread on this (https://twitter.com/runfaster2000/status/1119401606645080064), but I'm replying here since 280 characters isn't enough 😄.

**TL;DR;** I would like to see some numbers showing that there is a perf benefit (for a range of application types) to staying in 32-bit mode by default, as this is not my expectation. I would also like to see memory usage numbers and for us to determine if that is because of actually higher memory usage or because of more liberal settings in the JIT/GC that are taking advantage of the increased address space.

I will always buy that you can see less memory usage by running 32-bit. Pointers are half the size and most other things remain the same, so 64-bit applications will naturally see some increased memory usage. As compared to a native app, this can be slightly worse on .NET since we have overhead of the various information the JIT and GC have to track; in addition to user types. However, I don't believe that in common applications this is particularly egregious. I also do not buy that there is a perf benefit to remaining on a 32-bit application.

Every architecture/platform has an ABI (Application Binary Interface) that defines things like how parameters are passed between methods, which CPU registers must be saved by the caller/callee method, default alignment of resources, etc. 32-bit has been around a lot longer and CPUs had a lot less resources when the ABI was defined. This means that you have more scenarios under which values need to be stored and retrieved from memory in order to preserve these limited CPU resources (namely CPU registers). It also means that, since not every CPU can be guaranteed to have these newer resources, sometimes tradeoffs have to happen when passing values around. An example of this is floating-point values being returned on the x87 FPU stack; even if modern apps never use it otherwise. It also means that operating on data that is larger than 32-bit can be more complicated and require more instructions. x64 CPUs on the other hand brought a new baseline to the table in terms of available CPU resources, instructions that were available, etc. One of the primary benefits is that it doubled the number of available integer and floating-point registers and made SSE/SSE2 instructions a baseline.

On 32-bit Windows, the default calling convention is `__cdecl` for regular C/C++ code and `__stdcall` for most Win32 APIs. Due to the limited number of registers, all values are passed by pushing them onto the stack (to memory). This means that you have a forced memory write and read for every parameter you pass (and memory access, even cached, is not "cheap"). Values are returned in register when they are less than 8-bytes and as a hidden pointer to memory otherwise. For value types, this commonly means a shadow copy happens (the value type is copied to a location in memory, and a pointer to that location is passed to the callee). For reference types, this means you have a double indirection to lookup the contents. For floating-point values, these are returned on the x87 FPU stack, which isn't otherwise used in modern applications and generally requires additional work to get it back into the SIMD registers that are used. There are also other various things about the 32-bit ABI that make it less than ideal. You only have three registers to save values in across calls, otherwise values you need to preserve also have to be spilled to memory so the callee method doesn't overwrite them. Also, anytime you need to work with 64-bit data (`System.Int64` or `System.UInt64`) you have to take up two registers to hold the data, you have to do more operations to compute the result, and you you have to work harder to allow atomic operations when multithreading is important.

On 64-bit Windows, the default calling convention is generally referred to as "fastcall" and it takes advantage of the additional CPU resources that were introduced. Now, rather than passing values by storing them to memory, it will pass the first four arguments in register (completely eliding the need for a memory access). Value types larger than 8-bytes still need to be shadow-copied and floating-point values are passed and returned directly in the SIMD registers rather than the x87 FPU stack. There are now 3 more integer registers and 10 more floating-point registers to save values in that the callee won't overwrite and the callee itself has 4 more integer registers in which to use, allowing it to spill less stuff to memory. Working with 64-bit data is also much simpler, as everything becomes a single operation and atomicity is easier to provide.

Overall, this means that even though you might be taking up slightly more memory in your app; common scenarios can involve less overall memory access. Memory access itself is expensive (compared to register access; and is cheap compared to disk IO, which is cheap compared to network IO), especially when you have a cache miss, so reducing the number of times you have to read/write memory can be important. An example of this is if you were to take something like: https://github.com/tannergooding/demoapp, which is a .NET Core 3.0 and WPF-based demo app that renders 3D models. If you run this app in 32-bit vs 64-bit mode you'll see that, for something like the `landscape` model, there is a ~21% increase in memory (going from `~150mb` to `~183mb`) but a 41% perf increase (going from `~50fps` to `~70fps` average); I wrote this app for the MVP summit to demo hardware intrinsics and tested on a 4k monitor and an AMD Ryzen 1800X machine.

Likewise, I believe Roslyn underwent various testing/benchmarking and it was found that there was no perf degredation from running the compiler in 32-bit vs 64-bit mode. I would expect that many other applications will see similar (either a perf increase or no perf degredation due to being run in 64-bit mode). Many native and UWP apps on my box are 64-bit by default and show similar results when compared to their 32-bit counterparts.

jkotas: The overall performance will be better on 64-bit for some apps and it will be better on 32-bit for other apps. It is probably close to a wash on average, depends on your benchmark mix. Performance is not going to help to decide this.

I think we should be looking at the other factors to decide the default:

- 64-bit is more secure, e.g. much better address space randomization, other new OS and hardware security features may be 64-bit only. 
- 64-bit is conceptually simpler. You do not have to think about whether 32-bit is enough for your app.
- The industry is on path to drop and deprecate 32-bit support.
- 32-bit is more compatible with existing .NET Framework default.

Looking at this list, I would think 64-bit would be better forward-looking default.

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+* Only architecture-specific (including self-contained) applications should have EXEs.
+* The default for `--self-contained` is bad for usability. We should fix that for 3.0+ projects.
+
+Note: The term *architecture* is used throughout this document. It is intended to mean the unique combination of an operating system and CPU, like Windows X64, or Linux ARM64. It is equivalent to runtime identifier (RID).
+
+Note: There are separate challenges on building libraries that have native dependencies. That issue is not covered in this document, but is a problem that will be discussed in a following document.
+
+## Current Behavior
+
+The default `dotnet build` behavior in .NET Core 3.0 Preview 4 is the following:
```
KathleenDollard: Can you clarify whether this is also the .NET Core 2.2 behavior in the doc. IOW, is this a preview over preview or release over release change.

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+`dotnet build` should have three modes:
+
+* Architecture-agnostic, framework-dependent, application that includes native dependencies for all supported architectures and no EXE
+* Achitecture-specific, framework-dependent, application that includes an EXE and native dependencies for a single targeted architecture
+* Architecture-specific, self-contained, application that includes EXE and native dependencies for a single targeted architecture
+
+The proposal is:
+
+* Retain the first behavior as the default.
+* Enable the second behavior as opt-in. Enable this behavior in Windows client application templates.
```
KathleenDollard: If I understand your intent, change "Enable" to "Default to"

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+Hello World!
+```
+
+Notice the inclusion of native dependencies for Linux, macOS and two Windows architectures. At the same time, the application includes an executable that will only run on Windows x64. If the application is launched with `dotnet cpumath.dll`, it will then run on all of the architectures supported by the native dependencies.
+
+## Proposed Behavior
+
+`dotnet build` should have three modes:
+
+* Architecture-agnostic, framework-dependent, application that includes native dependencies for all supported architectures and no EXE
```
KathleenDollard: Are there any development scenarios where missing the EXE would be a detriment?

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+* Achitecture-specific, framework-dependent, application that includes an EXE and native dependencies for a single targeted architecture
+* Architecture-specific, self-contained, application that includes EXE and native dependencies for a single targeted architecture
+
+The proposal is:
+
+* Retain the first behavior as the default.
+* Enable the second behavior as opt-in. Enable this behavior in Windows client application templates.
+* Maintain the third behavior as opt-in (change defaults for `-r` and `--self-contained` for 3.0+).
+
+The first behavior is most similar to the .NET Core 2.2 behavior. It is a good behavior because it builds an application with no assumptions and produces artifacts with the broadest scope and utility. Many developers will want the second behavior. It needs to be very easy to opt-in to that behavior, either via the CLI or with a single property in a template.
```
KathleenDollard: Maybe clarify that the template is doing a simple change in the `csproj` (if I understand you), because users can also do that to existing apps.

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+The following are more specific goals to improve CLI usability:
+
+* `build` and `publish` should have the same capability, except where there is a good scenario reason why a task should only be made available via `publish`.
+* Defaults should be the same between `build` and `publish`.
+* All common scenarios should work without specifying a RID if you are happy targeting the current machine.
+* We should strongly avoid any breaking changes.
+
+### Proposed Application Types
+
+For the purposes of this document, there are three types of applications:
```
KathleenDollard: This feels repetitive from above and you are already adding details below. Maybe a table or sub bullets would make it easier to see how these three align

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+#### Framework-dependent architecture-netural applications
+
+Maintain framework-dependent architecture-netural applications as the default for `build`.
+  
+* Produces an application with native dependencies for all supported architectures.
+* Remove the capability to produce EXEs for this type of build.
+* If EXEs are desired for this build type, consider for a later release. It's a hard scenario to get right due to filename conflicts.
+
+ASP.NET and console templates should accept this default.
```
KathleenDollard: This is important. Suggest:

The templates for ASP.NET and console applications will use this as the default.

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+#### Framework-dependent architecture-specific applications
+
+Expose a new opt-in build type, called architecture-specific framework-dependent.
+
+* Produces an application with native dependencies and an EXE for a single specified architecture.
+* This build type can be selected via
+  * MSBuild: `<ArchitectureSpecific>true</ArchitectureSpecific>`
+  * CLI: `--architecture-specific` or `-a`
+* A RID must be specified (or it is an error). Any of the following options is legal to specify a RID:
```
KathleenDollard: Why wouldn't we use the current OS to find a default RID?

proposed/building-arch-specific-applications.md
KathleenDollard: If it's 32 bit, we will be explaining it forever. And that default assumes the perf difference is both measurable and permanent.

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+The `win-x86` RID is proposed as the default to match the 32-bit default for .NET Framework projects. There is a perf benefit and it will be more compatible with existing .NET Framework projects that may have 32-bit only dependencies. It does mean that we need to install the Windows 32-bit .NET Core runtime by default (the 32-bit SDK is not needed). Visual Studio should make it straightforward to switch the default RID to `win-x64`.
+
+Note: It is possible that the RID should be specified a different way that better aligns with the Visual Studio configutation manager. This document doesn't take an opinion on that. it is a detail to work out.
+
+Alternative idea for specifying architecture:
+
+* MSBuild: `<ArchitectureSpecific>RID</ArchitectureSpecific>`
+
+This alternative idea requires one instead of two lines, but also replaces the existing `RuntimeIdentifier` property, which can be viewed as good or bad.
```
KathleenDollard: Bad, imo.

RID is sensible for a few scenarios. Something special for this scenario is not progress.

proposed/building-arch-specific-applications.md
nguerrera: It's worth noting that this is proposed as a template default, not a hidden magic choice in the build. I think that's great and it means there's no "forever" element to the choice here. We can always change templates if circumstances change.

jnm2: The explanation of "why is it on me to remember to make this explicit" would be the only forever element.

jkotas: > proposed as a template default

Does that mean that the following is no longer going to work?

1. Download 64-bit .NET Core SDK
2. dotnet new console
3. dotnet run

If the template defaults to 32-bit, is this going to fail with "you need to install 32-bit runtime" error?

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+Note: There are separate challenges on building libraries that have native dependencies. That issue is not covered in this document, but is a problem that will be discussed in a following document.
+
+## Current Behavior
+
+The default `dotnet build` behavior in .NET Core 3.0 Preview 4 is the following:
+
+* Framework-dependent application
+* Architecture-specific executable (matches architecture of the SDK )
+* Native dependencies for all achitectures
```
Lakritzator:
```suggestion
* Native dependencies for all architectures
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+This combination of characteristics, particularly the last two points, isn't entirely coherent. It gives you an executable for one architecture, but native dependencies for all achitectures. That means that the application will run on all architecture supported by the native dependencies (if any) but that an executable is provided for only one architecture and the `dotnet` launcher needs to be used on all other architectures.
```
Lakritzator:
```suggestion
This combination of characteristics, particularly the last two points, isn't entirely coherent. It gives you an executable for one architecture, but native dependencies for all architectures. That means that the application will run on all architecture supported by the native dependencies (if any) but that an executable is provided for only one architecture and the `dotnet` launcher needs to be used on all other architectures.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+Many .NET Framework developers unfamiliar with .NET Core will start using the newer environment with .NET Core 3.0 and will expect a similar experience as they have been used to, particularly for those that only ever intend to target Windows with Windows Forms and WPF applications. Very few of them will expect Linux or macOS binaries in their build output. We need to deliver an experience they expect.
+
+### Goals
+
+The following broad goals should be satisfied for building applications with native dependencies:
+
+* .NET CLI defaults (for architecture and native dependencies) are intuitive and align with key user scenarios.
+* .NET CLI opt-in arguments (for architecture and native dependencies) have good usability for key user scenarios, with a bias to short and memorable arguments and good defaults for those arguments.
+* It is straightforward to transform CLI opt-in arugments (for architecture and native dependencies) that enable your project use-case into msbuild properties in your project file, resulting in a custom default for that project.
```
Lakritzator:
```suggestion
* It is straightforward to transform CLI opt-in arguments (for architecture and native dependencies) that enable your project use-case into msbuild properties in your project file, resulting in a custom default for that project.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+* It is straightforward to produce deterministic builds, independent of OS or architecture of the build machine.
+* It is easy and intutive to share applications with others or deploy them to other machines (like a Rasperry Pi, or a Windows ARM64 laptop) as part of a fluid workflow.
```
Lakritzator:
```suggestion
* It is easy and intuitive to share applications with others or deploy them to other machines (like a Rasperry Pi, or a Windows ARM64 laptop) as part of a fluid workflow.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+* **Architecture-specific, framework-dependent** -- Applications of this kind target and can only run on a single architecture, like Windows x64 or Linux ARM32. They require an installed runtime to run. They carry one set of dependent native libraries for a single target architecture. They include an executable launcher for the given architecture. All application binaries are contained in a single directory.
+* **Architecture-specific, self-contained** -- Applications of this kind target and can only run on a single architecture, like Windows x64 or Linux ARM32. They carry a runtime and one set of dependent native libraries for a single target architecture. They include an executable launcher for the given architecture. All application, runtime and native binaries are contained in a single directory.
+
+### Proposed CLI argument and MSBuild property changes
+
+The following changes are proposed for .NET Core 3.0. `build` will be the specific verb discussed in the following text. The behavior is intended to be the same for `publish` unless specifically called out.
+
+#### Framework-dependent architecture-netural applications
```
Lakritzator:
```suggestion
#### Framework-dependent architecture-neutral applications
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+Maintain framework-dependent architecture-netural applications as the default for `build`.
```
Lakritzator:
```suggestion
Maintain framework-dependent architecture-neutral applications as the default for `build`.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+WPF and Windows Forms templates should opt-in to this build type, with the following pattern:
+
+```xml
+<ArchitectureSpecific>true</ArchitectureSpecific>
+<RuntimeIdentifier>win-x86</RuntimeIdentifier>
+```
+
+The `win-x86` RID is proposed as the default to match the 32-bit default for .NET Framework projects. There is a perf benefit and it will be more compatible with existing .NET Framework projects that may have 32-bit only dependencies. It does mean that we need to install the Windows 32-bit .NET Core runtime by default (the 32-bit SDK is not needed). Visual Studio should make it straightforward to switch the default RID to `win-x64`.
+
+Note: It is possible that the RID should be specified a different way that better aligns with the Visual Studio configutation manager. This document doesn't take an opinion on that. it is a detail to work out.
```
Lakritzator:
```suggestion
Note: It is possible that the RID should be specified a different way that better aligns with the Visual Studio configuration manager. This document doesn't take an opinion on that. it is a detail to work out.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+Maintain the existing self-contained behavior as-is for .NET Core 2.2 apps and earlier, but change it for 3.0 apps and later.
+
+The behavior where specifying a runtime (via `-r`) is convenient but confusing and unfortunate given that we consider framework-dependent applications to be the default. The proposal for framework-dependent architecture-specific apps makes self-contained opt-in when a RID is specific. It would be better to do that for all build types, making the system easier to understand.
+
+* For 1.x and 2.x application, `-r RID` assumes `--self-contained true`
+* For 3.0+ application, `-r RID` assumes `--self-contained false` -- `--self-contained` or `--self-contained true` must be specified to produce a self-contained application.
+
+## Exploration of current behavior for building .NET Core applications
+
+The following sections explore the current behavior for building .NET Core applications. It is provided for informational purposes only and isn't needed to understand the propopsal.
```
Lakritzator:
```suggestion
The following sections explore the current behavior for building .NET Core applications. It is provided for informational purposes only and isn't needed to understand the proposal.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+
+### Architecture-neutral, framework-dependent applications with native dependencies
+
+Architectural neutral applications with native dependencies must contain native assets for all supported operating systems and architectures. This means that this type of applications needs to include multiple builds of the same native library, in order to provide a compatible asset on each native environment on which the application is supported to run.
+
+The architecture-neutral model depends on three characteritics: an installed .NET Core runtime (for that environment) on the target machine, multiple variants of dependent native binaries carried with the application (the .NET assembly binder know how to pick the correct one), application binaries are composed of architectural-neutral byte code (IL) as opposed to pre-compiled code (using crossgen). This is a great model for cases where a developer wants to deploy the same build of an application to machines that use multiple operating systems and chips. It isn't a good option for developers that use conditional compilation that targets operating system or chip.
```
Lakritzator:
```suggestion
The architecture-neutral model depends on three characteristics: an installed .NET Core runtime (for that environment) on the target machine, multiple variants of dependent native binaries carried with the application (the .NET assembly binder know how to pick the correct one), application binaries are composed of architectural-neutral byte code (IL) as opposed to pre-compiled code (using crossgen). This is a great model for cases where a developer wants to deploy the same build of an application to machines that use multiple operating systems and chips. It isn't a good option for developers that use conditional compilation that targets operating system or chip.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,806 @@
...
+Expose a new opt-in build type, called architecture-specific framework-dependent.
+
+* Produces an application with native dependencies and an EXE for a single specified architecture.
+* This build type can be selected via
+  * MSBuild: `<ArchitectureSpecific>true</ArchitectureSpecific>`
+  * CLI: `--architecture-specific` or `-a`
+* A RID must be specified (or it is an error). Any of the following options is legal to specify a RID:
+  * MSBuild: `<RuntimeIdentifier>RID</RuntimeIdentifier>`
+  * CLI: `--runtime RID`, or `-r RID`
+  * MSBuild: `<RuntimeIdentifier>$(NETCoreSdkRuntimeIdentifier)</RuntimeIdentifier>` (means: use RID of SDK; this already works)
```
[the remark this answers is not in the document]
richlander: There is a CLI option suggested. Did you miss it?

proposed/building-arch-specific-applications.md
richlander: I have switched the default to 64-bit in the proposal based on this (and other) feedback.

proposed/building-arch-specific-applications.md
richlander: It's mostly the same as 2.2. I clarified in an update to the doc.

proposed/building-arch-specific-applications.md
richlander: Yes. Having an EXE is nice for launching an application from explorer, for example, even for web servers.

proposed/building-arch-specific-applications.md
richlander: You are right. Template wasn't the right term.

proposed/building-arch-specific-applications.md
richlander: I did repeat myself. I intentionally created a summary at the start and then a more indepth-version.  I streamlined the earlier version a bit to help.

proposed/building-arch-specific-applications.md
richlander: Fixed.

proposed/building-arch-specific-applications.md
richlander: It means that the build will produce different results on different machines, and some configurations will produce the wrong results. Some people will build Windows Forms applications on a Linux x64 machine, but will still want Windows x64 assets. "Current Arch" should be opt-in. It is documented in the spec.

proposed/building-arch-specific-applications.md
richlander: I agree. I changed the text.

richlander:
Thanks for all the feedback. 

I changed the default for Windows client applications to **Windows x64**.

richlander:
@KathleenDollard -- I cannot reply directly to your [big comment](https://github.com/dotnet/designs/pull/65#pullrequestreview-228928481) (it was in your approval). That's fine, but just means I need to reply in an unconnected way to your original comment.

You wrote this:

> I'd also like to see us consider downplaying the description "Framework Dependent". I think we should take as a message that it hasn't really caught on. I'd like to see these three:

```xml
<ArchitectureSpecific>[bool]</ArchitectureSpecific>
<SelfContained>[bool]</SelfContained>
<RuntimeIdentifier>[RID]</RuntimeIdentifier>`
```

We have been trying to keep the project files as lean as possible. While I agree setting key settings explicitly will help clarify behavior, it also gives up on the value of defaults. I'd like to elide defaults from the project file if at all possible. Thoughts?

KathleenDollard:
@richlander 

I think you could use entries like I suggested with defaults. I did not mean to imply that they all needed to always be there. 

I know we talk a lot about lean project files, but I think the reason for that is that we want project flies that are easily consumed by humans.

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,835 @@
...
+* **Architecture-specific, framework-dependent** -- Applications of this kind target and can only run on a single architecture, like Windows x64 or Linux ARM32. They require an installed runtime to run. They carry one set of dependent native libraries for a single target architecture. They include an executable launcher for the given architecture. All application binaries are contained in a single directory. Applications of this kind are launched via the native exectuable launcher that comes with the application. The launcher is responsible for locating a (typically) centrally installed .NET Core runtime.
```
Lakritzator:
```suggestion
* **Architecture-specific, framework-dependent** -- Applications of this kind target and can only run on a single architecture, like Windows x64 or Linux ARM32. They require an installed runtime to run. They carry one set of dependent native libraries for a single target architecture. They include an executable launcher for the given architecture. All application binaries are contained in a single directory. Applications of this kind are launched via the native executable launcher that comes with the application. The launcher is responsible for locating a (typically) centrally installed .NET Core runtime.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,835 @@
...
+* **Architecture-specific, self-contained** -- Applications of this kind target and can only run on a single architecture, like Windows x64 or Linux ARM32. They carry a runtime and one set of dependent native libraries for a single target architecture. They include an executable launcher for the given architecture. All application, runtime and native binaries are contained in a single directory. Applications of this kind are launched via the native exectuable launcher that comes with the application. The launcher discovers and uses the runtime in the same directory. It does not attempt to use any other runtime.
```
Lakritzator:
```suggestion
* **Architecture-specific, self-contained** -- Applications of this kind target and can only run on a single architecture, like Windows x64 or Linux ARM32. They carry a runtime and one set of dependent native libraries for a single target architecture. They include an executable launcher for the given architecture. All application, runtime and native binaries are contained in a single directory. Applications of this kind are launched via the native executable launcher that comes with the application. The launcher discovers and uses the runtime in the same directory. It does not attempt to use any other runtime.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,835 @@
...
+
+* MSBuild: `<ArchitectureSpecific>RID</ArchitectureSpecific>`
+
+This alternative idea requires one instead of two lines, but also replaces the existing `RuntimeIdentifier` property for only architecture-specific apps, which is likely confusing.
+
+#### Self-contained architecture-specific applications
+
+Maintain the existing self-contained behavior as-is for .NET Core 2.2 apps and earlier, but change it for 3.0 apps and later.
+
+The behavior of specifying a runtime (via `-r`) currently results ins self-contained applications by default. That's a confusing experience given that we consider framework-dependent applications to be the default. The proposed beavhior for framework-dependent architecture-specific apps makes self-contained opt-in. It would be better to do that for all build types, making the overall system easier to understand. We cannot reasonably do that for existing projects, but we can do it for 3.0 and later projects.
```
Lakritzator:
```suggestion
The behavior of specifying a runtime (via `-r`) currently results ins self-contained applications by default. That's a confusing experience given that we consider framework-dependent applications to be the default. The proposed behavior for framework-dependent architecture-specific apps makes self-contained opt-in. It would be better to do that for all build types, making the overall system easier to understand. We cannot reasonably do that for existing projects, but we can do it for 3.0 and later projects.
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,835 @@
...
+
+## Default architecture for Windows client applications
+
+This proposal defines 64-bit as the default architecture for Windows client applications. This is a departure from .NET Framework, which defines 32-bit as the default architecture, using the `32-bit preferred` MSIL setting. Given that .NET Core uses native executable launchers, there is no analogous concept as 32-preferred. We must choose 32-bit or 64-bit as the default for Windows applications.
+
+There are some performance advantages of 32-bit, however there are also performance disadantages. There are significant security advantages of 64-bit. Assuming performance is a wash across all potential client applications that will be built using .NET Core, then security advantages alone make choosing 64-bit as the default a sound choice. A developer can still choose 32-bit for their specific application. More context: [@tannergooding](https://github.com/dotnet/designs/pull/65#discussion_r277171851), [@jkotas](https://github.com/dotnet/designs/pull/65#discussion_r277172570).
```
Lakritzator:
```suggestion
There are some performance advantages of 32-bit, however there are also performance disadvantages. There are significant security advantages of 64-bit. Assuming performance is a wash across all potential client applications that will be built using .NET Core, then security advantages alone make choosing 64-bit as the default a sound choice. A developer can still choose 32-bit for their specific application. More context: [@tannergooding](https://github.com/dotnet/designs/pull/65#discussion_r277171851), [@jkotas](https://github.com/dotnet/designs/pull/65#discussion_r277172570).
```

proposed/building-arch-specific-applications.md
```diff
@@ -0,0 +1,835 @@
...
+
+### Enabling Windows client applications that support 32-bit and 64-bit environments
+
+We have seen several requests to enable applications that run in both 32-bit and 64-bit Windows environments, like `32-bit preferred`. This is possible, but not convenient. The following are the primary challenges with enabling that scenario:
+
+* It requires two launchers, for each of 32-bit and 64-bit architectures. They either need to have different names so that they can co-exist in the same directory, or live in different directories if they have the same name. It is unclear how end-users should be directed to the correct executable or how a setup program should be configured to select the correct executable.
+* It requires application setups to pre-req both the 32-bit and 64-bit runtimes and install the correct one.
+* Native dependencies for just Windows 32- and 64-bit should be included in the build output, but no other architectures.
+* It dissallows the use crossgen to produce ahead-of-time compiled application binaries as a means of improving startup performance. Crossgen is not provided as a tool that can be run on client machines.
```
Lakritzator:
```suggestion
* It disallows the use crossgen to produce ahead-of-time compiled application binaries as a means of improving startup performance. Crossgen is not provided as a tool that can be run on client machines.
```

terrajobst:
@richlander said this can be close because it has been superseded.

PR #67, opened 2019-04-30
SingleFile: Design doc update

swaroop-sridhar:
The implementation of stage 1 of single-file publishing is now complete.
This chages updates the docs with:
- Additional design/implementation details
- Changes in property names and configuration options.


accepted/single-file/extract.md
```diff
@@ -4,112 +4,115 @@ When running the single-file, the host will need to extract some bundled files r
...
+- The `<base>/<app>/<proc-id>`  is now considered app-root, and execution continues.
+
+### Extraction Configurations
+
+The following environment variables influence the extraction mechanism:
 
-For .Net Core 3.0, we propose that we start with the simple implementation that extracts necessary files to temporary locations on every run at startup. Based on customer feedback, we can implement other options discussed above.
\ No newline at end of file
+* `DOTNET_BUNDLE_EXTRACT_BASE_DIR`  The base directory within which the files embedded in a single-file app are extracted, as explained in sections above. This directory can be set up an admin-only writable location if necessary.
+* `DOTNET_BUNDLE_EXTRACT_ALL`:  Apps may choose to have all of their dependencies extracted to disk at runtime (instead of loading certain files directly from the bundle) by setting this variable.
```
jkotas: What is the scenario when/how the app will use this environment variable? If the app wants to choose, shouldn't it be a config setting instead?

swaroop-sridhar: I think it is useful to have settings in both forms:
* A option in `runtimeconfig.json`, since this is the best way to configure shipping apps.
* An environment variable, to assist in debugging scenarios, or if the customer wants to explicitly control where all extracted files land.

In the current preview, all embedded files are spilled to disk upfront, even before configuration files are processed. Therefore, we only have the environment variables. In the upcoming version, configuration files will be processed directly from the bundle. I'll add runtimeconfig options once this is done.

accepted/single-file/extract.md
```diff
@@ -4,112 +4,115 @@ When running the single-file, the host will need to extract some bundled files r
...
 
-#### Cleanup
+### Extraction Location
 
-The cleanup of extracted files in the install-location will be manual in this version. We can consider adding `dotnet CLI` commands for cleanup in future. However, future versions are expected to spill fewer artifacts to disk, making the cleanup commands a lower priority feature.
+The desirable characteristics of the extraction directory are:
 
-#### Pros
+- Should be configurable
+- Should be semi-permeant: Not a directory that is lost frequently (ex: on machine reboot), but considered recyclable storage, since any subsequent run can re-extract.
```
sbomer:
```suggestion
- Should be semi-permanent: Not a directory that is lost frequently (ex: on machine reboot), but considered recyclable storage, since any subsequent run can re-extract.
```

accepted/single-file/extract.md
```diff
@@ -4,112 +4,115 @@ When running the single-file, the host will need to extract some bundled files r
...
-#### Cons
+* `<base>` is 
 
-- Complex implementation.
-- Unsuitable for customer scenarios that cannot tolerate persistent disk state after the app terminates.
-- Cleanup is manual.
+  * `DOTNET_BUNDLE_EXTRACT_BASE_DIR` environment variable, if set.
+  * If not, defaults to 
+    * `%TEMP%\.net ` on Windows
+    * `$TMPDIR\.net` if `$TMPDIR` is set (Posix conforming OSes including Mac)
```
sbomer:
```suggestion
    * `$TMPDIR/.net` if `$TMPDIR` is set (Posix conforming OSes including Mac)
```

accepted/single-file/extract.md
swaroop-sridhar: @jkotas does this address your concern? Thanks.

jkotas: The design docs should capture what we plan to build, not the point-in-time state. I think doc should say that the environment variable is meant just for debugging. The design doc suggests that it is something we would recommend to deploy apps with. I do not think we want to recommending the apps with this.

BTW: What happens when I run the app without the env variable (only the non-bundleable files get expanded) and then run it again with env variable? Is the second run going to crash because of the app will expect all files to be on disk, but they won't be there?

swaroop-sridhar: OK Thanks @jkotas, I'll update the doc with config settings.

DOTNET_BUNDLE_EXTRACT_ALL means that all the files are always be extracted.
I think this flag is only useful for debugging.

swaroop-sridhar: @jkotas: I've addressed the feedback in this change: 2b7f4a6d841de30581e8dfdbb8aca4e39d968b17 Thanks.

PR #69, opened 2019-05-06
Add proposal for global.json updates

KathleenDollard:


accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+
+* Due to #2, simply installing a patch on the machine will not update the machine to use the safer version.
+  * The impact of this is mitigated by work we did in .NET Core 3.0 Preview 3 to replace earlier patch versions.
+* Due to #3, users can not guarantee they are pinned to a specific version.
+
+## Customer problems to address
+
+* User wants to install a preview but only use it with a subset of projects.
+  * Due to Rule #1, a `global.json` is the only way to opt out of a preview in the common case of the preview being the highest version on the machine.
+  * The `global.json` the user addss must include a specific version.
```
jkotas:
```suggestion
  * The `global.json` the user adds must include a specific version.
```

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+  * There is no mechanism to require at least a specific version, but otherwise use the latest, or the latest in a range.
+  * Upgrading to a newer SDK will provide no benefit to projects tied to an earlier version.
+* User wants to know what version of the SDK is being used.
+  * `global.json` is a subtle, hidden and easily forgotten marker – programmers may not understand the downsides of leaving a global.json that was suggested in a blog post in place – or they may forget it.
+* User rarely wants the risk of running the unsupported scenario of multiple SDK versions in a solution build.
+  * There is nothing that encourages users to place `global.json` in the repository root or the same level as the solution. Using different versions of the SDK for different projects in the solution is not a supported scenario.
+* Hosted CI, including Azure DevOps and Kudu, need to provide a small number of SDKs.
+  * If there is no patch of the major/minor/feature version (xyz in `x.y.znn`) present, the build fails, causing friction between user and host needs.
+
+## Proposal C
```
richlander: It is a bit strange to see Proposal C here and A and B as alternatives. That must be for historical reasons. Seems like an opportunity to switch the names of the proposal. Also, if there is only one proposal in the main part of the doc, maybe it doesn't need a name at all.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
+# Proposal for global.json updates for preview and roll-forward
+
+---
+**NOTE:** 
+
+SDK version numbers do not use semantic versioning. [Docs has an article on SDK versioning](https://docs.microsoft.com/en-us/dotnet/core/versions/).
+
+---
+
+`global.json` allows users to pin the version of the SDK they are using. This was added to .NET Core as a way for users to roll-back to a previous version per project until a fix could be provided if there a problem occurred. While our guidance remains to use it only for special circumstances, users now use it for numerous reasons.
```
richlander: A nit, but this is a bit of mixing of metaphors. "pinning" and "roll-back" are different semantics and it's best to avoid evoking both of those terms in the same explanation. I think of global.json as "pinning". I don't think it has any "roll-back" semantics, at least with my definition of "roll-back", which I think of as being automatic and not manual.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+Given x.y.znn
+
+* x is the major version
+* y is the minor version
+* z is the feature band
+* nn is the patch
+
+## Current behavior
+
+The SDK resolvers (in both the CLI and VS) search from the current location (project in VS) up the directory hierarchy until they encounter a `global.json` file. When the resolver finds one, _it stop looking_ and attempts to resolve the SDK version specified in that file. This happens even if this `global.json` does not specify an SDK version.
```
richlander: This paragraph aims to define the experience in terms of both the CLI and VS, but then defined the current location in terms of VS. I assume current location is equivalent to CWD. Is that right?

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+
+This proposal parallels [runtime roll-forward behavior as initially described](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md)} and [recent improvements](https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md). All of the features proposed here may not be included in Phase 1. It will include at least `ignorePreview` and `"rollforward": "disable"`
+
+This is called Proposal C to differentiate it from two earlier proposals that were discarded but are included at the end of this document for reference.
+
+### Details
+
+The runtime options allow for three basic approaches: find nearest, find latest and find exact. 
+
+To avoid a backward compatibility issue, the default recovery strategy when a match is not found needs to match the old `dotnet` behavior. This defines the defaults, and the default behavior differs between runtime.
```
richlander: Is that required? Or put another way, are there some behaviors that are more important to maintain and others that are more borderline? For example, the [behavior where a lower version of the SDK is used than requested](https://github.com/dotnet/designs/pull/69/files#diff-52992d6b81ad6f9528f44e8bd5f44910R37) seems like a break we might be able to accept.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+The runtime options allow for three basic approaches: find nearest, find latest and find exact. 
+
+To avoid a backward compatibility issue, the default recovery strategy when a match is not found needs to match the old `dotnet` behavior. This defines the defaults, and the default behavior differs between runtime. 
+
+* Runtime: From the [recent improvements](https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md) document "_In all cases except `disable` the highest available patch version is selected._"
+* SDK: For backwards compatibility, the additonal option `patch` is offered and is the default.
+  * If the specified patch version is found, it is used.
+  * In all other cases (except `disable`), the highest patch within the appropriate feature band is used. 
+
+The `global.json` schema to offers support for all of the identified scenarios that can be fixed without altering how we find `global.json`:
```
richlander: This sentence is a bit hard to parse.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+
+* `sdk`
+  * `version`
+    * The lowest SDK the resolver would select - any SDKs with lower versions are ignored.
```
richlander: "would" or "should" or "permitted"? "would" is a bit weak.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+    * Identical to today - wildcards are not supported.
+    * Default: empty
+  * `ignorePreview`
+    * `true` or `false`
+    * Default: `false` to more closely match today's behavior.
+    * Alternate names: `releaseOnly` or `useReleaseOnly`
+  * `rollForward`
+    * To avoid some of the repetitiveness, I do not restate in these rules that version numbers below that specified are always ignored. 
+    * `patch`
+      * If the requested major, minor, feature band, and patch if found, use it.
```
richlander:
```suggestion
      * If the requested major, minor, feature band, and patch is found, use it.
```

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+      * Otherwise, use the lowest higher minor, it's lowest feature band, and the highest patch within that feature band.
+    * `major` 
+      * If the requested major, minor, feature band, and patch is found, use it.
+      * Otherwise, if the requested major, minor and feature band is found, use the highest patch in that band.
+      * Otherwise, if the requested major and minor is found, use the highest patch in the _lowest_ feature band within that major and minor version.
+      * Otherwise, if the requested major is found, use the  _lowest_ minor within that major version, and it's lowest feature band, and it's highest patch.
+      * Otherwise, use the lowest higher major, it's lowest minor, and it's lowest feature band, and the highest patch within that feature band.
+    * `latestPatch`
+      * Even if the requested major, minor, feature band and patch is found, do not use it unless it matches rules below.
+      * Select the highest available patch version that matches the specified major, minor and feature band versions.
```
richlander: This wording can result in selecting a lower version than specified.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+* `sdk`
+  * `version`
+    * The lowest SDK the resolver would select - any SDKs with lower versions are ignored.
+    * Identical to today - wildcards are not supported.
+    * Default: empty
+  * `ignorePreview`
+    * `true` or `false`
+    * Default: `false` to more closely match today's behavior.
+    * Alternate names: `releaseOnly` or `useReleaseOnly`
+  * `rollForward`
```
richlander: I didn't see mention of which one of these policies is the default. It would also be good to see which policies align with existing behaviors.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+    "rollForward": [entry in table header]
+  }
+}
+```
+
+| Available SDKs                                       | patch     | feature | minor   | major   | latestPatch | latestFeature | latestMinor | latestMajor | disable |
+|------------------------------------------------------|-----------|---------|---------|---------|-------------|---------------|-------------|-------------|---------|
+| 2.1.500                                              | fail      | fail    | fail    | fail    | fail        | fail          | fail        | fail        | fail    |
+| 2.1.501, 2.1.503                                     | _2.1.501_ | 2.1.501 | 2.1.501 | 2.1.501 | _2.1.503_   | 2.1.503       | 2.1.503     | 2.1.503     | 2.1.501 |
+| 2.1.503, 2.1.505, 2.1.601, 2.2.101, 3.0.100          | 2.1.503   | 2.1.503 | 2.1.503 | 2.1.503 | 2.1.505     | 2.1.601       | 2.2.101     | 3.0.100     | fail    |
```
richlander: I believe that .503 should be replaced by .505 for this row. .503 would not be selected by any of the policies.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+| 2.1.601, 2.1.604, 2.1.702, 2.2.101, 2.2.203, 3.0.100 | fail      | 2.1.601 | 2.1.601 | 2.1.601 | fail        | 2.1.702       | 2.2.203     | 3.0.100     | fail    |
```
richlander: I believe that .601 should be replaced by .604 for this row (fourth row). .501 would not be selected by any of the policies.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+| 2.2.101, 2.2.203, 3.0.100                            | fail      | fail    | 2.2.101 | 2.2.101 | fail        | fail          | 2.2.203     | 3.0.100     | fail    |
+| 3.0.100, 3.1.102                                     | fail      | fail    | fail    | 3.0.100 | fail        | fail          | fail        | 3.1.102     | fail    |
+
+Or, if it is easier to understand with logic (bail out of logic on fail or select):
+
+1. If not `latest`, and there is a match, select it.
```
richlander: What is the definition of `latest`?

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+1. If we get here for `disable`, then fail.
+1. Exclude any lower versions.
+1. If `minor` or `latestMinor`, exclude any that don't match major version. 
+1. If `feature` or `latestFeature`, exclude any that don't match major and minor version.
+1. If `patch` or `latestPatch`, exclude any that don't match major, minor and feature band version.
+1. If there is one left, select it.
+1. If any of the `latest` options are used, select the highest. 
+1. Otherwise, select the lowest major, and within that lowest minor, and within that lowest feature band. Then select the highest patch in that feature band.  
+
+## Customer problems addressed
```
richlander: I have a scenario where there is a global.json at the root of the repo and I don't want to align with it, but I also don't want to specify anything. I just want the default behavior. What should I do in that case?

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+  * User opts out in projects where preview should not be used by adding `ignorePreview: true` to those projects, or
+  * User adds `ignorePreview: true` in a `global.json` higher in their file system, and adds `ignorePreview: false` in a `global.json` in specific projects.
+* User wants to always use a specific SDK patch version, and receive an error if it is not on the machine.
+  * User adds `rollForward: disable` to a `global.json`
+* User wants to always run the latest patch of a particular SDK feature band.
+  * User adds `rollForward: latestPatch` and the lowest version (with patch) they want to a `global.json`
+* User includes a feature that requires an SDK that corresponds to a SDK feature band or higher, and wants to run the best available tools.
+  * User adds `rollForward: latestMajor` and the lowest version (with patch) they want to a `global.json`
+* User wants to know what version of the SDK is being used.
+  * Not addressed in this proposal: `global.json` will still be a subtle, hidden and easily forgotten marker
```
richlander: It is odd to see multiple instances of "not address in this proposal" in a section called "Customer problems addressed".

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+{
+  "sdk": {
+    "version": "2.2.100",
+    "ignorePreview": false,
+    "rollForward": "patch"
+  }
+}
+```
+
+The common `global.json` for this case is likely to be:
```
richlander: What does "common" mean?

richlander: It might be easier to read these examples if they were reversed ... minimal syntax and then the fully-specified syntax to clarify. I find this approach a bit confusing, and it also isn't clear which is recommended for practical use.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+{
+  "sdk": {
+    "ignorePreview": true,
+  }
+}
+```
+
+The user may include their preference for previews high in their directory structure, and then individual project can elect different behavior with a new `global.json`. This will continue to follow today's rules. In particular: if there is a `global.json` file for a different reason, like a custom SDK, the higher `global.json` will not be used. 
+
+### 4 Select highest available, higher than specified
```
richlander: The term "higher than specified" is confusing. It makes it sound like there is a policy where a version higher than the one specified will be selected. From reading the doc, I don't think there is such a policy.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,790 @@
...
+    "rollForward": "latestMinor"
+  }
+}
+```
+
+### 10. Highest within arbitrary upper and lower bound
+
+If the user wanted to use any SDK higher than the one listed but ignore previews, they could use the following. 
+
+Proposal C
```
richlander: I was confused by this text.

KathleenDollard:
Nothing like pushing a proposal public to result in you waking up in the middle of the night realizing a fatal flaw in the design. 

The .NET Core SDK can only roll forward across patches because the "major"/"minor"/feature band portions of the version are non-linear. The proposal as written would result in rolling forward to an earlier toolset, which is particularly flawed in use latest. For example, 2.1.600 (MSBuild 16) would roll forward to 2.2.100 (MSBuild 15.n), which could break things. 

It's not in the scope of this proposal to look at .NET Core version numbers again. I'll simplify and repost the proposal with options for exact, no preview, current behavior and latest patch.

KathleenDollard:
Thanks to @jkotas and @richlander for early feedback. I'll include your comments in the new proposal.

PR #71, opened 2019-05-13
`global.json` updates

KathleenDollard:
Add features for 

- Preview
- Exact
- Latest within feature band
- Latest with lower bound


tannergooding:
I think one of the scenarios that is missing here, but that could easily be addressed, is allowing the user to specify a local install directory to check.

That is, for CI, some repos will always download a local SDK, regardless of what is available on box. And, while we have the ability to include additional search paths by modifying the current environment, there is no way to configure that to "just work" for a clean enlistment where the user is directly running `dotnet build`.

I would propose that you should be able to specify an additional directory to check for dotnet installs as part of the `global.json`. For example, specifying ` { "sdk": { "localDirectory": "./artifacts/dotnet" } }` (or a better name) would cause `dotnet` to first try to resolve an SDK/runtime installed under `localDirectory`.

As extensions, you could also have an option that blocks searching in the default locations (i.e. only look under `localDirectory` and not on the path) and an option that indicates whether or not downloading a missing sdk should be allowed (currently this requires CI environments to pull down the `dotnet-install` script and pass in much of the information that already exists in the `global.json`).

This functionality, combined with the new functionality proposed in the OP, would allow a repository to ensure that an appropriate SDK version is used and also have that integrate with their repo infrastructure such that they don't require users to use `build.cmd/sh` scripts to work successfully.

KathleenDollard:
We have this in mind for a future update, but need to keep this one simple. 

In other words: this is the first step of a journey.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+* z is the feature band
+* nn is the patch
+
+## Current behavior
+
+The SDK resolvers (in both the CLI and VS) search from the current location (project in VS) up the directory hierarchy until they encounter a `global.json` file. When the resolver finds one, _it stop looking_ and attempts to resolve the SDK version specified in that file. This happens even if this `global.json` does not specify an SDK version.
+
+When a `global.json` is encountered and it specifies an SDK version, the SDK resolvers attempt to locate that specific SDK patch version:
+
+1. When the resolver finds no `global.json` or finds one that does not specify SDK version, the latest SDK is used, regardless of whether it is preview or stable.
```
jaredpar: This doesn't seem to reflect the current behavior. Consider that when building from VS the preview SDK is only used from a preview VS.

Or is this correct in that an RTM VS will pick a preview SDK if there is no global.json but refuse to use a preview SDK if there is a global.json that specifies it?

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+2. When the resolver finds a `global.json`, the specific SDK patch version it specifies is used if it is found.
+3. When the resolver finds a `global.json`, and the SDK it specifies can't be found, the resolver looks for a close match:
+    * Prior to .NET Core 2.1
+      * The highest patch version for the SDK feature band is used _regardless of whether it is higher than what is specified_. For example, if `global.json` specifies 2.2.103 is specified and is not available, and 2.2.101 is the highest SDK on the machine, it will be used.
+    * In .NET Core 2.1 and higher
+      * The highest patch version for the SDK feature band is used _assuming it is higher than what is specified_. For example, if `global.json` specifies 2.2.103 is specified, and 2.2.101 is the highest SDK on the machine, an error will occur.
```
jaredpar: If I'm reading this correct we get the "greater than or equal to SDK version" feature by just specifying the version up to the feature brand level? That seems to be how the document reads but at the same time the behavior I'm describing has always seemed to be an ask vs. something that was already there.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+
+* `sdk`
+  * `version`
+    * The lowest SDK the resolver would select - any SDKs with lower versions are ignored.
+    * Wildcards are not supported.
+    * Default: empty
+  * `ignorePreview`
+    * `true` or `false`
+    * Default: `false` to match previous behavior.
+    * Alternate names: `releaseOnly` or `useReleaseOnly`
```
jaredpar: Consider also `usePreview`. Feel like a positive declarative property about the behavior desired here is more natural.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+    * This is new behavior.
+  * `rollForward`
```
jaredpar: What would be the impact of declaring `rollForward` but not specifying `sdk.version`? It seems this property is only meaningful if `sdk.version` is specified correct?

accepted/global-json-updates.md
nguerrera: Correct. This reflects the behavior of `dotnet`, but not VS. VS passes in a flag to the same underlying to control when previews are used or not. This flag defaults based on whether VS is a preview and can be overridden by a checkbox in VS. Having a bool in the global.json essentially makes this flag available without the checkbox.

So, `dotnet` can respect it too, and you can set up previews to be used in VS irrespective of checkboxes with this.

accepted/global-json-updates.md
nguerrera: The behavior today is exact version if it exists, else latest from feature band.

accepted/global-json-updates.md
nguerrera: I like usePreview.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+### 6. Select highest available with unusual combination
+
+If the user wants to use any SDK as high or higher than the one listed but ignore previews, they could use the following:
+
+```json
+{
+  "sdk": {
+    "version": "3.0.100-Pre",
+    "ignorePreview": true,
+    "rollForward": "highest"
```
nguerrera: Should this be:
```suggestion
    "rollForward": "latest"
```

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+    * Wildcards are not supported.
+    * Default: empty
+  * `ignorePreview`
+    * `true` or `false`
+    * Default: `false` to match previous behavior.
+    * Alternate names: `releaseOnly` or `useReleaseOnly`
+    * This is new behavior.
+  * `rollForward`
+    * Lower version numbers are always ignored. 
+    * `patch`
```
jaredpar: Is there a functional difference between the following: 

Option 1:

``` json
"sdk": {
  "version": "3.0.1"
}
```

Option 2

``` json
"sdk": {
  "version": "3.0.100",
  "rollForward": "highest"
}
```

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+* Due to #2, simply installing a patch on the machine will not update the machine to use the safer version.
+  * The impact of this is mitigated by work we did in .NET Core 3.0 Preview 3 to replace earlier patch versions.
+* Due to #3, users can not guarantee they are pinned to a specific version.
+
+## Proposal
+
+### Roll-forward
+
+.NET Core SDK "roll-forward" is problematic because version numbers are not linear. For example: 2.1.600 contains MSBuild 16 while 2.2.100 contains MSBuild 15.9. This is not a new problem, since it already occurs when there is no `global.json`. However, we do not want to roll-forward in scenarios that might surprise the user. Because of this, roll-forward will be allowed only in limited scenarios.
+
```
nguerrera: I don't think this is a major concern moving forward since we will not be using multiple SDKs per VS version (only multiple, runtimes, templates, targeting packs), etc. 
 
I also don't see how limiting the options helps. latest still has a problem in this case.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,315 @@
...
+      * This is new behavior.
+    * `latest`
+      * Even if the requested major, minor, feature band and patch is found, do not use it unless it matches rules below.
+      * Select the highest available major version, and select it's highest minor version, and that minor's highest feature band, and that feature band's highest patch.
+      * Using latest with a version is new behavior. Otherwise this matches previous behavior when there is no `global.json` or no version is specified.
+    * Defaults: 
+      * Defaults are defined to support backward compatibility issues.
+      * If a version is specified in the `global.json`, the default is `patch` to match previous behavior.
+      * If there is no `global.json` or a version is not specified, the default is`latest`, again to match previous behavior.
+
```
nguerrera: latest is a big hammer and patch, latest patch, and disable are too rigid in the face of upgrades/sdk clean up. I think we should mirror the minor and major options of the runtime. They are conservative and do not go to latest always. 

I also don't think we should be too concerned about using latestPatch by default even if exact is found. Patches are always highly compatible and contain security fixes. You should have to opt out of them instead of opting in to them.

My suggestion would be to have the same rollForward options as the runtime, but consider version Major.Minor.(Band * 100) + Patch. To be Major.(Minor * 100 + Band).Patch.

I would further suggest changing the default rollForward to Major. I don't consider this a significant breaking change. You will get security patches by default, and otherwise turn error cases into finding nearest match.

Since we will be stop leaving SDKs behind in VS, this will be a much more compatible default. Only users that explicitly opt in to more conservative roll forward will get hard broken by upgrades or sdk clean up.

accepted/global-json-updates.md
KathleenDollard: I'm assuming @nguerrera comment resolves @jaredpar 's. This is not the same as greater or equal, which is a feature added later in this proposal.

accepted/global-json-updates.md
KathleenDollard: I'm confused. There is no 3.0.1 patch of the SDK. Is there a typo in your comment?

accepted/global-json-updates.md
```diff
@@ -97,46 +92,55 @@ The `global.json` schema to offers support for all of the identified scenarios t
...
+    * `patch` **[[(I've gone back and forth on "patch" or "legacy" here)]]**
       * If the requested major, minor, feature band, and patch if found, use it.
       * Otherwise, use the highest patch within the specified major, minor and feature band.
+      * This recreates the previous SDK selection criteria.
     * `feature`
-      * If the requested major, minor, feature band, and patch is found, use it.
-      * Otherwise, if the requested major, minor and feature band is found, use the highest patch in that band.
-      * Otherwise, use the lowest higher feature band, and use the highest patch within that feature band.
+      * If the requested major, minor, feature band, is found, use the highest available patch in that band.
+      * Otherwise, use the lowest higher major/minor/feature band, use the highest patch within that feature band.
```
peterhuene: I don't get the intent of the new language and this wasn't my take away from this morning's meeting.

If I'm a user that specifies that I would like to roll-forward on `feature`, then I would expect dotnet to **not** cross a major and minor version (i.e. the major and minor must match what's requested).  I would expect it to roll forward to the next available feature band and the latest patch within that feature band, as per previously describe.

Same comment can be repeated for the changes to the `minor` and `major` policies.

nguerrera: I think it would help to use the the same language as https://github.com/dotnet/designs/blob/8fb85d0a79f33b1548e884a8ee5b85319b5946a1/accepted/runtime-binding.md#rollforward

My understanding is that we will have the same behavior, and the only change is making it four levels instead of three (feature added) and what appears like a patch is semver is actually (feature * 100) + patch. If there is another difference, then I have misunderstood something and that would change what I said in the meeting.

KathleenDollard: I really struggled with understanding that language and how long it would make this document. But a) @peterhuene 's comment is correct, my language is incorrect at present and b) @nguerrera 's indicates that my language is not solving the clarity issue. I'll take another look.

peterhuene: Thanks, Kathleen.  I just wanted to double check as that is how I implemented it.

I'm using your grid as basis for some automated tests, plus others covering `allowPrerelease`, parsing of the JSON, etc.

Currently if there is an error in the `global.json` I emit a detailed error message of what is wrong and print out a message that the latest installed SDK (with or without prereleases depending on how the resolver was invoked) will be used.  That's what we previously did for errors, although errors other than a parse error were completely silent by default (required tracing to diagnose).

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+The `global.json` schema to offers support for all of the identified scenarios that can be fixed without altering how we find `global.json`:
+
+* `sdk`
+  * `version`
+    * The lowest SDK the resolver would select - any SDKs with lower versions are ignored.
+    * Identical to today - wildcards are not supported.
+    * Default: empty
+  * `usePreview`
+    * `true` or `false`
+    * Default: `true` to match today's behavior.
```
nguerrera: We should state that for VS, default is to honor what VS does for previews as described above. And Global.json should win over VS on this.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+* User rarely wants the risk of running the unsupported scenario of multiple SDK versions in a solution build.
+  * There is nothing that encourages users to place `global.json` in the repository root or the same level as the solution. Using different versions of the SDK for different projects in the solution is not a supported scenario.
+* Hosted CI, including Azure DevOps and Kudu, need to provide a small number of SDKs.
+  * If there is no patch of the major/minor/feature version (xyz in `x.y.znn`) present, the build fails, causing friction between user and host needs.
+
+## Proposal
+
+This proposal parallels [runtime roll-forward behavior as initially described](https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/roll-forward-on-no-candidate-fx.md)} and [recent improvements](https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md). All of the features proposed here may not be included in Phase 1. It will include at least `usePreview` and `"rollforward": "disable"`
+
```
nguerrera: We're doing the full set below now, right. Do we need the disclaimer about phase 1?

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+  * `usePreview`
+    * `true` or `false`
+    * Default: `true` to match today's behavior.
+  * `rollForward` (The `version` must be specified if `rollForward` is specified, except `latestMajor`.)
+    * There are four basic intents for roll-forward:
+      * `patch`: legacy behavior.
+      * `feature`, `minor`, `major`: highest patch of the nearest version within specified limits.
+      * `latestPatch`, `latestFeature`, `latestMinor`, `latestMajor`: highest patch of the highest available version within specified limits.
+      * `disable`: use exactly the version specified.
+    * `patch` **[[(I've gone back and forth on "patch" or "legacy" here)]]**
```
nguerrera: I like `patch`. It's not exactly analogous to the other non-latest as it fallback to latest instead of falling back to nearest, but close enough. We can document the difference.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+
+### `global.json` containing a version number
+
+**Breaking Change** This default represents a breaking change. We took this change because we believe the new behavior will cause no change or be an improvement.  
+
+Previously, if the specified patch version was present, it was used even if a higher patch number was installed. In discussing this with users, ignoring the updated patch was not the anticipated behavior. Also, the previous default did not roll forward across feature, minor or major bands. This could result in failure when a usable SDK existed on the machine.
+
+The new behavior is that the highest patch within the feature band will be used even if the requested version is available, unless the `rollForward` is set to `disable`.
+
+Also, the nearest higher SDK will be used if there is no version that matches the requested major, minor and feature. This behavior can be changed by specifying other options for the `rollForward` property.
```
nguerrera: Is it nearest higher SDK or is it basically RollForward=major by default? The difference I see is in the case where there are multiple patches.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+### 6. Select highest available with unusual combination
+
+If the user wants to use any SDK as high or higher than the one listed but ignore previews, they could use the following:
+
+```json
+{
+  "sdk": {
+    "version": "3.0.100-Pre",
+    "usePreview": false,
+    "rollForward": "highestMajor"
```
nguerrera:
```suggestion
    "rollForward": "latestMajor"
```

accepted/global-json-updates.md
vitek-karas: I think the default should map to some well defined roll forward setting - it should not be yet another way of doing things - that's super confusing. I don't know which is the best default... currently it is basically `patch`. Changing it to `major` feels like a rather big change. But I guess we make different guarantees about SDK backward compatibility across major versions than we do for the product...

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+* Runtime: From the [recent improvements](https://github.com/dotnet/designs/blob/master/accepted/runtime-binding.md) document" _In all cases except `disable` the highest available patch version is selected._"
+
+The `global.json` schema to offers support for all of the identified scenarios that can be fixed without altering how we find `global.json`:
+
+* `sdk`
+  * `version`
+    * The lowest SDK the resolver would select - any SDKs with lower versions are ignored.
+    * Identical to today - wildcards are not supported.
+    * Default: empty
+  * `usePreview`
```
vitek-karas: In SemVer the correct term is pre-release. In the framework roll forward we even have an environment variable which uses the term "pre-release" for this.

I think this needs a much more detailed description of the desired behavior. Again in framework roll forward we actually spent a lot of time on this: https://github.com/dotnet/core-setup/blob/master/Documentation/design-docs/framework-version-resolution.md#pre-release-versions

In short:
* If the `version` is a release version, should we try to fulfill the request with release first and only then fallback to pre-release? (even is `usePreview` is true).
* What happens if `version` is pre-release but `usePreview=false`
* If `usePreview=false` and we can't find any version and are about to fail - should we consider pre-release versions in this case to avoid failing?

In the below text you always state that we pick the latest patch in the given range (except for disable). We should be explicit what it means for pre-release versions. Do we always pick latest pre-release for a given patch? We actually went back and forth on this for frameworks and ended up having a somewhat in-between solution:
* If the `version` is pre-release, and there is an exact match, use it - don't roll to latest patch/pre-release - otherwise it's really hard to test several different pre-release versions side-by-side. Also we tend to treat pre-release versions as breaking (P4 versus P8 - there might be breaking changes there), so it's better to pick "closest" in this case, to avoid breaking people. I know that we mostly mitigate this by uninstalling other previews, but still - even for internal testing this is important.

accepted/global-json-updates.md
nguerrera: VS will be cleaning up after itself moving forward and will only have a current SDK with optional runtimes, templates, etc. for older version. As such, using recent SDK with projects targeting older runtimes is mainline scenario. We should not break it as a general rule. There will be exceptions and those can be handled with an explicit strengthening in global.json IMHO. The current strictness by default causes a mess of having to keep a bazillion sdks to open up different projects. And we're not going to be leaving SDKs behind anymore.

nguerrera: > There will be exceptions and those can be handled with an explicit strengthening in global.json IMHO.

Or you can install older SDK and major will still prefer it. In the case where you hit a break. 

I do think that the sdk compatibility is different than runtime compatibility. We have always (meaning before .NET Core even) given customers latest tools (Visual Studio et al) to use on their existing projects. I think this should always work by default.

accepted/global-json-updates.md
peterhuene: The setting is `allowPrerelease`.  This part is out of date.

peterhuene: To answer questions:

> If the version is a release version, should we try to fulfill the request with release first and only then fallback to pre-release? (even is usePreview is true).

If I specify `version = 2.2.300, rollForward = latestMajor, allowPrerelease = true` and have `2.2.505` and `3.0.100-preview7` installed, I would be quite surprised if it prefers 2.2.505 over 3.0.100-preview7.  I don't think the version should have semantic meaning regarding whether or not prereleases are matched against (see next answer).

> What happens if version is pre-release but usePreview=false

Currently it silently allows prerelease versions.  My changes make this a warning saying we're ignoring your `usePreview` setting and proceed to allow prereleases.  Personally, I don't like either of these behaviors as to me `version` is a *minimum* bound for the roll-forward policy and shouldn't otherwise have semantic meaning.

> If usePreview=false and we can't find any version and are about to fail - should we consider pre-release versions in this case to avoid failing?

I don't think so.  If I instruct my tools to not consider prerelease versions then it should do so.

> Do we always pick latest pre-release for a given patch?

As this design is currently going to be implemented, yes.  Users should be able to explicitly use a specific prerelease patch level by using either the `patch` or `disable` policies.  Generally users don't install preview servicing releases and our testers don't test multiple preview servicing releases side by side, so I think the impact here would be low.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+    * `true` or `false`
+    * Default: `true` to match today's behavior.
+  * `rollForward` (The `version` must be specified if `rollForward` is specified, except `latestMajor`.)
```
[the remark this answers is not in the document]
peterhuene: I disagree only because there would be no way to represent the current default behavior if global.json were not present (or present and not have sdk settings) in global.json without allowing a version-less `latestMajor`.  But it's not a strong-held disagreement since I don't know how important that is.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,369 @@
...
+    * There are four basic intents for roll-forward:
```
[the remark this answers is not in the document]
peterhuene: Currently the implementation isn't case-insensitive, but if the runtime does it this way, the SDK should to (personally not a big fan of case-insensitive enums).

I'll fix the implementation PR.

accepted/global-json-updates.md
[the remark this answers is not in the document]
KathleenDollard: I think it is essential that we be able to represent the default behavior with an explicit entry (latestMajor with no version). This is the only way to override a setting higher in the directory structure.

accepted/global-json-updates.md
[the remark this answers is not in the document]
KathleenDollard: We should match the runtime here. Thanks @vitek-karas for the catch.

accepted/global-json-updates.md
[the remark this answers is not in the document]
nguerrera: Does version=0.0.0, rollForward=latestMajor not work? I think allowing version to be omitted for latestMajor is fine, but wondering if it's true that it's impossible. Certainly using 0.0.0 is not obvious.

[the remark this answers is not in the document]
peterhuene: That's a good point.  I could see a scenario where someone might want some subtree to not respect the SDK settings from further up the source tree and just use the default "without a global.json" behavior.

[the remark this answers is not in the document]
peterhuene: You're right, version `0.0.0` would work and I agree that it would be non-obvious.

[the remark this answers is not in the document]
peterhuene: Given that works, perhaps we should update the design to express the default "no global.json present" behavior with a version of `0.0.0` and treat a missing version when `rollForward` is specified as an error for consistency?  I'm happy either way, personally.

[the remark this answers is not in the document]
vitek-karas: I do agree that `0.0.0` is not discoverable... I guess the original is OK then.

accepted/global-json-updates.md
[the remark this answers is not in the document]
vitek-karas: Context: We allowed case-insensitive for framework roll forward because that setting can be specified in several different ways. For JSON it makes sense to use the same casing you have here. For command that is less obvious but might be OKish, but for environment variables it just didn't look right.

accepted/global-json-updates.md
vitek-karas: OK - makes sense, I just think this should be explicitly stated - if nothing else so that there's a doc which can be compared against tests/actual behavior.

accepted/global-json-updates.md
```diff
@@ -0,0 +1,374 @@
...
+
+## Examples to illustrate roll-forward options
+
+The following table considers various SDK version combinations that may be available on the machine, and the outcome of the following `global.json`:
+
+```json
+{
+  "sdk": {
+    "version": "2.1.501",
+    "rollForward": [entry in table header]
```
sdmaclea: This syntax is confusing as the '[' and ']' signify a json dictionary. This would be my preference.
```suggestion
    "rollForward": "<entry in table header>"
```

accepted/global-json-updates.md
```diff
@@ -0,0 +1,374 @@
...
+| 2.1.700, 2.2.100, 2.2.103     | 2.2.100 | 2.2.103     | 2.2.103 | 2.2.103     | 2.2.103 | fail    | 2.2.100 | 2.2.100 |
+| 2.1.700, 2.2.103, 3.1.100-Pre | 2.2.103 | 3.1.100-Pre | 2.2.103 | 3.1.100-Pre | 2.2.103 | fail    | fail    | 2.2.100 |
+| 2.1.700, 2.2.103, 3.1.100     | 2.2.103 | 3.1.100     | 3.1.100 | 3.1.100     | 3.1.100 | fail    | fail    | 2.2.103 |
+
+## Other proposals considered
+
+Prior to arriving at this proposal, two others were considered. Neither of these align well with roll-forward on the .NET Core Runtime.
+
+* An independent versioning scheme. It's largest downside is that it would be _another_ way users would have to think about specifying versioning behavior. 
+* An attempt to parallel NuGet. Definining [NuGet version ranges](https://docs.microsoft.com/en-us/nuget/reference/package-versioning) is relatively complex, and it didn't feel like a good fit.
```
sdmaclea: We may also want to consider a version blacklist.  A specific version is incompatible with a given project.  Triggers a compile exception, generate bad code ...

KathleenDollard: @sdmaclea Is there a specific scenario, or is this for completeness/in case we screw up. 

Since we're human, this makes sense to consider for a backlog item, but I don't want to make the current round of changes any more complex unless necessary.

sdmaclea: This is for completeness or in case we screw up.

We don't have a way to work around specific broken SDK releases. It is especially problematic when the latest release is broken.  Presumably the preferred solution here would be to use `disable` with a specific version.

vitek-karas:
@KathleenDollard any objections to merge this so that we have a doc we can point customers at?

PR #72, opened 2019-05-15
Single-File: Fix a typo in the usage section

swaroop-sridhar:
Fix the use of --single-file boolean switch


swaroop-sridhar:
cc: @mihaimyh

PR #74, opened 2019-05-18
Single-file: Add notes about accessing files from Bundle

swaroop-sridhar:
This change adds certain details about accessing the files
embedded within  asingle-file bundle without having to spill
them to files on disk.

### DIFF
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -182,7 +182,7 @@ We need to determine the semantics of current APIs such as `Assembly.Location` t
 
 #### `Assembly.Location`
 
-`Assembly.Location`  for a bundled assembly could return:
+There are a few options to consider for the `Assembly.Location`  property of a bundled assembly:
 
 * A fixed literal (ex: `null`) indicating that no actual location is available.
 * The simple name of the assembly (with no path).
@@ -191,18 +191,25 @@ We need to determine the semantics of current APIs such as `Assembly.Location` t
 * In the case of files spilled to disk (ex: ready-to-run assemblies) -- the actual location of the extracted file.
 * A configurable selection of the above, etc.
 
-Most of the app development can be agnostic to whether the app is published as single-file or not. However, the parts of the app that deal with physical locations of files need to be aware of the single-file packaging. 
-
-In the first implementation, we propose keeping the default behavior of `Assembly.Location`. That is, 
+We propose keeping the default behavior of `Assembly.Location`. That is, 
 
 * If the assembly is loaded directly from the bundle, return empty-string
 * If the assembly is extracted to disk, return the location of the extracted file.
 
-The semantics of assembly location should be tuned further based on user feedback, and such that it aligns with other bundling technologies in .Net Core ecosystem.
+Most of the app development can be agnostic to whether the app is published as single-file or not. However, the parts of the app that deal with physical locations of files need to be aware of the single-file packaging. 
 
 #### `AppContext.BaseDirectory`
 
-If files are extracted out to disk, the `AppContext.BaseDirectory` is set to the directory where files are extracted. If all files in a bundle can be processed directly without need for extraction (an extraction directory is not created), and the `AppContext.BaseDirectory` is set to the directory where the AppHost resides.
+A few options to consider for the `AppContext.BaseDirectory` are:
+
+* The directory where embedded files are extracted out: This option enables easy access to content files spilled to disk. However, if no files need to be extracted to disk for an application bundle, there's no extraction directory. 
+* The extraction-directory if files are extracted, the AppHost directory otherwise. The limitation to this approach is that the value of  `AppContext.BaseDirectory`  changes with the way applications are packaged/executed.
+* The directory where `AppHost` binary resides (always).
+
+We propose that `AppContext.BaseDirectory`  should always be set to the directory where the `AppHost` bundle resides.  This scheme doesn't provide an obvious mechanism to access the contents of the extraction directory -- by design. The recommended method for accessing content files from the bundle are: 
+
+- Do not bundle application data files into the single-exe; instead them next to the bundle. This way, the application binary is a single-file, but not the whole application.
+- Embed data files as managed resources into application binary, and access them via resource management [APIs](<https://docs.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getmanifestresourcestream?view=netcore-3.0>). 
 
 ## Testing
 
--- a/accepted/single-file/design.md
+++ b/accepted/single-file/design.md
@@ -143,10 +143,6 @@ On Startup, the AppHost checks if it has embedded files. If so, it
 
 An app may choose to only embed some files (ex: due to licensing restrictions) and expect to pickup other dependencies from application-launch directory, nuget packages, etc. In order to resolve assemblies and native libraries, the embedded resources are probed first, followed by other probing paths. 
 
-#### App_Root Setting
-
-If files are extracted out to disk, the corehost `App_Root` is set to the directory where files are extracted. If all files in a bundle can be processed directly without need for extraction (an extraction directory is not created), and the `App_Root` is set to the directory where the AppHost resides.
-
 ### New API
 
 In this section, we propose adding a few APIs to facilitate common operations on bundled-apps.
@@ -182,7 +178,11 @@ We can also provide an abstraction that abstracts away the physical location of 
 
 ### Existing API
 
-We need to determine the semantics of current APIs such as `Assembly.Location` that return the information about an assembly's location on disk. For example, `Assembly.Location`  for a bundled assembly could return:
+We need to determine the semantics of current APIs such as `Assembly.Location` that return the information about an assembly's location on disk. 
+
+#### `Assembly.Location`
+
+`Assembly.Location`  for a bundled assembly could return:
 
 * A fixed literal (ex: `null`) indicating that no actual location is available.
 * The simple name of the assembly (with no path).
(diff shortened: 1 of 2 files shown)


accepted/single-file/design.md
```diff
@@ -135,28 +135,17 @@ Once the single-file-publish tooling is added to the publish pipeline, other sta
...
+* Memory maps the entire bundle file.
+* Extracts the necessary files (ex: native binaries) to disk (if any) as explained in this [document](extract.md). 
+* Sets up data-structures so other components can access files embedded directly, as explained in this [document](direct.md)
 
 #### Dependency Resolution
 
 An app may choose to only embed some files (ex: due to licensing restrictions) and expect to pickup other dependencies from application-launch directory, nuget packages, etc. In order to resolve assemblies and native libraries, the embedded resources are probed first, followed by other probing paths. 
 
-The communication between the host and the runtime will continue to be through properties such as `TRUSTED_PLATFORM_ASSEMBLIES`, `NATIVE_DLL_SEARCH_DIRECTORIES`, etc.  These properties will be populated with entries from the bundled files (using a special marker) and the extracted files.
+#### App_Root Setting
```
jkotas: `App_Root` is a name of internal implementation detail. The design should talk about observable behavior. Should this talk about AppBase?

accepted/single-file/design.md
```diff
@@ -135,28 +135,17 @@ Once the single-file-publish tooling is added to the publish pipeline, other sta
...
-
-Where
-
-* `il.dll` is a pure managed assembly that is loaded directly from the bundle
-* `r2r.dll` is a ready-to-run assembly that is bundled and extract out to disk
-* `dep.dll` is an assembly that is not bundled with the app,
-
-the TPA with single-exe publish will be (assuming `?` is the bundle marker):
-`?/il.dll`; `/var/tmp/.net/app/e53xf3/r2r.dll`; `/usr/local/nuget/dep.dll`
+If files are extracted out to disk, the corehost `App_Root` is set to the directory where files are extracted. If all files in a bundle can be processed directly without need for extraction (an extraction directory is not created), and the `App_Root` is set to the directory where the AppHost resides.
```
jkotas: I know I may have suggested this during our discussions, but I am not sure whether it is a good idea after thinking about it some more. It creates non-desirable coupling where extracting even a single file will change the setting globally.

accepted/single-file/direct.md
```diff
@@ -0,0 +1,85 @@
...
+// Returns:
+//  S_OK
+
+CORECLR_HOSTING_API(coreclr_register_bundle_reader,
+                    bool (*bundle_reader)(const char* name, 
+                                          const void** buffer, 
+                                          size_t* size));
+```
+
+`HostPolicy` communicates the location of files residing within the bundle with a special bundle-marker (`:`) in the CoreCLR properties such as `TRUSTED_PLATFORM_ASSEMBLIES`.
```
jkotas: `:` is path separator on Unix.

swaroop-sridhar: The `:` can additionally be used for the bundle marker, right? We need to parse the input correctly. 
For example: `:il1.dll::il2.dll:/path/r2r.dll`. I've fixed the example in this section to use the correct separator.

accepted/single-file/design.md
swaroop-sridhar: The above formulation was based on the constraints that the application's files should be accessible from `AppContext.BaseDirectory`. However, when the all files are loaded from within the bundle, there is no extraction-directory for the `AppContext.BaseDirectory` to point at.

We have a few options::
* Always set `AppContext.BaseDirectory` to the directory containing the AppHost. Content extraction directory can be accessed (when applicable) via APIs such as those proposed here: https://github.com/dotnet/corefx/issues/37435 

* When running from bundle, set `AppContext.BaseDirectory`  to a distinct value, such as `null`, empty-string, or `:`.

* Always create an extraction directory (or have a default location when there are no files to be extracted) and point `AppContext.BaseDirectory`  to that directory.

* Set `AppContext.BaseDirectory` as described in this PR. But this has the shortcoming that the `AppContext.BaseDirectory` value changes based on the contents of the bundle -- which may be surprising for developers.

This problem is similar to the issue of setting `Assembly.Location` value.

swaroop-sridhar: I've updated the PR to discuss the issue in terms of `AppContext.BaseDirectory`. Thanks.

jkotas: > Content extraction directory can be accessed (when applicable) via APIs such as those proposed 

I am worried about providing first class access to the extraction directory. Ideally, we would not have the extraction directory at all and we will hopefully get to a point in future where most apps do not need the extraction directory. Thus, the extraction directory should better be a hidden implementation detail. What about if we start with this: 

- Do not bundle application data files and leave them next to the .exe. (ie do what ASP.NET does today). Your app binary is a single file, but the app as a whole is not a single file.
- Or, attach data files to your app as managed resources. Access the data files via APIs for accessing managed resources.

accepted/single-file/direct.md
jkotas: I guess you can design a scheme where you overload meaning of `:`. 

Note that the TRUSTED_PLATFORM_ASSEMBLIES list is parsed using a simple `string.Split` call in number of places today, e.g.: https://github.com/adrianoc/cecilifier/blob/71a0fe93c3da12f4eb0c969db3ff01935eca6b6b/Cecilifier.Core.Tests/Framework/Utils.cs#L12 . They are not going to work anymore if you overload meaning of `:`.

The current TPA list design has corner case bugs like https://github.com/dotnet/core-setup/issues/4141. This will add more corner case bugs to it. I am wondering whether we should fix these corner cases as well while we are on it.

accepted/single-file/design.md
swaroop-sridhar: @jkotas, I've updated the PR, wrt `AppContext.Basedirectory` and `Assembly.Location` in this change https://github.com/dotnet/designs/pull/74/commits/9cde7edd390f74213e0d281ddc5d465107841ab0. Please take a look at it.

accepted/single-file/direct.md
swaroop-sridhar: The only reason I picked overloaded meaning for `:` is to not preclude another character from use within paths. I'm happy to use a different character (say `#`) -- I'm not sure if you're suggesting the change ?

jkotas: Since we are going change the format or meaning of `TRUSTED_PLATFORM_ASSEMBLIES`, I am suggesting that it would be useful to fix the other known problems with it - where it does not allow certain characters.

It looks like you are going to have to add callbacks to read the file. Would it make sense to handle everything using the callbacks? If the callback is registered, the runtime would not look at the `TRUSTED_PLATFORM_ASSEMBLIES` list at all. It would get us onto a path to solve all escaping problems with `TRUSTED_PLATFORM_ASSEMBLIES`.

The other alternatives I can think are escaping scheme for `TRUSTED_PLATFORM_ASSEMBLIES` that allows *any* characters; or start passing around more elaborate structures than just plain strings.

accepted/single-file/design.md
swaroop-sridhar: Under these rules, certain apps that build artifacts (ex: asp.net using `deps.json` to read compiler configurations) will need to adapt to the situation with additional work -- ex: pack the `deps.json` file as a managed resource as well as bundled file, or device a different protocol to communicate the information without touching the `deps.json` file.

accepted/single-file/direct.md
swaroop-sridhar: How about we say that the host passes two properties to the runtime (both strings using standard separator for the platform):
* `TRUSTED_BUNDLED_ASSEMBLIES` containing assemblies in the bundle and
* `TRUSTED_PLATFORM_ASSEMBLIES` containing remaining files on disk.

The TPA-binder will look for an assembly in `TRUSTED_BUNDLED_ASSEMBLIES` first and use a special call-back if found. Otherwise, it processes `TRUSTED_PLATFORM_ASSEMBLIES` as usual.
Similarly for `PLATFORM_RESOURCE_ROOTS`.

I think this change will cause minimal churn in the host-runtime handshake, while implementing the load-bundled-assemblies feature.

I agree that passing the information as structures or callbacks for TPA and related properties is a good idea. But we'll need to continue to support hosts that provide TPA in the original form. So it may introduce a bit of complexity, which is best handled independently.

jkotas: > The TPA-binder will look for an assembly in TRUSTED_BUNDLED_ASSEMBLIES first and use a special call-back if found.

I do not understand what is the benefit of having both `TRUSTED_BUNDLED_ASSEMBLIES` and the callback. The callback alone should be enough, no?

swaroop-sridhar: So, the host keeps `TRUSTED_BUNDLED_ASSEMBLIES` on it's side and returns a pointer to the PE image if found, right? Yes, that will work.

jkotas: Also, nit - I would not use `TRUSTED` in any new names. It is left-over from Silverlight and there is nothing trusted about these in .NET Core.

swaroop-sridhar: Thanks @Jan, I've updated the PR with these notes in this change: https://github.com/dotnet/designs/pull/74/commits/7c2e77a525b071f573ccc5866545d5d7ed6bd519

accepted/single-file/direct.md
```diff
@@ -0,0 +1,78 @@
...
+
+CORECLR_HOSTING_API(coreclr_register_bundle_reader,
+                    bool (*bundle_reader)(const char* name, 
+                                          const void** buffer, 
+                                          size_t* size));
+```
+
+`HostPolicy` maintains a list of bundled assemblies internally, and communicates the assemblies on disk (unbundled or extracted) to the runtime in `TRUSTED_PLATFORM_ASSEMBLIES`.
+
+In order to resolve an assembly, the TPA-Binder first checks if it is available in the bundle (by invoking the `bundle_reader` callback). If the assembly is available in the bundle, the runtime loads the binary from memory via PELoader. If the assembly is not available in the bundle, the runtime continues to resolve it using the TPA-list.
```
jkotas: It may be useful to think through 

Do we consider the `bundle_reader` callback to be a stable method, or do we consider it an implementation detail that we can change it at will? If it is the earlier, we may want to think a bit more about the signature to make sure that it will work well for AOT content. If it is the later, we can start with the one you are proposing.

swaroop-sridhar: `bundler_reader` is designed to be a matter of implementation, as a communication between apphost and the runtime. 

I checked with @janvorli, who said that the signature should be fine wrt loading ready-to-run images.

accepted/single-file/direct.md
jan: Great job, @swaroop-sridhar

PR #76, opened 2019-05-21
Single-File: Extraction configurations as build-time properties.

swaroop-sridhar:
This commit changes the way the location where files in a bundle
are extracted can be influenced at build time.

Originally, they were written as properties in `runtimeconfig.json` file,
which is complex to handle at runtime, because the apphost needs to
extract resources before `hostfxr` process `runtimeconfig.json`
(because hostfxr itself needs to be extracted).

Instead the configurations are now build-time properties, which can be
written to and processed from the bundle-headers.

### DIFF
--- a/accepted/single-file/extract.md
+++ b/accepted/single-file/extract.md
@@ -112,9 +112,14 @@ During startup, the Apphost:
 
 ### Extraction Configurations
 
-The following build-properties can be set to influence the extraction mechanism:
+The following settings can be used to configure the extraction mechanism:
 
-* `ExtractBaseDir` The base directory within which the files embedded in a single-file app are extracted, as explained in sections above. This directory can be set up an admin-only writable location if necessary.
-* `ExtractAllFiles`:  Extract all embedded dependencies in each run (instead of loading certain files directly from the bundle) .
+* Extract all embedded dependencies in each run (instead of loading certain files directly from the bundle) 
+  * Build time: Set the property `ExtractAllFiles= true`
+  * Run-time: Set the environment variable `DOTNET_BUNDLE_EXTRACT_ALL_FILES=true`
+* Set the base directory within which the files embedded in a single-file app are extracted, as explained in sections above. 
+  * Run-time: Set the `DOTNET_BUNDLE_EXTRACT_BASE_DIR` environment variable to the full path.
+  * There is no build-time configuration, since this is a path-setting on the machine where the app is run.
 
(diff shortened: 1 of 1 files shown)


accepted/single-file/extract.md
```diff
@@ -112,7 +112,7 @@ During startup, the Apphost:
 
 ### Extraction Configurations
 
-The following switches can be set in the application configuration file (`runtimeconfig.json`) to influence the extraction mechanism:
+The following build-properties can be set to influence the extraction mechanism:
 
 * `ExtractBaseDir` The base directory within which the files embedded in a single-file app are extracted, as explained in sections above. This directory can be set up an admin-only writable location if necessary.
```
jkotas: I do not think that `ExtractBaseDir` makes sense as a property to burn into the bundle.

swaroop-sridhar: I spoke with @jkotas about this.

The concern here is that:
* The path `ExtractBaseDir` is client-machine specific, and should not be configured at build time.
* For controlled environments where the target path is known, it can be set on the target machines by using the environment variable.

So, the proposal is to:
* Let `ExtractBaseDir` be only configurable at runtime via the  `DOTNET_BUNDLE_EXTRACT_BASE_DIR` 
* ExtractAllFiles will be a compile-time flag burned into the bundle-headers.

swaroop-sridhar: CC: @igloo15: wrt your comment here 
https://github.com/dotnet/coreclr/issues/20287#issuecomment-491094236 

I believe your scenario can work with configuring the extraction-base directory at runtime, while it may cost some additional work to configure the target machines.

igloo15: My only concern is using environment variables is tricky it requires me to write a bundler/starter script for the bundled exe that executes the exe and sets the environment variable. Thus I no longer can distribute just the exe which is my intention. I can't preconfigure the machine with the environment variable as the machine is a vm that is spun up from an image I don't control. 

I understand the issues with client-machine specific paths related to ExtractBaseDir but allowing maybe local to exe extraction might be nice. Where the files are extracted to the location of the bundled exe

swaroop-sridhar: @jkotas I've updated the text, per your recommendation. Thanks.

swaroop-sridhar: @igloo15, I'll think about adding the option to extract the files to the same directory. 
* This option is not suitable in all customer scenarios, since the directory where the app exists may not be writable, but it definitely a safer setting than writing in the full path.
* In the current version, extracting to an existing directory is not straight-forward; but this is a matter of implementation.

PR #88, opened 2020-01-03
single-file: Use unpredictable directory names on Unix

lpereira:


accepted/single-file/extract.md
```diff
@@ -78,11 +78,14 @@ For a single-file app, the extraction directory is `<base>/<app>/<bundle-id>`
 * `<base>` is 
 
   * `DOTNET_BUNDLE_EXTRACT_BASE_DIR` environment variable, if set.
-  * If not, defaults to 
-    * `%TEMP%\.net ` on Windows
-    * `$TMPDIR/.net` if `$TMPDIR` is set (Posix conforming OSes including Mac)
-    * Otherwise `/var/tmp/.net` (Ubuntu)  if the directory exists.
-    * Otherwise  `/tmp/.net` 
+  * If not, on Windows, defaults to `%TEMP%\.net`
+  * On Unix-like systems, where multiple users may use a single system, an approach that removes the possibility of name collisions and other users creating files to prevent an application to start (by a malicious user creating a predictable directory name) is used instead:
```
swaroop-sridhar: For Unix extraction directory, I think we should clarify that:
* In .net 3, the extraction directory is:
    * `$TMPDIR/<user-ID>/.net` if `$TMPDIR` is set. If not, 
    * `/var/tmp/<user-ID>/.net` if that exists, otherwise 
    * `/tmp/<user-ID>/.net` if that exists 
* In .net 5, the proposed new scheme is ...

accepted/single-file/extract.md
```diff
@@ -78,11 +78,14 @@ For a single-file app, the extraction directory is `<base>/<app>/<bundle-id>`
...
+     * The directory to extract the bundle is created with `mkdtemp()`, using the `$TMPDIR` environment variable (if set), `/var/tmp/` (if exists, because it survives reboots), falling back to `/tmp` (does not survive reboots, it's often a ramdisk) in the template (e.g. `/var/tmp/dotnet-<app>-XXXXXX`);
+     * Symbolic link is created in `~/.cache/dotnet/<app>/<bundle-id>` pointing to the directory created by `mkdtemp()`;
+         * `~/.cache` might be in another directory if `$XDG_CACHE_HOME` is set. See [the XDG spec for information](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html).
```
swaroop-sridhar: It would be good to write this as a declarative spec. So, I'd write something like:
If `$XDG_CACHE_HOME` is set, create a symbolic link at `$XDG_CACHE_HOME/.net`
Otherwise  `~/.cache/.net`


So, replace "`~/.cache/` might be " to "is".

accepted/single-file/extract.md
lpereira: Done

accepted/single-file/extract.md
lpereira: Done

accepted/single-file/extract.md
```diff
@@ -78,11 +78,20 @@ For a single-file app, the extraction directory is `<base>/<app>/<bundle-id>`
...
+     * For .NET 3, the extraction directory follows the following order:
+         * `${TMPDIR}/.net/${UID}` if `${TMPDIR}` is set; otherwise,
+         * `/var/tmp/.net/${UID}` if that exists; otherwise,
+         * `/tmp/.net/${UID}` if that exists; fails otherwise.
+     * For .NET 5, the proposal is to follow this scheme:
+         * The directory to extract the bundle is created with `mkdtemp()`, using the `$TMPDIR` environment variable (if set), `/var/tmp/` (if exists, because it survives reboots), falling back to `/tmp` (does not survive reboots, it's often a ramdisk) in the template (e.g. `/var/tmp/dotnet-<app>-XXXXXX`);
+         * A symbolic link to the directory created by `mkdtemp()` is created in a predictable location:
+             * If `${XDG_CACHE_HOME}` is set, the symlink is created under `${XDG_CACHE_HOME}/.cache/dotnet/<app>/<bundle-id>` (See [the XDG spec for information](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)); otherwise,
```
tmds: @swaroop-sridhar @lpereira @jkotas we've been setting DOTNET_BUNDLE_EXTRACT_BASE_DIR to `$XDG_CACHE_HOME/dotnet_bundle_extract` in the Fedora and RHEL .NET Core 3.x installers.

A default under `XDG_CACHE_HOME` would be good. I'm not sure about the gains of symlinking to `/var/tmp/`+`mkdtemp`.
Putting everything under `XDG_CACHE_HOME` avoids any multi-user issue.

cc @omajid

lpereira: We can change this to extract under `XDG_CACHE_HOME` if it's set -- and not provide any symlinks -- but go with the symlink route if it's not set.  Not all environments will have `XDG_CACHE_HOME` set (for instance, container images and whatnot), and this has to work regardless. I'll noodle on this.

The main issue, though, is that the cache directory isn't always cleaned up, whereas in most cases, the system will take care of temporary directories, be it because they were `tmpfs` in the first place, or because a cronjob will delete old files from persistent temporary directories.  I worry that the directory under `XDG_CACHE_HOME` will be polluted with old versions of applications that are not being used any longer; stale symlinks on the other hand are easy and cheap to clean up.

swaroop-sridhar: Yes, if `$DOTNET_BUNDLE_EXTRACT_BASE_DIR` is set, no other directory will be used.
As @lpereira  said, the default directory has been within `$TMPDIR` for the desire of cleanup. Thanks.

PR #93, opened 2020-02-26
Add feature proposal for AsnReader/AsnWriter

bartonjs:


galvesribeiro:
This would be awesome to be added to the CLR. I see practical use cases while developing EMV kernels that deal with PIN cryptography.

SteveSyfuhs:
This would be a very useful addition. The AsnReader/Writer in their current form work quite well so much so that I forked it into a separate library for it's use. I would love to get rid of that, and just use what's built in to the framework. 

Some things I found a bit annoying and necessary to tweak though:

 * String Encoding -- It doesn't support GeneralString. Short of explicitly adding support, it would be nice to have a way to load in customized encodings so callers can fill in the gaps as necessary (GeneralString in *most* use cases simply follows the IA5 encoding, but that assumption is dangerous).
 * GeneralizedTime -- It would be nice to set omitFractionalSeconds globally. It's a bit tedious to indicate everywhere when the protocol explicitly requires it or not.
 * Code generation -- Oh wouldn't it be delightful if I could just take an ASN.1 schema and convert it to code instead of manually converting to to XML first? ;)

bartonjs:
> Code generation -- Oh wouldn't it be delightful if I could just take an ASN.1 schema and convert it to code instead of manually converting to to XML first? ;)

Code generation is called out as a non-goal for this spec (the asn.xslt isn't included with it), deferring it to later.  An ASN.1 -> C# compiler would also be a nice feature; or even a slim version that runs with the new (still under development) C# compile/codegen project.  But they can be done in terms of this feature set.

> GeneralizedTime -- It would be nice to set omitFractionalSeconds globally.

How would you want that to work?

```C#
partial class AsnWriter
{
    // Maybe an optional ctor parameter instead of a settable property so the state can't get changed up by a helper method?
    public bool OmitFractionalSeconds { get; set; }
 
    public void WriteGeneralizedTime(DateTimeOffset value) => WriteGeneralizedTime(value, OmitFractionalSeconds);
    public void WriteGeneralizedTime(DateTimeOffset value, bool omitFractionalSeconds) => throw null;
}
```

> String Encoding -- It doesn't support GeneralString.

Partially because I don't understand all the stuff in the table about escape codes and the like.  I'm not opposed to adding it, if someone can convince me they understand the encoding :smile:.  It's obviously "IA5 + some stuff", and "+ some stuff" that I can't understand scares me in security-relevant code.

>  it would be nice to have a way to load in customized encodings so callers can fill in the gaps as necessary 

What would you want that to look like?  It's currently possible by using PeekTag / PeekContentBytes; or TryReadPrimitiveCharacterStringBytes and throwing it at a custom Encoding object.

SteveSyfuhs:
Passing `OmitFractionalSeconds` through the constructor seems like it would open up an odd precedent, but I do agree it should be readonly. An `AsnOptions { OmitFractionalSeconds = true }` might be another option too, that way any future things can be added without modifying the signature. Then the implementation could be

```csharp
partial class AsnWriter
{
    public AsnWriter(AsnOptions options)
    {
	OmitFractionalSeconds = options.OmitFractionalSeconds;
    }
	
    public bool OmitFractionalSeconds { get; }
 
    public void WriteGeneralizedTime(DateTimeOffset value) => WriteGeneralizedTime(value, OmitFractionalSeconds);
    public void WriteGeneralizedTime(DateTimeOffset value, bool omitFractionalSeconds) => throw null;
}
```

I can't speak to what an effective general-use implementation of GeneralString encoding would look like unfortunately and I agree that fudging it in security paths is bad. For reference here's how [Kerberos uses this](https://tools.ietf.org/html/rfc4120#section-5.2.1):

> In practice, many implementations treat GeneralStrings as if they
> were 8-bit strings of whichever character set the implementation
> defaults to, without regard to correct usage of character-set
> designation escape sequences.  The default character set is often
> determined by the current user's operating system-dependent locale.
> At least one major implementation places unescaped UTF-8 encoded
> Unicode characters in the GeneralString.  This failure to adhere to
> the GeneralString specifications results in interoperability issues
> when conflicting character encodings are utilized by the Kerberos
> clients, services, and KDC.
> 
> This unfortunate situation is the result of improper documentation of
> the restrictions of the ASN.1 GeneralString type in prior Kerberos
> specifications.
> 
> The new (post-RFC 1510) type KerberosString, defined below, is a
> GeneralString that is constrained to contain only characters in
> IA5String.
> 
>   KerberosString  ::= GeneralString (IA5String)

This might be where the `AsnOptions` object becomes more useful. Adding a dictionary property that lets you overload encoding types:

```csharp
AsnOptions options = new AsnOptions() 
{
   Encodings = new Dictionary<UniversalTagNumber, Encoding> {
      [UniversalTagNumber.GeneralString] = new GeneralStringEncoding()
   }
};

...

public class GeneralStringEncoding : IA5Encoding { ... }
```

Another option is adding a flag `bool TreatGeneralStringAsIA5Encoding`, though that makes future compatibility awkward if the encoding is actually implemented.

terrajobst:
@bartonjs 

Please update the filename to be in `accepted/2020` instead

bartonjs:
@terrajobst Isn't that where it is already? (accepted/2020/\[featurename\]/\[featuername\].md)

terrajobst:
Didn’t check — just merged my PR and assumed you need to update 😀

terrajobst:
Uploaded the notes for our first review [here](https://github.com/dotnet/apireviews/pull/113)

filipnavara:
Not sure where it is appropriate to comment but I guess this is the place. The API review touched on the namespace question. I am personally not a big fan of `System.Protocols[.Asn1]`. I think of ASN.1  more as a "data format" which is used by [communication] protocol(s), not a protocol by itself.

There's already `System.Text` namespace, how about `System.Binary.Asn1`?

shaggygi:
This is interesting as BACnet is described by ASN.1.
For example, http://bacnetbill.blogspot.com/2009/10/bacnet-tagging-rules.html

bartonjs:
@shaggygi It looks like BACnet uses the ASN.1 description language, but uses an encoding that's not compatible with the BER family of encodings for ASN.1 values.

It's a little unclear if BACnet's encoding would even be structurally compatible with the reader here.  It /might/ work, but it might be better off with its own similar structure.

terrajobst:
@bartonjs, should this be merged now? We're done with the API review, right?

bartonjs:
> We're done with the API review, right?

The exiting note from round 2 was effectively "look into replacing AsnValueReader with a collection of stateless statics" (working name: `public static class AsnDecoder`).

I'm happy with merging this now and opening a new PR to edit it for the AsnValueReader => AsnDecoder changes; or we can leave this and I'll just update it when I'm confident on the flow... whatever works best.

bartonjs:
API shape update from API review (rounds 1 and 2) as well as some of the open questions:

```diff
diff --git a/asnref.before.cs b/asnref.after2.cs
index 4264537c496..0345bc21493 100644
--- a/asnref.before.cs
+++ b/asnref.after.cs
@@ -1,345 +1,250 @@
 // Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 // ------------------------------------------------------------------------------
 // Changes to this file must follow the https://aka.ms/api-review process.
 // ------------------------------------------------------------------------------
 
+using System.Collections;
 using System.Numerics;
 
-namespace System.Security.Cryptography.Asn1
+namespace System.Formats.Asn1
 {
     public struct Asn1Tag : IEquatable<Asn1Tag>
     {
         private int _dummyPrimitive;
         public static readonly Asn1Tag Boolean;
         public static readonly Asn1Tag ConstructedBitString;
         public static readonly Asn1Tag ConstructedOctetString;
-        public static readonly Asn1Tag EndOfContents;
         public static readonly Asn1Tag Enumerated;
         public static readonly Asn1Tag GeneralizedTime;
         public static readonly Asn1Tag Integer;
         public static readonly Asn1Tag Null;
         public static readonly Asn1Tag ObjectIdentifier;
         public static readonly Asn1Tag PrimitiveBitString;
         public static readonly Asn1Tag PrimitiveOctetString;
         public static readonly Asn1Tag Sequence;
         public static readonly Asn1Tag SetOf;
         public static readonly Asn1Tag UtcTime;
         public Asn1Tag(TagClass tagClass, int tagValue, bool isConstructed = false) { throw null; }
         public Asn1Tag(UniversalTagNumber universalTagNumber, bool isConstructed = false) { throw null; }
         public bool IsConstructed { get { throw null; } }
         public TagClass TagClass { get { throw null; } }
         public readonly int TagValue { get { throw null; } }
         public Asn1Tag AsConstructed() { throw null; }
         public Asn1Tag AsPrimitive() { throw null; }
         public int CalculateEncodedSize() { throw null; }
+        public static Asn1Tag Decode(ReadOnlySpan<byte> source, out int bytesConsumed) { throw null; }
         public int Encode(Span<byte> destination) { throw null; }
         public bool Equals(Asn1Tag other) { throw null; }
         public override bool Equals(object? obj) { throw null; }
         public override int GetHashCode() { throw null; }
         public bool HasSameClassAndValue(Asn1Tag other) { throw null; }
         public static bool operator ==(Asn1Tag left, Asn1Tag right) { throw null; }
         public static bool operator !=(Asn1Tag left, Asn1Tag right) { throw null; }
         public override string ToString() { throw null; }
         public static bool TryDecode(ReadOnlySpan<byte> source, out Asn1Tag tag, out int bytesConsumed) { throw null; }
         public bool TryEncode(Span<byte> destination, out int bytesWritten) { throw null; }
     }
+    public sealed class AsnContentException : Exception
+    {
+        public AsnContentException() { }
+        public AsnContentException(string message) { }
+        public AsnContentException(string message, Exception inner) { }
+    }
-    public ref struct AsnValueReader
+    public static class AsnDecoder
     {
-        private object _dummy;
-        private int _dummyPrimitive;
-        public AsnValueReader(ReadOnlySpan<byte> data, AsnEncodingRules ruleSet) { throw null; }
-        public bool HasData { get { throw null; } }
-        public readonly AsnEncodingRules RuleSet { get { throw null; } }

-        public ReadOnlySpan<byte> PeekContentBytes() { throw null; }
-        public ReadOnlySpan<byte> PeekEncodedValue() { throw null; }
-        public Asn1Tag PeekTag() { throw null; }
-        public ReadOnlySpan<byte> ReadEncodedValue() { throw null; }
+        public static Asn1Tag ReadEncodedValue(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int contentOffset, out int contentLength, out int bytesConsumed) { throw null; }
+        public static bool TryReadEncodedValue(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out Asn1Tag tag, out int contentOffset, out int contentLength, out int bytesConsumed) { throw null; }

-        public bool ReadBoolean() { throw null; }
-        public bool ReadBoolean(Asn1Tag expectedTag) { throw null; }
+        public static bool ReadBoolean(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }

-        public BigInteger ReadInteger() { throw null; }
-        public BigInteger ReadInteger(Asn1Tag expectedTag) { throw null; }
+        public static BigInteger ReadInteger(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public ReadOnlySpan<byte> ReadIntegerBytes() { throw null; }
-        public ReadOnlySpan<byte> ReadIntegerBytes(Asn1Tag expectedTag) { throw null; }
+        public static ReadOnlySpan<byte> ReadIntegerBytes(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadInt16(out short value) { throw null; }
-        public bool TryReadInt16(Asn1Tag expectedTag, out short value) { throw null; }
-        public bool TryReadInt32(out int value) { throw null; }
-        public bool TryReadInt32(Asn1Tag expectedTag, out int value) { throw null; }
+        public static bool TryReadInt32(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int value, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadInt64(out long value) { throw null; }
-        public bool TryReadInt64(Asn1Tag expectedTag, out long value) { throw null; }
+        public static bool TryReadInt64(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out long value, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadInt8(out sbyte value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadInt8(Asn1Tag expectedTag, out sbyte value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt16(Asn1Tag expectedTag, out ushort value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt16(out ushort value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt32(Asn1Tag expectedTag, out uint value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt32(out uint value) { throw null; }
+        [CLSCompliantAttribute(false)] public static bool TryReadUInt32(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out uint value, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt64(Asn1Tag expectedTag, out ulong value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt64(out ulong value) { throw null; }
+        [CLSCompliantAttribute(false)] public static bool TryReadUInt64(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out ulong value, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadUInt8(out byte value) { throw null; }
-        public bool TryReadUInt8(Asn1Tag expectedTag, out byte value) { throw null; }

-        public byte[] ReadBitString(out int unusedBitCount) { throw null; }
-        public byte[] ReadBitString(Asn1Tag expectedTag, out int unusedBitCount) { throw null; }
+        public static byte[] ReadBitString(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int unusedBitCount, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyBitStringBytes(Asn1Tag expectedTag, Span<byte> destination, out int unusedBitCount, out int bytesWritten) { throw null; }
-        public bool TryCopyBitStringBytes(Span<byte> destination, out int unusedBitCount, out int bytesWritten) { throw null; }
+        public static bool TryReadBitString(ReadOnlySpan<byte> source, Span<byte> destination, AsnEncodingRules ruleSet, out int unusedBitCount, out int bytesConsumed, out int bytesWritten, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadPrimitiveBitStringValue(out int unusedBitCount, out ReadOnlySpan<byte> value) { throw null; }
-        public bool TryReadPrimitiveBitStringValue(Asn1Tag expectedTag, out int unusedBitCount, out ReadOnlySpan<byte> value) { throw null; }
+        public static bool TryReadPrimitiveBitString(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int unusedBitCount, out ReadOnlySpan<byte> value, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }

+        public static BitArray ReadNamedBitList(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public Enum ReadNamedBitListValue(Asn1Tag expectedTag, Type tFlagsEnum) { throw null; }
-        public Enum ReadNamedBitListValue(Type tFlagsEnum) { throw null; }
+        public static Enum ReadNamedBitListValue(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, Type tFlagsEnum, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public TFlagsEnum ReadNamedBitListValue<TFlagsEnum>() where TFlagsEnum : struct { throw null; }
-        public TFlagsEnum ReadNamedBitListValue<TFlagsEnum>(Asn1Tag expectedTag) where TFlagsEnum : struct { throw null; }
+        public static TFlagsEnum ReadNamedBitListValue<TFlagsEnum>(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) where TFlagsEnum : Enum { throw null; }

-        public byte[] ReadOctetString() { throw null; }
-        public byte[] ReadOctetString(Asn1Tag expectedTag) { throw null; }
+        public static byte[] ReadOctetString(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyOctetStringBytes(Asn1Tag expectedTag, Span<byte> destination, out int bytesWritten) { throw null; }
-        public bool TryCopyOctetStringBytes(Span<byte> destination, out int bytesWritten) { throw null; }
+        public static bool TryReadOctetString(ReadOnlySpan<byte> source, Span<byte> destination, AsnEncodingRules ruleSet, out int bytesConsumed, out int bytesWritten, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadPrimitiveOctetStringBytes(out ReadOnlySpan<byte> contents) { throw null; }
-        public bool TryReadPrimitiveOctetStringBytes(Asn1Tag expectedTag, out ReadOnlySpan<byte> contents) { throw null; }
+        public static bool TryReadPrimitiveOctetString(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out ReadOnlySpan<byte> value, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }

-        public void ReadNull() { }
-        public void ReadNull(Asn1Tag expectedTag) { }
+        public static void ReadNull(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }

-        public Oid ReadObjectIdentifier() { throw null; }
-        public Oid ReadObjectIdentifier(Asn1Tag expectedTag) { throw null; }
+        public static string ReadObjectIdentifier(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public string ReadObjectIdentifierAsString() { throw null; }
-        public string ReadObjectIdentifierAsString(Asn1Tag expectedTag) { throw null; }

-        public ReadOnlySpan<byte> ReadEnumeratedBytes() { throw null; }
-        public ReadOnlySpan<byte> ReadEnumeratedBytes(Asn1Tag expectedTag) { throw null; }
+        public static ReadOnlySpan<byte> ReadEnumeratedBytes(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public Enum ReadEnumeratedValue(Asn1Tag expectedTag, Type tEnum) { throw null; }
-        public Enum ReadEnumeratedValue(Type tEnum) { throw null; }
+        public static Enum ReadEnumeratedValue(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, Type tEnum, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public TEnum ReadEnumeratedValue<TEnum>() where TEnum : struct { throw null; }
-        public TEnum ReadEnumeratedValue<TEnum>(Asn1Tag expectedTag) where TEnum : struct { throw null; }
+        public static TEnum ReadEnumeratedValue<TEnum>(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) where TEnum : Enum { throw null; }

-        public AsnValueReader ReadSequence() { throw null; }
-        public AsnValueReader ReadSequence(Asn1Tag expectedTag) { throw null; }
+        public static void ReadSequence(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int contentOffset, out int contentLength, out int bytesConsumed, Asn1Tag? expectedTag) { throw null; }
-        public AsnValueReader ReadSetOf(bool skipSortOrderValidation = false) { throw null; }
-        public AsnValueReader ReadSetOf(Asn1Tag expectedTag, bool skipSortOrderValidation = false) { throw null; }
+        public static void ReadSetOf(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int contentOffset, out int contentLength, out int bytesConsumed, bool skipSortOrderValidation = false, Asn1Tag? expectedTag = default) { throw null; }

-        public DateTimeOffset ReadUtcTime(int twoDigitYearMax = 2049) { throw null; }
-        public DateTimeOffset ReadUtcTime(Asn1Tag expectedTag, int twoDigitYearMax = 2049) { throw null; }
+        public static DateTimeOffset ReadUtcTime(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, int twoDigitYearMax = 2049, Asn1Tag? expectedTag = default) { throw null; }

-        public DateTimeOffset ReadGeneralizedTime(bool disallowFractions = false) { throw null; }
-        public DateTimeOffset ReadGeneralizedTime(Asn1Tag expectedTag, bool disallowFractions = false) { throw null; }
+        public static DateTimeOffset ReadGeneralizedTime(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }

-        public string ReadCharacterString(Asn1Tag expectedTag, UniversalTagNumber encodingType) { throw null; }
-        public string ReadCharacterString(UniversalTagNumber encodingType) { throw null; }
+        public static string ReadCharacterString(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, UniversalTagNumber encodingType, out int bytesConsumed, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyCharacterString(Asn1Tag expectedTag, UniversalTagNumber encodingType, Span<char> destination, out int charsWritten) { throw null; }
-        public bool TryCopyCharacterString(UniversalTagNumber encodingType, Span<char> destination, out int charsWritten) { throw null; }
+        public static bool TryReadCharacterString(ReadOnlySpan<byte> source, Span<char> destination, AsnEncodingRules ruleSet, UniversalTagNumber encodingType, out int bytesConsumed, out int charsWritten, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyCharacterStringBytes(Asn1Tag expectedTag, UniversalTagNumber encodingType, Span<byte> destination, out int bytesWritten) { throw null; }
-        public bool TryCopyCharacterStringBytes(UniversalTagNumber encodingType, Span<byte> destination, out int bytesWritten) { throw null; }
+        public static bool TryReadCharacterStringBytes(ReadOnlySpan<byte> source, Span<byte> destination, AsnEncodingRules ruleSet, Asn1Tag expectedTag, out int bytesWritten, out int bytesConsumed) { throw null; }
-        public bool TryReadPrimitiveCharacterStringBytes(Asn1Tag expectedTag, UniversalTagNumber encodingType, out ReadOnlySpan<byte> contents) { throw null; }
-        public bool TryReadPrimitiveCharacterStringBytes(UniversalTagNumber encodingType, out ReadOnlySpan<byte> contents) { throw null; }
+        public static bool TryReadPrimitiveCharacterStringBytes(ReadOnlySpan<byte> source, AsnEncodingRules ruleSet, Asn1Tag expectedTag, out ReadOnlySpan<byte> value, out int bytesConsumed) { throw null; }

-        public void ThrowIfNotEmpty() { }
     }
     public enum AsnEncodingRules
     {
         BER = 0,
         CER = 1,
         DER = 2
     }
     public class AsnReader
     {
         public AsnReader(ReadOnlyMemory<byte> data, AsnEncodingRules ruleSet) { }
+        public AsnReader(ReadOnlyMemory<byte> data, AsnReaderOptions options) { }
         public bool HasData { get { throw null; } }
         public AsnEncodingRules RuleSet { get { throw null; } }
         public ReadOnlyMemory<byte> PeekContentBytes() { throw null; }
         public ReadOnlyMemory<byte> PeekEncodedValue() { throw null; }
         public Asn1Tag PeekTag() { throw null; }
         public ReadOnlyMemory<byte> ReadEncodedValue() { throw null; }
         public void ThrowIfNotEmpty() { }

-        public bool ReadBoolean() { throw null; }
-        public bool ReadBoolean(Asn1Tag expectedTag) { throw null; }
+        public bool ReadBoolean(Asn1Tag? expectedTag = default) { throw null; }

-        public BigInteger ReadInteger() { throw null; }
-        public BigInteger ReadInteger(Asn1Tag expectedTag) { throw null; }
+        public BigInteger ReadInteger(Asn1Tag? expectedTag = default) { throw null; }
-        public ReadOnlyMemory<byte> ReadIntegerBytes() { throw null; }
-        public ReadOnlyMemory<byte> ReadIntegerBytes(Asn1Tag expectedTag) { throw null; }
+        public ReadOnlyMemory<byte> ReadIntegerBytes(Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadInt16(out short value) { throw null; }
-        public bool TryReadInt16(Asn1Tag expectedTag, out short value) { throw null; }
-        public bool TryReadInt32(out int value) { throw null; }
-        public bool TryReadInt32(Asn1Tag expectedTag, out int value) { throw null; }
+        public bool TryReadInt32(out int value, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadInt64(out long value) { throw null; }
-        public bool TryReadInt64(Asn1Tag expectedTag, out long value) { throw null; }
+        public bool TryReadInt64(out long value, Asn1Tag? expectedTag = default) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadInt8(out sbyte value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadInt8(Asn1Tag expectedTag, out sbyte value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt16(Asn1Tag expectedTag, out ushort value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt16(out ushort value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt32(Asn1Tag expectedTag, out uint value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt32(out uint value) { throw null; }
+        [CLSCompliantAttribute(false)] public bool TryReadUInt32(out uint value, Asn1Tag? expectedTag = default) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt64(Asn1Tag expectedTag, out ulong value) { throw null; }
-        [CLSCompliantAttribute(false)] public bool TryReadUInt64(out ulong value) { throw null; }
+        [CLSCompliantAttribute(false)] public bool TryReadUInt64(out ulong value, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadUInt8(out byte value) { throw null; }
-        public bool TryReadUInt8(Asn1Tag expectedTag, out byte value) { throw null; }

-        public byte[] ReadBitString(out int unusedBitCount) { throw null; }
-        public byte[] ReadBitString(Asn1Tag expectedTag, out int unusedBitCount) { throw null; }
+        public byte[] ReadBitString(out int unusedBitCount, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyBitStringBytes(Asn1Tag expectedTag, Span<byte> destination, out int unusedBitCount, out int bytesWritten) { throw null; }
-        public bool TryCopyBitStringBytes(Span<byte> destination, out int unusedBitCount, out int bytesWritten) { throw null; }
+        public bool TryReadBitString(Span<byte> destination, out int unusedBitCount, out int bytesWritten, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadPrimitiveBitStringValue(out int unusedBitCount, out ReadOnlyMemory<byte> value) { throw null; }
-        public bool TryReadPrimitiveBitStringValue(Asn1Tag expectedTag, out int unusedBitCount, out ReadOnlyMemory<byte> value) { throw null; }
+        public bool TryReadPrimitiveBitString(out int unusedBitCount, out ReadOnlyMemory<byte> value, Asn1Tag? expectedTag = default) { throw null; }

+        public BitArray ReadNamedBitList(Asn1Tag? expectedTag = default) { throw null; }
-        public Enum ReadNamedBitListValue(Asn1Tag expectedTag, Type tFlagsEnum) { throw null; }
-        public Enum ReadNamedBitListValue(Type tFlagsEnum) { throw null; }
+        public Enum ReadNamedBitListValue(Type tFlagsEnum, Asn1Tag? expectedTag = default) { throw null; }
-        public TFlagsEnum ReadNamedBitListValue<TFlagsEnum>() where TFlagsEnum : struct { throw null; }
-        public TFlagsEnum ReadNamedBitListValue<TFlagsEnum>(Asn1Tag expectedTag) where TFlagsEnum : struct { throw null; }
+        public TFlagsEnum ReadNamedBitListValue<TFlagsEnum>(Asn1Tag? expectedTag = default) where TFlagsEnum : Enum { throw null; }

-        public byte[] ReadOctetString() { throw null; }
-        public byte[] ReadOctetString(Asn1Tag expectedTag) { throw null; }
+        public byte[] ReadOctetString(Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyOctetStringBytes(Asn1Tag expectedTag, Span<byte> destination, out int bytesWritten) { throw null; }
-        public bool TryCopyOctetStringBytes(Span<byte> destination, out int bytesWritten) { throw null; }
+        public bool TryReadOctetString(Span<byte> destination, out int bytesWritten, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryReadPrimitiveOctetStringBytes(out ReadOnlyMemory<byte> contents) { throw null; }
-        public bool TryReadPrimitiveOctetStringBytes(Asn1Tag expectedTag, out ReadOnlyMemory<byte> contents) { throw null; }
+        public bool TryReadPrimitiveOctetString(out ReadOnlyMemory<byte> contents, Asn1Tag? expectedTag = default) { throw null; }

-        public void ReadNull() { }
-        public void ReadNull(Asn1Tag expectedTag) { }
+        public void ReadNull(Asn1Tag? expectedTag = default) { }

-        public ReadOnlyMemory<byte> ReadEnumeratedBytes() { throw null; }
-        public ReadOnlyMemory<byte> ReadEnumeratedBytes(Asn1Tag expectedTag) { throw null; }
+        public ReadOnlyMemory<byte> ReadEnumeratedBytes(Asn1Tag? expectedTag = default) { throw null; }

-        public Enum ReadEnumeratedValue(Asn1Tag expectedTag, Type tEnum) { throw null; }
-        public Enum ReadEnumeratedValue(Type tEnum) { throw null; }
+        public Enum ReadEnumeratedValue(Type tEnum, Asn1Tag? expectedTag = default) { throw null; }
-        public TEnum ReadEnumeratedValue<TEnum>() where TEnum : struct { throw null; }
-        public TEnum ReadEnumeratedValue<TEnum>(Asn1Tag expectedTag) where TEnum : struct { throw null; }
+        public TEnum ReadEnumeratedValue<TEnum>(Asn1Tag? expectedTag = default) where TEnum : Enum { throw null; }

-        public DateTimeOffset ReadGeneralizedTime(bool disallowFractions = false) { throw null; }
-        public DateTimeOffset ReadGeneralizedTime(Asn1Tag expectedTag, bool disallowFractions = false) { throw null; }
+        public DateTimeOffset ReadGeneralizedTime(Asn1Tag? expectedTag = default) { throw null; }

-        public Oid ReadObjectIdentifier() { throw null; }
-        public Oid ReadObjectIdentifier(Asn1Tag expectedTag) { throw null; }
+        public string ReadObjectIdentifier(Asn1Tag? expectedTag = default) { throw null; }
-        public string ReadObjectIdentifierAsString() { throw null; }
-        public string ReadObjectIdentifierAsString(Asn1Tag expectedTag) { throw null; }

-        public AsnReader ReadSequence() { throw null; }
-        public AsnReader ReadSequence(Asn1Tag expectedTag) { throw null; }
+        public AsnReader ReadSequence(Asn1Tag? expectedTag = default) { throw null; }
-        public AsnReader ReadSetOf(bool skipSortOrderValidation = false) { throw null; }
-        public AsnReader ReadSetOf(Asn1Tag expectedTag, bool skipSortOrderValidation = false) { throw null; }
+        public AsnReader ReadSetOf(bool skipSortOrderValidation, Asn1Tag? expectedTag = default) { throw null; }
+        public AsnReader ReadSetOf(Asn1Tag? expectedTag = default) { throw null; }

-        public DateTimeOffset ReadUtcTime(int twoDigitYearMax = 2049) { throw null; }
-        public DateTimeOffset ReadUtcTime(Asn1Tag expectedTag, int twoDigitYearMax = 2049) { throw null; }
+        public DateTimeOffset ReadUtcTime(int twoDigitYearMax, Asn1Tag? expectedTag = default) { throw null; }
+        public DateTimeOffset ReadUtcTime(Asn1Tag? expectedTag = default) { throw null; }

-        public string ReadCharacterString(Asn1Tag expectedTag, UniversalTagNumber encodingType) { throw null; }
-        public string ReadCharacterString(UniversalTagNumber encodingType) { throw null; }
+        public string ReadCharacterString(UniversalTagNumber encodingType, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyCharacterString(Asn1Tag expectedTag, UniversalTagNumber encodingType, Span<char> destination, out int charsWritten) { throw null; }
-        public bool TryCopyCharacterString(UniversalTagNumber encodingType, Span<char> destination, out int charsWritten) { throw null; }
+        public bool TryReadCharacterString(Span<char> destination, UniversalTagNumber encodingType, out int charsWritten, Asn1Tag? expectedTag = default) { throw null; }
-        public bool TryCopyCharacterStringBytes(Asn1Tag expectedTag, UniversalTagNumber encodingType, Span<byte> destination, out int bytesWritten) { throw null; }
-        public bool TryCopyCharacterStringBytes(UniversalTagNumber encodingType, Span<byte> destination, out int bytesWritten) { throw null; }
+        public bool TryReadCharacterStringBytes(Span<byte> destination, Asn1Tag expectedTag, out int bytesWritten) { throw null; }
-        public bool TryReadPrimitiveCharacterStringBytes(Asn1Tag expectedTag, UniversalTagNumber encodingType, out ReadOnlyMemory<byte> contents) { throw null; }
-        public bool TryReadPrimitiveCharacterStringBytes(UniversalTagNumber encodingType, out ReadOnlyMemory<byte> contents) { throw null; }
+        public bool TryReadPrimitiveCharacterStringBytes(Asn1Tag expectedTag, out ReadOnlyMemory<byte> contents) { throw null; }
     }
+    public struct AsnReaderOptions
+    {
+        private int _dummyPrimitive;
+        public AsnReaderOptions(AsnEncodingRules ruleSet) { throw null; }
+        public AsnEncodingRules RuleSet { get { throw null; } }
+        public bool SkipSetSortOrderVerification { get { throw null; } set { } }
+        public int UtcTimeTwoDigitYearMax { get { throw null; } set { } }
+    }
-    public sealed class AsnWriter : IDisposable
+    public sealed class AsnWriter
     {
         public AsnWriter(AsnEncodingRules ruleSet) { }
         public AsnEncodingRules RuleSet { get { throw null; } }
+        public void CopyTo(AsnWriter destination) { }
-        public void Dispose() { }
         public byte[] Encode() { throw null; }
         public int GetEncodedLength() { throw null; }
         public void Reset() { }
         public bool TryEncode(Span<byte> destination, out int bytesWritten) { throw null; }
+        public bool EncodedValueEquals(AsnWriter other) { throw null; }
-        public bool ValueEquals(ReadOnlySpan<byte> other) { throw null; }
+        public bool EncodedValueEquals(ReadOnlySpan<byte> other) { throw null; }
-        public void WriteEncodedValue(ReadOnlySpan<byte> preEncodedValue) { }
+        public void WriteEncodedValue(ReadOnlySpan<byte> value) { }

-        public void WriteBoolean(Asn1Tag tag, bool value) { }
-        public void WriteBoolean(bool value) { }
+        public void WriteBoolean(bool value, Asn1Tag? tag = default) { }

-        public void WriteInteger(Asn1Tag tag, long value) { }
-        public void WriteInteger(long value) { }
+        public void WriteInteger(long value, Asn1Tag? tag = default) { }
-        public void WriteInteger(Asn1Tag tag, BigInteger value) { }
-        public void WriteInteger(BigInteger value) { }
+        public void WriteInteger(BigInteger value, Asn1Tag? tag = default) { }
-        public void WriteInteger(Asn1Tag tag, ReadOnlySpan<byte> value) { }
-        public void WriteInteger(ReadOnlySpan<byte> value) { }
+        public void WriteInteger(ReadOnlySpan<byte> value, Asn1Tag? tag = default) { }
-        [CLSCompliantAttribute(false)] public void WriteInteger(Asn1Tag tag, ulong value) { }
-        [CLSCompliantAttribute(false)] public void WriteInteger(ulong value) { }
+        [CLSCompliantAttribute(false)] public void WriteInteger(ulong value, Asn1Tag? tag = default) { }
-        public void WriteIntegerUnsigned(Asn1Tag tag, ReadOnlySpan<byte> value) { }
-        public void WriteIntegerUnsigned(ReadOnlySpan<byte> value) { }
+        public void WriteIntegerUnsigned(ReadOnlySpan<byte> value, Asn1Tag? tag = default) { }

-        public void WriteBitString(Asn1Tag tag, ReadOnlySpan<byte> bitString, int unusedBitCount = 0) { }
-        public void WriteBitString(ReadOnlySpan<byte> bitString, int unusedBitCount = 0) { }
+        public void WriteBitString(ReadOnlySpan<byte> bitString, int unusedBitCount = 0, Asn1Tag? tag = default) { }

+        public void WriteNamedBitList(BitArray value, Asn1Tag? tag = default) { }
-        public void WriteNamedBitList(Asn1Tag tag, object enumValue) { }
-        public void WriteNamedBitList(object enumValue) { }
+        public void WriteNamedBitList(Enum value, Asn1Tag? tag = default) { }
-        public void WriteNamedBitList<TEnum>(Asn1Tag tag, TEnum enumValue) where TEnum : struct { }
-        public void WriteNamedBitList<TEnum>(TEnum enumValue) where TEnum : struct { }
+        public void WriteNamedBitList<TEnum>(TEnum value, Asn1Tag? tag = default) where TEnum : Enum { }

-        public void WriteOctetString(Asn1Tag tag, ReadOnlySpan<byte> octetString) { }
-        public void WriteOctetString(ReadOnlySpan<byte> octetString) { }
+        public void WriteOctetString(ReadOnlySpan<byte> octetString, Asn1Tag? tag = default) { }
-        public void PushOctetString() { }
-        public void PushOctetString(Asn1Tag tag) { }
+        public Scope PushOctetString(Asn1Tag? tag = default) { throw null; }
-        public void PopOctetString() { }
-        public void PopOctetString(Asn1Tag tag) { }
+        public void PopOctetString(Asn1Tag? tag) { }

-        public void WriteNull(Asn1Tag tag) { }
-        public void WriteNull() { }
+        public void WriteNull(Asn1Tag? tag = default) { }

-        public void WriteObjectIdentifier(Asn1Tag tag, ReadOnlySpan<char> oidValue) { }
-        public void WriteObjectIdentifier(ReadOnlySpan<char> oidValue) { }
+        public void WriteObjectIdentifier(ReadOnlySpan<char> oidValue, Asn1Tag? tag = default) { }
-        public void WriteObjectIdentifier(Asn1Tag tag, Oid oid) { }
-        public void WriteObjectIdentifier(Oid oid) { }
-        public void WriteObjectIdentifier(Asn1Tag tag, string oidValue) { }
-        public void WriteObjectIdentifier(string oidValue) { }
+        public void WriteObjectIdentifier(string oidValue, Asn1Tag? tag = default) { }

-        public void WriteEnumeratedValue(Asn1Tag tag, object enumValue) { }
-        public void WriteEnumeratedValue(object enumValue) { }
+        public void WriteEnumeratedValue(Enum value, Asn1Tag? tag = default) { }
-        public void WriteEnumeratedValue<TEnum>(Asn1Tag tag, TEnum enumValue) where TEnum : struct { }
-        public void WriteEnumeratedValue<TEnum>(TEnum enumValue) where TEnum : struct { }
+        public void WriteEnumeratedValue<TEnum>(TEnum value, Asn1Tag? tag = default) where TEnum : Enum { }

-        public void PushSequence(Asn1Tag tag) { }
-        public void PushSequence() { }
+        public Scope PushSequence(Asn1Tag? tag = default) { throw null; }
-        public void PopSequence(Asn1Tag tag) { }
-        public void PopSequence() { }
+        public void PopSequence(Asn1Tag? tag = default) { }

-        public void PushSetOf(Asn1Tag tag) { }
-        public void PushSetOf() { }
+        public Scope PushSetOf(Asn1Tag? tag = default) { throw null; }
-        public void PopSetOf(Asn1Tag tag) { }
-        public void PopSetOf() { }
+        public void PopSetOf(Asn1Tag? tag = default) { }

-        public void WriteGeneralizedTime(Asn1Tag tag, DateTimeOffset value, bool omitFractionalSeconds = false) { }
-        public void WriteGeneralizedTime(DateTimeOffset value, bool omitFractionalSeconds = false) { }
+        public void WriteGeneralizedTime(DateTimeOffset value, bool omitFractionalSeconds = false, Asn1Tag? tag = default) { }

-        public void WriteUtcTime(Asn1Tag tag, DateTimeOffset value, int twoDigitYearMax) { }
-        public void WriteUtcTime(DateTimeOffset value, int twoDigitYearMax) { }
+        public void WriteUtcTime(DateTimeOffset value, int twoDigitYearMax, Asn1Tag? tag = default) { }
-        public void WriteUtcTime(Asn1Tag tag, DateTimeOffset value) { }
-        public void WriteUtcTime(DateTimeOffset value) { }
+        public void WriteUtcTime(DateTimeOffset value, Asn1Tag? tag = default) { }

-        public void WriteCharacterString(Asn1Tag tag, UniversalTagNumber encodingType, ReadOnlySpan<char> str) { }
-        public void WriteCharacterString(UniversalTagNumber encodingType, ReadOnlySpan<char> str) { }
+        public void WriteCharacterString(UniversalTagNumber encodingType, ReadOnlySpan<char> str, Asn1Tag? tag = default) { }
-        public void WriteCharacterString(Asn1Tag tag, UniversalTagNumber encodingType, string str) { }
-        public void WriteCharacterString(UniversalTagNumber encodingType, string str) { }
+        public void WriteCharacterString(UniversalTagNumber encodingType, string value, Asn1Tag? tag = default) { }

+        public readonly struct Scope : IDisposable
+        {
+            private readonly object _dummy;
+            private readonly int _dummyPrimitive;
+            public void Dispose() { }
+        }
     }
-    public enum TagClass : byte
+    public enum TagClass
     {
         Universal = 0,
-        Application = (byte)64,
+        Application = 64,
-        ContextSpecific = (byte)128,
+        ContextSpecific = 128,
-        Private = (byte)192
+        Private = 192
     }
     public enum UniversalTagNumber
     {
         EndOfContents = 0,
         Boolean = 1,
         Integer = 2,
         BitString = 3,
         OctetString = 4,
         Null = 5,
         ObjectIdentifier = 6,
         ObjectDescriptor = 7,
         External = 8,
         InstanceOf = 8,
         Real = 9,
         Enumerated = 10,
         Embedded = 11,
         UTF8String = 12,
         RelativeObjectIdentifier = 13,
         Time = 14,
         Sequence = 16,
         SequenceOf = 16,
         Set = 17,
         SetOf = 17,
         NumericString = 18,
         PrintableString = 19,
         T61String = 20,
         TeletexString = 20,
         VideotexString = 21,
         IA5String = 22,
         UtcTime = 23,
         GeneralizedTime = 24,
         GraphicString = 25,
         ISO646String = 26,
         VisibleString = 26,
         GeneralString = 27,
         UniversalString = 28,
         UnrestrictedCharacterString = 29,
         BMPString = 30,
         Date = 31,
         TimeOfDay = 32,
         DateTime = 33,
         Duration = 34,
         ObjectIdentifierIRI = 35,
         RelativeObjectIdentifierIRI = 36
     }
 }
```

onyxmaster:
Looking forward to this very much. I had a need for ASN.1 I/O to implement OCSP-related services for our internal PKI and having this would help immensely.

bartonjs:
The implementation of this feature is already in late-stage progress at https://github.com/dotnet/runtime/pull/36729

scott-xu:
What about making types under `System.Security.Cryptography.Pkcs.Asn1` and `System.Security.Cryptography.X509Certificates.Asn1` be public as well?
I ask because I want to verify a SignedCMS which is very similar with https://github.com/dotnet/runtime/blob/master/src/libraries/System.Security.Cryptography.Pkcs/src/System/Security/Cryptography/Pkcs/Rfc3161TimestampToken.cs#L292
However, `X509IssuerSerial`, `EssCertId` and `EssCertIdV2` etc. are internal.

PR #94, opened 2020-03-03
Updated folder structure in accept to be year based

terrajobst:
The folder structure under `accepted` has become a bit unwieldy. In other systems (such as our blog and our SharePoint) we've used a year based system, which helps seeing the forest for the trees. Let's do this here too?


terrajobst:
Merging, since nobody objected

PR #97, opened 2020-03-13
Add spec for MinimumOSVersion.

mhutch:
This spec describes a standard mechanism for libraries to express the availability of APIs on different platforms and different versions of those platforms using attributes and MSBuild properties, and how those annotations can be used to improve the developer experience at multiple levels. It's based on the model already present in Xamarin.iOS.

I expect parts to be split into other documents and refined but this should be a good starting point.

### DIFF
--- a/accepted/2020/minimum-os-version/minimum-os-version.md
+++ b/accepted/2020/minimum-os-version/minimum-os-version.md
@@ -1,0 +1,246 @@
+# .NET 5 OS API Versioning
+
+# Introduction
+
+The format for [.NET 5 target framework monikers (TFMs)](https://github.com/dotnet/designs/pull/92) includes a version component that represents a revision of the operating system (OS) binding APIs, for example `net5.0-ios15.0` would include a binding to the iOS 15.0 APIs. However, it is a common pattern for apps and libraries to support running on older versions of the OS by using runtime checks to guard their calls so that they do not call APIs that are not available. This applies today for the Xamarin platforms (iOS, Android, Mac etc) and for UWP.
+
+This document proposes standard mechanisms and patterns for runtime API checks and compatibility that can be used across all .NET platforms to increase consistency of developer experience.
+
+# OS Versions
+
+There are three different OS version values that may be relevant for apps and libraries: the *OS API Version*, the *OS Target Version*, and the *OS Minimum Version*. It is important to distinguish and clarify these versions, as they have very different purposes.
+
+## OS API Version
+
+The OS API Version is part of the target framework, and affects which framework reference assemblies the app is compiled against. It does not affect which OS versions the app can run on (for well behaved platforms where backwards ABI compatibility is preserved). The primary reason for an app project to target an old OS API version is to allow the project to continue to build on older versions of the SDK that do not support the more recent API versions. A secondary reason may be to limit the API surface shown in IntelliSense so that it’s easier to avoid newer APIs, though this use case could perhaps be better supported in future by IDE experiences such allowing filtering IntelliSense by platform version.
+
+However, for libraries that are distributed as NuGets, it is useful to be able to multi-target across multiple API versions. This allows NuGets to access features available in newer API versions while continuing to ship updates for consumption by apps and libraries that (for whatever reason) target older API versions.
+
+By making the OS API version part of the target framework moniker (TFM), we take advantage of all the existing TFM-based mechanisms for multi-targeting builds against multiple versions of the API, selecting multi-targeted builds of a library from NuGets, and validating binary compatibility of references.
+
+The OS API version has major, minor, servicing and binding revision components. This takes the following form:
+
+```
+{major}[.{minor}[.{servicing}]][-r{bindingRevision}]
+```
+
+ When a component is not specified, this is equivalent to it being zero. The major, minor and servicing components of the version should match the OS version. The binding revision may be incremented if the OS binding is updated independently of OS updates. For example, if `net5.0-ios15.0` is released with some missing bindings, it could be followed by a release `net5.0-ios15.0-r1` that adds them.
+
+## OS Target Version
+
+The OS Target Version is used on certain platforms to enable backwards compatibility behaviors so that old apps continue to run correctly when API behaviors change in newer OS versions. For example, when an app that targeted Android 9.0 was run on Android 10.0, the OS might “quirk” the API behavior to continue to match the 9.0 behavior, but apps that targeted 10.0 or newer would get the newer API behavior.
+
+For now, handling of this version is left out of the scope of this document. Although it would be good to have a standard pattern in how this is defined across platforms, it’s not a universal concept, and doesn’t easily map into useful tooling. It’s also an app-level setting, and not useful for libraries except perhaps for warning of potential incompatibility, but that would have a very high false positive rate.
+
+## OS Minimum Version
+
+The OS Minimum Version is the minimum version of the OS that an app or library supports running on. If it is not specified, it should default to the OS API Version. If an app or library author specifies an OS Minimum Version, they are declaring that they are taking responsibility for ensuring that the app can run on that version and higher by using runtime checks to guard calls to potentially unavailable APIs. This allows apps and libraries to progressively “light up” features on newer OS versions while continuing to run downlevel.
+
+We will make this concept first class and propagate the value throughout the build system and NuGet consistently, which will enable us to build analyzers and other experiences to help developers follow this pattern. This tooling will be detailed later in this document.
+
+# API Annotations
+
+To enable tooling to provide a good experience, the reference assemblies must be annotated with information about the availability of the APIs, as is current done for Xamarin Apple platforms.
+
+The following attributes are proposed for the `System.Runtime.Versioning` namespace. They are based on the Xamarin attributes, but they do not have the `Deprecated` availability kind (it’s redundant), they rename the `Unavailable` availability kind to `Removed` (for consistency), and they do not have the ability to differentiate availability based on architecture (reduction of scope).
+
+```csharp
+[AttributeUsage(AttributeTargets.All, AllowMultiple=true)]
+public abstract class PlatformAvailabilityAttribute
+{
+    internal PlatformAvailabilityAttribute(
+    string platformIdentifier,
+    int majorVersion, int minorVersion, int servicingVersion, int bindingRevision
+    string message
+    ) {}
+    public AvailabilityKind AvailabilityKind { get; }
+    public string Message { get; }
+    public string PlatformIdentifier { get; }
+    public Version OSVersion { get; }
+    public int BindingRevision { get; }
+}
+
+public enum AvailabilityKind
+{
+    Introduced,
+    Obsoleted,
+    Removed
+}
+```
+
+The platform string is expected to have the same values as used in the `TargetPlatformIdentifier` portion of platform-specific TFMs. Having it in the availability attributes may seem redundant as it will also be burned into assembly attributes for platform-specific reference assemblies. However, explicitly specifying the attribute is useful for platform-agnostic `net5.0` reference assemblies that have methods that are only available on some platforms e.g. due to having platform-specific implementations via bait-and-switch. This may also improve source readability when multi-targeting as it will remove the need to use `#if` around the annotations.
+
+Constants for the platform identifiers could perhaps be exposed on a `PlatformIdentifier` type.
+
+The values declared on this attribute are expected to be accessible by extracting constructor arguments from metadata. However, for readability it’s expected to be applied using subclasses with more specific names. To enforce this, it has an internal constructor and its subclasses are sealed.
+
+The concrete subclasses are `IntroducedAttribute`, `ObsoletedAttribute` and `RemovedAttribute` subclasses, which take of the following form:
+
+```csharp
+public sealed class IntroducedAttribute : AvailabilityAttribute
+{
+    public IntroducedAttribute(
+    string platform,
+    int majorVersion,
+    int minorVersion = 0,
+    int servicingVersion = 0,
+    int bindingRevision = 0,
+    string message = null
+    ) : base(platform, majorVersion, minorVersion, servicingVersion, bindingRevision, message) {}
+}
+```
+
+These attributes are used as follows:
+
+```csharp
+[Introduced(PlatformIdentifier.macOS, 10, 2)]
+[Deprecated(PlatformIdentifier.macOS, 10, 9)]
+[Removed(PlatformIdentifier.macOS, 10, 12, 1)]
+public class Foo : NSObject { ... }
+```
+
+The purpose of keeping ‘removed’ APIs and annotating them with `RemovedAttribute` is to allow apps and libraries to continue to use these APIs in fallback paths for older OS versions. This is a pattern that exists on Apple platforms.
+
+If and when C# support target typing for enums, we could add convenience overloads that would improve readability:
+
+```csharp
+[Introduced (.macOS, 10, 1)]
+[Obsoleted (.macOS, 10, 8)]
+[Removed (.macOS, 10, 12)]
+```
+
+***NOTE****: These attributes would only be needed on reference assemblies. They would not be needed at runtime. They could be removed by the linker, and we could also strip them from the implementation assemblies that we ship ourselves.*
+
+# Runtime Checks
+
+***NOTE****: See* [*this issue*](https://github.com/dotnet/runtime/issues/33331) *for a more active/current discussion of runtime check APIs.*
+
+The following method will be added to the `RuntimeInformation` class in the `System.Runtime.InteropServices` namespace:
+
+```csharp
+public static bool CheckOSVersion(
+    int majorVersion, int minorVersion = 0, int servicingVersion = 0
+);
+```
+
+It is used as follows:
+
+```csharp
+if (RuntimeInformation.CheckOSVersion(15)) {
+    // use 15.0+ APIs
+} else {
+    // gracefully degrade
+}
+```
+
+There will be an additional overload that checks the OS as well as the OS version, for use in portable libraries:
+
+```csharp
+public static bool RuntimeInformation.CheckOS(
+    string platformIdentifier,
+    int majorVersion, int minorVersion = 0, int servicingVersion = 0
+);
+```
+
+As with the availability attributes, there could be an additional overload that used enums and target typing to improve readability.
+
+***NOTE****: The linker, JIT and/or AOT compiler could treat these method as intrinsics with constant values, which would enable it to eliminate unnecessary fallback code when consuming portable libraries or libraries with a lower OS minimum version than the app itself.*
+
+# MSBuild Properties
+
+The `Microsoft.Net.Sdk` targets will extract the OS API Version component from the `TargetFramework` into the `TargetPlatformVersion` MSBuild property and burn it into an assembly attribute. Tools that use this value are expected to access it from the MSBuild property or from the assembly attribute.
+
+Project files may specify a `MinimumPlatformVersion`, and if they do not it will default to the `TargetPlatformVersion` value. The `MinimumPlatformVersion` must not be higher than the `TargetPlatformVersion`, and this will be validated at the start of the build.
+
+It is recommended that platforms define more specifically named versions of this property of the form `Minimum{PlatformIdentifier}Version`, for example `MinimumIOSVersion` or `MinimumWindowsVersion`, and that their targets assign this value to `MinimumOSVersion` if it is set. This will simplify use of the property multi-targeting.
+
+> NOTE: would `{PlatformIdentifier}MinimumVersion` be better? It would sort/group more easily with other OS-specific properties.
+
+> NOTE: Using the existing `TargetPlatformVersion` property to represent the API version may be confusing for platforms such as Android that allow having different values for the target version and API version..
+
+# Assembly Level Minimum Version
+
+We cannot provide useful errors or warnings at an assembly level when a project references a library with a higher minimum version, as it may guard usage of that library with runtime checks. 
+
+However, it still useful to burn the minimum platform version into the assembly or NuGet package as a a baseline value for the classes and members that are not explicitly annotated. It can also be displayed as an information value in the user experience, for example in the NuGet package manager or on NuGet.org, although those are out of the scope of this document.
+
+## Assembly Attribute
+
+The minimum OS version will be burned into the assembly at build time using an assembly attribute. This attribute will be in the `System.Runtime.Versioning` namespace and will take the following form:
+
+```csharp
+[AttributeUsage(AttributeTargets.Assembly, AllowMultiple=true)]
+public sealed class OSMinimumVersion
+{
+    public OSMinimumVersion(
+    string platformIdentifier,
+    int majorVersion, int minorVersion, int servicingVersion
+    ) {}
+    public string PlatformIdentifier { get; }
+    public Version Version { get; }
+}
+```
+
+As with `PlatformAvailabilityAttribute` , the `PlatformIdentifier` property is included so that platform-agnostic assemblies can participate in platform minimum version annotations and checks.
+
+The MSBuild step that generates and injects TFM assembly attributes will be updated to generate this attribute when the `MinimumPlatformVersion` and `TargetPlatformIdentifier` properties are set. 
+
+Platform-neutral assemblies, i.e. assemblies that target a platform-neutral TFM such as `net5`, will have to declare the attribute explicitly.
+
+> NOTE: This seems a bit inconsistent. As a user, I’d expect `Minimum{OSName}Version` properties to work in platform-neutral assemblies too, but handling all those properties when building platform-neutral assemblies would be problematic as it would centralize the definitions of all of the platform-specific `MinimumVersion` properties into the* `Microsoft.Net.Sdk` targets.
+>
+>Perhaps we could use items and metadata instead of platform-specific properties, e.g. `<PlatformInfo Include="ios" MinimumVersion="15.0" />`. For platform-specific projects, we could allow the simpler `MinimumPlatformVersion` property by mapping it into one of these items, e.g. `<PlatformInfo Include="$(TargetPlatformIdentifier)" Version="$(MinimumPlatformVersion)" Condition="'$(TargetPlatformIdentifier)' != ''" />`.
+
+## NuGet Manifest
+
+When generating a NuGet package using MSBuild, any platform minimum version constraints expressed in the project file will be listed in the `.nuspec`  manifest in a `<platformInfo>` child of the `<metadata>` element. This element and all of its children are optional. They expose the same information as the assembly annotation attributes, but avoid the need for consumers to read the assembly metadata. This could potentially be validated and/or automated by NuGet..
+
+Within the `<platformInfo>` there are child elements corresponding to the TFM short names (used by NuGet as the folder name) of the reference assemblies in the NuGet. If the element represents a platform-specific TFM (e.g. `net5-ios14.0` or `net6-android9.0`) then it must have a `minimumVersion` attribute representing the minimum platform version for that platform:
+
+```xml
+<metadata>
+    <platformInfo>
+    <net5-ios15.0 minimumVersion="13.0" />
+    <net5-mac10.15 minimumVersion="10.11" />
+    </platformInfo>
+</metadata>
+```
+
+A reference assembly for a platform-neutral TFM (e.g. `net5` or `net6`) may still need to express platform-specific minimum versions if the implementations are platform-specific (aka ‘bait and switch’). As before, the element matches the TFM of the reference assembly, but this time there is no `minimumVersion`. Instead there are sub-elements that provide `minimumVersion` values for platforms using the `TargetPlatformIdentifier` of those platforms to identify them:
+
+```xml
+<metadata>
+    <platformInfo>
+    <net6>
+        <ios minimumVersion="13.0">
+        <mac minimumVersion="11.0">
+    </net6>
+    </platformInfo>
+</metadata>
+```
+
+Using these mechanisms, a NuGet package can declare minimum platform versions for any or all of its reference assemblies.
+
+# Availability Inheritance
+
+Members and nested types that do not have explicit availability annotations will inherit the values from their containing types. Types that do not have `Introduced` annotations will be treated as if they were introduced in the minimum platform version of the assembly itself.
+
+# IDE Experience
+
+Quick Info and IntelliSense tooltips will display the minimum platform version for any member with a minimum platform version higher than the current project’s minimum platform version.
+
+We may also want to consider some kind of sorting, grouping or colorization based hints in the IntelliSense list.
+
+# Analyzers
+
+There will be a Roslyn analyzer that will squiggle calls to any method or property with a minimum platform version higher than the current project’s minimum platform version, unless either of the following is true:
+
+
+- the call is guarded by a `CheckOS` or `CheckOSVersion` call against a version that is greater than or equal to the member’s minimum platform version
+- the caller member has a minimum platform version that is equal to or higher than the callee
+
+# See Also
+
+- [Swift](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html) `[@available](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html)` [attribute](https://docs.swift.org/swift-book/ReferenceManual/Attributes.html)
+- [Xamarin.iOS/Mac](https://docs.microsoft.com/en-us/dotnet/api/objcruntime.availabilitybaseattribute?view=xamarin-ios-sdk-12) `AvailabilityBaseAttribute` and its subclasses
+
--- a/accepted/2020/minimum-os-version/minimum-os-version.md
+++ b/accepted/2020/minimum-os-version/minimum-os-version.md
@@ -109,11 +109,11 @@ If and when C# support target typing for enums, we could add convenience overloa
 [Removed (.macOS, 10, 12)]
 ```
 
-***NOTE****: These attributes would only be needed on reference assemblies. They would not be needed at runtime. They could be removed by the linker, and we could also strip them from the implementation assemblies that we ship ourselves.*
+> NOTE: These attributes would only be needed on reference assemblies. They would not be needed at runtime. They could be removed by the linker, and we could also strip them from the implementation assemblies that we ship ourselves.
 
 # Runtime Checks
 
-***NOTE****: See* [*this issue*](https://github.com/dotnet/runtime/issues/33331) *for a more active/current discussion of runtime check APIs.*
+> NOTE: See [this issue](https://github.com/dotnet/runtime/issues/33331) for a more active/current discussion of runtime check APIs.
 
 The following method will be added to the `RuntimeInformation` class in the `System.Runtime.InteropServices` namespace:
 
@@ -144,7 +144,7 @@ public static bool RuntimeInformation.CheckOS(
 
 As with the availability attributes, there could be an additional overload that used enums and target typing to improve readability.
 
-***NOTE****: The linker, JIT and/or AOT compiler could treat these method as intrinsics with constant values, which would enable it to eliminate unnecessary fallback code when consuming portable libraries or libraries with a lower OS minimum version than the app itself.*
+> NOTE: The linker, JIT and/or AOT compiler could treat these method as intrinsics with constant values, which would enable it to eliminate unnecessary fallback code when consuming portable libraries or libraries with a lower OS minimum version than the app itself.
 
 # MSBuild Properties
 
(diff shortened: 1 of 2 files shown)


mhutch:
cc @jonpryor @rolfbjarne @spouliot

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+## OS Minimum Version
+
+The OS Minimum Version is the minimum version of the OS that an app or library supports running on. If it is not specified, it should default to the OS API Version. If an app or library author specifies an OS Minimum Version, they are declaring that they are taking responsibility for ensuring that the app can run on that version and higher by using runtime checks to guard calls to potentially unavailable APIs. This allows apps and libraries to progressively “light up” features on newer OS versions while continuing to run downlevel.
+
+We will make this concept first class and propagate the value throughout the build system and NuGet consistently, which will enable us to build analyzers and other experiences to help developers follow this pattern. This tooling will be detailed later in this document.
+
+# API Annotations
```
marek-safar: It might be helpful to outline our guidelines for 3rd party TFM libraries. 
- should they always have the annotation
- or only when they directly call to platform api
- or only when they wrap annotated api
- etc

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+public enum AvailabilityKind
+{
+    Introduced,
+    Obsoleted,
+    Removed
+}
+```
+
+The platform string is expected to have the same values as used in the `TargetPlatformIdentifier` portion of platform-specific TFMs. Having it in the availability attributes may seem redundant as it will also be burned into assembly attributes for platform-specific reference assemblies. However, explicitly specifying the attribute is useful for platform-agnostic `net5.0` reference assemblies that have methods that are only available on some platforms e.g. due to having platform-specific implementations via bait-and-switch. This may also improve source readability when multi-targeting as it will remove the need to use `#if` around the annotations.
```
marek-safar: @terrajobst is the thinking to apply such attributes to core libraries as well? For example when OpenSSL requires specific underlying version, or security APIs relying on specific PAL ios/macOS version?

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+Constants for the platform identifiers could perhaps be exposed on a `PlatformIdentifier` type.
+
+The values declared on this attribute are expected to be accessible by extracting constructor arguments from metadata. However, for readability it’s expected to be applied using subclasses with more specific names. To enforce this, it has an internal constructor and its subclasses are sealed.
+
+The concrete subclasses are `IntroducedAttribute`, `ObsoletedAttribute` and `RemovedAttribute` subclasses, which take of the following form:
+
+```csharp
+public sealed class IntroducedAttribute : AvailabilityAttribute
```
rolfbjarne: The base class is `PlatformAvailabilityAttribute` from above: 

```suggestion
public sealed class IntroducedAttribute : PlatformAvailabilityAttribute
```

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+We may also want to consider some kind of sorting, grouping or colorization based hints in the IntelliSense list.
+
+# Analyzers
+
+There will be a Roslyn analyzer that will squiggle calls to any method or property with a minimum platform version higher than the current project’s minimum platform version, unless either of the following is true:
+
+
+- the call is guarded by a `CheckOS` or `CheckOSVersion` call against a version that is greater than or equal to the member’s minimum platform version
+- the caller member has a minimum platform version that is equal to or higher than the callee
```
spouliot: There's potential to have analyzers to ensure availability attributes are logical - but it brings some special cases.

E.g. when a new type is introduced in a hierarchy, which is not a breaking change, you can end up with

```csharp
[Introduced (.ios, 8,0)]
class NSFoo : NSObject {
   string Id {get;}
   string Name {get;}
}
```
becoming
```csharp
[Introduced (.ios, 14,0)]
class NSMiddle : NSObject {
   [Introduced (.ios, 8,0)]
   string Id {get;}
}

[Introduced (.ios, 8,0)]
class NSFoo : NSMiddle {
   string Name {get;}
}
```

and it's easy to think there's a mistake in `Middle` since a member was available well before it's type. While you can't create `NSMiddle` on iOS8 you can still use `NSFoo.Middle` (without warnings) on iOS8.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+If and when C# support target typing for enums, we could add convenience overloads that would improve readability:
+
+```csharp
+[Introduced (.macOS, 10, 1)]
+[Obsoleted (.macOS, 10, 8)]
+[Removed (.macOS, 10, 12)]
+```
+
+> NOTE: These attributes would only be needed on reference assemblies. They would not be needed at runtime. They could be removed by the linker, and we could also strip them from the implementation assemblies that we ship ourselves.
```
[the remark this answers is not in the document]
marek-safar: I don't think we even need to do that. The attributes will exist in reference assemblies only which are not processed by illinker.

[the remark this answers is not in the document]
spouliot: For the BCL itself it has not impact, but for platform assemblies (our main use case) we need the attributes in the implementation assemblies since they are heavily used by our test suite.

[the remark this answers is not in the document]
marek-safar: I think that's a very specific use case and size increase does not matter in this case. I also don't think everyone should pay the penalty of having the attributes in implementation libs just because it makes your testing simpler.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+# Runtime Checks
+
+> NOTE: See [this issue](https://github.com/dotnet/runtime/issues/33331) for a more active/current discussion of runtime check APIs.
+
+The following method will be added to the `RuntimeInformation` class in the `System.Runtime.InteropServices` namespace:
+
+```csharp
+public static bool CheckOSVersion(
```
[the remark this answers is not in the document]
mhutch: Yes, it should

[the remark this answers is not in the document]
jkotas: If we ignore the recommended way to do Windows OS version checks and make this respect the actual OS version, we should fix `Environment.OSVersion` to do the same to be consistent. (I do not have a problem with this - just saying that we should be consistent across the product.)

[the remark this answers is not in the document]
stephentoub: > we should fix Environment.OSVersion to do the same to be consistent. (I do not have a problem with this - just saying that we should be consistent across the product.)

I think we should just do that.  I know it's not the recommended solution, but so many times I've seen folks say that Environment.OSVersion is buggy and then go and just do the P/Invoke themselves (or parse some other string-based representation that's doing it).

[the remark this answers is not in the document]
jkotas: Opened https://github.com/dotnet/runtime/issues/33643

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+Project files may specify a `MinimumPlatformVersion`, and if they do not it will default to the `TargetPlatformVersion` value. The `MinimumPlatformVersion` must not be higher than the `TargetPlatformVersion`, and this will be validated at the start of the build.
+
+It is recommended that platforms define more specifically named versions of this property of the form `Minimum{PlatformIdentifier}Version`, for example `MinimumIOSVersion` or `MinimumWindowsVersion`, and that their targets assign this value to `MinimumOSVersion` if it is set. This will simplify use of the property multi-targeting.
+
+> NOTE: would `{PlatformIdentifier}MinimumVersion` be better? It would sort/group more easily with other OS-specific properties.
+
+> NOTE: Using the existing `TargetPlatformVersion` property to represent the API version may be confusing for platforms such as Android that allow having different values for the target version and API version..
```
stevenbrix: when would you want to specify a different target and API version?

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+As with the availability attributes, there could be an additional overload that used enums and target typing to improve readability.
+
+> NOTE: The linker, JIT and/or AOT compiler could treat these method as intrinsics with constant values, which would enable it to eliminate unnecessary fallback code when consuming portable libraries or libraries with a lower OS minimum version than the app itself.
+
+# MSBuild Properties
+
+The `Microsoft.Net.Sdk` targets will extract the OS API Version component from the `TargetFramework` into the `TargetPlatformVersion` MSBuild property and burn it into an assembly attribute. Tools that use this value are expected to access it from the MSBuild property or from the assembly attribute.
+
+Project files may specify a `MinimumPlatformVersion`, and if they do not it will default to the `TargetPlatformVersion` value. The `MinimumPlatformVersion` must not be higher than the `TargetPlatformVersion`, and this will be validated at the start of the build.
```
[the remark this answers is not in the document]
mhutch: These two specs were written in parallel after we had a discussion a few weeks ago - I don't think either name is canonical right now.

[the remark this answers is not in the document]
Nirmal4G: Those property names exist in Windows SDK for a while now. That's why I suggested using those. We can conceptually map those 3 version properties for all platforms, no matter what their platform specific behaviour is.

They differ and that's why we have different platforms 😐. But we don't have to bring that complexity into the user facing project file.

accepted/2020/minimum-os-version/minimum-os-version.md
Nirmal4G: Android always target the API levels. So `*MinVersion` -> becomes minimum API level, `*MaxVersion` becomes targeted API level and we can have `TargetPlatformVersion` point to either of these depending on the project or have new value that mirrors the behaviour of `TargetPlatformMaxVersionTested`.

Before you guys ask, yes, Windows SDK has more than 4 properties just for versions and 3 just for UWP project files!

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+It is recommended that platforms define more specifically named versions of this property of the form `Minimum{PlatformIdentifier}Version`, for example `MinimumIOSVersion` or `MinimumWindowsVersion`, and that their targets assign this value to `MinimumOSVersion` if it is set. This will simplify use of the property multi-targeting.
+
+> NOTE: would `{PlatformIdentifier}MinimumVersion` be better? It would sort/group more easily with other OS-specific properties.
```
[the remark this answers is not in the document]
Nirmal4G: When doing multi-targeting, we can condition the `TargetPlatform*Version` properties under `TargetFramework` (aka multi-targeting) property (still don't like the name, BTW, so confusing!). But for single-targeting projects, we don't need to because, you know, they only target one thing.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+# Assembly Level Minimum Version
+
+We cannot provide useful errors or warnings at an assembly level when a project references a library with a higher minimum version, as it may guard usage of that library with runtime checks. 
+
+However, it still useful to burn the minimum platform version into the assembly or NuGet package as a a baseline value for the classes and members that are not explicitly annotated. It can also be displayed as an information value in the user experience, for example in the NuGet package manager or on NuGet.org, although those are out of the scope of this document.
+
+## Assembly Attribute
+
+The minimum OS version will be burned into the assembly at build time using an assembly attribute. This attribute will be in the `System.Runtime.Versioning` namespace and will take the following form:
```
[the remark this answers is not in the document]
Nirmal4G: Why not let the Platform runtime (if there will be) take care of roll-forwarding just like the coreclr does now?

It'd be way simpler just to define `TargetPlatform` as analogues to `TargetFramework` attribute which is already there and let the `runtimeconfig` take care of the roll-forwarding.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+Within the `<platformInfo>` there are child elements corresponding to the TFM short names (used by NuGet as the folder name) of the reference assemblies in the NuGet. If the element represents a platform-specific TFM (e.g. `net5-ios14.0` or `net6-android9.0`) then it must have a `minimumVersion` attribute representing the minimum platform version for that platform:
+
+```xml
+<metadata>
+    <platformInfo>
+    <net5-ios15.0 minimumVersion="13.0" />
+    <net5-mac10.15 minimumVersion="10.11" />
+    </platformInfo>
+</metadata>
```
Nirmal4G, line 206:
```suggestion
<metadata>
    <platforms framework="net5.0">
      <platform id="ios" minVersion="13.0" targetVersion="13.2" maxVersion="14.0">
      <platform id="mac" minVersion="10.10" targetVersion="10.11" maxVersion="10.14">
      <platform id="android" minVersion="18" targetVersion="21" maxVersion="24">
    </platforms>
    <platforms framework="net6.0">
      <platform id="ios" minVersion="13.1" targetVersion="13.3" maxVersion="14.0">
      <platform id="mac" minVersion="10.11" targetVersion="10.12" maxVersion="10.14">
      <platform id="android" minVersion="19" targetVersion="22" maxVersion="24">
    </platforms>
</metadata>
```

We can use this format for the platform neutral declarations too (see **below**). These are only needed if we have interop between native lib and a managed implementation but could be useful for binding libs too as they are essentially more or less the same!

`targetVersion` and `maxVersion` are totally optional. They could used to further manage the dependencies and resolve conflicts for large typical project that targets multiple platforms.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+    <platformInfo>
+    <net6>
+        <ios minimumVersion="13.0">
+        <mac minimumVersion="11.0">
+    </net6>
+    </platformInfo>
+</metadata>
+```
+
+Using these mechanisms, a NuGet package can declare minimum platform versions for any or all of its reference assemblies.
```
Nirmal4G, line 222: Why do we want to declare target platform for neutral assemblies? They are platform and runtime independent, no?

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+# Availability Inheritance
+
+Members and nested types that do not have explicit availability annotations will inherit the values from their containing types. Types that do not have `Introduced` annotations will be treated as if they were introduced in the minimum platform version of the assembly itself.
```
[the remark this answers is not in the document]
Nirmal4G: Can't we have separate assembly for Legacy APIs. I really like the concept of jQuery and jQuery Migrate which allows you to slowly migrate away from obsolete APIs. Sure, Analysers are good and all but when considering IoT and WASM targets, size and performance is everything.

So for E.g., if we can have APIs like string and string format overloads moved to legacy libs in favour of string interpolation APIs, that would reduce the bloat of the libs by a significant factor!!

> Please @ or # link the relevant issues if you can find them.

accepted/2020/minimum-os-version/minimum-os-version.md
Nirmal4G: We can use [Windows](https://docs.microsoft.com/en-us/windows/uwp/debug-test-perf/version-adaptive-code)'s existing version and api tagging or we could go with existing tags which are already there to expose them as such in a more .NET/WinMD way. Existing tools could be easily supported if we did follow latter idea.

> Try to use ideas as much as from the Windows platform teams. They are relevant to all other platforms out there.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+    ) : base(platform, majorVersion, minorVersion, servicingVersion, bindingRevision, message) {}
+}
+```
+
+These attributes are used as follows:
+
+```csharp
+[Introduced(PlatformIdentifier.macOS, 10, 2)]
+[Deprecated(PlatformIdentifier.macOS, 10, 9)]
+[Removed(PlatformIdentifier.macOS, 10, 12, 1)]
```
Nirmal4G, line 98: Why not use solutions like `WinMD` and `XLang` to generate metadata (bindings) about the platforms' APIs instead manually duplicating the Swift's attributes into .NET attributes?

That way bindings can be language/framework neutral and in unified format.

accepted/2020/minimum-os-version/minimum-os-version.md
Nirmal4G: Even for platform-agnostic assemblies using platform-specific implementations should match the public surface APIs (aka the ref assembly should be one and only one for any number of platform specific implementations!)

But that need not be the case for platform-specific assemblies that might have different public APIs for each platform.

accepted/2020/minimum-os-version/minimum-os-version.md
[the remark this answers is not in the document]
mhutch: @Nirmal4G that's somewhat orthogonal to this proposal, but check out https://github.com/dotnet/designs/blob/master/accepted/2020/better-obsoletion/better-obsoletion.md

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
+# .NET 5 OS API Versioning
+
+# Introduction
+
+The format for [.NET 5 target framework monikers (TFMs)](https://github.com/dotnet/designs/pull/92) includes a version component that represents a revision of the operating system (OS) binding APIs, for example `net5.0-ios15.0` would include a binding to the iOS 15.0 APIs. However, it is a common pattern for apps and libraries to support running on older versions of the OS by using runtime checks to guard their calls so that they do not call APIs that are not available. This applies today for the Xamarin platforms (iOS, Android, Mac etc) and for UWP.
```
richlander, line 5:
```suggestion
The format for [.NET 5 target framework monikers (TFMs)](https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md) includes a version component that represents a revision of the operating system (OS) binding APIs. For example, `net5.0-ios15.0` would include a binding to the iOS 15.0 APIs. However, it is a common pattern for apps and libraries to support running on older versions of an OS by using runtime checks to guard their calls so that they do not call APIs that are not available. This applies today for the Xamarin platforms (iOS, Android, Mac etc) and for UWP.
```

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+This document proposes standard mechanisms and patterns for runtime API checks and compatibility that can be used across all .NET platforms to increase consistency of developer experience.
```
richlander, line 7:
```suggestion
This document proposes a set of standard mechanisms and patterns for runtime API checks and compatibility that can be used across all .NET platforms to increase consistency of developer experience.
```

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+# OS Versions
+
+There are three different OS version values that may be relevant for apps and libraries: the *OS API Version*, the *OS Target Version*, and the *OS Minimum Version*. It is important to distinguish and clarify these versions, as they have very different purposes.
+
+## OS API Version
+
+The OS API Version is part of the target framework, and affects which framework reference assemblies the app is compiled against. It does not affect which OS versions the app can run on (for well behaved platforms where backwards ABI compatibility is preserved). The primary reason for an app project to target an old OS API version is to allow the project to continue to build on older versions of the SDK that do not support the more recent API versions. A secondary reason may be to limit the API surface shown in IntelliSense so that it’s easier to avoid newer APIs, though this use case could perhaps be better supported in future by IDE experiences such allowing filtering IntelliSense by platform version.
```
richlander, line 15: > The OS API Version is part of the target framework

Would be good to clarify with an example.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+However, for libraries that are distributed as NuGets, it is useful to be able to multi-target across multiple API versions. This allows NuGets to access features available in newer API versions while continuing to ship updates for consumption by apps and libraries that (for whatever reason) target older API versions.
```
richlander, line 17: Maybe too niche, but in a big shop, you could also use this same pattern for P2P.

accepted/2020/minimum-os-version/minimum-os-version.md
richlander: Yes, I would assume we would add these to `net5.0` TFM.

accepted/2020/minimum-os-version/minimum-os-version.md
richlander: This is a super simple and inexpensive mechanism. We don't want to take on heavy weight dependencies for it. It's also "OK" to duplicate attributes like this.

accepted/2020/minimum-os-version/minimum-os-version.md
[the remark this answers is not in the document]
Nirmal4G: Not the solution I was hoping for for but 😐!

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+public static bool CheckOSVersion(
+    int majorVersion, int minorVersion = 0, int servicingVersion = 0
+);
+```
+
+It is used as follows:
+
+```csharp
+if (RuntimeInformation.CheckOSVersion(15)) {
+    // use 15.0+ APIs
```
richlander, line 130: Have we thought through the potential for `MissingMethodException`? In the general case, metadata needs to be available to satisfy the JIT. This is not the case for Xamarin.iOS.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+} else {
+    // gracefully degrade
+}
+```
+
+There will be an additional overload that checks the OS as well as the OS version, for use in portable libraries:
```
richlander, line 136: Probably best to define "portable libraries" or use a different term. Is it "libraries that support multiple operating systems via runtime checks?" If yes, do we have a story for providing the compiler with metadata for multiple OSes, for that library?

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+    public string PlatformIdentifier { get; }
+    public Version Version { get; }
+}
+```
+
+As with `PlatformAvailabilityAttribute` , the `PlatformIdentifier` property is included so that platform-agnostic assemblies can participate in platform minimum version annotations and checks.
+
+The MSBuild step that generates and injects TFM assembly attributes will be updated to generate this attribute when the `MinimumPlatformVersion` and `TargetPlatformIdentifier` properties are set.
+
+Platform-neutral assemblies, i.e. assemblies that target a platform-neutral TFM such as `net5`, will have to declare the attribute explicitly.
```
richlander, line 188:
```suggestion
Platform-neutral assemblies, i.e. assemblies that target a platform-neutral TFM such as `net5.0`, will have to declare the attribute explicitly.
```

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+## NuGet Manifest
+
+When generating a NuGet package using MSBuild, any platform minimum version constraints expressed in the project file will be listed in the `.nuspec`  manifest in a `<platformInfo>` child of the `<metadata>` element. This element and all of its children are optional. They expose the same information as the assembly annotation attributes, but avoid the need for consumers to read the assembly metadata. This could potentially be validated and/or automated by NuGet..
+
+Within the `<platformInfo>` there are child elements corresponding to the TFM short names (used by NuGet as the folder name) of the reference assemblies in the NuGet. If the element represents a platform-specific TFM (e.g. `net5-ios14.0` or `net6-android9.0`) then it must have a `minimumVersion` attribute representing the minimum platform version for that platform:
```
richlander, line 198:
```suggestion
Within the `<platformInfo>` element, there are child elements corresponding to the TFM short names (used by NuGet as the folder name) of the reference assemblies in the NuGet. If the element represents a platform-specific TFM (e.g. `net5-ios14.0` or `net6-android9.0`) then it must have a `minimumVersion` attribute representing the minimum platform version for that platform:
```

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+```xml
+<metadata>
+    <platformInfo>
+    <net5-ios15.0 minimumVersion="13.0" />
+    <net5-mac10.15 minimumVersion="10.11" />
```
richlander, line 204:
```suggestion
    <net5.0-ios15.0 minimumVersion="13.0" />
    <net5.0-mac10.15 minimumVersion="10.11" />
```

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+    </platformInfo>
+</metadata>
+```
+
+A reference assembly for a platform-neutral TFM (e.g. `net5` or `net6`) may still need to express platform-specific minimum versions if the implementations are platform-specific (aka ‘bait and switch’). As before, the element matches the TFM of the reference assembly, but this time there is no `minimumVersion`. Instead there are sub-elements that provide `minimumVersion` values for platforms using the `TargetPlatformIdentifier` of those platforms to identify them:
```
richlander, line 209:
```suggestion
A reference assembly for a platform-neutral TFM (e.g. `net5.0` or `net6.0`) may still need to express platform-specific minimum versions if the implementations are platform-specific (aka ‘bait and switch’). As before, the element matches the TFM of the reference assembly, but this time there is no `minimumVersion`. Instead there are sub-elements that provide `minimumVersion` values for platforms using the `TargetPlatformIdentifier` of those platforms to identify them:
```

richlander: Honest question ... do we have satisfactory support for bait-and-switch in dotnet pack today? If not, it should noted.

accepted/2020/minimum-os-version/minimum-os-version.md
richlander: This is covered a bit earlier in the doc. The vision is that neutral assemblies can still do guarded light-up across a set of OSes.

accepted/2020/minimum-os-version/minimum-os-version.md
richlander: I think this is a non-issue, and I'm thinking of challenges from the past. Any platform that supports runtime lightup needs to make sure that it is safe to do so. For example, on Windows, this means that a CSWinRT assembly that targets n Windows version needs to itself expose which APIs are safe to call on n-1. And that's what this spec is describing.

jonwis:
From the Windows Platform POV, using version numbers is only half the story.

Apps should use feature-detection and light-up through runtime metadata:
* [ApiInformation](https://docs.microsoft.com/en-us/uwp/api/Windows.Foundation.Metadata.ApiInformation)
* LoadLibrary+GetProcAddress 
* [IsApiSetImplemented](https://docs.microsoft.com/en-us/windows/win32/api/apiquery2/nf-apiquery2-isapisetimplemented) method.)

Some individual Windows API also have `IsSupported` methods that further refine the answer.  While MSDN does document "this API is available in Windows 10 Creator's Edition" there are times when an API has been serviced into an earlier version of the OS.  Windows is not a single linear edition.  Certain APIs are available (or functional) on Enterprise or Server editions.

We recognize that this matrix makes developing version-and-edition software challenging, and having a standardized approach like this is useful for declaring in code what's possible when.

Are applications ever expected to call the proposed APIs, or are they only for components to know if they can do their work, or only compile-time markup? I would not want to see an _application_ say:

```
if (RuntimeInformation.CheckWindowsVersion(WindowsVersions.Win10_1903)) {
    var x = new Component { ... };
    x.Muffin();
}
```

Instead, the _application_ should say:

```
if (Component.IsSupported()) {
    var x = new Component { ... };
}
```

And let the component do the light-up checks.  That way when the Component figures out a way to work on Windows 10 1803 the app doesn't have to find and fix the CheckWindowsVersions points.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,245 @@
...
+
+However, for libraries that are distributed as NuGets, it is useful to be able to multi-target across multiple API versions. This allows NuGets to access features available in newer API versions while continuing to ship updates for consumption by apps and libraries that (for whatever reason) target older API versions.
+
+By making the OS API version part of the target framework moniker (TFM), we take advantage of all the existing TFM-based mechanisms for multi-targeting builds against multiple versions of the API, selecting multi-targeted builds of a library from NuGets, and validating binary compatibility of references.
+
+The OS API version has major, minor, servicing and binding revision components. This takes the following form:
+
+```
+{major}[.{minor}[.{servicing}]][-r{bindingRevision}]
```
dsplaisted, line 24: We got rid of the binding revision component, right?  I think we said these are .NET Framework "Versions" though, ie they have up to 4 version components.

accepted/2020/minimum-os-version/minimum-os-version.md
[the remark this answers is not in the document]
mhutch: It's only intended for compile time usage

accepted/2020/minimum-os-version/minimum-os-version.md
mhutch: Added some very basic guidance on this

accepted/2020/minimum-os-version/minimum-os-version.md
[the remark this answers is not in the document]
mhutch: Added a note, but I'm not sure how we could make it easy to omit these from reference assemblies, though.

maxkatz6:
Will it work with WinRT API ?
Currently developers can access WinRT API with targeting UWP or referencing Microsoft.Windows.SDK.Contracts. And as I understand, it will be also available with targeting net5-win.
Also there is helper class ApiInformation, which is similar to RuntimeInformation from the spec:
https://docs.microsoft.com/en-us/uwp/api/windows.foundation.metadata.apiinformation?view=winrt-18362

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -0,0 +1,246 @@
...
+
+However, it still useful to burn the minimum platform version into the assembly or NuGet package as a a baseline value for the classes and members that are not explicitly annotated. It can also be displayed as an information value in the user experience, for example in the NuGet package manager or on NuGet.org, although those are out of the scope of this document.
+
+## Assembly Attribute
+
+The minimum OS version will be burned into the assembly at build time using an assembly attribute. This attribute will be in the `System.Runtime.Versioning` namespace and will take the following form:
+
+```csharp
+[AttributeUsage(AttributeTargets.Assembly, AllowMultiple=true)]
+public sealed class OSMinimumVersion
```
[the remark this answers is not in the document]
mhutch: Not sure tbh? It's not as closely related

mhutch:
> From the Windows Platform POV, using version numbers is only half the story.
> 
> Apps should use feature-detection and light-up through runtime metadata:
> 
> * [ApiInformation](https://docs.microsoft.com/en-us/uwp/api/Windows.Foundation.Metadata.ApiInformation)
> * LoadLibrary+GetProcAddress
> * [IsApiSetImplemented](https://docs.microsoft.com/en-us/windows/win32/api/apiquery2/nf-apiquery2-isapisetimplemented) method.)
> 
> Some individual Windows API also have `IsSupported` methods that further refine the answer. While MSDN does document "this API is available in Windows 10 Creator's Edition" there are times when an API has been serviced into an earlier version of the OS. Windows is not a single linear edition. Certain APIs are available (or functional) on Enterprise or Server editions.
> 
> We recognize that this matrix makes developing version-and-edition software challenging, and having a standardized approach like this is useful for declaring in code what's possible when.>
> Are applications ever expected to call the proposed APIs, or are they only for components to know if they can do their work, or only compile-time markup? I would not want to see an _application_ say: 
> ```
> if (RuntimeInformation.CheckWindowsVersion(WindowsVersions.Win10_1903)) {
>     var x = new Component { ... };
>     x.Muffin();
> }
> ```

Applications would be expected to call these APIs, though in some cases the JIT/AOT compiler could elide these checks.

> Instead, the _application_ should say:
> 
> ```
> if (Component.IsSupported()) {
>     var x = new Component { ... };
> }
> ```
> 
> And let the component do the light-up checks. That way when the Component figures out a way to work on Windows 10 1803 the app doesn't have to find and fix the CheckWindowsVersions points.

iOS and Android don't neatly support these kinds of checks though. The bindings are thin wrappers over the native APIs of the underlying platform, which in a given OS release may add things as granular as new helper methods on existing classes. We could probably we could probably break a lot of these new APIs down into feature/component/capability groups, but it would be a lot of error-prone manual annotation work for no real gain.

That said, I do think some kind of `RequiresCapabilityAttribute` annotations and `RuntimeInformation.HasCapability(capability)` check would be useful and complementary to the version checks.  Capability checks are often used for features that are dependent on hardware, or as a mechanism to abstract out the underlying OS in multi-platform libraries. We could consider bridging the two mechanisms with some kind of "*OS A, version B is known to support C,D,E capabilities*" metadata.

accepted/2020/minimum-os-version/minimum-os-version.md
mhutch: True! I'm not sure it's worth specifically calling it out though

accepted/2020/minimum-os-version/minimum-os-version.md
mhutch: No, we do not. That was one of the things that NuGetizer did well.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -133,48 +129,48 @@ if (RuntimeInformation.CheckOSVersion(15)) {
...
+***NOTE****: The linker, JIT and/or AOT compiler could treat these method as intrinsics with constant values, which would enable it to eliminate unnecessary fallback code when consuming portable libraries or libraries with a lower OS minimum version than the app itself.*
 
 # MSBuild Properties
 
 The `Microsoft.Net.Sdk` targets will extract the OS API Version component from the `TargetFramework` into the `TargetPlatformVersion` MSBuild property and burn it into an assembly attribute. Tools that use this value are expected to access it from the MSBuild property or from the assembly attribute.
 
-Project files may specify a `MinimumPlatformVersion`, and if they do not it will default to the `TargetPlatformVersion` value. The `MinimumPlatformVersion` must not be higher than the `TargetPlatformVersion`, and this will be validated at the start of the build.
+Project files may specify an `OSMinimumVersion`, and if they do not it will default to OS Version equivalent to the OS API version specified in the `TargetPlatformVersion` value. The `OSMinimumVersion` must not be higher than the `TargetPlatformVersion`, and this will be validated at the start of the build.
```
dsplaisted, line 151: How will we handle this if we update the bindings outside of the OS version.  IE we ship iOS bindings 15.0.1, but it still maps to iOS 15.0.  Does each workload need to keep a mapping between `TargetPlatformVersion` and `OSMinimumVersion` for that platform?

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -133,48 +129,48 @@ if (RuntimeInformation.CheckOSVersion(15)) {
...
 
-> NOTE: would `{PlatformIdentifier}MinimumVersion` be better? It would sort/group more easily with other OS-specific properties.
+These per-platform properties will not work when building platform-neutral assemblies, as that would require centralizing definitions of all of the platform-specific `MinimumVersion` properties into the `Microsoft.Net.Sdk` targets. We may at some point add some other way to specify `OSMinimumVersion` values for multiple OSes for a single platform-neutral assembly, perhaps using items, but for now such assemblies must use assembly attributes directly.
 
-> NOTE: Using the existing `TargetPlatformVersion` property to represent the API version may be confusing for platforms such as Android that allow having different values for the target version and API version..
+> ***NOTE***: `{PlatformIdentifier}MinimumVersion` was chosen over `Minimum{PlatformIdentifier}Version` as it it's more easily searchable and sorts/group more easily with other OS-specific properties.
```
dsplaisted, line 157: I think Windows already has a property defined that follows a different format: `TargetPlatformMinVersion`.  We should cover how that relates to the new properties.

accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -183,39 +179,37 @@ public sealed class OSMinimumVersion
...
+Like the API annotation attributes, the `OSMinimumVersionAttribute` attribute is not needed in implementation assemblies as it is not used at runtime.
 
 ## NuGet Manifest
 
-When generating a NuGet package using MSBuild, any platform minimum version constraints expressed in the project file will be listed in the `.nuspec`  manifest in a `<platformInfo>` child of the `<metadata>` element. This element and all of its children are optional. They expose the same information as the assembly annotation attributes, but avoid the need for consumers to read the assembly metadata. This could potentially be validated and/or automated by NuGet..
+When generating a NuGet package using MSBuild, any platform minimum version constraints expressed in the project file will be listed in the `.nuspec` manifest in a `<platformInfo>` child of the `<metadata>` element. This element and all of its children are optional. They expose the same information as the assembly annotation attributes, but avoid the need for consumers to read the assembly metadata. This could potentially be validated and/or automated by NuGet..
```
dsplaisted, line 192: Nit: extra period at end of paragraph.

accepted/2020/minimum-os-version/minimum-os-version.md
mhutch: Added a clarifying note - https://github.com/dotnet/designs/pull/97/commits/7c34d18720ec6c25374321ee8bb2a21e74eae99a /cc @terrajobst

accepted/2020/minimum-os-version/minimum-os-version.md
mhutch: Yes, that's unfortunate but necessary. Still thinking about whether to formally define a standard mechanism or leave it up to the platform targets.

Nirmal4G: Why not define a roll forward mechanism for bindings just like the runtime packages?

That would certainly remove these complications!!

mhutch: @Nirmal4G it's not that easy.

For example: from the TFM `net5-ios15.0` we can infer that the `OSMinimumVersion` default is `15.0`. However, suppose the TFM `net5-ios15.0.1` was a revision of the bindings rather than bindings for a new OS version, then its `OSMinimumVersion` would also be need to be `15.0`.

We could handle that in a general way by saying that only the first two components of the binding version are the OS version and the third component is the bindings revision. However, those kinds of conventions we need to be determined on a platform by platform basis, as different platforms have different versioning schemes.

Nirmal4G: This is still a problem in the current proposal with the TF having bindings revision. Do we expect people to target the same platform for each bindings revision, I don't think so. I didn't encounter any packages targeting specifically any platform with multiple bindings!!! Are there?

If so, Why not separate the bindings version into a separate property?

Say `PlatformPackageVersion` if we're using nuget as the delivery mechanism for the bindings and `PlatformBindingsRevision` if we're shipping bindings within the workload itself. If we're using both we could use the latter to set the former for resolving the implicit FrameworkReference.

accepted/2020/minimum-os-version/minimum-os-version.md
Nirmal4G: Apple platforms and Windows defines the same version for both API and OS. But Android has API levels. But do the MSBuild projects (and the Android SDK tools) need the OS version? If so, Why?

PR #99, opened 2020-03-20
Feature switch - ability to selectively disable framework functionality

vitek-karas:
Proposal for unifying pieces of existing functionality into a cohesive story for disabling parts of framework functionality.


accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
...
+### Size
+By disabling functionality which is otherwise always available the linker can trim more code from the app reducing its size.
+For example
+* Disable support for FTP in [`HttpRequest.Create`](https://docs.microsoft.com/en-us/dotnet/api/system.net.webrequest.create?view=netcore-3.1#System_Net_WebRequest_Create_System_String_) (this is an example of a factory API which based on a string picks up features)
+* Disable security algorithms which are not in use by the app. For example [`AsymmetricalAlgorithm.Create(string)`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.asymmetricalgorithm.create?view=netcore-3.1#System_Security_Cryptography_AsymmetricAlgorithm_Create_System_String_) can create a large list of algorithms (another factory API).
+
+### Security
+Disable functionality which is not recommended anymore for security reasons. The app may choose to disable such functionality and thus guarantee that nothing in the app can use it.
+For example:
+* Disable `BinaryFormatter`
```
blowdart: 🎈

blowdart:
I'd be interested to see how serializers cope with deserialization of types that are disabled, and if that can be "graceful".

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
...
+### Stability
+Using `AppContext.TryGetSwitch` as the underlying runtime store of these switches is consistent with the intent of that API (and the API has been present in the .NET libraries for a long time). Unfortunately there are no guarantees that the value will never change, in fact there's an API `AppContext.SetSwitch` which is designed to change the value at runtime. Fixing the values of these switches in the linker would make usage of the `SetSwitch` API ineffective and rather confusing.
+
+*Note that the existing compatibility switches in `CoreLib` already have this problem and the current design seems to be to ignore it. That is the value is read from `AppContext` the first time it's needed and cached.*
+
+
+## References
+
+Very similar functionality has been proposed before in [dotnet/designs#42](https://github.com/dotnet/designs/pull/42). The main differences proposed here:
+* Feature switches apply at runtime as well as in the linker. The previous proposal was only for linker.
```
MichalStrehovsky: The link to AppContext so that we can emulate this at runtime was in that proposal too: https://github.com/dotnet/designs/pull/42/files#diff-416341d2703781003936b3f17e39fd0eR89 😉

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
+# Feature switch
+The functionality available in the .NET Core frameworks is getting bigger and wider with every release. Applications typically don't need or want all of it. With technologies like the linker we have the ability to remove parts of the framework which the app doesn't use. These technologies have limitations base don existing public API patterns and code behavior, sometimes they can't determine what the app really needs. This proposal adds a way to explicitly disable functionality in the framework.
```
stephentoub:
```suggestion
The functionality available in the .NET libraries is getting bigger and wider with every release. Applications typically don't need or want all of it, which means when an app includes the libraries as part of it, app size is unnecessarily increased. With technologies like the linker, we have the ability to remove parts of the framework which the app doesn't use. These technologies have limitations based on existing public API patterns and code behavior, sometimes they can't determine what the app really needs. This proposal adds a way to explicitly disable functionality in the framework.
```

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Goals
+* Ability to declaratively turn-off functionality in the framework (and potentially any library/app)
+* Same behavior all the time - regardless of linker being used or not (`Debug` build behaves the same as published `Release` app).
```
stephentoub: "Same behavior all the time"... with regards to the functionality being turned on/off.  I'm clarifying that this isn't addressing the problem of the linker removing functionality that the app may have been relying on via reflection or some such means.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
...
+* Ability to set different defaults based on the type of application (Blazor, Xamarin, …)
+
+
+## Expected usage
+### Size
+By disabling functionality which is otherwise always available the linker can trim more code from the app reducing its size.
+For example
+* Disable support for FTP in [`HttpRequest.Create`](https://docs.microsoft.com/en-us/dotnet/api/system.net.webrequest.create?view=netcore-3.1#System_Net_WebRequest_Create_System_String_) (this is an example of a factory API which based on a string picks up features)
```
stephentoub: A nit on this specific example: I'd rather see the app stop using the legacy WebRequest entirely ;)

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
...
+Each feature switch should have a read-only static property defined and all the decisions about the feature's availability should be based on a simple condition using such property. Depending on the impact of the feature switch the property may or may not be a public API.
+
+Feature switches for large areas of functionality which may span multiple libraries should be exposed as public APIs to make it easy and consistent. For example if there would be a feature switch to disable support for HTTP2 there should be a public API property indicating the desired behavior. This is because it's likely that even libraries outside of core libraries may need to change their behavior based on the property. Having one property in this case would enable linker to trim branches across multiple libraries in a consistent way.
+
+Feature switches for localized functionality which is not likely to span multiple libraries can use only internal properties.
+
+*Note: `CoreLib` for example already has `LocalAppContextSwitches` which does basically exactly this. We could simply extend it to usage of feature switches not just compatibility switches.*
```
stephentoub: From a quick glance, it doesn't look like corelib's current structure for LocalAppContextSwitches would make it amenable to tiering optimizations that would let the JIT trim away "dead" code resulting from a switch being set.  It'd be nice if the recommended structure for these were amenable to that, e.g. that a static readonly bool would end up being the thing conditionals were based on.  This won't matter in many cases, but it could for checks that show up on hot paths.  GlobalizationMode.Invariant is a good example of this: it is accessed on hot paths, and it's just a property that returns a static readonly bool field.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
...
+
+*Question: It's unclear why `GlobalizationMode.Invariant` is not based on this behavior - but maybe it needs native runtime integration.*
+
+### Unified pattern for feature definition in MSBuild
+Introduce a standard pattern how to define these properties in the SDK. Basically a way to define a property and have it automatically passed through to `.runtimeconfig.json` as well as linker substitutions.
+
+This requires a mapping between the MSBuild property name, the full name of the feature switch for runtime configuration (which would show up in `.runtimeconfig.json`) and the read-only static property in the managed code which is used to branch the behavior.
```
stephentoub: How do you expect this mapping to surface and be exposed to the linker?

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,113 @@
...
+There are 3 different representations to each feature switch
+* The MSBuild property - for example `InvariantGlobalization` - this is used by developers to turn the feature on/off.
+* The runtime property in `.runtimeconfig.json` - for example `System.Globalization.Invariant` - this is used at runtime to alter the behavior of the code (via the IL property below)
+* The IL property which is used in the code - for example `System.Globalization.GlobalizationMode.Invariant, System.Private.CoreLib` - this is used by the linker to fix the value of the property and propagate that to remove unused code.
+
+The mapping between MSBuild property and runtime property pretty much has to be part of the SDK (in the MSBuild somewhere) as the SDK already generates the `.runtimeconfig.json` and this needs to be part of it. The second mapping to the IL property can take several options:
+* Also int he SDK - becomes an input to the ILLink task
+* Encoded as attributes in the BCL - we could add a new attribute to the IL properties to mark which runtime options they map to - linker can create the mapping from these attributes. In this case the ILLink task would take the runtime properties as input
```
stephentoub: This would mean the libraries would need to know about the MSBuild property names?

accepted/2020/feature-switch.md
vitek-karas: I was trying to say that the behavior is consistent regardless of linker being used or not - linker acts as an optimization, not a behavior changing tool. I'll try to rewrite this.

accepted/2020/feature-switch.md
vitek-karas: I know :-) - but it's such a nice example. I'd be happy to use a different one if we have it.

accepted/2020/feature-switch.md
vitek-karas: OK - I will change that - should I copy the GlobalizationMode.Invariant - or is there some typical pattern which JIT can figure out?

accepted/2020/feature-switch.md
stephentoub: Ok ;) There are a variety of factories or APIs that take options that configure what concrete instances are created, e.g. XmlReader/Writer.Create, Regex w/ RegexOptions.Compiled, etc.

accepted/2020/feature-switch.md
vitek-karas: The idea is to use the runtime property names in the attributes - for example:
```C#
internal static class GlobalizationMode
{
    [FeatureSwitch("System.Globalization.Invariant")]
    internal static bool Invariant { get; }
}
```

accepted/2020/feature-switch.md
vitek-karas: Sorry - totally missed that - fixed.

accepted/2020/feature-switch.md
vitek-karas: Good - I'll use Regex as an example.

stephentoub: https://github.com/dotnet/runtime/blob/a83598dd151628672b1d2ff72310cc8c2065a019/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L77-L81

accepted/2020/feature-switch.md
vitek-karas: Depending on the picked representation (see below) either the linker would be the one knowing about the mapping (attributes or pattern matching) or it would be directly given the result of the mapping. Linker should not be the component which has the mapping hard-coded - it should be hard-coded either in the SDK, or in the libraries code or nowhere (linker pattern matching).

accepted/2020/feature-switch.md
eerhardt: In my mind, the name that is the "glue" is the string that is in the `runtimeconfig.json`. This is the "Public API" for this switch from the runtime/library perspective. This is the string that the library passes into `AppContext.TryGetSwitch`, so the library already knows about this string.

On the MSBuild side, there is a step in-between the property name and the `runtimeconfig.json`, and that is the `RuntimeHostConfigurationOption` item.  `RuntimeHostConfigurationOption` is the information that gets put into the `runtimeconfig.json`. The mapping between the MSBuild property and the `RuntimeHostConfigurationOption` also lives in MSBuild:

https://github.com/dotnet/sdk/blob/36ef8b2aa8e5d579c921704bdab69a7407936889/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L341-L354

```xml
  <!--
    Defaults @(RuntimeHostConfigurationOption) items based on MSBuild properties.
    -->

    <RuntimeHostConfigurationOption Include="System.Globalization.Invariant"
                                    Condition="'$(InvariantGlobalization)' != ''"
                                    Value="$(InvariantGlobalization)" />

    <RuntimeHostConfigurationOption Include="System.GC.Server"
                                    Condition="'$(ServerGarbageCollection)' != ''"
                                    Value="$(ServerGarbageCollection)" />
```

vitek-karas: @eerhardt Exactly - "runtime property" == "property in `.runtimeconfig.json`". Sorry for the confusion in terminology.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,119 @@
...
+* Ability to set different defaults based on the type of application (Blazor, Xamarin, …)
+
+
+## Expected usage
+### Size
+By disabling functionality which is otherwise always available the linker can trim more code from the app reducing its size.
+For example
+* Disable support for compiled regular expressions. The constructor for regular expression [`Regex(string, RegexOptions)`](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.-ctor?view=netcore-3.1#System_Text_RegularExpressions_Regex__ctor_System_String_System_Text_RegularExpressions_RegexOptions_) uses the second parameter to determine if it should pre-compile the regular expression or interpret it each time. The code ot pre-compile the expression is relatively large and if the app doesn't use this feature it can be removed. But linker typically won't be able to determine that since it's a dynamic behavior. (This is already [conditionally enabled](https://github.com/dotnet/runtime/blob/a83598dd151628672b1d2ff72310cc8c2065a019/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L77-L81) only when dynamic code compilation is supported by the runtime.)
```
jkotas:
```suggestion
* Disable support for compiled regular expressions. The constructor for regular expression [`Regex(string, RegexOptions)`](https://docs.microsoft.com/en-us/dotnet/api/system.text.regularexpressions.regex.-ctor?view=netcore-3.1#System_Text_RegularExpressions_Regex__ctor_System_String_System_Text_RegularExpressions_RegexOptions_) uses the second parameter to determine if it should pre-compile the regular expression or interpret it each time. The code to pre-compile the expression is relatively large and if the app doesn't use this feature it can be removed. But linker typically won't be able to determine that since it's a dynamic behavior. (This is already [conditionally enabled](https://github.com/dotnet/runtime/blob/a83598dd151628672b1d2ff72310cc8c2065a019/src/libraries/System.Text.RegularExpressions/src/System/Text/RegularExpressions/Regex.cs#L77-L81) only when dynamic code compilation is supported by the runtime.)
```

vitek-karas:
@blowdart The fallback behavior depends on each feature, but I expect in most cases the code would throw. That said in some cases it's possible to fallback (`Regex` won't compile if it can't even if asked for).

vitek-karas:
@blowdart Curious which types you had in mind - where we would "disable a type" and then it's used in deserialization. Is the case for cryptography algorithms in some case (I don't know)?
The one case where we want to use this is `ResourceManager` which internally uses `BinaryFormatter` to read data from the resource stream in some cases. In this case the behavior would be to throw (there's no sensible fallback).

blowdart:
I don't know if I have particulars in mind, I'm just remember that we don't use obsolete on enums often because it breaks serialization. I just don't want to end up in the same place.

accepted/2020/feature-switch.md
jkotas: `GlobalizationMode.Invariant` is a fine example. Property backed by a readonly bool field is the key pattern.

The LocalAppContextSwitches pattern has ability to [disable caching](https://github.com/dotnet/runtime/blob/0a2f0e6da98a784601b37d19417b930a33b2c478/src/libraries/System.Private.CoreLib/src/System/LocalAppContextSwitches.Common.cs#L33). This is used in [a few tests](https://github.com/dotnet/runtime/search?q=DisableCaching&unscoped_q=DisableCaching). The ability to disable caching is what makes the pattern unfriendly to linking. I do not see a problem with giving up the disable caching hook and implement the tests in a different way.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Generate the right input for the linker in SDK
+Currently there's no integration of the substitutions feature in the linker with the `ILLink` task/targets. Introduce a standard way to  define the content of `substitutions.xml` from the SDK (MSBuild items/properties) and then implement the file generation in the `ILLink` task and pass it to the linker.
+
+*Or change linker to not use the XML format and instead take the input on command line.*
+
+
+## Open questions
+
+### Boolean only or more complex values
```
jkotas: I would stick with boolean values.

The problem with factory APIs can be solved by adding new APIs: The factory can be changed to create nothing or just the common things that everybody needs, and a new API can be introduced to teach the factory how to create more things.

We have a prior art with for this e.g. with encodings: `Encoding.RegisterProvider`

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,119 @@
...
+
+### Which components own the mapping between different representations of feature switches
+There are 3 different representations to each feature switch
+* The MSBuild property - for example `InvariantGlobalization` - this is used by developers to turn the feature on/off.
+* The runtime property in `.runtimeconfig.json` - for example `System.Globalization.Invariant` - this is used at runtime to alter the behavior of the code (via the IL property below)
+* The IL property which is used in the code - for example `System.Globalization.GlobalizationMode.Invariant, System.Private.CoreLib` - this is used by the linker to fix the value of the property and propagate that to remove unused code.
+
+The mapping between MSBuild property and runtime property pretty much has to be part of the SDK (in the MSBuild somewhere) as the SDK already generates the `.runtimeconfig.json` and this needs to be part of it. The second mapping to the IL property can take several options:
```
jkotas: Would it make sense to get rid of this mapping and switch to unified system where the name is same everywhere? What is the naming pattern we would unify on if we were to do this?

The existing cases can be handled as hard-coded one off, or we can take a breaking change for them in .NET 5.

vitek-karas: It would be hard to unify on the name of the MSBuild property - and I don't think it's the right thing to do anyway. But we could unify the runtime and link-time names. The runtime name can be arbitrary as long as it's unique because we will need code to call the `TryGetSwitch` anyway. The link-time name needs to be able to tell linker what to do, so ideally the full name of the property.

So in the example of the globalization invariant the unified name would be `System.Globalization.GlobalizationMode.Invariant`. Such a string would work for the runtime property as well (it would not work for MSBuild since it's not very user friendly and it's not a valid MSBuild property anyway).

There are some draw backs though:
* Inability to have a 1:many mapping between the runtime property and the IL properties - makes it even more important to make the IL property with the right visibility.
* Inability to specify the assembly for the IL property making it a bit ambiguous. That said this could be used to "hack around" the above limitation of missing 1:many mapping (more than one assembly could define the property with the same namespace/type/name)

accepted/2020/feature-switch.md
marek-safar: I think boolean and primitive integral types are what we should stick to for things to be simple for everyone.

accepted/2020/feature-switch.md
marek-safar: I think we should be careful about what we decide to expose. Make it easy to tweak any static bool property might cause developers and us pain when they decide to play with fire and become framework optimizers. My suggestion would be to expose this in two ways

1. Hidden
We'll use the properties (substitutions) to simplify our implementation. In that case, we need to ship the rules together with the framework assemblies (probably embedded resources) and they cannot be overwritten (think of big-endian for win-x64) and should not be available for public modification.

2. Visible
We expose a set of predefined well tested features like (globalization or serialization) and only they will have any meaning to the framework. We map them all into same MSBuild property or something similar and the features will be quite high-level and validated by us that they work as expected alone or when combined together.

So in both cases, the mapping should be part of the assembly resources as it can be RID specific or in an internal attribute if we want to start inventing another attribute set and add another linker pass to read them.

vitek-karas: If we think we need to make this explicit (as in only specific set of properties can be treated as consts by the linker) then I would very much like to use attributes for it. The overall number of them will be tiny and linker doesn't need a new pass - the `TypeMapStep` should be the right place.
Note that if we keep the support for `substitution.xml` in the linker in theory developers can play these games anyway. So it's just a matter of how "hidden" it is.
I personally would go with one of these two:
* Use the name of the IL property as the name of the runtime property - and only have a mapping from MSBuild property to this value. We would document the limited set of MSBuild properties which do have a mapping like this. Developers would have to play with the underlying runtime property set to get into dangerous places.
* Use attributes to describe the mapping from runtime property -> IL property

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,119 @@
...
+* Security algorithms which are not in use by the app. For example [`AsymmetricalAlgorithm.Create(string)`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.asymmetricalgorithm.create?view=netcore-3.1#System_Security_Cryptography_AsymmetricAlgorithm_Create_System_String_) can create a large list of algorithms (another factory API).
+
+### Security
+Disable functionality which is not recommended anymore for security reasons. The app may choose to disable such functionality and thus guarantee that nothing in the app can use it.
+For example:
+* `BinaryFormatter`
+* Some cryptography algorithms which are not considered secure anymore. For example [`HashAlgorithm.Create(string)`](https://docs.microsoft.com/en-us/dotnet/api/system.security.cryptography.hashalgorithm.create?view=netcore-3.1#System_Security_Cryptography_HashAlgorithm_Create_System_String_) can create SHA1
```
terrajobst: The crypto algorithm thing keeps coming but I don't think that's in the same boat as `BinaryFormatter`. SHA1 and MD5 might not be cryptographically secure, but many document formats and wire protocols require it. For example, if you want to read a Git repo, you need to use SHA1. I don't see us ever disabling those by default. From a linker stand point we probably don't save much either.

jkotas: You can always do a direct `SHA1.Create()` or `MD5.Create()` if you really want these outdated algorithms. This config won't affect that.

stephentoub: For example, SHA1 is required as part of the websocket handshake, and ClientWebSocket uses SHA1.Create directly.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,119 @@
+# Feature switch
+The functionality available in the .NET libraries is getting bigger and wider with every release. Applications typically don't need or want all of it, which means when an app includes the libraries as part of it, app size is unnecessarily increased. With technologies like the linker, we have the ability to remove parts of the framework which the app doesn't use. These technologies have limitations based on existing public API patterns and code behavior, sometimes they can't determine what the app really needs. This proposal adds a way to explicitly disable functionality in the framework.
+
+## Goals
+* The ability for developers to declaratively remove functionality in the framework, 3rd party libraries or final app
```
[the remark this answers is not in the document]
vitek-karas: > Libraries can fail with a sensible error message (i.e. detect when functionality they need was removed)

If the specific feature uses a public API property (as described below in the doc), then the library can easily check the value of that property and potentially fail if it requires the feature in question. In all cases it can check the runtime property value via `AppContext.TryGetSwitch`, but that is not a very nice experience. Maybe we should expose most (if not all) features via a public API property to remedy this.

> Application authors get better tooling to understand what their dependencies need

This is the hard problem. Basically the features switches are in some sense a workaround to tooling not being able to figure out the real dependencies of an application. Without them the tooling has to keep all possibilities open, but that can bring in lot of code. The feature switch is a way for the app to "help" the tooling do a better job at removing dependencies. If we could solve these automatically, then we would not need the feature switches.

We'll see what exact feature switches we'll end up with, I hope that most of them will be very easy to explain and thus easy to reason about.

accepted/2020/feature-switch.md
sbomer: Based on some questions that came up in https://github.com/mono/linker/pull/1047, I think we need more clarity on what this looks like for hidden switches in particularly. My understanding is that public features have a documented MSBuild property and feature name == appcontext string. What's an example of something we might make a hidden feature? I think endianness falls more into the category of https://github.com/dotnet/runtime/issues/31785 which is not a goal of the feature switches.

vitek-karas: @marek-safar to make the discussion easier - could you please provide a sample (probably from the Xamarin world) of a hidden "feature switch" which is not tied to a build of libraries, but would be provided by the Xamarin SDK for example?

As for how we provide the mapping:
* I personally like attributes better as they are naturally tied to the code.
* But I do see value in discussing the embedded XML as that would avoid another pass in the linker (which is pretty expensive).

Nirmal4G: > when they decide to play with fire and become framework optimizers

With great power comes great responsibility! 😎!

Could the concept of a feature switch be tied to API checks, essentially, they are availability checks to change behaviour at runtime and compile time? Is it not?

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,119 @@
...
+
+*Note that it seems likely that some of switches may require 1:many mapping because they're targeting existing code which uses multiple properties to determine the presence of the feature.*
+
+Ultimately some of these switches may also need VS UI integration, but for now most of the feature switches should be perfectly fine with only MSBuild properties.
+
+The ability to specify these from MSBuild projects as properties also means that different templates and SDKs can set different defaults - so for example Blazor or Xamarin SDKs may choose to turn off some of these features by default.
+
+### Generate the right input for the linker in SDK
+Currently there's no integration of the substitutions feature in the linker with the `ILLink` task/targets. Introduce a standard way to  define the content of `substitutions.xml` from the SDK (MSBuild items/properties) and then implement the file generation in the `ILLink` task and pass it to the linker.
```
Nirmal4G, line 79: We could have a separate SDK that supports manipulating IL, and all the things. There's one already but it's features are narrow!! We could expand the feature set of `Microsoft.NET.Sdk.IL` to include the capabilities that are discussed here?!

accepted/2020/feature-switch.md
Nirmal4G: Boolean and Enums are more than enough.

accepted/2020/feature-switch.md
Nirmal4G: > From a linker stand point we probably don't save much either.

Every byte saved... is every second (perf) gain!!! 😉

accepted/2020/feature-switch.md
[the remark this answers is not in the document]
vitek-karas: @Nirmal4G I probably don't understand what you mean by "transitive flow".

The feature switches should only be set by the application project (the one building the final app) - as that's the one which will run the linker. We don't really support running linker on anything else from the SDK right now.
Libraries and frameworks will expose feature switches by reacting to runtime properties and possibly also adding msbuild properties (to make it easier to use). I don't expect libraries (and definitely not frameworks) to actually set any feature switch - that's up to the app.

Similarly for runtime properties - these should only be set by the application (and the proposed feature switch MSBuild integration will make that automatic).

`runtimeconfig.dev` files serve different purpose (they are supposed to be the local machine configuration). I also don't think libraries/frameworks have any say to what goes into `runtimeconfig.dev`.

The above discussed "flow" of feature switches between framework/libraries/app is that if framework defines a feature switch, it should probably expose it as a public API (public static bool property) to which both libraries and app can react accordingly. Linker will see the usage of that property in all the components of the app and apply trimming accordingly as well.

accepted/2020/feature-switch.md
vitek-karas: @Nirmal4G I must admit I don't know what exactly you refer to by "API checks". The proposal effectively uses public static bool API properties as the tool which the app/library/framework can use to determine behavior and linker will use it to trim the code.

I do agree about power and responsibility, but if possible we should not make it easy to break things. So the ability to play god will still be there (via custom substitution.xml files), but it's not going to be as simple as defining a single MSBuild property.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Goals
+* The ability for developers to declaratively remove functionality in the framework, 3rd party libraries or final app
+* Consistent behavior across different build configurations - the app should get the exact same behavior whether the linker was used on it or not
+* Ability to set different defaults based on the type of application (Blazor, Xamarin, …)
+
+
+## Expected usage
+### Size
+By disabling functionality which is otherwise always available the linker can trim more code from the app reducing its size.
```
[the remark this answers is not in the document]
vitek-karas: What exactly will be a feature switch is yet to be determined. The one problem with having lot of them is testing - it makes the potential matrix of available configuration very large. So we'll probably try to keep the number small.

For the scenarios which you mention, ideally linker should be able to remove the code if you're not using it. Now making that a reality is pretty hard, but introducing a feature switch for those will be even harder.

accepted/2020/feature-switch.md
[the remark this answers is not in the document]
Nirmal4G: > I probably don't understand what you mean by "transitive flow"

_It was conceptual question about trimming by setting feature flags!_

Say if a lib depends on a feature and the app developer sets that switch to trim it from project and like that to several other libs or possibly in a framework pack. What will happen? The code that depends on the feature will fail at runtime.

What I was saying is to trim only when all the code paths from those dependences are safe from that feature flag! **Reflection** and compiled **RegEx** are few ones that create a problem with the linker.

accepted/2020/feature-switch.md
Nirmal4G: You can use MSBuild items.

accepted/2020/feature-switch.md
Nirmal4G: _Again, its philosophical. but does have a some context here._

We need that power when deploying .NET to resource constrained devices. e.g. IoT and Apple Watch, etc.

> if possible we should not make it easy to break things

True. I like having the file to change the linker behaviour. But even for simple ones I opt for the file way rather than MSBuild properties. MSBuild Items can be used instead too.

accepted/2020/feature-switch.md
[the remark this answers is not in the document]
Nirmal4G: > For the scenarios which you mention, ideally linker should be able to remove the code if you're not using it. Now making that a reality is pretty hard, but introducing a feature switch for those will be even harder.

True that. Still, it will be useful in AoT builds. I would not worry about compat since its a one time build. You are not going to manually patch the native images.

accepted/2020/feature-switch.md
MichalStrehovsky: > But I do see value in discussing the embedded XML as that would avoid another pass in the linker

What would be the new pass? My naive expectation is that the custom attribute would be processed when we do `RemoveUnreachableBlocks`: before evaluating the body of the method for its "constantness" in `FindConstantExpressionsMethods`, we would look for the attribute and treat it as a constant if it has one and the feature is being removed. Feature switch already depends on `RemoveUnreachableBlocks` running.

sbomer:
I added some short notes as discussed today. @vitek-karas @eerhardt @marek-safar PTAL.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Unified pattern for feature definition in MSBuild
+Introduce a standard pattern how to define these properties in the SDK. Basically a way to define a property and have it automatically passed through to `.runtimeconfig.json` as well as linker substitutions.
+
+This requires a mapping between the MSBuild property name, the full name of the feature switch for runtime configuration (which would show up in `.runtimeconfig.json`) and the read-only static property in the managed code which is used to branch the behavior. The mappings will be defined by [`RuntimeHostConfigurationOptions`](https://github.com/dotnet/sdk/blob/36ef8b2aa8e5d579c921704bdab69a7407936889/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L347) which includes an AppContext configuration name and its value based on the values of user-facing MSBuild properties, and by `substitutions.xml` files embedded in the assembly defining the features.
```
marek-safar, line 72: @sbomer @vitek-karas could this better specify what will be the format used inside csproj file? It would also be helpful to show the full configuration flow for one feature when the feature is controlled from csproj as oppose to.runtimeconfig.json

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,121 @@
...
+
+*Note that it seems likely that some of switches may require 1:many mapping because they're targeting existing code which uses multiple properties to determine the presence of the feature. The `substitutions.xml` will allow 1:many mappings if needed.*
+
+Ultimately some of these switches may also need VS UI integration, but for now most of the feature switches should be perfectly fine with only MSBuild properties.
+
+The ability to specify these from MSBuild projects as properties also means that different templates and SDKs can set different defaults - so for example Blazor or Xamarin SDKs may choose to turn off some of these features by default.
+
+### Generate the right input for the linker in SDK
+
+The names/values from `RuntimeHostConfigurationOptions` will be passed to the ILLink task, which will apply any feature substitutions defined in `substitutions.xml`. The substitutions file format will be extended to condition the substitutions based on the feature name/value.
```
marek-safar, line 82: Please specify which names/keys will be passed from RuntimeHostConfigurationOptions

accepted/2020/feature-switch.md
sbomer: I added some examples.

accepted/2020/feature-switch.md
sbomer: Clarified this.

accepted/2020/feature-switch.md
```diff
@@ -77,11 +77,60 @@ Ultimately some of these switches may also need VS UI integration, but for now m
...
 ### Generate the right input for the linker in SDK
 
-The names/values from `RuntimeHostConfigurationOptions` will be passed to the ILLink task, which will apply any feature substitutions defined in `substitutions.xml`. The substitutions file format will be extended to condition the substitutions based on the feature name/value.
+All names/values from `RuntimeHostConfigurationOptions` will be passed to the ILLink task, which will apply any feature substitutions defined in [`substitutions.xml`](https://github.com/mono/linker/blob/master/src/linker/README.md#using-custom-substitutions). The substitutions file format will be extended to condition the substitutions based on the feature name/value. Any `RuntimeHostConfigurationOptions` which do not have feature implementations in `substitutions.xml` will not result in any modifications to the IL.
 
-The assembly defining a feature will carry an embedded `substitutions.xml` that defines the implementation of the corresponding IL property when the feature switch is set.
+Example of a feature implementation:
+
+```xml
+<linker feature="System.Runtime.OptionalBehavior" value="false">
```
vitek-karas, line 100: This will need some tweaking as we'll need to support more than one feature per `substitution.xml`.
I assume the 'value' acts as sort of a condition - "If System.Runtime.OptionalBehavior==false then do this".

accepted/2020/feature-switch.md
```diff
@@ -77,11 +77,60 @@ Ultimately some of these switches may also need VS UI integration, but for now m
...
+```
+
+Example of setting a feature property from MSBuild.
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0</TargetFramework>
+    <!-- disable the OptionalBehavior feature -->
+    <OptionalBehavior>false</OptionalBehavior>
```
marek-safar, line 117: Where will the mapping between `System.Runtime.OptionalBehavior` and `OptionalBehavior` names exist?

vitek-karas: That's the `RuntimeHostConfigurationOption` item group: https://github.com/dotnet/designs/pull/99/files#diff-c53a2566ab58ccf823f487a837d80088R84-R86

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,170 @@
...
+
+Ultimately some of these switches may also need VS UI integration, but for now most of the feature switches should be perfectly fine with only MSBuild properties.
+
+The ability to specify these from MSBuild projects as properties also means that different templates and SDKs can set different defaults - so for example Blazor or Xamarin SDKs may choose to turn off some of these features by default.
+
+Example definition of a feature flag with SDK support in [`Microsoft.NET.Sdk.targets`](https://github.com/dotnet/sdk/blob/36ef8b2aa8e5d579c921704bdab69a7407936889/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets) or by SDK components:
+```xml
+<ItemGroup>
+    ...
+    <RuntimeHostConfigurationOption Include="System.Runtime.OptionalBehavior"
```
vitek-karas, line 84: It would be a bit better to use somewhat different name to explicitly show that the MSBuild property doesn't have to match the runtime property in any way. So for example `System.Runtime.CustomBehavior` - or something (not good with names)

sbomer: I changed them a bit - not sure if they are much better since names are hard. :)

accepted/2020/feature-switch.md
sbomer: I moved it to the `assembly` level in this example. I implemented it like you suggested in https://github.com/mono/linker/pull/1132/files - see if that looks ok to you.

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,170 @@
...
+
+### Generate the right input for the linker in SDK
+
+All names/values from `RuntimeHostConfigurationOptions` will be passed to the ILLink task, which will apply any feature substitutions defined in [`substitutions.xml`](https://github.com/mono/linker/blob/master/src/linker/README.md#using-custom-substitutions). The substitutions file format will be extended to condition the substitutions based on the feature name/value. Any `RuntimeHostConfigurationOptions` which do not have feature implementations in `substitutions.xml` will not result in any modifications to the IL.
+
+Example of a feature implementation:
+
+```xml
+<linker>
+  <assembly fullname="System.Runtime.FeatureDefiningAssembly" feature="System.Runtime.OptionalFeatureBehavior" featurevalue="false">
```
marek-safar, line 101: I think this should better be feature="OptionalUserFacingBehavior". Secondly the feature should be on <linker> node not at <assembly> to allow grouping across assemblies

accepted/2020/feature-switch.md
```diff
@@ -0,0 +1,170 @@
...
+</Project>
+```
+SDK authors as always need to ensure that logic is imported in the correct order to allow users to override these default values, if allowed.
+
+## Open questions
+
+### Boolean only or more complex values
+Should the feature switches always represent simple on/off statements and thus be represented as boolean properties? In some cases the simple boolean may necessitate presence of multiple similar feature switches. For example the security algorithms case described above. In the extreme each such algorithm would have its own feature switch - this would obviously create confusion. We could potentially allow feature switches to have string values from a predefined set to solve this problem.
+
+### Which components own the mapping between different representations of feature switches
```
eerhardt, line 140: I believe this "open question" is now answered above, and this section can be removed now.  (or kept, but marked clearly as "alternate design considerations")

accepted/2020/feature-switch.md
vitek-karas: Why do you think it should be based on the MSBuild property (`OptionalUserFacingBehavior`) and not on the runtime config property (`System.Runtime.OptionalFeatureBehavior`). Note that the implementation (C# of the library) should use `AppContext.TryGetSwitch("System.Runtime.OptionalFeatureBehavior")` (the runtime config property) in order to get consistent behavior between linked and non-linked versions of the app. And since the `substitution.xml` is very close to the actual implementation in the library I actually find it nice that they both talk about the same property.

In that sense the MSBuild is really only a syntax sugar for `csproj` (and potentially for VS UI one day).

vitek-karas: > Secondly the feature should be on node not at to allow grouping across assemblies

This I don't understand. Do you mean it should look something like:
```XML
<feature name="System.Runtime.OptionalFeatureBehavior" value="false">
  <assembly fullname="System.Runtime.FeatureDefiningAssembly">
    <type fullname="System.Runtime.FeatureDefiningType">
      <method signature="System.Boolean get_IsOptionalFeatureEnabled()" body="stub" value="false"/>
    </type>
  </assembly>
</feature>
```

accepted/2020/feature-switch.md
vitek-karas: Should be fixed now. I also moved one other "open question".

accepted/2020/feature-switch.md
sbomer: Thanks for cleaning up the ones I missed.

accepted/2020/feature-switch.md
marek-safar: Sorry for not being clear. I thought about something like
```
<linker feature="System.Runtime.OptionalFeatureBehavior" 
featurevalue="false">
	...
</linker>
```

I haven't thought about allowing feature on every node which I think settles this down.

vitek-karas: The one case I'm struggling with is if we allow non-boolean feature values. Then the current approach doesn't work that well (since there's only one feature value defined per node). But I guess we can live with that....?

sbomer: If we want to support, say, integer substitutions where the integer can vary over a wide range (and assuming we don't want to put every possible value into the xml), we'd need to extend the format anyway, something like:
```xml
<method signature="System.Int32 get_FeatureInt()" body="$(FeatureInt)" />
```
where we substitute the `$(FeatureInt)` with a value known at link time. In this case the extra indirection means we will be substituting values into the substitution definitions which will be substituted into IL :)

The feature conditions just make this more redundant because it's likely we'll use a pattern like:
```xml
<assembly feature="IntegerFeature" featurevalue="$(FeatureInt)">
...
  <method signature="System.Int32 get_FeatureInt()" body="$(FeatureInt)" />
...
</assembly>
```
...or maybe in the inner scope we could reference the `featurevalue` of the outer scope... this quickly turns into a "let's design an XML-based programming language" which we already have enough of in our ecosystem. ;)

I'm hoping we can use the boolean features to get off the ground quickly, and later settle on some kind of attribute-based feature definitions instead.

marek-safar: > I'm hoping we can use the boolean features to get off the ground quickly, and later settle on some kind of attribute-based feature definitions instead.

I agree. Let's not design another DSL, if there is a need to propagate the input value we could come up with something later

vitek-karas: Just asking ... perfectly OK to support just booleans.

sbomer: Sounds good. I'm glad you brought it up - it's definitely worth noting that it doesn't extend nicely to non-booleans.

PR #101, opened 2020-03-25
Security support for new .net 5 platforms

steveisok:


danmosemsft:
adding @wfurt for TLS angle

danmosemsft:
and @samsp-msft for boringssl background.

accepted/2020/mono-convergence/security-support-new-platforms.md
```diff
@@ -0,0 +1,90 @@
...
+
+### Android
+There is currently no platform level API’s that can work with Android, so any native dependency we take must be bundled with the individual application.
+
+**System.Security.Cryptography.Native**
+
+The current dependency on OpenSSL could work as there seems to be support for building on Android.  This is nice because it would seemingly ‘just fit’ with what we already have.  However, there are some things to take into consideration:
+
+* Do we fork a copy of OpenSSL and build from that?
+* What is our process for shipping security vulnerability updates?
```
stephentoub: If we have to ship something with an app, don't these questions/concerns apply to any choice?

accepted/2020/mono-convergence/security-support-new-platforms.md
```diff
@@ -0,0 +1,90 @@
...
+
+**System.Net.Security.Native**
+
+For authentication scenarios, we’ve relied on libgss to do the work.  It is currently unknown if we *can* support it on Android.  Some initial [discovery](https://github.com/dotnet/runtime/issues/32680#issuecomment-599307356) has been started by a community member and would need to be expanded on / validated.
+
+To move forward, we should try to decide on the best approach.
+
+Do we:
+* Investigate libgss further and flush out all the dependencies / challenges?
+* Bring most / all functionality into a managed implementation?  What does that look like?  Who would do it?
```
stephentoub: What does Xamarin do today?

accepted/2020/mono-convergence/security-support-new-platforms.md
steveisok: Yep - I'll try to restate.

accepted/2020/mono-convergence/security-support-new-platforms.md
steveisok: We now use SocketsHttpHandler as the default, so the same as corefx.  The previous way was a custom implementation that is pretty old.

accepted/2020/mono-convergence/security-support-new-platforms.md
samsp-msft: I was aware that Google Play Services includes cronet & boringSSL, but they seem to only be exposed via a Java cronet API - https://developer.android.com/guide/topics/connectivity/cronet - which probably doesn't help us.

accepted/2020/mono-convergence/security-support-new-platforms.md
```diff
@@ -0,0 +1,90 @@
...
+For `System.Security.Cryptography.Native.Apple` (OSX):
+
+[src/libraries/Native/Unix/System.Security.Cryptography.Native.Apple] (https://github.com/dotnet/runtime/tree/master/src/libraries/Native/Unix/System.Security.Cryptography.Native.Apple)
+
+### Android
+There is currently no platform level API’s that can work with Android, so any native dependency we take must be bundled with the individual application.
+
+**System.Security.Cryptography.Native**
+
+The current dependency on OpenSSL could work as there seems to be support for building on Android.  This is nice because it would seemingly ‘just fit’ with what we already have.  However, there are some things to take into consideration:
```
jkotas: What does Xamarin do today for this one?

accepted/2020/mono-convergence/security-support-new-platforms.md
stephentoub: This isn't just about HttpClient, this is about all of System.Security.Cryptography.

accepted/2020/mono-convergence/security-support-new-platforms.md
stephentoub: Last I knew, Google's documented stance on BoringSSL was that it was unsupported, could change its ABI at any time, etc... exactly what you want when taking a dependency on a security library ;-)  If there's a supported/maintained variant, that's then more interesting.

bartonjs: We really, really, really want to avoid shipping a cryptography library with the product.  AFAIK, it adds a whole lot of government paperwork; and it creates a very complicated servicing model for both us and the applications built using it.

If there's any library in the box that we can reliably use, that's the best choice almost without regard to how painful it is to initially use it.

samsp-msft: I think you'd need to use the conscrypt APIs - which are higher level http APIs.
This may help: https://groups.google.com/forum/#!topic/conscrypt/T4c2YCY91qc.

accepted/2020/mono-convergence/security-support-new-platforms.md
```diff
@@ -0,0 +1,90 @@
+# Security support for new .NET 5 platforms
+
+## Overview
+Currently in .NET Core, we have a solid [strategy](https://github.com/dotnet/corefx/blob/master/Documentation/architecture/cross-platform-cryptography.md) for dealing with cryptography on Linux and OSX.  At the native level, `System.Security.Cryptography.Native` relies upon OpenSSL for Linux and `System.Security.Cryptography.Native.Apple` relies upon platform specific API’s for OSX.
```
kartheekp-ms: > https://github.com/dotnet/corefx/blob/master/Documentation/architecture/cross-platform-cryptography.md

This link is broken. May be you are referring to [this link](https://github.com/dotnet/runtime/blob/master/docs/design/features/cross-platform-cryptography.md) instead.

danmosemsft:
cc @ericstj

danmosemsft:
@steveisok who owns the next action here -- just checking we're not all waiting for each other 😺

danmosemsft:
@steveharter

steveisok:
@danmosemsft I think the doc is good to go.  Unless determining what we're going to do about OpenSSL is a pre-requisite?

filipnavara:
What's the plan about this doc? Meanwhile the Android part was implemented through JNI calls and iOS through Apple APIs. Most of the text seems outdated in this regard.

steveisok:
Hmm, thought this was in.  Lost in the .NET 5 - 6 shuffle.  It is a bit dated - I can give it some updates as we've made some progress.

PR #103, opened 2020-03-26
Platform Specific HttpClient Support 

steveisok:


danmosemsft:
For stuff that relates to WASM specific implementation like this, should we have a review from someone on Blazor?

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+Follow proposal #2b and hook up existing AndroidClientHandler using `Activator.CreateInstance (``"``Xamarin.Android.Net.AndroidClientHandler, Mono.Android``"``)`. The whole AndroidClientHandler implementation will remain inside `xamarin/xamarin-android`.
+
+Only RID specific version of System.Net.Http will be built as there won’t be any new API available as part of `dotnet/runtime` version.
+
+The code will most likely have to include special linker annotations to keep the dependency on the constructor inside external “unknown” assembly.
+
+### WebAssembly
+Follow proposal #5 and port WasmHttpMessageHandler to dotnet/runtime together with wasm interop layer.
+
+The public implementation of WasmHttpMessageHandler will be packaged into special NuGet which will be create to hold only this public type. The interop layer needed to support this handler will be also included into the NuGet package. To match existing naming it will be called
```
karelz, line 435:
```suggestion
The public implementation of WasmHttpMessageHandler will be packaged into special NuGet which will be created to hold only this public type. The interop layer needed to support this handler will be also included into the NuGet package. To match existing naming it will be called
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+* System.Net.Http.WebAssemblyHttpHandler
+
+## Future plans
+If we manage to extract Objective-C or Java interop into special subcomponents we might be able to switch to solution which will allow us to ship and test everything inbox (inside dotnet/runtime).
+
+### Xamarin.iOS/Xamarin.tvOS/Xamarin.watchOS
+Follow proposal #5 and port [NSUrlSessionHandler](https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/NSUrlSessionHandler.cs) to `dotnet/runtime` repo and extend existing interop layer to include bridge to platform specific APIs to support the implementation. The sources will be located under [src/libraries/System.Het.Http/](https://github.com/dotnet/runtime/tree/master/src/libraries/System.Net.Http/).
```
karelz, line 443:
```suggestion
Follow proposal #5 "All HttpHandlers are fully implemented inbox" and port [NSUrlSessionHandler](https://github.com/xamarin/xamarin-macios/blob/master/src/Foundation/NSUrlSessionHandler.cs) to `dotnet/runtime` repo and extend existing interop layer to include bridge to platform specific APIs to support the implementation. The sources will be located under [src/libraries/System.Het.Http/](https://github.com/dotnet/runtime/tree/master/src/libraries/System.Net.Http/). 
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+### WebAssembly
+Follow proposal #5 and port WasmHttpMessageHandler to dotnet/runtime together with wasm interop layer.
+
+The public implementation of WasmHttpMessageHandler will be packaged into special NuGet which will be create to hold only this public type. The interop layer needed to support this handler will be also included into the NuGet package. To match existing naming it will be called
+
+* System.Net.Http.WebAssemblyHttpHandler
+
+## Future plans
+If we manage to extract Objective-C or Java interop into special subcomponents we might be able to switch to solution which will allow us to ship and test everything inbox (inside dotnet/runtime).
```
karelz, line 440:
```suggestion
If we manage to extract Objective-C or Java interop into special subcomponents we might be able to switch to solution which will allow us to ship and test everything inbox (inside `dotnet/runtime` repo).
```

danmosemsft:
@stephentoub can you please review (or sign off if you did already)?

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+Extend HttpClient to be configurable via static configuration type.
+```c#
+public static class HttpClientHandlers
+{
+  public static HttpClientHandler DefaultPlatformHandler { get; internal set; }
+}
+```
+
+```
```
karelz, line 109:
```suggestion
```c#
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+```c#
+// System.Net.Http.HandlerFactory.dll
+public static class HttpHandlerFactory
+{
+  // It could be of object type not to take dependency on System.Net.Http
+  public HttpClientHandler Create () => throw new NotImplementedException();
+}
+```
+
+```
```
karelz, line 257:
```suggestion
```c#
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+### Handlers in Detail
+**SocketsHttpHandler Handler**
+
+The handler which is purely managed and behaves generally the same across all platforms except WatchOS (sockets are not available). The downside of using it is mostly felt in performance compared to the native handlers.
+
+**CFNetwork Handler**
+
+The CFNetwork handler is based in the old Network Stack API provided by Apple. This handler, while it has better performance, lacks features compared with the other ones. It is mostly around for legacy reasons such as:
+
+* Support old servers that do not work with NSUrlSessionHandler
```
spouliot, line 60: I don't recall server incompatibilities, maybe SSLv2 ?

`CFNetworkHandler` mostly exists because `NSUrlSession` API came later (and it was too late to remove without breaking code).

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+* Support old devices that do not support the newer NSUrlSessionHandler.
```
spouliot, line 61: More like **old OS**.
Xamarin.iOS supports iOS 7+ where CFNetwork is available.
However `NSUrlSession` API came with iOS 8+.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+**NSUrlSession Handler**
+
+The NSUrlSession handler is the most modern of the native handlers provided. This handler has great performance since it relies on the OS and provides caching and other advanced features such as proxy support. The downside of using the handler comes in behaviour differences between the other platforms.
```
spouliot, line 65: The native (NSUrlSession and CFNetwork) also works great with cellular and, when needed, switching back and forth between the wifi/cellular antennas.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+**SocketsHttpHandler Handler**
+
+It’s not supported at all due to lack of Sockets on the platform. Changing it to throw PNSE for every public member will maintain API compatibility and reduce size.
+
+
+## .NET 5 Adoption Proposals
+Platform-specific handlers are implemented on top of the enhanced platform interop layer. They are in most cases relatively complex even with such abstractions available. It’s possible to implement them fully using basic interop mechanics like pinvoke but that would be tedious and very error-prone due to duplication of all the interop subtle details Xamarin and WebAssembly SDK built over the years. We would also need to introduce Xamarin namespace and build TFM specific versions of relevant assemblies as well as convert all publicly exposes types to primitive types or types which derive only from BCL types to avoid second-level dependencies which might not be desirable.
+
+In general, keeping the platform dependant handlers away from .NET Core BCL and having them available only as part of binding APIs in optional SDKs (e.g. Xamarin.iOS) seems like the most effective integration.
```
spouliot, line 97: ❤️

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+  {
+    platformInstance = Type.GetType ("HttpClientHandlerFactory").GetMethod("Create").Invoke () as HttpClientHandler;
+  }
+}
+```
+
+**Disadvantages**
+
+* Costly and slow reflection infrastructure would be pulled in.
+* Need to add PreserveDependency to all possible target assemblies.
```
spouliot, line 243: They are only issues if this was the only option.
If you use reflection (and I rather not) for debug/linker-less builds then you save more time (linking) than the execution time (for reflection).

For release/optimized builds you want the linker and that removes the disadvantages.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+// System.Net.Http.dll
+public HttpClientHandler ()
+{
+#if iOS
+  platformInstance = Xamarin.iOS.HttpHandleFactory.Create();
+#else
+...
+#endif
+}
+```
```
spouliot, line 283: **Disadvantages**

* The assembly `System.Net.Http.dll` is different for each platform

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+**Disadvantages**
+
+* Single type assembly is against our design practices
+* A circular dependency between System.Net.Http.Private.dll and System.Net.Http.dll might exist in optional components
+* API contract would live in different assembly than implementation
+* Reliance on more dotnet publish to figure out right implementation assembly
+* Adding public member to HttpClientHandler would be a breaking change
+* HttpClientHandler acts as a wrapper in most SDKs and the code would have to be duplicated
+* Platform specific implementation would need to write their own copy of proxy logic
```
spouliot, line 326: That's a breaking change for SDK since they already have the handlers (plural for XI/XM).
The solution must allow the selection of `CFNetworkHander` and `NSUrlSessionHandler` (today).
We might be able to kill `CFNetworkHander` (at some point, e.g. Apple dropping iOS7 support) but Apple might decide to _gift us_ something new (forcing us to support more than one handler again).

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+#else if ..
+
+#endif
+```
+
+**Disadvantages**
+
+* dotnet/runtime will need to include reference assemblies of all optional components to be able to build ref and impl assemblies.
+* Public HttpClientHandler API shape might be different between TFM as it’s controlled by optional components
+* Updating HttpClientHandler public surface for optional component will require build a new version of dotnet/runtime
```
spouliot, line 373: How is that gonna be tested ?

What happen when a new native API is added and support must be added back to `dotnet` and backported to current releases (in sync with other changes made in the SDK) ?

It might require a newer Xcode (than the one currently used) and that can bring its own fun (like the issue we had with Xcode 11.4).

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+**Conclusion**
+
+The big advantage of this proposal is that the default HttpClientHandler would be have the most optimal implementation on every platform (performance and size wise). The availability of only platform specific properties on HttpClientHandler has its pros & cons as well.
+
+### Proposal #5: All HttpHandlers are fully implemented inbox
+
+Implement all platform-specific handlers fully inside dotnet/runtime using platform-specific interop libraries. We would extend low-level interop API which and use it to implement the managed API for HttpClientHandler and other types which are exposed from it.
+
+There is already a foundation for CF interop at [https://github.com/dotnet/runtime/tree/master/src/libraries/Common/src/Interop/OSX](https://github.com/dotnet/runtime/tree/master/src/libraries/Common/src/Interop/OSX). This will need to be extended to include Objective-C bridge for necessary APIs.
```
spouliot, line 383: It's not an extension to the current interop. What you call "CF interop" is C-based, with **manual** ref-counted memory, and p/invoke support (for C) are already there.

For ObjC you need things are much more complex - we have a runtime/tooling (the SDK) that just do that (the rest are just bindings, what we're talking to move here). Doing a submodule of `xamarin-macios` sounds counter-productive.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+**Disadvantages**
+
+* Implementation of HttpClientHandler might require implementing more dependent type (e.g. Android.Net.SSLSocketFactory)
+* Problems with handling platform-specific exceptions
+* Not all decorations used for Xamarin bindings will be available (e.g. [iOS (7,0)])
+* Starting from scratch for Android
+
+**Conclusion**
+
+No build magic and more shared code prospect, as well as the ability to test HttpClient changes easily across all platforms, are very appealing.
```
spouliot, line 396: It won't ease testing as you won't have the (ObjC for XI/XM) runtime. This is creating a new cyclic dependency as you'll need the SDK to create something going back to a different/newer SDK.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+​​  public partial class HttpClientHandler
+​​  {
+​​    internal static HandlerType DefaultHandlerType { get; } = HttpClientHandlerType.PlatformNative;
+​​  }
+}
+```
+
+The implementation has to use the property in the way to allow the ILLinker to remove the unnecessary dependency of unused handler when the type is not used elsewhere.
+
+### Xamarin.iOS/Xamarin.tvOS
```
spouliot, line 416: Xamarin.watchOS too

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+Follow proposal #2b and hook up existing `NSUrlSessionHandler` using `Activator.CreateInstance ("Foundation.NSUrlSessionHandler, Xamarin.iOS/Xamarin.tvOS")​`. The actual implementation will remain inside xamarin/xamarin-macios​ in repo.
+
+The plan ignores CFNetworkHandler and other legacy handlers. They will stay in `xamarin/xamarin-macios` and will be available only when constructed manually. The implementation logic inside HttpClientHandler will rely on DefaultHandlerType selection to set the implicit handler behaviour between NSUrlSessionHandler and SocketHttpHandler.
```
spouliot, line 419: `CFNetworkHandler` is the only native option for iOS7.x

danmosemsft:
@SteveSandersonMS do we need review from Blazor person? Or is this more of a Mono/Webasm implementation detail? I do'nt know where the boundary lies.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+**Example**
+```c#
+// This should use the default message handler for the application. Either as
+// preselected by the framework or set in the Project Options for the project.
+
+HttpClient client = new HttpClient();
+
+// Same behavour should be observed when using following code
+HttpClient client = new HttpClient(new HttpClientHandler());
```
stephentoub, line 21: I assume there are at least some properties on HttpClientHandler that won't work on all the platforms?  Might be worth calling that out here, that while `new HttpClient()` and `new HttpClient(new HttpClientHandler())` should work fine and identically, `new HttpClient(new HttpClientHandler() { SomeProperty = ... })` may not be supported everywhere, presumably throwing PlatformNotSupportedException where it's not.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+There are several reasons why using `SocketHttpHandler` is not desirable or not even possible on some platforms
+
+* Lack of platform native TLS support for better performance
+* Bloats the application size
+* Performance and power consumption impact of managed implementation
+* Integration with the platform settings
+* Unavailability of Socket APIs
```
stephentoub, line 30:
```suggestion
* Integrated TLS. A TLS implementation integrated with the native HTTP stack can lead to better performance in various situations.
* Application size. Including an HTTP stack when the mobile platform already includes one can lead to an unnecessarily larger application size.
* Unavailability of Socket APIs.  Some platforms simply don't expose the socket primitives necessary to implement SocketsHttpHandler.
* Power consumption. Native HTTP stacks on various mobile platforms are able to more effectively manage power consumption as it relates to things like when communication devices are enabled or powered down.
* Platform settings. Some platform-level settings aren't exposed in a way that SocketsHttpHandler could view them but are directly integrated with the native HTTP stack.
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+* [cachePolicy](https://developer.apple.com/documentation/foundation/nsurlrequest/cachepolicy/useprotocolcachepolicy)
+* [allowsConstrainedNetworkAccess](https://developer.apple.com/documentation/foundation/nsmutableurlrequest/3325676-allowsconstrainednetworkaccess)
+* [allowsExpensiveNetworkAccess](https://developer.apple.com/documentation/foundation/nsmutableurlrequest/3325677-allowsexpensivenetworkaccess)
+* [networkServiceType](https://developer.apple.com/documentation/foundation/nsmutableurlrequest/1412378-networkservicetype)
+* [ReadTimeout](https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setReadTimeout(int))
+* [AllowUserInteraction](https://docs.oracle.com/javase/7/docs/api/java/net/URLConnection.html#setAllowUserInteraction(boolean))
+
+## Xamarin.iOS/tvOS/watchOS
+There are 3 handlers available where **NSUrlSession** handler is the default. Except for watchOS where only NSUrlSessionHandler is supported.
```
stephentoub, line 45:
```suggestion
There are 3 handlers available (SocketsHttpHandler, CFNetworkHandler, and NSUrlSessionHandler), with **NSUrlSessionHandler** as the default (except on watchOS, where the only supported handler is NSUrlSessionHandler).
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+The handler is only preselected as the default with [UI options available](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/http-stack#selecting-an-httpclient-stack) using VS Project Settings dialog to change them. This is required to “fix” existing HttpClient code which was written with the assumption that the default handler will be the one working best on the platform.
+
+The ILLinker is used to remove the unnecessary handlers from the final binary in this case.
+
+### Handlers in Detail
+**SocketsHttpHandler Handler**
+
+The handler which is purely managed and behaves generally the same across all platforms except WatchOS (sockets are not available). The downside of using it is mostly felt in performance compared to the native handlers.
```
stephentoub, line 54:
```suggestion
The handler, which is purely managed, and behaves generally the same across all platforms except WatchOS (sockets are not available). The downside of using it is mostly felt in performance compared to the native handlers.
```

stephentoub, line 54: Do we have data on the performance difference?

stephentoub, line 54: It's a managed implementation, but for a few very specific things, mainly related to getting proxy information from the OS, we do P/Invoke.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+The users can choose from two handlers with AndroidClientHandler being the default and preferred over socket-based handler. The setting of which handler to use is also available via [Project Settings](https://docs.microsoft.com/en-us/xamarin/android/app-fundamentals/http-stack?context=xamarin%2Fcross-platform&tabs=windows) as for Xamarin.iOS.
+
+### Handlers in Detail
+**SocketsHttpHandler Handler**
+
+The handler which is purely managed and behaves generally the same across all platforms. The same downsides as iOS apply here as well.
+
+**AndroidClientHandler Handler**
+
+The AndroidClientHandler is a bridge for Java.Net.HttpURLConnection. Not only does it provide a better performance, but it also helps drive a more consistent Android experience when it comes to networking.
```
stephentoub, line 77: Here as well I'd be interested in understanding the performance difference (e.g. throughput, not power).

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+* It handles only HttpClient API and we’d need to “obsolete” manual HttpClientHandler construction.
+* Initialization of static state from Xamarin SDKs could be ugly.
+
+**Conclusion**
+
+REJECTED.  It only supports `new HttpClient()`, not `new HttpClient(new HttpClientHandler())`.
+
+### Proposal #2: Convert HttpClientHandler to Proxy pattern
+Make HttpClientHandler a partial class, make all its members virtual, and on iOS/Android have the implementation of the class wrap an HttpClientHandler instance field. The instantiation of the underlying instance will have the flexibility to be controlled by platform implementation.
```
stephentoub, line 131: Why "partial"?  Making it a partial class is just a compiler trick to support various ways of structuring the code; all of the partial components would still need to be compiled in at the same time.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+
+**Located in dotnet/runtime repo**
+```c#
+public class HttpClientHandler
+{
+  readonly HttpClientHandler platformInstance;
+
+  public HttpClientHandler ()
```
stephentoub, line 139: I believe we'd also need a new protected ctor? e.g. the one the derived type below is passing `true` to.

accepted/2020/mono-convergence/platform-specific-httpclient.md
stephentoub: > The assembly System.Net.Http.dll is different for each platform

That's already the case.  We build System.Net.Http.dll for Windows, for Linux, and for macOS.  While 99.9% of the code is common between them, there's a tiny bit of code in each that does P/Invokes for the given OS to look up things like proxy settings.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+Follow proposal #2b and hook up existing AndroidClientHandler using `Activator.CreateInstance (``"``Xamarin.Android.Net.AndroidClientHandler, Mono.Android``"``)`. The whole AndroidClientHandler implementation will remain inside `xamarin/xamarin-android`.
+
+Only RID specific version of System.Net.Http will be built as there won’t be any new API available as part of `dotnet/runtime` version.
+
+The code will most likely have to include special linker annotations to keep the dependency on the constructor inside external “unknown” assembly.
+
+### WebAssembly
+Follow proposal #5 and port WasmHttpMessageHandler to dotnet/runtime together with wasm interop layer.
+
+The public implementation of WasmHttpMessageHandler will be packaged into special NuGet which will be created to hold only this public type. The interop layer needed to support this handler will be also included into the NuGet package. To match existing naming it will be called
```
stephentoub, line 435: Based on the API review we did of WasmHttpMessageHandler, this case may end up being revisited, mainly because the settings it was exposing are meant to change per request.

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Removed the text

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Remove the text

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Incorporated into the description

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: No (changed the text to not mention performance).

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Fixed

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: I don't think this matters for netcoreapp side

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: That's not what the text is saying. There would be no dependency on XI runtime all needed code would exist inside dotnet/runtime.

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Tweaked the text to make it clear the code would be independent to xamarin-macios

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: > How is that gonna be tested ?

Same as other proposals depending on implementation in other repo.

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: The proposal allows any logic to exist in SDKs so they could create any handler

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: That's what the text says already `Costly and slow reflection infrastructure would be pulled in`

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+  {
+    throw new PlatformNotSupportedException();    
+  }
+}
+```
+
+**Disadvantages**
+
+* Requires ILLinker to always run which slows down build process.
+* Changing internal state of HttpClientHandler will be tricky as it can break hook used by ILLinker.
```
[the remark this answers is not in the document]
marek-safar: You are comparing different setups. Today you ship the implementation with your SDK that won't be the case in .net5 and you will be poking inside "unknown" assembly

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+  private static HttpClientHandler CreateNewInstance ()
+  {
+    throw new PlatformNotSupportedException();    
+  }
+}
+```
+
+**Disadvantages**
+
+* Requires ILLinker to always run which slows down build process.
```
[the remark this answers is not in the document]
marek-safar: That would be another proposal, feel free add it to the list if you think it's important

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+* Testing will have to be done at the integration level.
+* The existing handlers would need to be changed to derive from common type the proxy could operate on (HttpClientHandler, an interface, or something else).
+* All HttpClientHandler properties will have to be virtual or we’ll have to introduce IHttpClientHandler which would also solve the recursion problem.
+* Every method and property will have to have implementation override otherwise will hang in recursion / stack overflow.
+* Factory for creating the platform instance won’t be part of public contract.
+* Adding new public members will be difficult without IHttpClientHandler.
+
+**Conclusion**
+
+This is the cheapest option to implement which meets our needs as it’s closest to what we Mono and Xamarin use today.
```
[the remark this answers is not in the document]
marek-safar: That not does satisfy the need from libraries team to support `new HttpClientHandler()`

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,452 @@
...
+```
+
+**Disadvantages**
+
+* It handles only HttpClient API and we’d need to “obsolete” manual HttpClientHandler construction.
+* Initialization of static state from Xamarin SDKs could be ugly.
+
+**Conclusion**
+
+REJECTED.  It only supports `new HttpClient()`, not `new HttpClient(new HttpClientHandler())`.
```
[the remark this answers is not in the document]
marek-safar: Please see `Required Scenarios` section for what has to be supported

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Included property sample

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: We don't have such data but it might be possible to gather them once we have XA working on net5

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Fixed

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Fixed, `true` was left-over from the previous version

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: The design here does not go into such details so that should be fine but I updated the names to reflect the latest naming

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,456 @@
...
+The handler is only preselected as the default with [UI options available](https://docs.microsoft.com/en-us/xamarin/cross-platform/macios/http-stack#selecting-an-httpclient-stack) using VS Project Settings dialog to change them. This is required to “fix” existing HttpClient code which was written with the assumption that the default handler will be the one working best on the platform.
+
+The ILLinker is used to remove the unnecessary handlers from the final binary in this case.
+
+### Handlers in Detail
+**SocketsHttpHandler Handler**
+
+It's a managed implementation, but for a few very specific things, mainly related to getting proxy information from the OS, it does P/Invoke. The handler behaves generally the same across all platforms where sockets are available. The downside of using it is mostly felt in networking connection switching when needed, for examaple when switching back and forth between the wifi/cellular antennas.
+
+It's a managed implementation, but for a few very specific things, mainly related to getting proxy information from the OS, we do P/Invoke.
```
rolfbjarne, line 59: The sentence `It's a managed implementation, but for a few very specific things, mainly related to getting proxy information from the OS, we do P/Invoke.` is duplicated.

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,456 @@
...
+
+**CFNetwork Handler**
+
+The CFNetwork handler is based in the old Network Stack API provided by Apple. This handler, while it has better performance, lacks features compared with the other ones.
+
+**NSUrlSession Handler**
+
+The NSUrlSession handler is the most modern of the native handlers provided. This handler has great performance since it relies on the OS and provides caching and other advanced features such as proxy support. The downside of using the handler comes in behaviour differences between the other platforms.
```
rolfbjarne, line 67: This sounds better:

```suggestion
The NSUrlSession handler is the most modern of the native handlers provided. This handler has great performance since it relies on the OS and provides caching and other advanced features such as proxy support. The downside of using this handler comes in behaviour differences between the other platforms.
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
```diff
@@ -0,0 +1,456 @@
...
+
+This handler is a special bridge between C# and Javascript. As Wasm matures, the shape of the handler and what it supports will change significantly and we could also rewrite it to be WASI based.
+
+**SocketsHttpHandler Handler**
+
+It’s not supported at all due to lack of Sockets on the platform. Changing it to throw PNSE for every public member will maintain API compatibility and reduce size.
+
+
+## .NET 5 Adoption Proposals
+Platform-specific handlers are implemented on top of the enhanced platform interop layer. They are in most cases relatively complex even with such abstractions available. It’s possible to implement them fully using basic interop mechanics like pinvoke but that would be tedious and very error-prone due to duplication of all the interop subtle details Xamarin and WebAssembly SDK built over the years. We would also need to introduce Xamarin namespace and build TFM specific versions of relevant assemblies as well as convert all publicly exposes types to primitive types or types which derive only from BCL types to avoid second-level dependencies which might not be desirable.
```
rolfbjarne, line 97:
```suggestion
Platform-specific handlers are implemented on top of the enhanced platform interop layer. They are in most cases relatively complex even with such abstractions available. It’s possible to implement them fully using basic interop mechanics like P/Invoke but that would be tedious and very error-prone due to duplication of all the interop subtle details Xamarin and WebAssembly SDK built over the years. We would also need to introduce Xamarin namespace and build TFM specific versions of relevant assemblies as well as convert all publicly exposes types to primitive types or types which derive only from BCL types to avoid second-level dependencies which might not be desirable.
```

accepted/2020/mono-convergence/platform-specific-httpclient.md
marek-safar: Fixed

danmosemsft:
@marek-safar is this good to merge now?

PR #104, opened 2020-03-27
Add spec for workload resolution in MSBuild

dsplaisted:
This spec covers some of the implementation for optional SDK workloads in .NET 5.  The overall spec is in #100.

### DIFF
--- a/accepted/2020/workloads/workload-resolvers.md
+++ b/accepted/2020/workloads/workload-resolvers.md
@@ -37,7 +37,7 @@ For .NET 5, iOS, Android, and WPF/Windows Forms project will all use the same `M
 
 ## Workload guidance
 
-The bulk of functionality from orkloads will be contained in workload "packs", which will be the unit of delivery for workloads.  This will include MSBuild targets and tasks.  The assets from the packs will be laid out under the dotnet root folder, probably in `packs/<Pack ID>/<Pack Version>` or a similar folder.
+The bulk of functionality from workloads will be contained in workload "packs", which will be the unit of delivery for workloads.  This will include MSBuild targets and tasks.  The assets from the packs will be laid out under the dotnet root folder, probably in `packs/<Pack ID>/<Pack Version>` or a similar folder.
 
 Workloads will also provide a manifest package which describes which packs are necessary for which workloads.  The manifest will specify the name and version for each pack that is available for a given .NET SDK release band, as well as which packs are required for each logical workload.  The manifest for workloads will be present even when the workloads it describes are not installed.
 
@@ -197,7 +197,7 @@ The .NET SDK (Microsoft.NET.Sdk) should:
 
 ### VS behavior
 
-When the VS project system loads a project, it should check the evaluation resort for `MissingWorkloadPack` items.  If there are any, then it should raise the VS event to veto the project load and trigger the in-product acquisition experience for those workloads.
+When the VS project system loads a project, it should check the evaluation result for `MissingWorkloadPack` items.  If there are any, then it should raise the VS event to veto the project load and trigger the in-product acquisition experience for those workloads.
 
 In the same way that the SDK error generation does, the project system should map from the missing workload packs to the missing workloads via the workload manifests.  Additionally, it will need to map from the workload IDs that the .NET SDK uses to the right IDs that VS will use to install the workloads.  That mapping should either be defined in the workload manifest or in another file that is part of the same packgae.
 
--- a/accepted/2020/workloads/workload-resolvers.md
+++ b/accepted/2020/workloads/workload-resolvers.md
@@ -9,7 +9,7 @@ In .NET 5, we will add support for iOS and Android.  The .NET SDK (formerly know
 
 ## Experiences
 
-- Projects using workloads just work if the workloads are installed
+- Projects dependent on a workload just work if the workload is installed
 - When opening a project in Visual Studio where the project depends on a workload that is not installed, Visual Studio will use the in-product acquisition experience to pop up a dialog box that notifies the developer what workloads are required, where they can click through to install the workload.
 - When building a project from the command line where the project depends on a workload that is not installed, a friendly error message will be generated specifying which workloads are required, and how to install them (ie by running the VS installer)
 - Improve existing experiences involving SDK resolvers
@@ -39,13 +39,13 @@ For .NET 5, iOS, Android, and WPF/Windows Forms project will all use the same `M
 
 The bulk of functionality from workloads will be contained in workload "packs", which will be the unit of delivery for workloads.  This will include MSBuild targets and tasks.  The assets from the packs will be laid out under the dotnet root folder, probably in `packs/<Pack ID>/<Pack Version>` or a similar folder.
 
-Workloads will also provide a manifest package which describes which packs are necessary for which workloads.  The manifest will specify the name and version for each pack that is available for a given .NET SDK release band, as well as which packs are required for each logical workload.  The manifest for workloads will be present even when the workloads it describes are not installed.
+Workloads will also provide a manifest package which describes which packs are necessary for which workloads.  The manifest will specify the name and version for each pack that is available for a given .NET SDK feature band, as well as which packs are required for each logical workload.  The manifest for workloads will be present even when the workloads it describes are not installed.
 
 ### Workload Manifest Targets Files
 
 In addition to the manifest file itself (likely in json format), the workload manifest package will include a `WorkloadManifest.targets` file which should import the necessary .targets files from the workload pack if a project depends on the workload.  The `WorkloadManifest.targets` files from all workload manifests will be imported into all projects that use Microsoft.NET.Sdk.
 
-The `WorkloadManifest.targets` should use MSBuild Sdk imports to import .targets files from an workload pack.  This should be done conditionally based on properties that determine whether the workload should be used.  For example, the Android workload manifest targets file might look like this:
+The `WorkloadManifest.targets` should use MSBuild Sdk imports to import .targets files from a workload pack.  This should be done conditionally based on properties that determine whether the workload should be used.  For example, the Android workload manifest targets file might look like this:
 
 ```xml
 <Import Project="Sdk.targets" Sdk="Xamarin.Android.Sdk"
@@ -96,7 +96,7 @@ This would import the iOS targets from the Xamarin.iOS.Sdk.15.0 workload pack if
 
 Workloads may also provide MSBuild logic in an `AutoImport.props` file in the Sdk folder of the workload pack which will be imported before the body of the project file.  We expect this to be use for default globs, for example to specify that all .xaml files should be included as `Page` items.
 
-However, `AutoImport.props` files from workloads are subject to restrictions and must be carefully authored.  This is because since they need to be imported before the body of the project is evaluated, they can't be included conditionally based on the properties in the project that define whether the workload is in use.  So the `AutoImport.props` files from all workload packs will be imported for all projects using Microsoft.NET.Sdk.
+However, `AutoImport.props` files from workloads are subject to restrictions and must be carefully authored.  This is because since they need to be imported before the body of the project is evaluated, they can't be included conditionally based on the properties in the project that define whether the workload is in use.  So the `AutoImport.props` files from all workload packs will be imported for all projects using `Microsoft.NET.Sdk`.
 
 MSBuild uses multiple phases for evaluation, and properties and imports are evaluated in a phase before items.  This means it is possible to have conditions on item or `ItemGroup` elements in a .props file that depend on properties that are defined in the body of the project or in .targets files.
 
@@ -177,7 +177,7 @@ However, if the requested SDK name is `Microsoft.NET.SDK.WorkloadAutoImportProps
 The .NET SDK (Microsoft.NET.Sdk) should:
 
 - Set the `WorkloadManifestRoot` property to a path where workload manifest packages for the current version band will be laid out.
-  - This should be `<DOTNET ROOT>/workloadmanifests/<VERSION BAND>/`
+  - This should be `<DOTNET ROOT>/workloadmanifests/<FEATURE VERSION BAND>/`
   - For example, `c:\Program Files\dotnet\workloadmanifests\5.0.100\`
   - Workload manifest packages should be laid out in folders under the root corresponding to the name of the workload manifest, but should not include a version in the path.  They should be updated in-place.  This allows the SDK to import all of the active `WorkloadManifest.targets` files with a simple wildcard import
 - Include the following .targets import:
@@ -199,7 +199,7 @@ The .NET SDK (Microsoft.NET.Sdk) should:
 
 When the VS project system loads a project, it should check the evaluation result for `MissingWorkloadPack` items.  If there are any, then it should raise the VS event to veto the project load and trigger the in-product acquisition experience for those workloads.
 
-In the same way that the SDK error generation does, the project system should map from the missing workload packs to the missing workloads via the workload manifests.  Additionally, it will need to map from the workload IDs that the .NET SDK uses to the right IDs that VS will use to install the workloads.  That mapping should either be defined in the workload manifest or in another file that is part of the same packgae.
+In the same way that the SDK error generation does, the project system should map from the missing workload packs to the missing workloads via the workload manifests.  Additionally, it will need to map from the workload IDs that the .NET SDK uses to the right IDs that VS will use to install the workloads.  That mapping should either be defined in the workload manifest or in another file that is part of the same package.
 
 QUESTION: Will it be OK to put this mapping of .NET workload IDs to VS workload/component IDs in the workload manifest packages?  Or do we need to supply it some other way?
 
--- a/accepted/2020/workloads/workload-resolvers.md
+++ b/accepted/2020/workloads/workload-resolvers.md
@@ -104,7 +104,7 @@ So workloads may include MSBuild items in the `AutoImport.props` file if they ar
 
 Note that in contrast to the `WorkloadManifest.targets` file, which is part of the workload manifest package, the `AutoImport.props` files are part of the workload packs (under the Sdk folder).  This is so that they can be imported only if the workload is installed, rather than imported for all workloads whether they are installed or not.  This reduces the impact of a bug in an `AutoImport.props` file, and means that if there is an issue in the .props file, it can be worked around by uninstalling the workload.
 
-As an example, here is a slightly simplified version of what the .props file for WPF might look like:
+As an example, here is a slightly simplified version of what the `AutoImport.props` file for WPF might look like:
 
 ```xml
 <Project>
--- a/accepted/2020/workloads/workload-resolvers.md
+++ b/accepted/2020/workloads/workload-resolvers.md
@@ -160,10 +160,9 @@ We will create an MSBuild SDK resolver to handle workloads.  Its behavior should
 - Look up the Name of the requested SDK in the workload manifests, matching to the workload pack IDs
 - If a matching workload pack is found in a workload manifest, then get its version number, and use that version number to generate the path where that workload pack would be installed.  This should be the following, or similar: `<DOTNET ROOT>/packs/<PACK ID>/<PACK VERSION>/Sdk`
 - If the workload pack is installed (ie the path (except the final `Sdk`) exists), then return the generated path
-- If the workload pack is not installed, then the returned `SdkResult` should:
-  - Have `Success` set to true
-  - Return no SDK paths
-  - Include a `MissingWorkloadPack` item to add to the evaluation result.  The identity should be the name of the workload pack / MSBuild SDK that was requested.  It should have `Version` metadata set to the version that was read from the workload manifest.
+- If the workload pack is not installed, then the resolver should return a successful `SdkResult` which:
+  - Has no resolved SDK Paths
+  - Includes a `MissingWorkloadPack` item to add to the evaluation result.  The identity should be the name of the workload pack / MSBuild SDK that was requested.  It should have `Version` metadata set to the version that was read from the workload manifest.
 
 However, if the requested SDK name is `Microsoft.NET.SDK.WorkloadAutoImportPropsLocator`, then the resolver should instead behave as follows:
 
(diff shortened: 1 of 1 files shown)


accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,238 @@
...
+- [Target Framework Names in .NET 5](https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md)
+- [NET SDK workloads overview](https://github.com/dotnet/designs/blob/workloads/accepted/2020/workloads/workloads.md) ([PR](https://github.com/dotnet/designs/pull/100))
+
+## Experiences
+
+- Projects using workloads just work if the workloads are installed
+- When opening a project in Visual Studio where the project depends on a workload that is not installed, Visual Studio will use the in-product acquisition experience to pop up a dialog box that notifies the developer what workloads are required, where they can click through to install the workload.
+- When building a project from the command line where the project depends on a workload that is not installed, a friendly error message will be generated specifying which workloads are required, and how to install them (ie by running the VS installer)
+- Improve existing experiences involving SDK resolvers
+  - Improve experience in VS when opening a project with a project.json that specifies an SDK that can't be found
```
akoeplinger:
```suggestion
  - Improve experience in VS when opening a project with a global.json that specifies an SDK that can't be found
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,238 @@
...
+## Project files
+
+For .NET 5, iOS, Android, and WPF/Windows Forms project will all use the same `Microsoft.NET.Sdk` MSBuild SDK in the project file.  The `TargetFramework` will be used to specify the operating system, and a project targeting Windows will use the `UseWPF` or `UseWindowsForms` properties to opt in to WPF and/or Windows Forms.  For example, we plan for an Android project to look something like this:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net5.0-android42</TargetFramework>
```
akoeplinger: nit: Android API level 42 doesn't exist and since this section doesn't explain the versioning I'd pick an existing API level like `android21` to reduce confusion

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,238 @@
...
+  </PropertyGroup>
+
+</Project>
+```
+
+## Workload guidance
+
+The bulk of functionality from orkloads will be contained in workload "packs", which will be the unit of delivery for workloads.  This will include MSBuild targets and tasks.  The assets from the packs will be laid out under the dotnet root folder, probably in `packs/<Pack ID>/<Pack Version>` or a similar folder.
```
akoeplinger:
```suggestion
The bulk of functionality from workloads will be contained in workload "packs", which will be the unit of delivery for workloads.  This will include MSBuild targets and tasks.  The assets from the packs will be laid out under the dotnet root folder, probably in `packs/<Pack ID>/<Pack Version>` or a similar folder.
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,238 @@
...
+
+- Include the following import in one of its .props files:
+
+```xml
+<Import Project="AutoImport.props" Sdk="Microsoft.NET.SDK.WorkloadAutoImportPropsLocator">
+```
+
+### VS behavior
+
+When the VS project system loads a project, it should check the evaluation resort for `MissingWorkloadPack` items.  If there are any, then it should raise the VS event to veto the project load and trigger the in-product acquisition experience for those workloads.
```
akoeplinger:
```suggestion
When the VS project system loads a project, it should check the evaluation result for `MissingWorkloadPack` items.  If there are any, then it should raise the VS event to veto the project load and trigger the in-product acquisition experience for those workloads.
```

Nirmal4G:
What happened to the Sdks concept?. Instead of introducing a new one, we should expand the capabilities of the Sdk concept.

There are already Framework, MetaPackages and TargetingPacks. We should use those to provide a modular SDK.

For E.g.:

`<Vendor>.<Platform>.Sdk` -> meta SDK for a platform provides a composite set of SDKs that provide various technology targets to build on.

Let's take an example of Windows platform. It has WPF, WinForms, UAP, WinNT APIs
 you can essentially have...

`Microsoft.Windows.Sdk`, `Microsoft.Windows.Universal.Sdk` (UAP) and `Microsoft.Windows.Desktop.Sdk` (WindowsForms + WPF)

similarly for other platforms, you can have `Xamarin.Apple.Sdk`, `Xamarin.Android.Sdk`.

Common cross IDE support tools can also be provided via these packages as well in the form of a format `<Vendor>.<Platform/Tech>.Tools.Sdk` or `ProjectSystem.Sdk` or something similar.

You can also generate the `AutoImport.<props/targets>` in `MSBuildProjectExtensionsPath` based on the target in the project file if the user doesn't have the required Sdks/Frameworks installed so that It can be resolved from NuGet and placed under `~\.dotnet\sdks\`, `~\.dotnet\packs` or `~\.dotnet\frameworks`, `~\.dotnet\libraries` and `~\.dotnet\tools`.

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+In .NET 5, we will add support for iOS and Android.  The .NET SDK (formerly known as the .NET Core SDK) will be able to build projects targeting iOS and Android.  However, the .NET SDK will not be a monolithic SDK with support for all possible project types.  Rather, the iOS and Android support (and eventually more pieces) will be delivered as optional SDK workloads which may or may not be installed.  This proposal covers how we will use MSBuild SDK resolvers to hook up the build logic from workloads at build time, and to handle scenarios where a workload required to build a project is not installed.
+
+## Links
+
+- [Target Framework Names in .NET 5](https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md)
+- [NET SDK workloads overview](https://github.com/dotnet/designs/blob/workloads/accepted/2020/workloads/workloads.md) ([PR](https://github.com/dotnet/designs/pull/100))
+
+## Experiences
+
+- Projects using workloads just work if the workloads are installed
```
richlander, line 12:
```suggestion
- Projects dependent on a workload just work if the workload is installed
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+  </PropertyGroup>
+
+</Project>
+```
+
+## Workload guidance
+
+The bulk of functionality from workloads will be contained in workload "packs", which will be the unit of delivery for workloads.  This will include MSBuild targets and tasks.  The assets from the packs will be laid out under the dotnet root folder, probably in `packs/<Pack ID>/<Pack Version>` or a similar folder.
+
+Workloads will also provide a manifest package which describes which packs are necessary for which workloads.  The manifest will specify the name and version for each pack that is available for a given .NET SDK release band, as well as which packs are required for each logical workload.  The manifest for workloads will be present even when the workloads it describes are not installed.
```
richlander, line 42: @KathleenDollard asked us to use "feature band" as the primary term.

```suggestion
Workloads will also provide a manifest package which describes which packs are necessary for which workloads.  The manifest will specify the name and version for each pack that is available for a given .NET SDK feature band, as well as which packs are required for each logical workload.  The manifest for workloads will be present even when the workloads it describes are not installed.
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+### Workload Manifest Targets Files
+
+In addition to the manifest file itself (likely in json format), the workload manifest package will include a `WorkloadManifest.targets` file which should import the necessary .targets files from the workload pack if a project depends on the workload.  The `WorkloadManifest.targets` files from all workload manifests will be imported into all projects that use Microsoft.NET.Sdk.
```
richlander, line 46: Why not call it `Workload.targets`? It seems that the manifest is an implementation detail.

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+The `WorkloadManifest.targets` should use MSBuild Sdk imports to import .targets files from an workload pack.  This should be done conditionally based on properties that determine whether the workload should be used.  For example, the Android workload manifest targets file might look like this:
```
richlander, line 48:
```suggestion
The `WorkloadManifest.targets` should use MSBuild Sdk imports to import .targets files from a workload pack.  This should be done conditionally based on properties that determine whether the workload should be used.  For example, the Android workload manifest targets file might look like this:
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+<Project>
+    <Import Project="Sdk.targets" Sdk="Xamarin.Android.Sdk" />
+    <Import Project="Sdk.targets" Sdk="Xamarin.Android.Aot.Sdk" Condition="'$(UseAot)'=='true'" />
+    <!--
+    Import any other sdks based on project state.
+    -->
+</Project>
+```
+
+In this example, the iOS and Android workloads both share the same workload manifest.  In addition to `WorkloadManifest.targets`, the manifest package includes separate targets files specific to iOS and Android.  The more complex logic for which workload packs to import from is guarded behind a simple `TargetPlatformIdentifier` check in the `WorkloadManifest.targets`.
```
richlander, line 78: > In this example, the iOS and Android workloads both share the same workload manifest.

I assume we would not recommend this. Is it even a good idea to show this as an example?

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+This indicates that the `Sdk.targets` file should be imported from the `Xamarin.Android.Sdk` workload pack if the `TargetPlatformIdentifier` is `android`.  The version of the `Xamarin.Android.Sdk` workload pack to use will be the one defined in the workload manifest json file.
+
+Because the workload manifest targets files will be imported by all projects, it is important to try to guard against bugs in these targets from causing issues with projects that don't intend to use the workload at all.  One way of reducing risk can be to split the logic into multiple targets files, where the targets files with more complex logic are only imported from the entry point workload manifest targets file based on a simple condition.  For example:
+
+```xml
+<!-- WorkloadManifest.targets (entry point): -->
+
+<Project>
+    <Import Project="Xamarin.Manifest.Android.Framework.targets" Condition="'$(TargetPlatformIdentifier)'=='android'" />
```
richlander, line 63: The condition in uniform in the example here and the one above. I'm not an MSBuild person, but it is hard for me to understand the value of this second option if the condition is constant.

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+</Project>
+```
+
+This would import the iOS targets from the Xamarin.iOS.Sdk.15.0 workload pack if the `TargetPlatformVersion` is 15.0 or higher, and from the Xamarin.iOS.Sdk workload pack otherwise.
+
+### Workload props files
+
+Workloads may also provide MSBuild logic in an `AutoImport.props` file in the Sdk folder of the workload pack which will be imported before the body of the project file.  We expect this to be use for default globs, for example to specify that all .xaml files should be included as `Page` items.
+
+However, `AutoImport.props` files from workloads are subject to restrictions and must be carefully authored.  This is because since they need to be imported before the body of the project is evaluated, they can't be included conditionally based on the properties in the project that define whether the workload is in use.  So the `AutoImport.props` files from all workload packs will be imported for all projects using Microsoft.NET.Sdk.
```
richlander, line 99:
```suggestion
However, `AutoImport.props` files from workloads are subject to restrictions and must be carefully authored.  This is because since they need to be imported before the body of the project is evaluated, they can't be included conditionally based on the properties in the project that define whether the workload is in use.  So the `AutoImport.props` files from all workload packs will be imported for all projects using `Microsoft.NET.Sdk`.
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+MSBuild uses multiple phases for evaluation, and properties and imports are evaluated in a phase before items.  This means it is possible to have conditions on item or `ItemGroup` elements in a .props file that depend on properties that are defined in the body of the project or in .targets files.
+
+So workloads may include MSBuild items in the `AutoImport.props` file if they are appropriately conditioned to only activate when the workload is in use.  They should not set MSBuild properties, as there is no way to set those properties only if the workload is in use.  If workload wants to set a default value for a property, then it should do so in its .targets file with a condition to set the property if it is not already set.
```
richlander, line 103:
```suggestion
Workloads may include MSBuild items in the `AutoImport.props` file if they are appropriately conditioned to only activate when the workload is in use.  They should not set MSBuild properties, as there is no way to set those properties only if the workload is in use.  If workload wants to set a default value for a property, then it should do so in its .targets file with a condition to set the property if it is not already set.
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+Note that in contrast to the `WorkloadManifest.targets` file, which is part of the workload manifest package, the `AutoImport.props` files are part of the workload packs (under the Sdk folder).  This is so that they can be imported only if the workload is installed, rather than imported for all workloads whether they are installed or not.  This reduces the impact of a bug in an `AutoImport.props` file, and means that if there is an issue in the .props file, it can be worked around by uninstalling the workload.
```
richlander, line 105: This paragraph and the one before it seems like they duplicate some of the same content and could be streamlined a bit.

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+### New resolver behavior
+
+We will create an MSBuild SDK resolver to handle workloads.  Its behavior should be the following:
+
+- Look up the Name of the requested SDK in the workload manifests, matching to the workload pack IDs
+- If a matching workload pack is found in a workload manifest, then get its version number, and use that version number to generate the path where that workload pack would be installed.  This should be the following, or similar: `<DOTNET ROOT>/packs/<PACK ID>/<PACK VERSION>/Sdk`
+- If the workload pack is installed (ie the path (except the final `Sdk`) exists), then return the generated path
+- If the workload pack is not installed, then the returned `SdkResult` should:
+  - Have `Success` set to true
```
richlander, line 164: When would `Success` be false?

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+- Find all installed workload packs from those manifests
+- From the installed packs, find all of the packs that have an `Sdk/AutoImport.props` file in them
+- Return the Sdk path for all of the packs that had the `Sdk/AutoImport.props` file in them
+
+### SDK behavior
+
+The .NET SDK (Microsoft.NET.Sdk) should:
+
+- Set the `WorkloadManifestRoot` property to a path where workload manifest packages for the current version band will be laid out.
+  - This should be `<DOTNET ROOT>/workloadmanifests/<VERSION BAND>/`
```
richlander, line 180:
```suggestion
  - This should be `<DOTNET ROOT>/workloadmanifests/<FEATURE VERSION BAND>/`
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+```xml
+<Import Project="AutoImport.props" Sdk="Microsoft.NET.SDK.WorkloadAutoImportPropsLocator">
+```
+
+### VS behavior
+
+When the VS project system loads a project, it should check the evaluation result for `MissingWorkloadPack` items.  If there are any, then it should raise the VS event to veto the project load and trigger the in-product acquisition experience for those workloads.
+
+In the same way that the SDK error generation does, the project system should map from the missing workload packs to the missing workloads via the workload manifests.  Additionally, it will need to map from the workload IDs that the .NET SDK uses to the right IDs that VS will use to install the workloads.  That mapping should either be defined in the workload manifest or in another file that is part of the same packgae.
```
richlander, line 202: > Additionally, it will need to map from the workload IDs that the .NET SDK uses to the right IDs that VS will use to install the workloads.  That mapping should either be defined in the workload manifest or in another file that is part of the same packgae.

We should make VS Installer just another installer, and avoid making the .NET SDK and the VS Installer coupled. I think we should start with the mapping file idea from the start. If we cannot afford that in .NET 5.0, we should intend to fix that for 6.0.

```suggestion
In the same way that the SDK error generation does, the project system should map from the missing workload packs to the missing workloads via the workload manifests.  Additionally, it will need to map from the workload IDs that the .NET SDK uses to the right IDs that VS will use to install the workloads.  That mapping should either be defined in the workload manifest or in another file that is part of the same package.
```

accepted/2020/workloads/workload-resolvers.md
```diff
@@ -0,0 +1,237 @@
...
+
+When we add support for `dotnet bundle restore`, we will need to be able to run a target that will gather all of the workloads required by a project and its project references.  We will implement this via a new entry point target that walks the project graph and collects the `MissingWorkloadPack` items from each one.
+
+## Other experience improvements
+
+The additional features we plan to add to MSBuild SDK resolvers will also enable us to make the following improvements in experiences that are not directly related to workloads.
+
+## NuGet SDK Resolver
+
+MSBuild includes a [NuGet-based MSBuild SDK resolver](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2019#how-project-sdks-are-resolved).  It will be used if a version number is specified for the SDK, either in the SDK attribute or in a global.json file.
```
richlander, line 216: > It will be used if a version number is specified for the SDK, either in the SDK attribute or in a global.json file.

I don't quite get this (I'm probably just missing context).  I was expecting this to say "If you had an sdk reference other than MS.Net.Sdk".

accepted/2020/workloads/workload-resolvers.md
dsplaisted: It's part of the workload manifest and is imported by all projects.  I think with the name `Workload.targets`, it would be easy to get it confused with the targets which are in the workload and only imported by the workload manifest targets for projects that use the workload.

@mhutch @wli3 @rainersigwald What do you think?  I don't have a strong opinion on the naming here.

accepted/2020/workloads/workload-resolvers.md
dsplaisted: If iOS and Android share components, then they may have to share the same workload manifest (because you can't have one workload depend on a different version of a workload pack than another workload for the same band).

@mhutch, what are your thoughts?

accepted/2020/workloads/workload-resolvers.md
dsplaisted: If you start to have complex logic to determine which workload packs are imported, then this pattern helps reduce risk that bugs in that logic will impact projects which don't use your workload at all.

accepted/2020/workloads/workload-resolvers.md
wli3: I prefer longer and descriptive name

WorkloadManifest.targets and Workload.targets both don't really mean too much other than it is related to workload or manifest while everything in the manifest folder is

 We should start from  "a file that should import the necessary .targets files from the workload pack if a project depends on the workload". target  and summarize this sentence to a few words.

WorkloadPackTargetEntry.target?
WorkloadPackTarget.target?
ImportWorkloadPackTaget.target?

accepted/2020/workloads/workload-resolvers.md
dsplaisted: It would be set to false if there was some sort of error (for example a malformed manifest or something).

accepted/2020/workloads/workload-resolvers.md
dsplaisted: You have to specify a version number in order for the NuGet SDK resolver to kick in.  You can specify that either directly in the Sdk attribute: `<Project Sdk="MSBuild.Sdk.Extras/2.0.54">`, or in global.json:

```json
{
  "msbuild-sdks": {
    "MSBuild.Sdk.Extras": "2.0.54"
  }
}
```

accepted/2020/workloads/workload-resolvers.md
richlander: K. Might be good to call that out somewhere. I read "false" is for exceptional cases, while the case you call out is an expected case.

accepted/2020/workloads/workload-resolvers.md
richlander: I get it. The proposal as-is makes sense to me, and I appreciate the explanation.

accepted/2020/workloads/workload-resolvers.md
richlander: Good point. My concern is that Android and iOS are expected to have updates on different schedules. How do we achieve the following goals:

* lower costs for us -- avoid re-testing iOS for Android updates, and vice versa
* avoid UX challenges for customers -- iOS customers do not get update notifications for Android updates, and vice versa

These might both be .NET 6.0 challenges given our scoping for .NET 6.0, but they are descriptive of my primary concerns.

accepted/2020/workloads/workload-resolvers.md
richlander: It sounds like the conditions might become more complex over time (even within this release).

I don't have anything to offer here. Just a question.

accepted/2020/workloads/workload-resolvers.md
richlander: I see. That seems like overly precise wording.

```suggestion
MSBuild includes a [NuGet-based MSBuild SDK resolver](https://docs.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2019#how-project-sdks-are-resolved).  It will be used if a custom SDK is specified, as an SDK name and version number, either in the SDK attribute in the project file or in a global.json file.
```

accepted/2020/workloads/workload-resolvers.md
mhutch: I would not expect iOS and Android to have the same manifest. But iOS, macOS, tvOS, and watchOS might as they are built out of the same codebase.

PR #109, opened 2020-04-09
Clarify precedences of TFMs

terrajobst:
We need to decide what precedences we want the new TFMs to have in relation to the old ones. For example, consider this scenario:

* Project: `net5.0-ios`
* Package:
  * `net5.0`
  * `xamarin.ios`

Both `net5.0` and `xamarin.ios` are applicable for the project. Which assets should NuGet prefer?

### DIFF
--- a/accepted/2020/net5/net5.md
+++ b/accepted/2020/net5/net5.md
@@ -280,6 +280,86 @@ We'll have the following TFMs:
 |                 | (+everything else inherited from net5.0)                   |                                   |
 | Tizen, Unity... | Will follow the Xamarin model                              |                                   |
 
+### Precedences
+
+We need to decide what precedences we want the new TFMs to have in relation to
+the old ones as well as in relation with the new universally portable TFM
+`net5.0` (and higher).
+
+Let's look at two examples:
+
+* **Example A**
+  * Project targets `net6.0-ios`
+  * Package offers
+    * `net6.0`
+    * `net5.0-ios`
+* **Example B**
+  * Project targets `net6.0-ios`
+  * Package offers
+    * `net6.0`
+    * `xamarin.ios`
+
+There are effectively two options:
+
+1. Prefer platform-specific assets, regardless how old their version is.
+2. Prefer the assets that are closer in version number.
+
+Example | Option (1)    | Option (2)
+--------|---------------|--------------
+A       | `net5.0-ios`  | `net6.0`
+B       | `xamarin.ios` | `net6.0`
+
+We prefer Option (2). @mhutch described it best:
+
+> I prefer the Option (2) because it allows re-converging to platform-neutral:
+> in both A and B, if `net6.0` added functionality that meant that the iOS
+> specific implementation was no longer necessary, then the package would not
+> need to have `net6.0-ios` assets.
+>
+> If the `net5.0-ios` assembly has iOS-specific _functionality_, then when the
+> package adds `net6.0` it should also add `net6.0-ios`, else it feels to me
+> like a broken package.
+
+@dsplaisted also added this:
+
+> One of the main benefits of the target framework changes in .NET 5 and on is
+> that it aligns the version numbers of the platform specific targets
+> (`net5.0-ios`) with the portable target (`net5.0`). This makes it easy to
+> understand the version compatibility rules. It also means that we can have a
+> simple and easy understand rule for choosing the best asset that is more
+> likely to produce a result that matches the intent of a package author.
+
+And finally @ericstj added that the current model (where we prefer
+platform-specific assets, regardless of version number) isn't helpful:
+
+> Option (2) makes sense to me. I don't recall a place where we've actually
+> preferred that RIDs had precedence over TFM. Usually it's a problem and
+> results in bloated packages. I like reversing the precedence here.
+
+This leads to these rules:
+
+1. Select list of applicable assets
+   * This will include `netX.Y`, `netX.Y-*`, and `xamarin.*` if they are
+     considered compatible
+   * This will not include assets that use an OS/OS version that is incompatible
+     with the project.
+2. Group `netX.Y*` by X.Y and sort groups from highest to lowest
+3. If the first group contains both a `netX.Y` and a `netX.Y-*` or a
+   `xamarin.*`, pick `netX.Y-*`. If multiple entries for `netX.Y-*` exist, pick
+   the highest version.
+4. If no groups were produced, pick the appropriate TFM based on existing rules
+   (this covers .NET Standard, PCLs, and general asset target fallback).
+
+Then intention is:
+
+1. We slice the package assets to find the latest .NET version that is
+   compatible with the project.
+2. If the latest .NET version provides both a platform-specific- and a portable
+   asset, we'll prefer the platform-specific asset.
+3. If multiple platform-specific assets exist, we'll use the one with the
+   highest compatible OS version.
+4. Stay backwards compatible with existing .NET Standard rules.
+
 ### OS versions
 
 The TFM will support an optional OS version, such as:


clairernovotny:
Based on the existing guide of platform being more specific and thus winning, I’d expect the xamarin.ios one to be picked.

mhutch:
I'm not sure the existing behaviour really translates directly to .NET 5, given that historically xamarin TFMs were always "newest".

If I had a NuGet containing `net6.0`, `net5.0-ios`, and `net5.0` assets, and I referenced it from a `net6.0-ios` project, I think it would make the most sense to get `net6.0` assets. This would be useful if `net6.0` added portable functionality that meant the library author was able to remove their platform-specific code. If the library author still needs iOS specific code with `.net6.0`, they should add `net6.0-ios` when they add `net6.0`.

clairernovotny:
@mhutch I think we may need a deeper discussion around this as I'm not sure I understand. 

Based on the way NuGet has always worked to date, and it's certainly possible to change it, I'd expect platform specific to "always win" over a newer version. Thus, if a package contained `net6.0`, `net5.0`, `net5.0-ios` assets, and I have a `net6.0-ios` project, I'd expect it to get `net5.0-ios` as that's the "better match"/platform specific one.

That's how it works today as well, where a `net48` project will resolve to a `net1.0` asset even if the package has a `netstandard2.0` target as well.

dsplaisted:
I think we should go with what @mhutch is suggesting, which is to some degree a break from the past as @clairernovotny points out.

The reason to do so is because in the past there was no relationship between the version numbers of the more platform specific target and the less platform specific target.  IE, there's nothing in the system that tells you that a `netstandard2.0` asset is "newer" than a `net45` asset.  So we go with the more platform specific asset, which is `net45`.  On top of that, as @mhutch points out, the Xamarin target frameworks didn't even version.

The big change in the new TargetFramework model is that the version numbers all align.  So we do know that `net6.0` is "newer" than `net5.0-ios`, and I think we should prefer `net6.0` in this case.

terrajobst:
OK, I think we should separate the discussion into two aspects:

1. Precedence between `netA.B` and `xamarin.xxx` when referenced from `netC.D-xxx` and C.D >= A.B 
2. Precedence between `netA.B-xxx` and `netC.D` and when referenced from `netE.F-xxx` and A.B < C.D and C.D <= E.F

I assumed for (2) we follow the current model, which is preferring `netA.B-xxx`. If we want to prefer the close version no matter the OS suffix, then I think we can apply this for (1) as well. But I think it would be counter intuitive to treat them differently.

@ericstj @rrelyea what are your thoughts on this?

mhutch:
I think that `net5.0-ios` should prefer `net5.0` over `xamarin.ios`. I just used `net6.0` to illustrate how this decision would impact future versions.

I would not suggest we change any of the precedence between existing frameworks, obviously.

clairernovotny:
> I think that net5.0-ios should prefer net5.0 over xamarin.ios. I just used net6.0 to illustrate how this decision would impact future versions.

I would agree with that. I perhaps misunderstood the earlier comment as suggesting that a `net6.0-ios` project would choose a `net6.0` asset that's in a package over a `net5.0-ios` one if both are present. Are we on the same page here that in this case, the `net5.0-ios` asset would be used?

dsplaisted:
> > I think that net5.0-ios should prefer net5.0 over xamarin.ios. I just used net6.0 to illustrate how this decision would impact future versions.
> 
> I would agree with that. I perhaps misunderstood the earlier comment as suggesting that a `net6.0-ios` project would choose a `net6.0` asset that's in a package over a `net5.0-ios` one if both are present. Are we on the same page here that in this case, the `net5.0-ios` asset would be used?

It sounds like @terrajobst may be on the same page, but I'm not and it doesn't sound like @mhutch  is either.

I do think we should be consistent.  To me, Xamarin.iOS should be treated roughly the same as a pre `net5.0` TargetFramework that had an `-ios` suffix, ie something like `netcoreapp3.1-ios`.  And my opinion is that we should choose the later .NET version over a platform-specific target framework for an earlier .NET version, because as @mhutch points out, a later version of .NET may add portable functionality that make platform-specific code no longer necessary.

richlander:
This is a good conversation.

I think it depends a lot on what we're planning to do with `xamarin-ios` next. Will be add 5.0 APIs or leave it at the 3.1 era? Same question about C#8 vs C# 9. I guess now that I think about it, we should keep it at the 3.1 era. That starts to paint the picture like `netcoreapp3.1-ios`, as has been suggested.

accepted/2020/net5/net5.md
```diff
@@ -279,6 +279,60 @@ We'll have the following TFMs:
...
+1. Treat existing Xamarin TFMs as more specific
+2. Treat `net5.0` as more specific
+
+It seems easier to explain that `net5.0-android` and `net5.0-ios`  are simply
+renamed versions of `xamarin.android` and `xamarin.ios` respectively.  Both are
+platform specific. `net5.0` is a portable TFM, similar to `netcoreapp` and
+`netstandard`. It seems generally more logical to first look for platform
+specific TFMs and then fall back to portable TFMs:
+
+1. First look for the same TFM
```
nkolev92, line 306: fwiw, I don't think there's any mapping that does otherwise :)

terrajobst: Yep, just listed for completeness :-)

terrajobst:
@mhutch

> I think that `net5.0-ios` should prefer `net5.0` over `xamarin.ios`. I just used `net6.0` to illustrate how this decision would impact future versions.

Let me repeat what I heard to make sure we're understanding each other.

Example A:

* Project targets `net6.0-ios`
* Package offers
  * `net6.0`
  * `net5.0-ios`

Example B:

* Project targets `net6.0-ios`
* Package offers
  * `net6.0`
  * `xamarin.ios`

* It seems both @mhutch and @dsplaisted are suggesting that in Example A and B NuGet should select `net6.0`.
* It seems @clairernovotny would prefer NuGet to pick `net5.0-ios` in Example A and `net6.0` in Example B.
* My proposal in this PR suggests NuGet would pick `net5.0-ios` and `xamarin.ios`.

Is that a fair summary?

If so, I can live with the proposal of @mhutch and @dsplaisted because it's self consistent: we pick the closer version, no matter the OS suffix. I don't like @clairernovotny's proposal because we have a mixture of rules on .NET 5 based platforms, and sometimes we pick the closer one and sometimes we pick the OS specific one.

dsplaisted:
@terrajobst Yes, I you've accurately captured my opinion (and I think the two examples are quite helpful to clarify the options).

clairernovotny:
@terrajobst I'm not suggesting that in example B that I'd want to pick `net6.0` over `xamarin.ios`. I thought I was saying the opposite.

> My proposal in this PR suggests NuGet would pick net5.0-ios and xamarin.ios.

This is my understanding and expectation based on how NuGet has always worked to-date, and behavior that makes sense to me.

terrajobst:
@clairernovotny 

> @terrajobst I'm not suggesting that in example B that I'd want to pick `net6.0` over `xamarin.ios`. I thought I was saying the opposite.

Cool, then you're in agreement with what the proposal currently says, which means there are only two options being discussed here.

@mhutch @dsplaisted could you describe in a bit more detail why you want the new model? @clairernovotny is right; it's a departure from how NuGet works today (how NuGet picks between .NET Standard and and platform specific TFMs). We should motivate this properly.

dsplaisted:
I think we have the model today not because it was the only real option we had.  Consider this example:

* Project targets `net48`
* Package offers
  * `net35`
  * `netstandard2.0`

A package author might expect the .NET Standard asset to be selected in this case, as it can use much more modern APIs.  However, there was no sane rule we could write that would do this, as we would have had to keep a mapping of what version of each platform corresponded to what version of .NET Standard.  And it probably wouldn't be clear what we should do for versions that released around the same time.  So we had to go with a rule that was simple and easy to understand.

One of the main benefits of the target framework changes in .NET 5 and on is that it aligns the version numbers of the platform specific targets (`net5.0-ios`) with the portable target (`net5.0`).  This makes it easy to understand the version compatibility rules.  It also means that we can have a simple and easy understand rule for choosing the best asset that is more likely to produce a result that matches the intent of a package author.

Another reason to change the model is because I think it aligns more with how we are handling the OS version selection.  Consider this example:

* Project targets `net6.0-ios15`
* Package offers
  * `net6.0-ios14`
  * `net5.0-ios15`

Now this seems like it may be a misauthored package.  Nevertheless, we need to decide which assets to select, and I believe what we decided is that it would choose `net6.0-ios14`.  IE the version of .NET would "win" over a better OS version when selecting the best asset.  I think overall the rule is simpler to understand (and maybe implement) if in both cases we first select the best `net` version and then look at the target platform and version.

mhutch:
I prefer the model I suggested because it allows re-converging to platform-neutral: in both A and B, if `net6.0` added functionality that meant that the iOS specific implementation was no longer necessary, then the package would not need to have `net6.0-ios` assets.

If the `net5.0-ios` assembly has iOS-specific *functionality*, then when the package adds `net6.0` it should also add `net6.0-ios`, else it feels to me like a broken package.

ericstj:
Makes sense to me.  I don't recall a place where we've actually preferred that RIDs had precedence over TFM.  Usually it's a problem and results in bloated packages.  I like reversing the precedence here.

terrajobst:
Cool, this makes sense to me. I'll update the proposal to reflect this.

terrajobst:
@mhutch @dsplaisted @clairernovotny @ericstj please re-review

accepted/2020/net5/net5.md
```diff
@@ -279,6 +279,73 @@ We'll have the following TFMs:
...
+
+This leads to these rules:
+
+* Select list of applicable assets
+  * This will include `netX.Y`, `netX.Y-*`, and `xamarin.*` if they are
+    considered compatible
+* If there is an exact match for the project's TFM, pick that
+  * Exact match includes TFI, TFV, TPI and TPV
+* Group `netX.Y*` by X.Y and sort groups from highest to lowest
+* If the first group contains both a `netX.Y` and a `netX.Y-*` or a `xamarin.*`,
```
dsplaisted, line 346: I think this is backwards.  As I read this currently, it would prefer `net5.0` over `net5.0-ios`.

terrajobst: > I think this is backwards. As I read this currently, it would prefer `net5.0` over `net5.0-ios`.

That's intended. Rule 2 makes sure that if you're in `net6.0-ios`, it will use `net6.0-ios` if the package offers it. But if the package offers a `net5.0-ios` and a `net6.0`, it will pick `net6.0` based on rule 4.

> 1. Select list of applicable assets
>    * This will include `netX.Y`, `netX.Y-*`, and `xamarin.*` if they are considered compatible
> 2. If there is an exact match for the project's TFM, pick that
>    * Exact match includes TFI, TFV, TPI and TPV
> 3. Group `netX.Y*` by X.Y and sort groups from highest to lowest
> 4. If the first group contains both a `netX.Y` and a `netX.Y-*` or a `xamarin.*`, pick `netX.Y`

Does that make sense?

bartonjs:
* **Example C**
  * Project targets `net6.0-ios`
  * Package offers  
    * `netstandard2.0`
    * `net5.0-ios`
    * `xamarin.ios`

net5.0-ios?

* **Example D**
  * Project targets `net6.0-ios`
  * Package offers  
    * `netstandard2.0`
    * `xamarin.ios`

xamarin.ios?

accepted/2020/net5/net5.md
dsplaisted: No, I'm not getting it.  You can't rely on just exact matches.  What about this case, for example?

  * Project targets `net6.0-ios`
  * Package offers  
    * `net5.0-ios`
    * `net5.0`

Rule 2 wouldn't apply as it's not an exact match, and rule 4 would mean it would prefer `net5.0`, as far as I can tell.

I think in needs to pick `netX.Y-*` over `netX.Y` as long as they are already compatible.

terrajobst: I confused myself. Rule (1) already filters out everything that isn't compatible and rule (2) will group by .NET version. So I think you're correct with your initial suggestion. Fixed.

terrajobst:
> * **Example C**
>   
>   * Project targets `net6.0-ios`
>   * Package offers
>     * `netstandard2.0`
>     * `net5.0-ios`
>     * `xamarin.ios`
> 
> net5.0-ios?

I believe that's the answer we want (and the current rules will produce that).

> * **Example D**
>   
>   * Project targets `net6.0-ios`
>   * Package offers
>     * `netstandard2.0`
>     * `xamarin.ios`
> 
> xamarin.ios?

I believe that's the answer we want (and the current rules will produce that).

> * **Example E**
>   
>   * Project targets `net6.0-ios`
>   * Package offers
>     * `netstandard2.1`
>     * `xamarin.ios`
> 
> netstandard2.1? (It's "newer"?)

I don't think netstandard2.1 is newer as the latest Mono-based Xamarin has support for .NET Standard 2.1 as well. So I think they are in the same bad, which is why I believe we want to pick `xamarin.ios` here, which the current rules would produce too.

terrajobst:
Reviewed in today's MSBuild/SDK/.NET sync. We concluded this is what we want.

PR #111, opened 2020-04-14
First draft of an analyzer for capability APIs

terrajobst:
This design isn't ready yet. I only created the draft PR so that I can point folks to it.

### DIFF
--- a/accepted/capability-checks/capability-checks.md
+++ b/accepted/capability-checks/capability-checks.md
@@ -1,0 +1,234 @@
+# Capability APIs
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+Thoughts I shared on Twitter:
+
+* This morning it happened. I had an epiphany under the shower to make
+  platform-checks much better via a generalized approach for capability APIs.
+  Basically, instead of doing actual platform checks, the analyzer would push you
+  call capability APIs.
+* The nice thing is that this also works for features that aren't OS specific,
+  but more of a combination of runtime and environment, such as whether you can
+  JIT.
+* Imagine in web assembly where we discover certain features wouldn't work. It's
+  hard to annotate an API with "works everywhere but here" in a fashion that
+  doesn't break consumers when the environment changes.
+
+## Scenarios and User Experience
+
+## Requirements
+
+### Goals
+
+### Non-Goals
+
+## Design
+
+### Open Questions
+
+#### How can we reduce the noise?
+
+Basically, we should make sure that people building applications don't have to
+perform capability checks if the answer is statically known. For example,
+registry will work on Windows, so folks building a WinForms/WPF app shouldn't
+have to suppress them. One option is add typical capabilities to the assembly,
+but that's messy. Another option is tag capabilities with platform annotations
+that basically say "on this platform, you don't have to ask for capability
+checks". Then it's only a matter of flowing the platform information from the
+build into the analyzer.
+
+#### Should this analyzer be-opt in?
+
+This might be useful for .NET Standard library developers and much less useful
+for people building .NET Core apps. We might also want different defaults.
+
+#### Should we provide an assert-style attribute?
+
+This would behave the same was tagging the assemblies, modules, types, or
+members with the `[RequiredXxxCapability]` but without requiring callers to
+check it. It's basically a suppression for a specific capability.
+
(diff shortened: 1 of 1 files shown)


accepted/capability-checks/capability-checks.md
```diff
@@ -0,0 +1,234 @@
...
+have to suppress them. One option is add typical capabilities to the assembly,
+but that's messy. Another option is tag capabilities with platform annotations
+that basically say "on this platform, you don't have to ask for capability
+checks". Then it's only a matter of flowing the platform information from the
+build into the analyzer.
+
+#### Should this analyzer be-opt in?
+
+This might be useful for .NET Standard library developers and much less useful
+for people building .NET Core apps. We might also want different defaults.
```
vitek-karas, line 44: Some capabilities will be less interesting for .NET Core apps, but some are as applicable as for .NET Standard libraries (I can still try to use Registry APIs when building app for Linux, it should warn). For that reason I think this should be on-by-default if we can make it performant enough.

accepted/capability-checks/capability-checks.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### Should we provide an assert-style attribute?
+
+This would behave the same was tagging the assemblies, modules, types, or
+members with the `[RequiredXxxCapability]` but without requiring callers to
+check it. It's basically a suppression for a specific capability.
```
vitek-karas, line 50: If it looks like suppression and behaves like suppression -> it might as well be a suppression 😄  - Why would not `SuppressMessage` (or the new `UnconditionalSuppressMessage`) work here? Maybe I don't understand the "assertion" pattern.

accepted/capability-checks/capability-checks.md
```diff
@@ -0,0 +1,234 @@
...
+
+#### Should different capabilities have different diagnostic ids?
+
+This probably doesn't work in Roslyn as the diagnostic set needs to be known
+statically.
+
```
vitek-karas, line 56: In general I agree, but it would be nice if we could make at least some with distinct IDs. If nothing else it would behave nicer in combination with suppression and warning settings in IDEs.
So for example those we define in the framework could come with distinct IDs.

accepted/capability-checks/capability-checks.md
```diff
@@ -0,0 +1,234 @@
...
+        var value = key.GetValue("foo");
+
+        /*
+        using (var key = Registry.CurrentUser.OpenSubKey(@"SoftwareFabrikamAssetManagement"))
+        {
+            if (key?.GetValue("LoggingDirectoryPath") is string configuredPath)
+                return configuredPath;
+        }
+        */s
+    }
```
vitek-karas, line 100: What if the code looks like:
```C#
if (Registry.IsSupported)
{ // read from registry }
else
{ throw new Exception("This feature requires registry"); }
```

The analyzer would probably mark this as "perfectly fine", but in reality it's as broken as before. Should we consider some kind of detection of this "Anti-pattern"?

accepted/capability-checks/capability-checks.md
```diff
@@ -0,0 +1,234 @@
...
+
+    // Would also mark
+    // - Other *Builder types
+    // - AppDomain.DefineDynamicAssembly
+    // - ILGenerator
+}
+```
+
+## Q & A
```
vitek-karas, line 234: Should we possibly fold the proposed "linker unfriendly" attribute into this as well?: https://github.com/dotnet/runtime/issues/33862
It looks like a good fit.

vitek-karas, line 234: What is the handling of issues inside dependencies (NuGets mostly)? For example with the registry APIs above:
I write an app which itself doesn't call any of the registry APIs, but it uses a NuGet which does call them. Running the analyzer on the app itself will not reveal any problems, but the app is still broken on Linux.

vitek-karas, line 234: We should also consider how is this related to the feature switch functionality: https://github.com/dotnet/designs/blob/master/accepted/2020/feature-switch.md

In short - linker could remove the "false" branches in lot of cases (reducing code size) - assuming this is integrated correctly.

terrajobst:
It doesn't seem we're pursuing this for now. If scenarios requires this, we might come back to it.

PR #113, opened 2020-04-20
JSON Serializer recommendations for 5.0

steveharter:
Based on a prior internal review and feedback, putting various information together to discuss.

### DIFF
--- a/proposed/SerializerGoals5.0.md
+++ b/proposed/SerializerGoals5.0.md
@@ -21,7 +21,7 @@ The options to consider, in priority order include:
 
 This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
 
-The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, JIL and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
+The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, Jil and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost (but in many cases has the fastest steady-state performance).
 
 There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general sceanrios, although we do not want to noticably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
 
@@ -50,7 +50,7 @@ Primarily this is for faster downloads of the runtime and applications.
 
 The runtime and BCL can be made smaller by removing Reflection.Emit support.
 
-Although outside the scope of this document, an application can be made smaller by running ILLinker (e.g. "tree-shaking"). Serializer scenarios are generally unsuited for linking but can be somewhat mitigated by using metadata to declare what members to preserve. The code-gen approaches in this document, depending on implementation, may help with this since they generate direct calls to members that are detectable by the linker.
+Although outside the scope of this document, an application can be made smaller by running ILLinker (e.g. "tree-shaking"). Serializers generated at runtime are generally problematic for linking but can be somewhat mitigated by using metadata to declare what members to preserve. The code-gen approaches in this document, depending on implementation, may help with this since they generate direct calls to members that are detectable by the linker.
 
 # Serializer design background
 ## Runtime code dependencies
@@ -131,7 +131,7 @@ A (prototype)[https://github.com/layomia/jsonconvertergenerator] was created for
 It used a simple (non-Roslyn) code generator using the existing custom Value converter model. Various serializer features (null handling, property lookup on deserialization, reference handling, async support, use of custom converters, etc) would need to be baked into the generated code. Thus the resuling code generation will be large, complex and not servicable. At the extreme (with no new helper methods exposed the serializer), each converter would need to implement all of the various features of the serializer.
 
 ## Metadata provider code-gen
-A run-time only prototype (not the actual generator) was created for this by @steveharter (see sample generated code below). It makes public the existing internal metadata classes including `JsonClassInfo` and `JsonPropertyInfo`.
+A run-time only [prototype](https://github.com/steveharter/runtime/tree/ExtConverters) which does not include the actual code generator was created by @steveharter (see sample generated code below). It makes public the existing internal metadata classes including `JsonClassInfo` and `JsonPropertyInfo`.
 
 It allows for minimal code gen and overlaps with other requested features including property ordering, before and after callbacks, and programmatic reading and writing of metadata.
 
@@ -230,7 +230,7 @@ public class WeatherForecast
 </details>
 
 ## Object converter + code-gen
-This is similar to "Metadata provider code-gen" but includes exposing a new type of "object converter". This type of converter, however, exposes the raw reader and writer and thus has performance issues with Streams due to requiring "read-ahead" which means the JSON is ensured to be complete to the end of the current JSON level and not encounter end-of-buffer scenarios within a converter.
+This is similar to "Metadata provider code-gen" but includes exposing a [new type of "object converter"](https://github.com/dotnet/runtime/issues/1562) for POCOs. This type of converter, however, exposes the raw reader and writer and thus has a slight performance degradation with Streams due to requiring "read-ahead". Read-ahead means the serializer ensures the JSON that will be obtained by the converter through the reader will not encounter end-of-buffer.
 
 # Code-Gen
 For the options that use code-gen there are several issues to discuss; this is covered in [other documentation](https://gist.github.com/layomia/5221e73374fd74dc360e74731dacfdb5) but one open issue is to determine what Types to add code-gen for. The two basic approaches are automatic and explicit:
@@ -297,5 +297,5 @@ The "Metadata code-gen" option does not require "Reflection features" although i
 
 **The recommended plan for 5.0 (in sequence):**
 1. *Required*: implement the required features for the "Minimal" option.
-2. *Optional*: Implement "Reflection features". If the result is noticably slower in the steady-state scenarios then we may need to consider contingency plans such as options to control whether to use IL emit or not -- i.e. startup vs. steady-state tradeoffs.
+2. *Optional*: Implement "Reflection features". If the result is noticeably slower in the steady-state scenarios then we may need to consider contingency plans such as options to control whether to use IL emit or not -- i.e. startup vs. steady-state tradeoffs.
 3. *Optional*: pending performance or overlap with other scenarios, implement "Metadata code-gen". This avoids all usages of System.Reflection including looking up properties and attributes. There may or may not be generated code for getters, setters or constructors depending on the performance of "Reflection features" and linker "tree shaking" requirements.
--- a/proposed/SerializerGoals5.0.md
+++ b/proposed/SerializerGoals5.0.md
@@ -1,6 +1,6 @@
 # Overview and goals
 
-April 21th, 2020.
+April 28th, 2020.
 
 This document provides options and recommendations for 5.0 to handle overlapping serializer goals including reach, faster cold startup performance, reducing memory usage, and smaller size-on-disk.
 
@@ -117,7 +117,7 @@ Some early issues created to start tracking the work:
 - [Add deferred loading of System.Private.Uri assembly by the serializer](https://github.com/dotnet/runtime/issues/35029). Minor startup gains; low-hanging fruit.
 
 ## Reflection features
-As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct useage of runtime code generation, but not the indirect usage.
+As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct usage of runtime code generation, but not the indirect usage.
 
 Note that AOT code-gen options likely eliminate the need for new reflection features for those POCOs that are code-gen'd. However, we can't assume or force all POCOs to be code-gen'd.
 
@@ -126,7 +126,7 @@ Misc other thoughts:
     - Another fallback would apply to a code-gen option where the generated POCO code closes the generic collections at build time. i.e. instantiate (or reference) a `JsonConverter<List<int>>` directly in the generated code instead of relying on the serializer to close `JsonConverter<T>` to `JsonConverter<List<int>>`.
 
 ## Converter code-gen
-A (prototype)[https://github.com/layomia/jsonconvertergenerator] was created for this by @layomia. It used a simple (non-Roslyn) code generator using the existing custom Value converter model.
+A [prototype](https://github.com/layomia/jsonconvertergenerator) was created for this by @layomia. It used a simple (non-Roslyn) code generator using the existing custom Value converter model.
 
 Various serializer features would need to be baked into the generated code including:
 - Setting and getting properties.
--- a/proposed/SerializerGoals5.0.md
+++ b/proposed/SerializerGoals5.0.md
@@ -19,23 +19,23 @@ The options to consider, in priority order include:
 
 ## Faster startup performance
 
-This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
+This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently.
 
 The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, Jil and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost (but in many cases has the fastest steady-state performance).
 
-There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general sceanrios, although we do not want to noticably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
+There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general scenarios, although we do not want to noticably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
 
 <details>
   <summary>5.0 master startup benchmarks (click to expand)</summary>
 
-A small POCO was used (`LoginViewModel` benchmark test class).
+A small POCO was used (`LoginViewModel` benchmark test class). Measurements were taken "cold" from deserializing hard-coded JSON directly and from serializing a populated object directly meaning there was no previous loading of the serialization assembly or inspecting POCO types for custom attributes beforehand.
 
-| Serializer |  Serialize (us)  | Serialize ratio | Deserialize (us) | Deserialize ratio |
+| Serializer |  Serialize Mean (us)  | Serialize ratio | Deserialize Mean (us) | Deserialize ratio |
 | :-- | :-- | :-- | :-- | :--
-| **System.Text.Json** | 4,720 | 1.00 | 19,379 | 1.00 
-| **Json.NET** | 24,916 | 5.28 | 102,226 | 5.28 
-| **Utf8Json** | 7,290 | 1.54 | 106,817 | 5.51 
-| **Jil** | 66,456 | 14.08 | 170,548 | 8.80 
+| **System.Text.Json** | 27,887 | 1.00 | 26,864 | 1.00 
+| **Json.NET** | 81,872 | 2.94 | 80,307 | 2.99
+| **Utf8Json** | 69,681 | 2.50 | 69,383 | 2.58
+| **Jil** | 103,760 | 3.72 | 97,679 | 3.64
 </details>
 
 ## Reduced memory usage
@@ -43,7 +43,7 @@ Reducing private bytes (non-shared per-process) memory is important for Bing sce
 
 This can be done by:
 - Avoiding runtime code generation including Reflection.Emit or JITting.
-- Capturing POCO metadata in generated code instead of RAM.
+- Minimize cached state held by the serializer if it can call into generated code to obtain or process the values. For example, it is possible to write a JSON property name with hard-coded escaped values instead of allocating a string or `JsonEncodedText` instance to hold the value.
 
 ## Reduced size-on-disk
 Primarily this is for faster downloads of the runtime and applications.
@@ -104,12 +104,12 @@ Breaking this apart:
 **This option should be assumed as a prerequisite for all of the other options below.**
 
 Basic features include:
-- Avoid reflection emit when it is not supported
-- Improve caching
+- Avoid runtime code generation when it is not supported.
+- Improve caching.
 - Other smaller features TBD.
 
-Reflection emit is used both directly and indirectly by the serializer:
-  - Directly: used to generate getters, setters and calls to constructors.
+Runtime code generation is used:
+  - Directly: used to generate getters, setters and calls to constructors via Reflection.Emit APIs.
   - Indirectly: used to support strongly-typed converters for value types that avoid boxing. This includes usages of `Type.MakeGenericType` (and possibly `Type.MakeGenericMethod` pending additional research).
 
 Some early issues created to start tracking the work:
@@ -117,7 +117,7 @@ Some early issues created to start tracking the work:
 - [Add deferred loading of System.Private.Uri assembly by the serializer](https://github.com/dotnet/runtime/issues/35029). Minor startup gains; low-hanging fruit.
 
 ## Reflection features
-As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct usage of Reflection.Emit, but not the indirect usage.
+As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct useage of runtime code generation, but not the indirect usage.
 
 Note that AOT code-gen options likely eliminate the need for new reflection features for those POCOs that are code-gen'd. However, we can't assume or force all POCOs to be code-gen'd.
 
@@ -126,9 +126,23 @@ Misc other thoughts:
     - Another fallback would apply to a code-gen option where the generated POCO code closes the generic collections at build time. i.e. instantiate (or reference) a `JsonConverter<List<int>>` directly in the generated code instead of relying on the serializer to close `JsonConverter<T>` to `JsonConverter<List<int>>`.
 
 ## Converter code-gen
-A (prototype)[https://github.com/layomia/jsonconvertergenerator] was created for this by @layomia.
+A (prototype)[https://github.com/layomia/jsonconvertergenerator] was created for this by @layomia. It used a simple (non-Roslyn) code generator using the existing custom Value converter model.
 
-It used a simple (non-Roslyn) code generator using the existing custom Value converter model. Various serializer features (null handling, property lookup on deserialization, reference handling, async support, use of custom converters, etc) would need to be baked into the generated code. Thus the resuling code generation will be large, complex and not servicable. At the extreme (with no new helper methods exposed the serializer), each converter would need to implement all of the various features of the serializer.
+Various serializer features would need to be baked into the generated code including:
+- Setting and getting properties.
+- Calling the constructor, possibly with values from JSON.
+- Null handling (e.g. if a null property should be serialized).
+- Property lookup on deserialization (match JSON property name to Type property) potenially case-insensitive.
+- Property naming policies (such as camel-casing).
+- Escaping and unescaping property names and values (potentially using a custom escaper).
+- Object reference handling (to preserve object references).
+- Extension data (preserving umatched JSON during deserialization so it can be written during serialization).
+- Async support (supporting a mode that doesn't drain a Stream upfront).
+- Use of custom converters (where logic\code is not known so it can't be pushed to code-gen).
+- Composition of objects (`List<Dictionary<string, Poco>>` or `SalesOrder.Customer.Address.City`).
+- New features over time (default values, quoted numbers, ...)
+
+Thus the resuling code generation will be large, complex and not servicable. At the extreme (with no new helper APIs exposed), every converter would need to implement the various serializer features above. However, a given converter's generated code could be made very specific to its Type and the environment expected to execute in (known `JsonSerializerOptions` values and features used), so much of the logic could be omitted.
 
 ## Metadata provider code-gen
 A run-time only [prototype](https://github.com/steveharter/runtime/tree/ExtConverters) which does not include the actual code generator was created by @steveharter (see sample generated code below). It makes public the existing internal metadata classes including `JsonClassInfo` and `JsonPropertyInfo`.
@@ -184,9 +198,7 @@ public class WeatherForecast
                 HasSetter = true,
                 ShouldSerialize = true,
                 ShouldDeserialize = true,
-                // These delegates are nice that they work with aggressive linker
-                // but slower due to the extra hop (which may be avoided with
-                // new reflection features).
+                // These delegates are nice that they work with an aggressive linker.
                 Get = (obj) => { return ((WeatherForecast)obj).Date; },
                 Set = (obj, value) => { ((WeatherForecast)obj).Date = value; }
             });
--- a/proposed/SerializerGoals5.0.md
+++ b/proposed/SerializerGoals5.0.md
@@ -1,0 +1,281 @@
+# Overview and goals
+
+Apri 20th, 2020.
+
+This document provides options and recommendations for 5.0 to handle overlapping serializer goals including reach, faster cold startup performance, reducing memory usage, and smaller size-on-disk.
+
+## Reach
+Achieved by limiting or removing dependencies on Reflection.Emit and `Type.MakeGenericType`. This allows the serializer to be used on additional runtimes including [Xamarin iOS](https://github.com/dotnet/runtime/issues/31326).
+
+Ideally we add a high-performance emit replacement for targeted scenarios such as property setters, getters and constructors.
+
+Some serializers do not have a non-emit fallback. Although System.Text.Json has a fallback for emit, it does not have a fallback for its usage of `Type.MakeGenericType`.
+
+## Faster startup performance
+
+This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
+
+The serializer has fast startup compared to most other serializers including Newtonsoft's JSON.Net, JIL and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
+
+There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general sceanrios, although we do not want to noticably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
+
+<details>
+  <summary>5.0 master startup benchmarks</summary>
+
+A small POCO was used (`LoginViewModel` benchmark test class).
+
+| Serializer |  Serialize (us)  | Serialize ratio | Deserialize (us) | Deserialize ratio |
+| :-- | :-- | :-- | :-- | :--
+| **System.Text.Json** | 4,720 | 1.00 | 19,379 | 1.00 
+| **Json.NET** | 24,916 | 5.28 | 102,226 | 5.28 
+| **Utf8Json** | 7,290 | 1.54 | 106,817 | 5.51 
+| **Jil** | 66,456 | 14.08 | 170,548 | 8.80 
+</details>
+
+## Reduced memory usage
+Reducing private bytes (non-shared or per-process) memory is important for Bing scenarios amongst others.
+    - This can be done by reducing emit\`Type.MakeGenericType` and by capturing POCO metadata in generated code instead of RAM.
+
+## Reduced size-on-disk
+Primarily this is for faster downloads of the runtime and applications
+
+The runtime can be made smaller by removing emit support. Some runtimes including iOS do not support dynamic code generation (emit or jitting) due to policy and thus require AOT (Ahead-Of-Time compilation).
+
+The application can be made smaller by support aggressive linking behavior (e.g. "tree-shaking"). This can be done in a number of ways outside the scope of this document. However, approaches in this document that do code-gen may automatically handle this due to generated calls to members.
+
+# Serializer design background
+Today the serializer avoids using Reflection.Emit on non-.NET scenarios through the use of an #ifdef. However, the serializer still uses emit indirectly through `Type.MakeGenericType`. Note that Xamarin stubs `MakeGenericType` and tries to support these on iOS by pre-generating types via AOT but there are still issues with this (including not being able to determine all possible `<T>` variants).
+
+The existing serializer makes use of custom converters which is a class that knows how to implement a Read(...) and Write(...) method. There are 3 types of converters:
+1. Value. The only type of custom converters that can be implemented publically today (by the community).
+2. Collection. Used internally for IEnumerable Types.
+3. Object. Used internally for all other Types.
+
+The converters are stored on the `JsonSerializerOptions` in a dictionary keyed on the Type to convert. A single converter may know how to convert multiple Types.
+
+Internal converters use metadata for a given Type which is obtained from using reflection (to look for properties and JsonAttribute-derived attributes on a Type and its members). This metadata is then merged with values from the `JsonSerializerOptions` instance which is possible since options instance is immutable once the first (de)serialization occurs.
+
+The immutable semantics of the options also help in other ways:
+- The metadata could be embedded into code generation (to avoid the reflection cost of looking up the metadata).
+- The metadata could be used to control how code is generated if the generated code has branching or other logic that can be determined based upon the metadata (for example, having a custom property lookup mechanism for each Type).
+- Avoids locks and issues that would occur if another thread changes the options.
+
+A custom Value converter (implemented by the community) does not have access to the metadata since the metadata types are internal. This has been an issue for some scenarios and opening up the metadata APIs has been requested.
+
+The built-in converters do have access to metadata. The metadata is maintained internally by a `JsonClassInfo` object for every type. For Object converters (meaning non-Value and non-Collection converters) there is also an instance of a `JsonPropertyInfo` object for every property.
+
+# Startup performance costs
+The overhead of deserializing a simple 4-property POCO is around **22ms** for the first run (<1ms for second run). This was measured in 5.0 master as of 3/27/2020 and tested on an Intel i7 3.2GHz.
+
+Breaking this apart:
+- System (non-serializer related) (~5ms)
+  - ~2ms due to a one-time hit of loading and initializing SSE2 hardware intrinsics (in System.Runtime.Intrisics). SSE2 support was added during 5.0 to improve peformance of escaping JSON during serialization. It is commonly used elsewhere, however, so this can normally be subtracted from raw serializer numbers.
+  - ~3ms for loading\initializing other assembly dependencies including using the Reader for the first time. There is some additional research here to determine where this hit is coming from, as 3.1 does not have this overhead. It may be that SSE2 contributes more than 2ms overhead.
+- Serializer (~17ms)
+  - ~5ms for main API overhead and initialization of the `JsonSerializerOptions`. This includes cache creation for built-in converters. Also, this is also the amount of time it takes to use a custom converter (that uses the writer) for the test POCO.
+    - Includes ~1ms for creating the built-in `System.Uri` converter (loads the `System.Private.Uri.dll` assembly).
+  - ~3ms due to direct and indirect use Reflection.Emit.
+    - Direct: Reflection.Emit for property setters\getters and constructors.
+    - Indirect: usage of `Type.MakeGenericType` and `Type.MakeGenericMethod` so converters for value types can avoid boxing.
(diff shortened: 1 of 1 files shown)


proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
+# Overview and goals
+
+Apri 20th, 2020.
+
+This document provides options and recommendations for 5.0 to handle overlapping serializer goals including reach, faster cold startup performance, reducing memory usage, and smaller size-on-disk.
+
+## Reach
+Achieved by limiting or removing dependencies on Reflection.Emit and `Type.MakeGenericType`. This allows the serializer to be used on additional runtimes including [Xamarin iOS](https://github.com/dotnet/runtime/issues/31326).
+
+Ideally we add a high-performance emit replacement for targeted scenarios such as property setters, getters and constructors.
```
jkotas, line 10: This would address reach only if *everything* can be done with these.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Reduced size-on-disk
+Primarily this is for faster downloads of the runtime and applications
+
+The runtime can be made smaller by removing emit support. Some runtimes including iOS do not support dynamic code generation (emit or jitting) due to policy and thus require AOT (Ahead-Of-Time compilation).
+
+The application can be made smaller by support aggressive linking behavior (e.g. "tree-shaking"). This can be done in a number of ways outside the scope of this document. However, approaches in this document that do code-gen may automatically handle this due to generated calls to members.
+
+# Serializer design background
+Today the serializer avoids using Reflection.Emit on non-.NET scenarios through the use of an #ifdef. However, the serializer still uses emit indirectly through `Type.MakeGenericType`. Note that Xamarin stubs `MakeGenericType` and tries to support these on iOS by pre-generating types via AOT but there are still issues with this (including not being able to determine all possible `<T>` variants).
```
jkotas, line 47: > non-.NET scenarios

Everything we do is .NET. This needs to better name.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+## Reduced memory usage
+Reducing private bytes (non-shared or per-process) memory is important for Bing scenarios amongst others.
+    - This can be done by reducing emit\`Type.MakeGenericType` and by capturing POCO metadata in generated code instead of RAM.
+
+## Reduced size-on-disk
+Primarily this is for faster downloads of the runtime and applications
+
+The runtime can be made smaller by removing emit support. Some runtimes including iOS do not support dynamic code generation (emit or jitting) due to policy and thus require AOT (Ahead-Of-Time compilation).
+
+The application can be made smaller by support aggressive linking behavior (e.g. "tree-shaking"). This can be done in a number of ways outside the scope of this document. However, approaches in this document that do code-gen may automatically handle this due to generated calls to members.
```
jkotas, line 44: > This can be done in a number of ways outside the scope of this document.

Those other ways are either conservative or give you unpredictable results. Not very useful.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+| :-- | :-- | :-- | :-- | :--
+| **System.Text.Json** | 4,720 | 1.00 | 19,379 | 1.00 
+| **Json.NET** | 24,916 | 5.28 | 102,226 | 5.28 
+| **Utf8Json** | 7,290 | 1.54 | 106,817 | 5.51 
+| **Jil** | 66,456 | 14.08 | 170,548 | 8.80 
+</details>
+
+## Reduced memory usage
+Reducing private bytes (non-shared or per-process) memory is important for Bing scenarios amongst others.
+    - This can be done by reducing emit\`Type.MakeGenericType` and by capturing POCO metadata in generated code instead of RAM.
```
jkotas, line 37:
```suggestion
    - This can be done by reducing emit or `Type.MakeGenericType` and by capturing POCO metadata in generated code instead of RAM.
```

to avoid confusing the markdown.

proposed/SerializerGoals5.0.md
jkotas, line 47: > the serializer still uses emit indirectly through `Type.MakeGenericType`

This is not really emit. I would describe `Type.MakeGenericType` as requiring runtime code generation or interpreter. 

It may be useful to use Reflection.Emit instead of emit throughout the code.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+    }
+}
+```
+</details>
+
+## Object converter + code-gen
+This is simular to "Metadata provider code-gen" but includes exposing a new type of "object converter". This type of converter, however, exposes the raw reader\writer and has performance issues with Streams due to requiring read-ahead.
+
+# Code-Gen
+For the options that use code-gen there are several issues to discuss; this is covered in other documention but one open issue is to determine what Types to add code-gen for. The two basic approaches are automatic and explicit:
```
jkotas, line 216:
```suggestion
For the options that use code-gen there are several issues to discuss; this is covered in other documentation but one open issue is to determine what Types to add code-gen for. The two basic approaches are automatic and explicit:
```

jkotas, line 216: Are there links for the other documentation?

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+- Automatic options:
+  - Look to calls to the serialzer and code-gen types passed to it (and all other Types reachable those roots).
```
jkotas, line 218:
```suggestion
  - Look to calls to the serializer and code-gen types passed to it (and all other Types reachable those roots).
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
+# Overview and goals
+
+Apri 20th, 2020.
+
+This document provides options and recommendations for 5.0 to handle overlapping serializer goals including reach, faster cold startup performance, reducing memory usage, and smaller size-on-disk.
+
+## Reach
+Achieved by limiting or removing dependencies on Reflection.Emit and `Type.MakeGenericType`. This allows the serializer to be used on additional runtimes including [Xamarin iOS](https://github.com/dotnet/runtime/issues/31326).
```
marek-safar, line 8:
```suggestion
Achieved by limiting or removing dependencies on Reflection.Emit and `Type.MakeGenericType`. This allows the serializer to be used on AOT optimized platforms including [Xamarin iOS](https://github.com/dotnet/runtime/issues/31326).
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+
+Some serializers do not have a non-emit fallback. Although System.Text.Json has a fallback for emit, it does not have a fallback for its usage of `Type.MakeGenericType`.
+
+## Faster startup performance
+
+This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
+
+The serializer has fast startup compared to most other serializers including Newtonsoft's JSON.Net, JIL and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
+
+There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general sceanrios, although we do not want to noticably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
```
marek-safar, line 20: Typo in `sceanrios`

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+</details>
+
+## Reduced memory usage
+Reducing private bytes (non-shared or per-process) memory is important for Bing scenarios amongst others.
+    - This can be done by reducing emit\`Type.MakeGenericType` and by capturing POCO metadata in generated code instead of RAM.
+
+## Reduced size-on-disk
+Primarily this is for faster downloads of the runtime and applications
+
+The runtime can be made smaller by removing emit support. Some runtimes including iOS do not support dynamic code generation (emit or jitting) due to policy and thus require AOT (Ahead-Of-Time compilation).
```
marek-safar, line 42:
```suggestion
The runtime and BCL can be made smaller by removing Reflection.Emit support. Some platforms including iOS do not support fast dynamic code generation using JIT and thus require AOT (Ahead-Of-Time compilation) with interpreter fallback which increases the runtime size.
```

proposed/SerializerGoals5.0.md
marek-safar, line 44:
```suggestion
The application can be made smaller by running ILLinker (e.g. "tree-shaking"). This can be done in a number of ways outside the scope of this document. However, approaches in this document that do code-gen will improve the code analysis due to generated direct calls to members.
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+- The metadata could be embedded into code generation (to avoid the reflection cost of looking up the metadata).
+- The metadata could be used to control how code is generated if the generated code has branching or other logic that can be determined based upon the metadata (for example, having a custom property lookup mechanism for each Type).
+- Avoids locks and issues that would occur if another thread changes the options.
+
+A custom Value converter (implemented by the community) does not have access to the metadata since the metadata types are internal. This has been an issue for some scenarios and opening up the metadata APIs has been requested.
+
+The built-in converters do have access to metadata. The metadata is maintained internally by a `JsonClassInfo` object for every type. For Object converters (meaning non-Value and non-Collection converters) there is also an instance of a `JsonPropertyInfo` object for every property.
+
+# Startup performance costs
+The overhead of deserializing a simple 4-property POCO is around **22ms** for the first run (<1ms for second run). This was measured in 5.0 master as of 3/27/2020 and tested on an Intel i7 3.2GHz.
```
marek-safar, line 68: Link to GH hast instead of `was measured in 5.0 master as of 3/27/2020 `

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+
+Note that AOT code-gen options likely eliminate the need for new reflection features for those POCOs that are code-gen'd. However, we can't assume or force all POCOs to be code-gen'd.
+
+Misc other thoughts:
+    - A fallback (not recommended currently) is to change the internal converters and box for value types (and generic collections of value types). This will create many more short-lived allocs and decrease steady-state performance but allow the serializer to at least function in a "limp-mode" (roughly half of the throughput).
+    - Another fallback would apply to a code-gen option where the generated POCO code closes the generic collections at build time. i.e. instantiate (or reference) a `JsonConverter<List<int>>` directly in the generated code instead of relying on the serializer to close `JsonConverter<T>` to `JsonConverter<List<int>>`.
+
+## Converter code-gen
+A prototype was created for this by @layomia.
```
marek-safar, line 109: Could this be linked from this doc?

proposed/SerializerGoals5.0.md
steveharter: I'll add them.

Per @layomia see [publicly available markdown with design notes/questions/benchmarks](https://gist.github.com/layomia/5221e73374fd74dc360e74731dacfdb5) and [code-gen prototype](https://github.com/layomia/jsonconvertergenerator).

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+
+<details>
+  <summary>5.0 master startup benchmarks</summary>
+
+A small POCO was used (`LoginViewModel` benchmark test class).
+
+| Serializer |  Serialize (us)  | Serialize ratio | Deserialize (us) | Deserialize ratio |
+| :-- | :-- | :-- | :-- | :--
+| **System.Text.Json** | 4,720 | 1.00 | 19,379 | 1.00 
+| **Json.NET** | 24,916 | 5.28 | 102,226 | 5.28
```
MichalStrehovsky, line 30: What's the ReadyToRun setting for these measurements? If things were left at their defaults, System.Text.Json will be compiled with ReadyToRun because that's how we ship the library out of CoreFX and none of the other libraries will get that benefit. ReadyToRun has huge impact on startup time.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+<details>
+  <summary>5.0 master startup benchmarks</summary>
+
+A small POCO was used (`LoginViewModel` benchmark test class).
+
+| Serializer |  Serialize (us)  | Serialize ratio | Deserialize (us) | Deserialize ratio |
+| :-- | :-- | :-- | :-- | :--
+| **System.Text.Json** | 4,720 | 1.00 | 19,379 | 1.00 
+| **Json.NET** | 24,916 | 5.28 | 102,226 | 5.28 
+| **Utf8Json** | 7,290 | 1.54 | 106,817 | 5.51
```
MichalStrehovsky, line 31: Are the Utf8Json numbers with https://github.com/neuecc/Utf8Json/tree/master/src/Utf8Json.UniversalCodeGenerator? It would be interesting to see how much pregeneration helps Utf8Json here.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
+# Overview and goals
+
+Apri 20th, 2020.
```
jnm2, line 3:
```suggestion
April 20th, 2020.
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+
+Ideally we add a high-performance emit replacement for targeted scenarios such as property setters, getters and constructors.
+
+Some serializers do not have a non-emit fallback. Although System.Text.Json has a fallback for emit, it does not have a fallback for its usage of `Type.MakeGenericType`.
+
+## Faster startup performance
+
+This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
+
+The serializer has fast startup compared to most other serializers including Newtonsoft's JSON.Net, JIL and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
```
jnm2, line 18:
```suggestion
The serializer has fast startup compared to most other serializers including Newtonsoft's JSON.Net, Jil and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+
+## Converter code-gen
+A prototype was created for this by @layomia.
+
+It used a simple (non-Roslyn) code generator using the existing custom Value converter model. Various serializer features (null handling, property lookup on deserialization, reference handling, async support, use of custom converters, etc) would need to be baked into the generated code. Thus the resuling code generation will be large, complex and not servicable. At the extreme (with no new helper methods exposed the serializer), each converter would need to implement all of the various features of the serializer.
+
+## Metadata provider code-gen
+A run-time only prototype (not the actual generator) was created for this by @steveharter (see sample generated code below). It makes public the existing internal metadata classes including `JsonClassInfo` and `JsonPropertyInfo`.
+
+It allows for minimal code gen and overlaps with other requested features including property ordering, before\after callbacks and programmatic reading\writing of metadata.
```
jnm2, line 116: Backslash is not used for punctuation in English. Also, I'd recommend https://docs.microsoft.com/en-us/style-guide/punctuation/slashes:

> Don't use a slash as a substitute for *or*, like *product/service*. If the UI uses a slash in this way, follow the interface in describing the label, but use *or* to describe the action in text.

```suggestion
It allows for minimal code gen and overlaps with other requested features including property ordering, before and after callbacks and programmatic reading and writing of metadata.
```

jnm2: (+ several other locations)

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+
+            return properties;
+        }
+    }
+}
+```
+</details>
+
+## Object converter + code-gen
+This is simular to "Metadata provider code-gen" but includes exposing a new type of "object converter". This type of converter, however, exposes the raw reader\writer and has performance issues with Streams due to requiring read-ahead.
```
jnm2, line 213:
```suggestion
This is similar to "Metadata provider code-gen" but includes exposing a new type of "object converter." This type of converter, however, exposes the raw reader and writer and has performance issues with Streams due to requiring read-ahead.
```

proposed/SerializerGoals5.0.md
jnm2:
```suggestion
  - Look to calls to the seriailzer and code-gen types passed to it (and all other Types reachable from those roots).
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+  - ...
+
+Once we determine the runtime flow and format of the generated code, there is still much tooling to get the Roslyn-based code generator working.
+
+# Comparison of options
+Below is the listing of options; severe limimations are adorned with an &#x274C;.
+
+| | Minimal | Reflection features | Code-gen option: explicit | Code-gen option: metadata provider | Code-gen option: "object" custom converter |
+| :-- | :-- | :-- | :-- | :-- | :-- |
+| **Goal: reach (without code-gen)** | yes | no | N\A | N\A | N\A |
```
jnm2, line 233: Backslash is not used in English.

```suggestion
| **Goal: reach (without code-gen)** | yes | no | N/A | N/A | N/A |
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,281 @@
...
+| *Steady state regression?* | no | not expected | not expected but needs fast property lookup design | expected minor on property getter\setters | expected minor on property getter\setters |
+| *Fast Stream perf (no read-ahead)* | yes | yes | possible but hard &#x274C; | yes | possible but hard &#x274C; |
+| | | | | | |
+| **Goal: reduced memory usage** | 
+| *Private bytes reduction* | yes (less MakeGenericType) | yes (less Emit) | yes (metadata baked into code) | TBD | TBD|
+| | | | | | |
+| **Goal: reduced size-on-disk**
+| *Supports smaller runtime with no Emit* | no | yes | no | no | no
+| *Could support linker "tree shaking"* | no | possible for the default ctor only | yes | yes | yes
+| *POCO code gen size* | n\a | n\a | large &#x274C; | small | small |
```
jnm2, line 252: Consistent capitalization

```suggestion
| *POCO code gen size* | N/A | N/A | large &#x274C; | small | small |
```

proposed/SerializerGoals5.0.md
steveharter: I re-worded this area to provide the full suite of options to consider since there are combination of refactoring, emit replacement, etc that can be used together.

proposed/SerializerGoals5.0.md
steveharter: Yes. I added a bit more information here that serialization scenarios are not generally suitable for linking.

proposed/SerializerGoals5.0.md
steveharter: I made a pass through and removed emit-based terminology when it could also apply to any runtime code generation.

proposed/SerializerGoals5.0.md
steveharter: I'll link that the next time I run the benchmarks.

proposed/SerializerGoals5.0.md
steveharter: Done

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+This can be done by:
+- Avoiding runtime code generation including Reflection.Emit or JITting.
+- Capturing POCO metadata in generated code instead of RAM.
+
+## Reduced size-on-disk
+Primarily this is for faster downloads of the runtime and applications.
+
+The runtime and BCL can be made smaller by removing Reflection.Emit support.
+
+Although outside the scope of this document, an application can be made smaller by running ILLinker (e.g. "tree-shaking"). Serializer scenarios are generally unsuited for linking but can be somewhat mitigated by using metadata to declare what members to preserve. The code-gen approaches in this document, depending on implementation, may help with this since they generate direct calls to members that are detectable by the linker.
```
jkotas, line 53:
```suggestion
Although outside the scope of this document, an application can be made smaller by running ILLinker (e.g. "tree-shaking"). Serializers generated at runtime are generally problematic for linking but can be somewhat mitigated by using metadata to declare what members to preserve. The code-gen approaches in this document, depending on implementation, may help with this since they generate direct calls to members that are detectable by the linker.
```
The problems with IL linking are not specific to the scenario; they are specific to the particular implementation strategies.

proposed/SerializerGoals5.0.md
steveharter: cc @layomia 

For the "overhead" status I provided later (e.g. 22ms overhead) those were done with a public 5.0 build.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+5) If possible, add or extend runtime interpreter support for missing cases. Xamarin (and .NET Native) has an interpreter mode.
+6) Throw PlatformNotSupportedException for when there is no alternative.
+
+## Faster startup performance
+
+This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
+
+The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, JIL and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
+
+There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general sceanrios, although we do not want to noticably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
```
jkotas, line 26:
```suggestion
There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general scenarios, although we do not want to noticeably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+3) Generate code for POCOs that avoids using the unsupported features. This also has higher runtime performance than the loosely-typed fallbacks.
+4) Add a high-performance replacement for targeted scenarios including property getters, property setters, constructors and as a stretch `Type.MakeGenericType and .MakeGenericMethod`.
+5) If possible, add or extend runtime interpreter support for missing cases. Xamarin (and .NET Native) has an interpreter mode.
+6) Throw PlatformNotSupportedException for when there is no alternative.
+
+## Faster startup performance
+
+This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
+
+The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, JIL and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
```
jkotas, line 24:
```suggestion
The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, JIL and Utf8Json. Startup perf is mostly affected by generating IL - the JIL serializer is the most aggressive here and has the highest startup cost.
```

jkotas: Actually, https://github.com/kevin-montrose/Jil uses `Jil` casing, so it is what we should use throughout the document as well, I think

jkotas, line 24:
```suggestion
The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, Jil and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost.
```
Match casing used by https://github.com/kevin-montrose/Jil

proposed/SerializerGoals5.0.md
steveharter: PTAL @layomia

proposed/SerializerGoals5.0.md
steveharter: Yep. Thanks

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+The serializer has fast startup compared to most other serializers including Newtonsoft's Json.NET, Jil and Utf8Json. Startup perf is mostly affected by generating IL - the Jil serializer is the most aggressive here and has the highest startup cost (but in many cases has the fastest steady-state performance).
+
+There is no stated goal for 5.0 to make the steady-state CPU performance faster than it is now in 5.0 for general sceanrios, although we do not want to noticably regress steady-state performance while addressing the stated goals including improving startup performance. It is possible some code-gen options described later may have steady-state performance gains, but again that is not a stated goal.
+
+<details>
+  <summary>5.0 master startup benchmarks (click to expand)</summary>
+
+A small POCO was used (`LoginViewModel` benchmark test class).
+
+| Serializer |  Serialize (us)  | Serialize ratio | Deserialize (us) | Deserialize ratio |
```
layomia, line 33: The `Serialize` and `Deserialize` columns are reversed here.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+| **Utf8Json** | 7,290 | 1.54 | 106,817 | 5.51 
+| **Jil** | 66,456 | 14.08 | 170,548 | 8.80 
+</details>
+
+## Reduced memory usage
+Reducing private bytes (non-shared per-process) memory is important for Bing scenarios and other scenarios that have many process using the same POCOs or start and stop many processes. Shared memory can be amortized across many processes plus is already there when a new processes starts and thus requires no additional overhead to generate.
+
+This can be done by:
+- Avoiding runtime code generation including Reflection.Emit or JITting.
+- Capturing POCO metadata in generated code instead of RAM.
```
davidfowl, line 46: Excuse my ignorance here but what does this mean? Representing metadata in code (aka custom metadata) vs what?

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+
+## Reduced size-on-disk
```
davidfowl, line 48: Can we quantify this?

proposed/SerializerGoals5.0.md
layomia: The ReadyToRun setting for these benchmarks is `false`.

I ran them again with `ReadyToRun` set to true using a Release build of .NET 5 at https://github.com/dotnet/runtime/commit/38c2d5513c - https://github.com/layomia/jsonconvertergenerator/blob/00e81d5411175c8874528f2837bfbf3319637426/run_benchmarks.py#L31-L32

Deserialize LoginViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| System.Text.Json   | 4351      | 1.00  |
| Json.NET           | 2417      | 0.56  |
| Utf8Json           | 732       | 0.17  |
| Jil                | 45095     | 10.36 |

Serialize LoginViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| System.Text.Json   | 20885     | 1.00  |
| Json.NET           | 82531     | 3.95  |
| Utf8Json           | 71666     | 3.43  |
| Jil                | 72114     | 3.45  |

<details>
<summary>
See more benchmarks for other POCOs. The trend is we have faster start-up perf than the other serializers when serializing, but  we are slower when deserializing (except for Jil).
</summary>

---

`Default` = System.Text.Json

Deserialize LoginViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 45095     | 10.36 |
| Json.NET           | 2417      | 0.56  |
| Utf8Json           | 732       | 0.17  |
| Default            | 4351      | 1.00  |

Deserialize Location

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 55325     | 10.52 |
| Json.NET           | 3304      | 0.63  |
| Utf8Json           | 1377      | 0.26  |
| Default            | 5257      | 1.00  |

Deserialize IndexViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 80323     | 10.49 |
| Json.NET           | 5327      | 0.70  |
| Utf8Json           | 2867      | 0.37  |
| Default            | 7654      | 1.00  |

Deserialize MyEventsListerViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 84313     | 10.33 |
| Json.NET           | 5229      | 0.64  |
| Utf8Json           | 3554      | 0.44  |
| Default            | 8161      | 1.00  |

Deserialize CollectionsOfPrimitives

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 49745     | 3.80  |
| Json.NET           | 5249      | 0.40  |
| Utf8Json           | 2545      | 0.19  |
| Default            | 13087     | 1.00  |

Serialize LoginViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 72114     | 3.45  |
| Json.NET           | 82531     | 3.95  |
| Utf8Json           | 71666     | 3.43  |
| Default            | 20885     | 1.00  |

Serialize Location

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 71383     | 3.28  |
| Json.NET           | 84474     | 3.88  |
| Utf8Json           | 70136     | 3.23  |
| Default            | 21744     | 1.00  |

Serialize IndexViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 92385     | 3.54  |
| Json.NET           | 89281     | 3.42  |
| Utf8Json           | 81163     | 3.11  |
| Default            | 26090     | 1.00  |

Serialize MyEventsListerViewModel

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 116070    | 3.76  |
| Json.NET           | 91335     | 2.96  |
| Utf8Json           | 89143     | 2.89  |
| Default            | 30833     | 1.00  |

Serialize CollectionsOfPrimitives

| Test               | Mean (us) | Ratio |
|--------------------|-----------|-------|
| Jil                | 74404     | 2.53  |
| Json.NET           | 85696     | 2.92  |
| Utf8Json           | 76411     | 2.60  |
| Default            | 29367     | 1.00  |

</details>

proposed/SerializerGoals5.0.md
layomia, line 33: The `Serialize` and `Deserialize` columns here are reversed.

proposed/SerializerGoals5.0.md
layomia: No they aren't. I'll take a look and share numbers.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+**This option should be assumed as a prerequisite for all of the other options below.**
+
+Basic features include:
+- Avoid reflection emit when it is not supported
+- Improve caching
+- Other smaller features TBD.
+
+Reflection emit is used both directly and indirectly by the serializer:
+  - Directly: used to generate getters, setters and calls to constructors.
+  - Indirectly: used to support strongly-typed converters for value types that avoid boxing. This includes usages of `Type.MakeGenericType` (and possibly `Type.MakeGenericMethod` pending additional research).
```
jkotas, line 113:
```suggestion
Runtime code generation is used:
  - Directly: used to generate getters, setters and calls to constructors via Reflection.Emit APIs.
  - Indirectly: used to support strongly-typed converters for value types that avoid boxing. This includes usages of `Type.MakeGenericType` (and possibly `Type.MakeGenericMethod` pending additional research).
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+    - Direct: Reflection.Emit for getters, setters and constructors.
+    - Indirect: usage of `Type.MakeGenericType` and `Type.MakeGenericMethod` so converters for value types can avoid boxing.
+  - ~9ms usage of System.Reflection (and perhaps other items not yet accounted for) to initialize metadata. This includes inspecting each property for attributes.
+
+# General options to consider
+## Minimal
+**This option should be assumed as a prerequisite for all of the other options below.**
+
+Basic features include:
+- Avoid reflection emit when it is not supported
```
jkotas, line 107:
```suggestion
- Avoid runtime code generation when it is not supported
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+Reflection emit is used both directly and indirectly by the serializer:
+  - Directly: used to generate getters, setters and calls to constructors.
+  - Indirectly: used to support strongly-typed converters for value types that avoid boxing. This includes usages of `Type.MakeGenericType` (and possibly `Type.MakeGenericMethod` pending additional research).
+
+Some early issues created to start tracking the work:
+- [Avoid Type.MakeGenericType when (de)serializing objects](https://github.com/dotnet/runtime/issues/35028)
+- [Add deferred loading of System.Private.Uri assembly by the serializer](https://github.com/dotnet/runtime/issues/35029). Minor startup gains; low-hanging fruit.
+
+## Reflection features
+As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct usage of Reflection.Emit, but not the indirect usage.
```
jkotas, line 120:
```suggestion
As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct runtime code generation, but not the indirect usage.
```

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+3) Generate code for POCOs that avoids using the unsupported features. This also has higher runtime performance than the loosely-typed fallbacks.
+4) Add a high-performance replacement for targeted scenarios including property getters, property setters, constructors and as a stretch `Type.MakeGenericType and .MakeGenericMethod`.
+5) If possible, add or extend runtime interpreter support for missing cases. Xamarin (and .NET Native) has an interpreter mode.
+6) Throw PlatformNotSupportedException for when there is no alternative.
+
+## Faster startup performance
+
+This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently. It is also important for benchmarks including TechEmpower.
```
jkotas, line 22:
```suggestion
This is important for scenarios including [micro-services](https://github.com/dotnet/runtime/issues/1568) where processes start and exit frequently.
```
TechEmpower does not measure startup performance.

proposed/SerializerGoals5.0.md
```diff
@@ -0,0 +1,301 @@
...
+            properties.Add(new JsonPropertyInfo<DateTime>
+            {
+                Options = Options,
+                Converter = Serializer.DateTimeConverter,
+                NameAsString = nameof(WeatherForecast.Date),
+                HasGetter = true,
+                HasSetter = true,
+                ShouldSerialize = true,
+                ShouldDeserialize = true,
+                // These delegates are nice that they work with aggressive linker
```
jkotas, line 187: How can the reflection generated good be ever faster than this? I think that this is the fastest you can get. I do not see what kind of magic can beat that.

steveharter: I was thinking if we have a fast way to call the getter\setter directly such as what you can do with MethodInfo.CreateDelegate but without having to use reflection to get the MethodInfo for a property.

If we could do that, it would be 2x as fast as the delegate "hop" approach as just as fast as IL Emit.

Here's pretty much every way to call a property getter (except for expressions):
```
Direct:2.
Reflection:831.
Manual loosely typed object delegate:49. **(what I have in prototype)**
Manual strongly typed delegate:28.
Generated static method which calls property:43.
MethodInfo getter:18. **(would be nice to be able to do something like this)**
Il getter:28.
Expression getter:11.
```

```cs
using System;
using System.Diagnostics;
using System.Linq.Expressions;
using System.Reflection;
using System.Reflection.Emit;

namespace ConsoleApp20benchmark
{
    class Program
    {
        static void Main(string[] args)
        {
            const long Iterations = 10_000_000;

            var poco = new POCO();
            PropertyInfo pi = typeof(POCO).GetProperty("MyProperty");
            MethodInfo mi = pi.GetGetMethod();

            int value;

            var sw = new Stopwatch();

            {
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = poco.MyProperty;
                    Debug.Assert(value == 42);
                }
                sw.Stop();

                Console.WriteLine($"Direct:{sw.ElapsedMilliseconds}.");
            }

            {
                sw.Reset();
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = (int)mi.Invoke(poco, null);
                    Debug.Assert(value == 42);
                }
                sw.Stop();

                Console.WriteLine($"Reflection:{sw.ElapsedMilliseconds}.");
            }

            {
                Func<object, int> ManualGetter = (obj) =>
                {
                    return ((POCO)obj).MyProperty;
                };

                sw.Reset();
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = ManualGetter(poco);
                    Debug.Assert(value == 42);
                }
                sw.Stop();

                Console.WriteLine($"Manual loosely typed object delegate:{sw.ElapsedMilliseconds}.");
            }

            {
                Func<POCO, int> ManualStronglyTypedGetter = (obj) =>
                {
                    return obj.MyProperty;
                };

                sw.Reset();
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = ManualStronglyTypedGetter(poco);
                    Debug.Assert(value == 42);
                }
                sw.Stop();

                Console.WriteLine($"Manual strongly typed delegate:{sw.ElapsedMilliseconds}.");
            }

            {
                Func<POCO, int> ManualStronglyTypedGetter = (obj) => POCO.CallMyPropertyGetter(obj);

                sw.Reset();
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = ManualStronglyTypedGetter(poco);
                    Debug.Assert(value == 42);
                }
                sw.Stop();

                Console.WriteLine($"Generated static method which calls property:{sw.ElapsedMilliseconds}.");
            }

            {
                // Unforunately, we must use MethodInfo to get the direct delegate.
                Func<POCO, int> DelegateGetter = (Func<POCO, int>)mi.CreateDelegate(typeof(Func<POCO, int>));

                sw.Reset();
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = DelegateGetter(poco);
                    Debug.Assert(value == 42);
                }
                sw.Stop();

                Console.WriteLine($"MethodInfo getter:{sw.ElapsedMilliseconds}.");
            }

            {
                var dynamicMethod = new DynamicMethod(
                    mi.Name,
                    mi.ReturnType,
                    new[] { typeof(object) },
                    typeof(Program).Module,
                    skipVisibility: true);

                ILGenerator generator = dynamicMethod.GetILGenerator();
                generator.Emit(OpCodes.Ldarg_0);
                generator.Emit(OpCodes.Castclass, mi.DeclaringType); // to verify type
                generator.Emit(OpCodes.Callvirt, mi);
                generator.Emit(OpCodes.Ret);

                Func<POCO, int> IlGetter = (Func<POCO, int>)dynamicMethod.CreateDelegate(typeof(Func<,>).MakeGenericType(mi.DeclaringType, mi.ReturnType));

                sw.Reset();
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = IlGetter(poco);
                    Debug.Assert(value == 42);
                }
                sw.Stop();
                
                Console.WriteLine($"Il getter:{sw.ElapsedMilliseconds}.");
            }

            {
                ParameterExpression parameter = Expression.Parameter(pi.DeclaringType, pi.Name);
                MemberExpression property = Expression.Property(parameter, pi);
                Func<POCO, int> ExpressionGetter = (Func<POCO, int>) Expression.Lambda(property, parameter).Compile();

                sw.Reset();
                sw.Start();
                for (long l = 0; l < Iterations; l++)
                {
                    value = ExpressionGetter(poco);
                    Debug.Assert(value == 42);
                }
                sw.Stop();

                Console.WriteLine($"Expression getter:{sw.ElapsedMilliseconds}.");
            }
        }


        class POCO
        {
            public int MyProperty => 42;

            internal static int CallMyPropertyGetter(POCO obj)
            {
                return obj.MyProperty;
            }
        }
    }
}
```
```

Comment (jkotas) [reply proposed/SerializerGoals5.0.md:187]: > Here's pretty much every way to call a property getter (except for expressions):

These numbers look suspect. I think you are seeing a lot of noise from tiered JIT, and other first-time initializations. I have added a big for-loop over the whole Main method. Here is the difference between 1st and 10th iteration:

1st iteration:
```
Direct:3.
Reflection:1233.
Manual loosely typed object delegate:80.
Manual strongly typed delegate:52.
Generated static method which calls property:80.
MethodInfo getter:37.
Il getter:44.
Expression getter:24.
```

10th iteration:
```
Direct:3.
Reflection:1104.
Manual loosely typed object delegate:26.
Manual strongly typed delegate:23.
Generated static method which calls property:26.
MethodInfo getter:30.
Il getter:33.
Expression getter:20.
```

Comment (jkotas) [reply proposed/SerializerGoals5.0.md:187]: The 10th iteration numbers say that direct call is super fast (it should actually get optimized out completely in your benchmark and so you are just measuring how fast one can count to 10_000_000), the reflection Invoke is slow, and the rest is pretty much the same - the differences are in the noise range.

Comment (jkotas) [reply proposed/SerializerGoals5.0.md:187]: Also, the distribution would be different for something that returns value type: It would show that operating on `object`s has extra overhead due to boxing.

Comment (jnm2) [reply proposed/SerializerGoals5.0.md:187]: BenchmarkDotNet = instant pit of success from what I've heard.

Comment (steveharter) [reply proposed/SerializerGoals5.0.md:187]: An advantage of the "hop" approach, however, is that it will work with linkers since it has the direct calls to the setters\getters. The same can't be said for some of the other approaches.

Comment (jkotas) [reply proposed/SerializerGoals5.0.md:187]: I agree with that.

Comment (jkotas) [reply proposed/SerializerGoals5.0.md:187]: To be precise:
- It works with the linker if you can reliably identify the full of types involved in the serialization. 
- If you use the strongly typed wrappers, it still requires code generation at runtime (e.g. hidden behind MakeGenericType, etc.)

File: proposed/SerializerGoals5.0.md (hunk 24 above)
Comment (steveharter) [reply proposed/SerializerGoals5.0.md:46]: Updating this to say minimize the amount of cached metadata by the serializer if we can just call into generated code to get the values or process them (such as writing out an escaped property directly instead of caching a `JsonEncodedText`.

File: proposed/SerializerGoals5.0.md (hunk 30 above)
Comment (steveharter) [reply proposed/SerializerGoals5.0.md]: > I think you are seeing a lot of noise from tiered JIT

Yes thanks the numbers do even out after a few runs.

However, if we are measuring cold start we do care about the first run.

File: proposed/SerializerGoals5.0.md (hunk 29 above)
Comment (steveharter) [reply proposed/SerializerGoals5.0.md]: Removed. Thanks

File: proposed/SerializerGoals5.0.md (hunk 25 above)
Comment (steveharter) [reply proposed/SerializerGoals5.0.md:48]: I assume you mean minimizing RAM\heap usage? Or do you mean size-on-disk? I do not have numbers yet for either but can get some numbers depending on what you find useful.

I do know that with large numbers of POCOs + many properties + longer-than-you-think property names (which is common) the cached metadata becomes significant, although the memory used is probably less than what the CLR maintains for each Type. Also FWIW currently for every property the serializer tracks 3 variants of its name - which I am currently looking at removing 1.

File: proposed/SerializerGoals5.0.md (hunk 30 above)
Comment (jkotas) [reply proposed/SerializerGoals5.0.md]: If you would like to measure cold start, you need to measure it properly. I do not think that the numbers from the first iteration say anything about the cold start characteristics of the different options.

Comment (steveharter) [reply proposed/SerializerGoals5.0.md]: Alsothe "MethodInfo getter" or variant of that approach would be useful to remove our Emit usage, which we are tracking in https://github.com/dotnet/runtime/issues/23716.

Comment (jkotas) [reply proposed/SerializerGoals5.0.md]: For example, the expression getter should have by far the worst cold start characteristic because of it pulls in Expressions that are big expensive piece of code.

File: proposed/SerializerGoals5.0.md [hunk 31]
```diff
@@ -0,0 +1,313 @@
...
+Runtime code generation is used:
+  - Directly: used to generate getters, setters and calls to constructors via Reflection.Emit APIs.
+  - Indirectly: used to support strongly-typed converters for value types that avoid boxing. This includes usages of `Type.MakeGenericType` (and possibly `Type.MakeGenericMethod` pending additional research).
+
+Some early issues created to start tracking the work:
+- [Avoid Type.MakeGenericType when (de)serializing objects](https://github.com/dotnet/runtime/issues/35028)
+- [Add deferred loading of System.Private.Uri assembly by the serializer](https://github.com/dotnet/runtime/issues/35029). Minor startup gains; low-hanging fruit.
+
+## Reflection features
+As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct useage of runtime code generation, but not the indirect usage.
```
jkotas, line 120:
```suggestion
As described in the [System.Reflection roadmap](https://github.com/dotnet/runtime/issues/31895) it is possible to add new APIs that will eliminate common usages of Reflection.Emit including getters, setters and constructors. This will help with the direct usage of runtime code generation, but not the indirect usage.
```

proposed/SerializerGoals5.0.md
steveharter: Yes definitely the Emit and Expression approaches are out for cold start.

Since today we use Emit I think we likely have two options:
1) For AOT, a code-gen'd delegate approach as shown in the doc's prototype will be fast and work with linkers.
2) For non-AOT a new reflection feature is needed that does not have cold start issues but still has fast steady-state performance.

steveharter: For non-AOT, until the "new reflection feature" is available, I suggest we continue to do what we do today which is use Emit where it is supported and use the slow Reflection.Invoke as a fallback.

steveharter: Ideally we have one approach for both AOT and non-AOT but until we get further along with the "new reflection feature" I'm not sure if that will be feasible.

We also need to determine the priority of linker support, and\or whether the new reflection feature could support the linker in some way.

jkotas: > We also need to determine the priority of linker support

+1. I think there is no shared understanding of it today.

Would it make sense to simplify this to just two scenarios that we focus on?
- The dynamic mode that we have today
- The linker and AOT friendly mode

I understand you can have number of options in between, but I am worried that having many different options will be hard to explain.

PR #114, opened 2020-04-22
.NET 5 TFMs: Clarify behavior on mapping to long names

terrajobst:


accepted/2020/net5/net5.md
```diff
@@ -502,6 +502,53 @@ NuGet uses the long form `.NETFramework,Version=4.5` in the dependency groups.
...
+long name form.
+
+Since the friendly name has become the primary name that developers author in
+the developer experience (project file, NuGet folder names, etc) we'd like to
+standardize on that encoding. Unfortunately, NuGet itself isn't consistent with
+using the friendly name today. We generally don't want to break backwards
+compatibility, that is existing frameworks should be encoded in the way they are
+encoded today, be that friendly name, long name or some custom encoding.
+
+However, for frameworks in the .NET 5 era (and later) we want to use
```
heaths, line 531: Probably should be, "we want to consistent use".

terrajobst: Good call. Fixed.

accepted/2020/net5/net5.md
```diff
@@ -502,6 +502,72 @@ NuGet uses the long form `.NETFramework,Version=4.5` in the dependency groups.
...
+    * For other target frameworks use current encoding (i.e. long form)
+    * This allows old and new tooling to be used in the same repo
+* `.nuspec` files
+    * **Use friendly name for .NET 5 and higher**
+    * Use existing nuspec form (which is neither friendly- nor long name) for
+      other targets
+    * This will preserve compatibility with older clients, and can be considered
+      mostly an implementation detail of NuGet
+* `NuGetFramework` APIs
+    * `GetDotNetFrameworkName()` should continue to return what it does today –
```
nkolev92, line 559: Just want to make sure that we are on the same page, this is the expensive solution from NuGet perspective. 
The one where NuGet.org, PM UI & related will need to be updated to use this new value. 

Sounds like we don't need it for build, so we are all good :)

accepted/2020/net5/net5.md
```diff
@@ -502,6 +502,72 @@ NuGet uses the long form `.NETFramework,Version=4.5` in the dependency groups.
...
+      the long form of the target framework information, without platform
+      information
+        * We should consider using `[Obsolete]` for `GetDotNetFrameworkName()`,
+          and creating a new, separate API that always returns the short form
+          and is more usefully named.
+    * Add a new `GetPlatformName()` method which returns the long form of the
```
nkolev92, line 565: LGTM. All individual properties will be available as well I think. 
@zkat knows best.

nkolev92: nvm, @zkat already reviewed this :D 
GH betrayed me yet again.

accepted/2020/net5/net5.md
```diff
@@ -502,6 +502,72 @@ NuGet uses the long form `.NETFramework,Version=4.5` in the dependency groups.
...
+    * Use existing nuspec form (which is neither friendly- nor long name) for
+      other targets
+    * This will preserve compatibility with older clients, and can be considered
+      mostly an implementation detail of NuGet
+* `NuGetFramework` APIs
+    * `GetDotNetFrameworkName()` should continue to return what it does today –
+      the long form of the target framework information, without platform
+      information
+        * We should consider using `[Obsolete]` for `GetDotNetFrameworkName()`,
+          and creating a new, separate API that always returns the short form
```
dsplaisted, line 563: I think we already have the API that returns the short form: `GetShortFolderName`

accepted/2020/net5/net5.md
zkat: that wasn't my understanding D: see https://github.com/dotnet/designs/pull/114#discussion_r413378960

accepted/2020/net5/net5.md
```diff
@@ -502,6 +502,72 @@ NuGet uses the long form `.NETFramework,Version=4.5` in the dependency groups.
...
+* `.nuspec` files
+    * **Use friendly name for .NET 5 and higher**
+    * Use existing nuspec form (which is neither friendly- nor long name) for
+      other targets
+    * This will preserve compatibility with older clients, and can be considered
+      mostly an implementation detail of NuGet
+* `NuGetFramework` APIs
+    * `GetDotNetFrameworkName()` should continue to return what it does today –
+      the long form of the target framework information, without platform
+      information
```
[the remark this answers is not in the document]
dsplaisted: I think the code creating the assets and lock files needs to understand which format it wants to use.  The assets file will always use the short form, while the lock file will only use the short form for .NET 5 and higher.

It's not possible to return a value from `GetDotNetFrameworkName` that will work with all current consumers of the API, so I think we should choose consistent / simple behavior, ie return the long form of the FrameworkName but no platform info.

We could consider marking it obsolete to help consumers understand that they may need to update their code, but in some cases it will still be the right API to use, so I'm not sold on obsoleting it.

accepted/2020/net5/net5.md
terrajobst: Alright, will remove this bullet then

accepted/2020/net5/net5.md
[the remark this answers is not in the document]
terrajobst: Or, we could just update `GetDotNetFrameworkName()` to use the policy:

* If .NET 5 or higher -> return friendly name
* else current behavior

accepted/2020/net5/net5.md
terrajobst: OK, I'll resolve this in favor of the other thread

accepted/2020/net5/net5.md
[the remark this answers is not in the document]
zkat: would we also be applying this rule to assets.json? So instead of -always- being short, we use the long version for legacy? (I think this is a good idea anyway fwiw)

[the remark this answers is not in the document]
terrajobst: I would because it centralizes the policy and avoids having to do this ".NET 5 era" check in multiple places.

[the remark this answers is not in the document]
terrajobst: Looks this got quorum :-)

PR #118, opened 2020-04-29
.NET Runtime Form Factors

jkotas:


jkotas:
cc @richlander

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,205 @@
...
+
+- [SingleFilePublish is quite unusable](https://github.com/dotnet/corert/issues/7200#issuecomment-564778685)
+- [Current executing assembly location is wrong for single file published project](https://github.com/dotnet/runtime/issues/13531)
+- [PublishSingleFile with Windows Service doesn't work with config-file](https://github.com/dotnet/runtime/issues/32735)
+
+We expect that developers will continue to experience compatibility issues with the latest iteration of single-file form factor that we will deliver with .NET 5.0. This is rooted in the success and prevalence of the global runtime, which many libraries, tools and some of the .NET APIs have been specifically designed for. This situation makes it challenging to establish new form factors where compatibility issues, relative to the global runtime, can be observed and that hinder adoption. This topic is a common thread throughout this document.
+
+### Optimized for Size
+
+Binary size is a critical for many scenarios: client-side Blazor, Xamarian or lean micro-services.
```
rolfbjarne, line 60:
```suggestion
Binary size is a critical for many scenarios: client-side Blazor, Xamarin or lean micro-services.
```

EatonZ:
You might consider adding [this one](https://github.com/mono/linker/issues/772) to the _Optimized for Size_ section.

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,205 @@
...
+- [PublishSingleFile with Windows Service doesn't work with config-file](https://github.com/dotnet/runtime/issues/32735)
+
+We expect that developers will continue to experience compatibility issues with the latest iteration of single-file form factor that we will deliver with .NET 5.0. This is rooted in the success and prevalence of the global runtime, which many libraries, tools and some of the .NET APIs have been specifically designed for. This situation makes it challenging to establish new form factors where compatibility issues, relative to the global runtime, can be observed and that hinder adoption. This topic is a common thread throughout this document.
+
+### Optimized for Size
+
+Binary size is a critical for many scenarios: client-side Blazor, Xamarin or lean micro-services.
+
+[IL linking](https://github.com/mono/linker) is the key technology to significantly reduce the size of applications. It analyzes the application and removes unreachable code from the final binaries. Certain code patterns cannot always be reliably analyzed by the IL linker, which can causes optimized applications to fail at runtime.
```
sbomer, line 62:
```suggestion
[IL linking](https://github.com/mono/linker) is the key technology to significantly reduce the size of applications. It analyzes the application and removes unreachable code from the final binaries. Certain code patterns cannot always be reliably analyzed by the IL linker, which can cause optimized applications to fail at runtime.
```

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,206 @@
+# .NET Runtime Form Factors
+
+[Jan Kotas](https://github.com/jkotas) [Rich Lander](https://github.com/richlander)
```
benaadams, line 3:
```suggestion
[Jan Kotas](https://github.com/jkotas) | [Rich Lander](https://github.com/richlander)
```

mattzink:
This doesn't seem to address the "JIT on every startup is a waste of resources, JIT should only be required for dynamically emitted code" argument. Are there plans around that?

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,206 @@
...
+
+### IL Linking
+
+The [IL linker](https://github.com/mono/linker) can generate optimized applications that fail at runtime, due to not being able to reliably analyze various problematic code patterns (largely centered on reflection use). It is not possible or desirable to simply invest more to improve the linker, as a strategy. Instead, build-time source generators will be the recommended mitigation for arbitrary reflection use.
+
+The barrier for entry for building source generators is very high today. There is no standard solution and only a few libraries can afford to build a custom one. Razor pages, gRPC, WPF data binding, and Data-Contract serializers each have their own custom build-time source generator solutions today. First-class Roslyn support for [source generators](https://github.com/dotnet/roslyn/blob/master/docs/features/source-generators.md) will lower the barrier of entry for components that need to generate source code.
```
encrypt0r, line 129: > Razor pages, gRPC, WPF data binding, and Data-Contract serializers each have their own custom build-time source generator solutions today.

I feel like this should be UWP data binding, as WPF doesn't support compile time binding (x:Bind)

jkotas:
> This doesn't seem to address the "JIT on every startup is a waste of resources, JIT should only be required for dynamically emitted code" argument. 

It depends. If the JIT is included with the runtime, there are other situations other than dynamically emitted code where it is profitable or necessary to take advantage of it, e.g. dynamically reoptimizing hot code (tiered JITing), compiling rarely used generic instantiations that are not worth including it the AOT image, or supporting existing AOT-unfriendly features that depend on dynamic code generation indirectly.

> Are there plans around that?

The doc covers it in https://github.com/dotnet/designs/pull/118/files#diff-e9261a43b3e3f2325321c410c9e3bd38R98 and https://github.com/dotnet/designs/pull/118/files#diff-e9261a43b3e3f2325321c410c9e3bd38R145 .

richlander:
Relevant source generator blog link: https://devblogs.microsoft.com/dotnet/introducing-c-source-generators/

charlesroddie:
This is a brilliant overview and strategy.

One thing I don't get about the doc is the distinction between the two AOT sections:
- "Ahead-of-time (AOT) Compilation": mainly discusses source generators.
- "Native AOT Form Factors": discusses "statically linked binaries with minimal dependencies", extreme size/startup/performance requirements, and CoreRT.

It's not really clear what the main distinction is between these sections. My guess is that the first section is intended to describe AOT modes of officially supported .Net platforms, and the second section is intended to describe AOT addressing more extreme requirements. Similar high-level considerations apply to both (linkers, analyzers, reflection, codegen...).

I think a bit of moving around would help.

I would move these next to each other as they are currently separated by a webassembly section.

Perhaps the material should be:

- "Ahead-of-time (AOT) Compilation": source generators plus put the static linking stuff in here, and move in here the general advantages of AOT (startup, platform requirements, consistent performance).

- "Strict AOT": Describe extreme requirements, how they may only be able to be met by an AOT-focused runtime rather than an AOT mode of CoreCLR or Mono. (I assume that is the meaning of "native". "native" can easily be confusing as all AOT compiles to something native.)

RUSshy:
Long live to CoreRT!! 

One thing i see mentioned about reflection is just: [proposals for faster reflection APIs](https://github.com/dotnet/runtime/issues/23716)

Will reflection free mode stay?

What about compile time reflection?

I can see why Source generators has been proposed, but it's not solving that introspection issue

https://github.com/dotnet/csharplang/issues/2939

I think having something like D's trait is valuable and will benefit not only CoreRT but everything else, including mono and Xamarin

No need extra work for creating a source generator, instead just query for stuff you need to be done at compile time

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,206 @@
...
+
+#### Libraries Optimized for Size
+
+The .NET runtime libraries have been often fine-tuned for raw speed of server workloads so far, without regards of code size impact. For example, the implementation of UTF-8 encoding grew to thousands of lines through vectorization and other performance optimizations, even though the algorithm can be expressed in a few hundred lines of code.
+
+Going forward, we will consider binary size and speed to be of equal importance. The libraries implementation will be tailored for aggressive trimming. We will create tooling for developers contributing to .NET libraries that will calculate and report the binary size impacts for a given PR. The tooling will not only track basic binary size regression but also additional checks like the introduction of cyclic dependencies. Where warranted, we will introduce separate implementations of key algorithms optimized for size and speed that the specific form factors include as appropriate.
```
dquist, line 110:
```suggestion
Going forward, we will consider binary size and speed to be of equal importance. The libraries' implementation will be tailored for aggressive trimming. We will create tooling for developers contributing to .NET libraries that will calculate and report the binary size impacts for a given PR. The tooling will not only track basic binary size regression but also additional checks like the introduction of cyclic dependencies. Where warranted, we will introduce separate implementations of key algorithms optimized for size and speed that the specific form factors include as appropriate.
```

Added a possessive apostrophe on libraries.

KexyBiscuit:
Not sure if it fits here, but it's sad to see https://github.com/NETMF is gone. Will .NET picks this area up in the future?

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,206 @@
...
+
+We will not attempt to re-create application virtualization solutions such as [BoxedApp](https://www.boxedapp.com/docs/index.html) or (`App-V`)[https://docs.microsoft.com/en-us/windows/application-management/app-v/appv-for-windows] that can make unaware applications deployable as single-file executables by virtualizing the operating system. Application virtualization solutions are a unique domain of sophisticated technology, outside the realm of .NET runtime.
+
+Single-file compatibility analyzers will detect the use of APIs that are known to cause single file deployments to break, such as `Assembly.LoadFile(path)` or `Assembly.Location`. New APIs, such as [API that returns path to the launching .exe path](https://github.com/dotnet/runtime/issues/30448), will be introduced to simplify mitigation of these issues.
+
+Managed C++ is incompatible with the single file form factor and addressing this incompatibility in the tooling is prohibitively expensive. Projects written in managed C++ (such as part of WPF) would have to be rewritten in C# or other managed language to be single-file compatible.
```
nietras, line 143: @jkotas could you link to or expand on why this is the case?

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,206 @@
...
+
+The following feedback outlines customer-observed problems with the *single-file form* factor:
+
+- [SingleFilePublish is quite unusable](https://github.com/dotnet/corert/issues/7200#issuecomment-564778685)
+- [Current executing assembly location is wrong for single file published project](https://github.com/dotnet/runtime/issues/13531)
+- [PublishSingleFile with Windows Service doesn't work with config-file](https://github.com/dotnet/runtime/issues/32735)
```
nietras, line 54: The above issues are very important and why we can't use single file/self-contained. Being able to define the extraction location yourself (relative to single file) would help a lot and then use SetDllDirectory (on windows) instead of changing current directory would perhaps be simple changes that could help previous dotnet framework devs.

We do single file publish to some degree by simply embedding all managed assemblies into exe, and copy native dlls to e. g. `x64` sub-directories. This keeps the root clean (next to exe). Hence, much better with regard to perception.

accepted/2020/form-factors.md
```diff
@@ -0,0 +1,206 @@
...
+
+We will introduce a linker compatibility analyzer to detect code patterns that cannot be reliably analyzed by the linker. This will still be needed for libraries that do not or cannot adopt the source generation technique.
+
+### Single File
+
+The single-file form factor bundles all code into a single binary that the code is directly executed from. Single-file apps are easier to deploy and distribute. Many users have asked for this form factor. We will focus on producing traditional immutable single-file executables, and provide guidelines for application developers to target to successfully target the single-file form factor.
+
+We will not attempt to re-create application virtualization solutions such as [BoxedApp](https://www.boxedapp.com/docs/index.html) or (`App-V`)[https://docs.microsoft.com/en-us/windows/application-management/app-v/appv-for-windows] that can make unaware applications deployable as single-file executables by virtualizing the operating system. Application virtualization solutions are a unique domain of sophisticated technology, outside the realm of .NET runtime.
```
AustinWise, line 139: The App-V link currently renders like this:

(`App-V`)[https://docs.microsoft.com/en-us/windows/application-management/app-v/appv-for-windows]

The square braces and parens should be swapped:

```
[`App-V`](https://docs.microsoft.com/en-us/windows/application-management/app-v/appv-for-windows)
```

Which will render like this:

[`App-V`](https://docs.microsoft.com/en-us/windows/application-management/app-v/appv-for-windows)

accepted/2020/form-factors.md
jkotas: You are right that WPF does not use build time source generation for data bindings. WPF uses the build time source generation for other reason.

jkotas:
@charlesroddie 

> I would move these next to each other as they are currently separated by a webassembly section.

Thanks for the feedback! I have moved the two sections together, and also changed the content a little bit.

> Strict AOT

I would like to keep "native" in the name since the same adjective is used in comparable technologies (e.g. Kotlin/Native, dart2native, GraalVM Native Image).

@RUSshy

> Will reflection free mode stay?

Yes. We may experiment with making more of cheap reflection APIs work in this mode, so that e.g. [networking](https://github.com/dotnet/runtime/issues/33127) works.

> What about compile time reflection?

I expect that we will keep innovating in this space. I do not see a clear path for including compile time reflection into the C# language today. The issues in https://github.com/dotnet/csharplang/ are the best place to discuss this further.

@KexyBiscuit

> https://github.com/NETMF is gone. Will .NET picks this area up in the future?

[Wilderness Labs](https://www.wildernesslabs.co/) has kept .NET for microcontrollers going. The Community Supported Form Factors chapter from this doc sets the framework for how we can work together better.

accepted/2020/form-factors.md
jkotas: Managed C++ binaries contain both managed and native code together. We are not able to load them from memory like pure managed assemblies. They must be handled same way as native libraries and either be in its own file on disk or statically linked into the single .exe. Statically linking them into the single .exe is difficult. It would require special support in Visual Studio linker.

nietras: @jkotas this implies all Managed C++ assemblies are mixed mode assemblies does it not? This was what threw me off, it would be clearer if this was specifically mentioned, if that is pure managed c++ assemblies are not an issue :)

jkotas: `-clr:pure` mode of managed C++ is not supported for .NET Core. This is documented at https://docs.microsoft.com/en-us/dotnet/core/porting/cpp-cli#ccli-net-core-limitations .

I agree with you that `-clr:pure` would work better for this. This comes back to costs `-clr:pure` is used very rarely and it is much more expensive for the C++ team to maintain than `-clr`. It is the reason behind this.

nietras: @jkotas I have many questions 😅 I understand that pure was dropped, but you can still have pure managed c++ assemblies but built for Framework, which however still work in Core. I don't understand the exact requirements for that to be the case, but assumed this was the case for pure managed c++ assemblies, is that true? 

We have a lot of third-party dependencies (from hardware vendors) that are c++/cli only if these work on Core can we use Core. So just trying to wrap my head around the conditions and limitations here. 😊

jkotas: Yes, it will likely work. https://docs.microsoft.com/en-us/dotnet/core/porting/third-party-deps#net-framework-compatibility-mode talks about it.

PR #120, opened 2020-04-30
Add spec for workload manifests

mhutch:
Add spec for workload manifests, expanding on the concepts from the [workload overview](https://github.com/dotnet/designs/blob/master/accepted/2020/workloads/workloads.md).

### DIFF
--- a/accepted/2020/workloads/workload-manifest.md
+++ b/accepted/2020/workloads/workload-manifest.md
@@ -81,7 +81,7 @@ Template packs are NuGet packages to be used by the dotnet templating engine. Wh
 
 ## Tools Packs
 
-Tools packs are .NET global tools packages. Tools packs are installed to `dotnet/tools-packs/{package-id}/{package-version}/` and the latest version of each installed tool made available in the `PATH`. When invoking a tools via `dotnet` e.g. invoking `dotnet-foo` as `dotnet foo`, the runtime will use global.json and the manifest to determine the appropriate version to use, so it is recommended to use the prefix on tools that are shipped in tools packs.
+Tools packs are .NET global tools packages. Tools packs are installed to `dotnet/tools-packs/{package-id}/{package-version}/` and can be invoked via `dotnet` e.g. invoking `dotnet foo`, the runtime will use global.json and the manifest to determine the appropriate version of `dotnet-foo` to run.
 
 Tools packs are intended to be used for tools that are exposed part of the development experience. Tools that are only expected to be invoked by build targets should be distributed in an SDK pack.
 
@@ -99,7 +99,7 @@ Each manifest may have a `WorkloadManifest.targets` MSBuild file beside it that 
 
 ## Packaging
 
-Manifests are packaged and distributed as NuGet packages. They will use a new package type called `SdkManifest` so they cannot be accidentally used as a `PackageReference`. The manifest’s `sdk-manifest.json` and `WorkloadTargets.targets`  are in the root folder of the NuGet. The ID of the NuGet is `{manifest-id}-{sdk-band}` and the version of the NuGet is the manifest version.
+Manifests are packaged and distributed as NuGet packages. They will use the existing package type called `DotnetPlatform` so they cannot be accidentally used as a `PackageReference`. The manifest’s `sdk-manifest.json` and `WorkloadTargets.targets`  are in the root folder of the NuGet. The ID of the NuGet is `{manifest-id}-{sdk-band}` and the version of the NuGet is the manifest version.
 
 As manifests are distributed as NuGets, it’s possible to use preview versions of workloads by subscribing to preview feeds.
 
@@ -109,7 +109,7 @@ Manifests are installed to `dotnet/sdk/{sdk-band}-manifests/{manifest-id}/` in a
 
 The dotnet SDK would be expected to include baseline versions of the installed manifests for all manifests that are known and supported. Although any components listed in this baseline manifest need to work with that SDK version, it does not need to be fully up-to-date: it exists primarily for listing workloads that are available to be installed, and for its ID to be known by the SDK when fetching updated versions from NuGet.
 
-Any workload management operation (install, uninstall, repair update) must be an transactional operation that also updates the installed manifests to the latest available versions for that SDK band.
+Any workload management operation (install, uninstall, repair update) must be a transactional operation that also updates the installed manifests to the latest available versions for that SDK band.
 
 ## Advertising
 
@@ -117,7 +117,9 @@ The .NET tooling will automatically and opportunistically download updated versi
 
 # Format
 
-Manifests are json files. The toplevel is a JSON object, containing the following keys:
+Manifests are json files. Comments are supported, both `//` and `/* */` styles.
+
+The toplevel is a JSON object, containing the following keys:
 
 | Key | Type | Value | Required |
 |--|--|--|--|
@@ -196,7 +198,7 @@ A workload or framework reference that referenced `"foo.framework"` would get `"
 
 ## Example Platform SDK Manifest
 
-Here is a *hypothetical* example manifest. It's not prescriptive but demonstrates concepts an patterns that can be used by platform implementations. It is annotated with comments, although **comments are not permitted in real manifests**.
+Here is a *hypothetical* example manifest. It's not prescriptive but demonstrates concepts an patterns that can be used by platform implementations.
 
 ```json5
 {
@@ -365,7 +367,7 @@ Here is an manifest snippet. It only uses a single pack, but concrete examples w
         ]
     },
     "ios": {
-        "description": "Create, build and run iOS apps ",
+        "description": "Create, build and run iOS apps",
         "packs": [
             "Xamarin.iOS.Framework",
             "Xamarin.iOS.Templates",
@@ -394,6 +396,8 @@ Here is an manifest snippet. It only uses a single pack, but concrete examples w
 
 The only purpose of these side-by-side versions is to make it so that servicing updates to an SDK series do not unexpectedly bring in API changes. They are not intended for long term use. **When a new SDK series is released, the versioned workloads and their packs should be removed from the manifest**.
 
+As demonstrated in this example, template packs should not be side-by-side versioned within an SDK band. Templates should always target the newest framework, automatically opting newly created projects into the newer of the side-by-side versions of the workload. If there is a developer scenario where specifying the framework is important then it should be exposed as a parameter on the templates rather than a separate template.
+
 ## Expected SDK Resolution Behavior
 
 The MSBuild logic in the `WorkloadManifest.targets` that imports SDKs packs based on the `TargetPlatformVersion` MSBuild property should use it as a **lower bound** on the pack OS API version, and implement the following behaviors:
@@ -418,7 +422,6 @@ We will need to develop a standard pattern or helper logic for this to make sure
 An alternate option that was considered was to make SDK feature releases when platform API updates are needed. For example, if a new version of iOS was released after 5.0.100, we could make a 5.0.200 SDK release with the updated platform API.
 
 This has the advantage that it does not introduce any new mechanisms. However, it has several major disadvantages:
-
 
 - Platform API updates require making an SDK series release, which could be expensive
 - No compatibility checks and error experience when building project that uses newer APIs on SDK series that only has older APIs
--- a/accepted/2020/workloads/workload-manifest.md
+++ b/accepted/2020/workloads/workload-manifest.md
@@ -1,0 +1,427 @@
+# .NET SDK Workload Manifests
+
+# Overview
+
+.NET SDK workload manifests describe workloads that are available to be installed in the .NET SDK.  They provide version mappings that allow MSBuild targets and other SDK tools to resolve the packs that comprise these workloads from the SDK installation directory, and can be used to acquire, repair, and update the workloads and their components.
+
+This document describes the format and versioning of the manifest. The installation and update experience will be described in other documents.
+
+# Definitions
+## SDK Band
+
+.NET SDK versions take the form 3.1.100, 3.1.101, 3.1.200. The first two components of this version represent the major and minor version of the runtime in the SDK. The third component (the “hundreds”) is a more complex: the first digit represents the feature release of the SDK, and the last two digits represent the patch level of the SDK and runtime combined.
+
+In the context of this document we will refer to an *SDK band* as the common portion of the version of all the SDK releases that differ only by patch level. For example, 3.1.100 and 3.1.105 are in the 3.1.100 band, while 3.2.100 is in the 3.2.100 band and 3.1.203 is in the 3.1.200 band.
+
+## SDK Pack
+
+An SDK pack is a unit of packaging for the assets that make up the SDK. Packs have an ID and version and are installed into folders in the SDK directory.
+
+Packs will be installed using CLI tooling or by a native installer such as the Visual Studio installer. It is likely that packs will be distributed via NuGet at some point, so it is recommended that pack authors reserve NuGet IDs for their packs even if other distribution methods are used initially.
+
+SDK packs are used for construction, composition and distribution only, they are not exposed through the user experience.
+
+## SDK Workload
+
+An SDK workload is a semantic name describing a set of one or more SDK packs. When a developer wants to be able to build a particular kind of app, they would expect to install a workload. Workloads should correspond to user scenarios.
+
+Here are some examples of SDK workloads and the packs that they might include:
+
+| Workload ID | Content|
+|--|--|
+| xamarin-android | Framework, runtime, build logic, tools and templates for Xamarin.Android.|
+| xamarin-forms | Templates for Xamarin.Forms and the Xamarin.Forms NuGet package used by the templates.<br><br>This would also include any packs necessary to build and run Xamarin.Forms projects. That could be done by depending on the xamarin-ios, xamarin-android and uwp components, or perhaps instead some kind of lightweight Xamarin.Forms host. |
+| xamarin-forms-android | Templates for Xamarin.Forms and the Xamarin.Forms NuGet package used by the templates.<br><br>This would also include any packs necessary to build and run Xamarin.Forms projects on Android by depending on xamarin-android.|
+| xamarin-android-llvm | The LLVM AOT compiler for Android, omitted by default from the xamarin-android component because it’s large and not all apps use it.<br><br>It may make sense for this to depend on the xamarin-android workload as it’s useless by itself.|
+
+# Versioning
+
+## Manifest Versioning
+
+Manifests follows a similar pattern to SDK versioning: each SDK band has a corresponding manifest band. Each SDK uses only the manifest band corresponding to its own band. Manifest versions however are completely independent of SDK patch levels; an SDK should use the latest available versions of the manifests for its band.
+
+To make this relationship more explicit, the SDK band for a manifest is encoded into the manifest’s ID e.g. `Microsoft.NET.Sdk.Android.SdkManifest-5.0.200`. The “version” of a manifest package can then be then simply a monotonic integer value.
+
+## Pack Versioning
+
+A manifest contains a list of pack IDs and their corresponding versions. A single version is defined for each pack: the most up-to-date version of that pack that is expected to be used with that SDK band.
+
+This intentionally separates pack versions from component definitions. A pack may be included in multiple workloads, and this separation ensures that all workloads get a consistent version of each pack. There’s also a semantic separation: the workload definition list describes which packs are included in each workload, while the pack version list describes the set of pack versions that are supported together.
+
+There are some cases where a single workload may want to contain side-by-side multiple versions of a single pack, or multiple components may want different versions of a single pack. This may apply, for example, to major/minor runtime versions, For these cases an aliasing mechanism is provided: a meaningful alias may be defined and mapped to a different version of a pack that is already included in the list. For example, `Foo-1.x` could be defined as an alias to version `1.0.5` of pack `Foo`, while `Foo` itself could be mapped to `2.1.4`. This is similar to how Linux libraries support side-by-side.
+
+# Pack Kinds
+
+The following kind of packs are permitted in manifests: targeting and runtime packs, SDK packs, library packs, template packs, and tools packs.
+
+All packs are expected to be installed into the `dotnet` folder.
+
+## Targeting and Runtime Packs
+
+The [targeting and runtime packs](https://github.com/dotnet/designs/pull/50) introduced in .NET Core 3.0 are permitted in workloads. Targeting packs contain the framework assemblies that a project compiles against at build time, and runtime packs contain the framework and runtime binaries used to execute the project. They install into `dotnet/packs/{pack-id}/{pack-version}` and are resolved at build time using `FrameworkReference` MSBuild items.
+
+Framework references are expected to be added to a project automatically by the manifest’s MSBuild targets based on the TFM. There will be a mechanism that allows the targets to omit the pack version from the framework reference. Targets may add multiple targeting or runtime pack references, for example when building a `net5.0-ios` project they could add a reference to a pack containing the `net5.0` reference assemblies and a pack containing the reference assemblies for the iOS platform bindings.
+
+## SDK Packs
+
+SDK packs are MSBuild SDK NuGet packages, but they are installed to `dotnet/packs/{pack-id}/{pack-version}`. They are expected to be automatically referenced by the targets.
+
+## Library Packs
+
+Library packs are normal NuGet packages. When a library pack is installed, the nupkg package file is placed in `dotnet/library-packs/` but are *not* extracted. This location is used as a directory feed by NuGet.
+
+If ones of these packages gets used, it will be extracted into the global packages folder. If the SDK is updated and the library pack is removed or replaced with a newer version, projects that have already been created and restored will continue to be able to use the extracted copy.
+
+This is intended to be used to pre-download NuGet packages referenced by templates so that a workload can work offline after installation. It should be used sparingly i.e. only for NuGets referenced by the core templates with default or popular options.
+
+## Template Packs
+
+Template packs are NuGet packages to be used by the dotnet templating engine. When a template pack is installed installed, the nupkg package file is placed in `dotnet/template-packs/` but is *not* extracted. Visual Studio and `dotnet new` are expected to automatically detect when templates nupkgs are installed and uninstalled to this folder and update their template hives automatically.
+
+
+## Tools Packs
+
+Tools packs are .NET global tools packages. Tools packs are installed to `dotnet/tools-packs/{package-id}/{package-version}/` and the latest version of each installed tool made available in the `PATH`. When invoking a tools via `dotnet` e.g. invoking `dotnet-foo` as `dotnet foo`, the runtime will use global.json and the manifest to determine the appropriate version to use, so it is recommended to use the prefix on tools that are shipped in tools packs.
+
+Tools packs are intended to be used for tools that are exposed part of the development experience. Tools that are only expected to be invoked by build targets should be distributed in an SDK pack.
+
+# Manifest Files
+
+## Composition
+
+Multiple manifests may be present in a .NET SDK band. Each manifest has its own ID and they are versioned independently. Components and packs may not be duplicated in multiple manifests within a band, though manifests may reference components and packs in other manifests that are expected to be installed.
+
+Manifests are not an arbitrary extensibility mechanism; they are intended to provide some flexibility in the distribution and composition of a coherent whole. For example, there could be a *dotnet* manifest used to update in-box shared runtimes, a *xamarin-ios* manifest used to describe and update the Xamarin.iOS optional components, and a *uwp* manifest used to describe and update the UWP optional components. These could all be produced from separate repositories using separate build and release processes.
+
+At any given point in time, the latest available version of all the manifests for an SDK series must always be able to be combined into a single consistent whole (in memory, not persisted to disk).
+
+Each manifest may have a `WorkloadManifest.targets` MSBuild file beside it that is unconditionally imported into all projects. This targets file contains conditioned imports to add automatic referenced to SDK packs based on project properties, and its usage and recommended patterns are described in more detail in the [workload resolvers spec](/workload-resolvers.md).
+
+## Packaging
+
+Manifests are packaged and distributed as NuGet packages. They will use a new package type called `SdkManifest` so they cannot be accidentally used as a `PackageReference`. The manifest’s `sdk-manifest.json` and `WorkloadTargets.targets`  are in the root folder of the NuGet. The ID of the NuGet is `{manifest-id}-{sdk-band}` and the version of the NuGet is the manifest version.
+
+As manifests are distributed as NuGets, it’s possible to use preview versions of workloads by subscribing to preview feeds.
+
+## Installation
+
+Manifests are installed to `dotnet/sdk/{sdk-band}-manifests/{manifest-id}/` in an unpacked form. These are called the *installed manifests*.
+
+The dotnet SDK would be expected to include baseline versions of the installed manifests for all manifests that are known and supported. Although any components listed in this baseline manifest need to work with that SDK version, it does not need to be fully up-to-date: it exists primarily for listing workloads that are available to be installed, and for its ID to be known by the SDK when fetching updated versions from NuGet.
+
+Any workload management operation (install, uninstall, repair update) must be an transactional operation that also updates the installed manifests to the latest available versions for that SDK band.
+
+## Advertising
+
+The .NET tooling will automatically and opportunistically download updated versions of all manifests for the current SDK band and unpack them to `~/.dotnet/sdk-advertising/{sdk-band}/{manifest-id}/`. These user-local updated copies of the manifest are known as *advertising manifests.* The advertising manifests are used when listing workloads that are available to be installed, and for the build tooling to produce warnings that newer versions of installed components are available*.* They are **not** used in pack resolution.
+
+# Format
+
+Manifests are json files. The toplevel is a JSON object, containing the following keys:
+
+| Key | Type | Value | Required |
+|--|--|--|--|
+| `version` | int | The version of the manifest. Must match the version of the NuGet package that contains the manifest. | Yes |
+| `description` | string | Description of the content and/or purpose of the manifest. This is primarily for commenting and/or diagnostic purposes and is not expected to be surfaced in the UX. | No |
+| `workloads` | object | Workload definitions keyed by workload ID. | No |
+| `packs` | object | Pack definitions keyed by pack ID. | No |
+| `data` | object | Allows manifests to include arbitrary data without risk of conflict. | No |
+
+## Workload Definitions
+
+Workload definitions take the following form:
+
+| Key | Type | Value | Required |
+|--|--|--|--|
+| `abstract` | bool | If `true`, this workload can only be extended, and is never exposed directly as an installable workload. Default is `false`. | No |
+| `kind` | string | Either `build` or `dev`. Default is `dev`. | No |
+| `description` | int | User-visible description for the workload. | Yes if dev and non-abstract |
+| `packs` | string array | IDs of the packs that are included in the workload. | No |
+| `extends` | string array | IDs of workloads whose packs should be included in this workload. | No |
+| `platforms` | string array | Limits the workload and workloads that extend it to only be shown and installed on these host platforms. The strings are RIDs. | False |
+
+At least one of `extends` or `packs` is required. If a workload resolves to zero packs, which is possible when some packs are platform-specific, it is implicitly abstract. A workload may transitively include the same pack multiple times or extend the same workload multiple times, and they will be deduplicated. As a consequence, recursive `extends` references are technically permitted but redundant and although they may result in validation warnings they will not result in runtime errors.
+
+The `kind` allows structuring workloads into smaller pieces so that their download and install footprint on CI is smaller. `build` workloads should contain only the packs that are used to build projects. They do not need descriptions as they are not expected to be shown in the UX - they will only be used via a CI-specific UX such as `dotnet workload restore --build-only`.
+
+Note that `extends` is functionally a dependency system and a way to factor out common sets of packages from workloads. By analogy to package managers such as apt-get and NuGet, workloads are metapackages that only permit unversioned dependencies and packs are packages that are only installable transitively.
+
+## Pack Definitions
+
+Pack definitions take the following form:
+
+| Key | Type | Value | Required |
+|--|--|--|--|
+| `kind` | string | Type of the pack. Valid values are `sdk`, `framework`, `library`, `template`, `tool`. | Yes |
+| `version` | string | Version of the pack. | Yes |
+| `alias-to` | object | Alias this definition to the provided package ID | No |
+
+The `alias-to` key allows aliasing this pack definition to a different underlying pack ID on each host platform. The keys are host platform RIDs, and the values are pack IDs. The `"*"` platform ID may be used to alias a pack on all platforms.
+
+For example, an SDK pack might contain a compiler that's a native executable and hence depends on the host platform:
+
+```json
+"foo.sdk.compiler": {
+    "version": "1.0",
+    "kind": "sdk",
+    "alias-to": {
+        "osx-x64": "foo.sdk.compiler.mac",
+        "win-x64": "foo.sdk.compiler.windows"
+    }
+}
+```
+
+Workload definitions and SDK imports would refer to this as `"foo.sdk.compiler"`, but on Windows the pack that gets installed and resolved would be `"foo.sdk.compiler.windows"`. The `"foo.sdk.compiler.windows"` pack would be present on disk and its ID would be visible in resolved paths, but otherwise transparent and not able to be used directly.
+
+Another use of aliases is to install multiple version bands of the same pack by synthesizing new versioned pack IDs:
+
+```json
+"foo.framework": {
+    "version": "2.0.4",
+    "kind": "framework",
+    "alias-to": {
+        "*": "foo.framework",
+    }
+},
+"foo.framework.1": {
+    "version": "1.3.2",
+    "kind": "framework",
+    "alias-to": {
+        "*": "foo.framework",
+    }
+},
+```
+
+A workload or framework reference that referenced `"foo.framework"` would get `"foo.framework"` version `2.0.4`, while a workload or framework reference that referenced `"foo.framework.1"` would get `"foo.framework"` version `1.3.2`,
+
+## Example Platform SDK Manifest
+
+Here is a *hypothetical* example manifest. It's not prescriptive but demonstrates concepts an patterns that can be used by platform implementations. It is annotated with comments, although **comments are not permitted in real manifests**.
+
+```json
+{
+    "version": 5,
+    "workloads": {
+        // this is a dev workload that would typically be installed by a developer getting
+        // started with this platform. it's composed of several smaller build workloads
+        // plus a template pack for creating projects
+        "xamarin-android": {
+            "description": "Create, build and run Android apps",
+            "kind": "dev",
+            "packs": [
+                "Xamarin.Android.Templates",
+            ],
+            // on dev machines we expect to pre-install support
+            // for common device architectures
+            "extends": [
+                "xamarin-android-build",
+                "xamarin-android-build-armv7a",
+                "xamarin-android-build-x86"
+            ],
+        },
+        "xamarin-android-build": {
+            "description": "Build and run Android apps",
+            "packs": [
+                "Xamarin.Android.Sdk",
+                "Xamarin.Android.BuildTools",
+                "Xamarin.Android.Framework",
+                "Xamarin.Android.Runtime",
+                "Mono.Android.Sdk"
+            ]
+        },
+        // on CI machines, this will only be installed if
+        // the app actually targets the armv7a architecture
+        "xamarin-android-build-armv7a": {
+            "kind": "build",
+            "packs": [
+                "Mono.Android.Runtime.Armv7a",
+            ],
+            // the dependency is likely redundant in practice as any
+            // workload restore that resolves xamarin-android-build-armv7a
+            // will also resolve xamarin-android-build, but let's be explicit
+            "extends": [ "xamarin-android-build" ],
+        },
+        // on CI machines, this will only be installed if
+        // the app actually targets the x86 architecture
+        "xamarin-android-build-x86": {
+            "kind": "build",
+            "packs": [
+                "Mono.Android.Runtime.x86",
+            ],
+            "extends": [ "xamarin-android-build" ],
+        },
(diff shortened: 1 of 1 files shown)


accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+# Overview
+
+.NET SDK workload manifests describe workloads that are available to be installed in the .NET SDK.  They provide version mappings that allow MSBuild targets and other SDK tools to resolve the packs that comprise these workloads from the SDK installation directory, and can be used to acquire, repair, and update the workloads and their components.
+
+This document describes the format and versioning of the manifest. The installation and update experience will be described in other documents.
+
+# Definitions
+## SDK Band
+
+.NET SDK versions take the form 3.1.100, 3.1.101, 3.1.200. The first two components of this version represent the major and minor version of the runtime in the SDK. The third component (the “hundreds”) is a more complex: the first digit represents the feature release of the SDK, and the last two digits represent the patch level of the SDK and runtime combined.
```
richlander, line 12:
```suggestion
.NET SDK versions take the form 3.1.100, 3.1.101, 3.1.200. The first two components of this version represent the major and minor version of the runtime in the SDK. The third component (the “hundreds”) is more complex: the first digit represents the feature release of the SDK, and the last two digits represent the patch level of the SDK and runtime combined.
```

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+In the context of this document we will refer to an *SDK band* as the common portion of the version of all the SDK releases that differ only by patch level. For example, 3.1.100 and 3.1.105 are in the 3.1.100 band, while 3.2.100 is in the 3.2.100 band and 3.1.203 is in the 3.1.200 band.
+
+## SDK Pack
+
+An SDK pack is a unit of packaging for the assets that make up the SDK. Packs have an ID and version and are installed into folders in the SDK directory.
+
+Packs will be installed using CLI tooling or by a native installer such as the Visual Studio installer. It is likely that packs will be distributed via NuGet at some point, so it is recommended that pack authors reserve NuGet IDs for their packs even if other distribution methods are used initially.
```
richlander, line 20: > it is recommended that pack authors reserve NuGet IDs

Makes sense in the broad. Can't the SDK team just reserve the right top-level names, and call it good? We should be like a TLD owner.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+| xamarin-forms-android | Templates for Xamarin.Forms and the Xamarin.Forms NuGet package used by the templates.<br><br>This would also include any packs necessary to build and run Xamarin.Forms projects on Android by depending on xamarin-android.|
+| xamarin-android-llvm | The LLVM AOT compiler for Android, omitted by default from the xamarin-android component because it’s large and not all apps use it.<br><br>It may make sense for this to depend on the xamarin-android workload as it’s useless by itself.|
+
+# Versioning
+
+## Manifest Versioning
+
+Manifests follows a similar pattern to SDK versioning: each SDK band has a corresponding manifest band. Each SDK uses only the manifest band corresponding to its own band. Manifest versions however are completely independent of SDK patch levels; an SDK should use the latest available versions of the manifests for its band.
```
richlander, line 41: > each SDK band has a corresponding manifest band

The introduction of a manifest band adds too much cognitive load. 

> an SDK should use the latest available versions of the manifests for its band

This is clear. My take is that the SDK always uses the latest manifest for sdk band. It's possible there will be an advanced scenario to do something different (would be quite useful for testing).

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+To make this relationship more explicit, the SDK band for a manifest is encoded into the manifest’s ID e.g. `Microsoft.NET.Sdk.Android.SdkManifest-5.0.200`. The “version” of a manifest package can then be then simply a monotonic integer value.
+
+## Pack Versioning
+
+A manifest contains a list of pack IDs and their corresponding versions. A single version is defined for each pack: the most up-to-date version of that pack that is expected to be used with that SDK band.
```
richlander, line 47: > A single version is defined for each pack: the most up-to-date version of that pack that is expected to be used with that SDK band.

Unlikely. What's your plan for supporting the 5.0 and 6.0 SDKs (or 6.0 and 7.0) concurrently, as it relates to the packs that are referenced?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+This intentionally separates pack versions from component definitions. A pack may be included in multiple workloads, and this separation ensures that all workloads get a consistent version of each pack. There’s also a semantic separation: the workload definition list describes which packs are included in each workload, while the pack version list describes the set of pack versions that are supported together.
+
+There are some cases where a single workload may want to contain side-by-side multiple versions of a single pack, or multiple components may want different versions of a single pack. This may apply, for example, to major/minor runtime versions, For these cases an aliasing mechanism is provided: a meaningful alias may be defined and mapped to a different version of a pack that is already included in the list. For example, `Foo-1.x` could be defined as an alias to version `1.0.5` of pack `Foo`, while `Foo` itself could be mapped to `2.1.4`. This is similar to how Linux libraries support side-by-side.
```
richlander, line 51: It would be good to have an example to help motivate this additional complexity.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+All packs are expected to be installed into the `dotnet` folder.
+
+## Targeting and Runtime Packs
+
+The [targeting and runtime packs](https://github.com/dotnet/designs/pull/50) introduced in .NET Core 3.0 are permitted in workloads. Targeting packs contain the framework assemblies that a project compiles against at build time, and runtime packs contain the framework and runtime binaries used to execute the project. They install into `dotnet/packs/{pack-id}/{pack-version}` and are resolved at build time using `FrameworkReference` MSBuild items.
+
+Framework references are expected to be added to a project automatically by the manifest’s MSBuild targets based on the TFM. There will be a mechanism that allows the targets to omit the pack version from the framework reference. Targets may add multiple targeting or runtime pack references, for example when building a `net5.0-ios` project they could add a reference to a pack containing the `net5.0` reference assemblies and a pack containing the reference assemblies for the iOS platform bindings.
```
richlander, line 63: Maybe a little in the weeds, but it might be helpful to explain that some fx references may be handled by additional mechanisms. Here, I'm thinking of `UseWinforms` and `UseWPF`.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+SDK packs are MSBuild SDK NuGet packages, but they are installed to `dotnet/packs/{pack-id}/{pack-version}`. They are expected to be automatically referenced by the targets.
+
+## Library Packs
+
+Library packs are normal NuGet packages. When a library pack is installed, the nupkg package file is placed in `dotnet/library-packs/` but are *not* extracted. This location is used as a directory feed by NuGet.
+
+If ones of these packages gets used, it will be extracted into the global packages folder. If the SDK is updated and the library pack is removed or replaced with a newer version, projects that have already been created and restored will continue to be able to use the extracted copy.
+
+This is intended to be used to pre-download NuGet packages referenced by templates so that a workload can work offline after installation. It should be used sparingly i.e. only for NuGets referenced by the core templates with default or popular options.
```
richlander, line 75: Will this be the `NuGetFallback` folder that we used in 2.x or something different?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Template Packs
+
+Template packs are NuGet packages to be used by the dotnet templating engine. When a template pack is installed installed, the nupkg package file is placed in `dotnet/template-packs/` but is *not* extracted. Visual Studio and `dotnet new` are expected to automatically detect when templates nupkgs are installed and uninstalled to this folder and update their template hives automatically.
```
richlander, line 79: Open: Should template packs be special cased, relative to library packs? Maybe we just need a registry of template packs.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+
+## Tools Packs
+
+Tools packs are .NET global tools packages. Tools packs are installed to `dotnet/tools-packs/{package-id}/{package-version}/` and can be invoked via `dotnet` e.g. invoking `dotnet foo`, the runtime will use global.json and the manifest to determine the appropriate version of `dotnet-foo` to run.
```
richlander, line 84: Any thoughts on tool discovery at runtime?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+Tools packs are intended to be used for tools that are exposed part of the development experience. Tools that are only expected to be invoked by build targets should be distributed in an SDK pack.
+
+# Manifest Files
+
+## Composition
+
+Multiple manifests may be present in a .NET SDK band. Each manifest has its own ID and they are versioned independently. Components and packs may not be duplicated in multiple manifests within a band, though manifests may reference components and packs in other manifests that are expected to be installed.
```
richlander, line 92: > Multiple manifests may be present in a .NET SDK band

Makes sense, however, this is a bit ambiguous. This is the wording that would make more sense to me: "Multiple workloads might be installed for a given SDK band, which will result in a multiple manifests being installed."

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+Manifests are not an arbitrary extensibility mechanism; they are intended to provide some flexibility in the distribution and composition of a coherent whole. For example, there could be a *dotnet* manifest used to update in-box shared runtimes, a *xamarin-ios* manifest used to describe and update the Xamarin.iOS optional components, and a *uwp* manifest used to describe and update the UWP optional components. These could all be produced from separate repositories using separate build and release processes.
```
richlander, line 94: You almost get there, but it would be good to clarify the time dimension. They can be released at different times w/no coordination.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+At any given point in time, the latest available version of all the manifests for an SDK series must always be able to be combined into a single consistent whole (in memory, not persisted to disk).
```
richlander, line 96: > At any given point in time

I'm not sure I see the importance of this phrase.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+Each manifest may have a `WorkloadManifest.targets` MSBuild file beside it that is unconditionally imported into all projects. This targets file contains conditioned imports to add automatic referenced to SDK packs based on project properties, and its usage and recommended patterns are described in more detail in the [workload resolvers spec](/workload-resolvers.md).
```
richlander, line 98: Just a wording nit, but I think of the "workload having a `WorkloadManifest.targets` MSBuild file beside the manifest"

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Packaging
+
+Manifests are packaged and distributed as NuGet packages. They will use the existing package type called `DotnetPlatform` so they cannot be accidentally used as a `PackageReference`. The manifest’s `sdk-manifest.json` and `WorkloadTargets.targets`  are in the root folder of the NuGet. The ID of the NuGet is `{manifest-id}-{sdk-band}` and the version of the NuGet is the manifest version.
+
+As manifests are distributed as NuGets, it’s possible to use preview versions of workloads by subscribing to preview feeds.
```
richlander, line 104: Good scenario.

How would this work? Would you imagine you install a preview SDK, and it will look for a manifest package for a given workload in that same band (across registered feeds)? What would you expect if it doesn't find one?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Installation
+
+Manifests are installed to `dotnet/sdk/{sdk-band}-manifests/{manifest-id}/` in an unpacked form. These are called the *installed manifests*.
```
richlander, line 108: > unpacked form

Would be good to use the word "extracted" since that's what you use earlier.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+The dotnet SDK would be expected to include baseline versions of the installed manifests for all manifests that are known and supported. Although any components listed in this baseline manifest need to work with that SDK version, it does not need to be fully up-to-date: it exists primarily for listing workloads that are available to be installed, and for its ID to be known by the SDK when fetching updated versions from NuGet.
```
richlander, line 110: I'm wondering if there is value in the manifest being complete? Put another way, I'm wondering if a better model would be having a partial manifest in-box that never/rarely needs to be updated. There is no offline experience that we're trying to maintain.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+Any workload management operation (install, uninstall, repair update) must be a transactional operation that also updates the installed manifests to the latest available versions for that SDK band.
+
+## Advertising
+
+The .NET tooling will automatically and opportunistically download updated versions of all manifests for the current SDK band and unpack them to `~/.dotnet/sdk-advertising/{sdk-band}/{manifest-id}/`. These user-local updated copies of the manifest are known as *advertising manifests.* The advertising manifests are used when listing workloads that are available to be installed, and for the build tooling to produce warnings that newer versions of installed components are available*.* They are **not** used in pack resolution.
+
+# Format
+
+Manifests are json files. Comments are supported, both `//` and `/* */` styles.
```
richlander, line 120: Would be super useful to see the minimally-conformant manifest, according to these format rules. Once I see that, I'll ask what the UX or scenario would be for that minimally conformant manifest.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+| `extends` | string array | IDs of workloads whose packs should be included in this workload. | No |
+| `platforms` | string array | Limits the workload and workloads that extend it to only be shown and installed on these host platforms. The strings are RIDs. | False |
+
+At least one of `extends` or `packs` is required. If a workload resolves to zero packs, which is possible when some packs are platform-specific, it is implicitly abstract. A workload may transitively include the same pack multiple times or extend the same workload multiple times, and they will be deduplicated. As a consequence, recursive `extends` references are technically permitted but redundant and although they may result in validation warnings they will not result in runtime errors.
+
+The `kind` allows structuring workloads into smaller pieces so that their download and install footprint on CI is smaller. `build` workloads should contain only the packs that are used to build projects. They do not need descriptions as they are not expected to be shown in the UX - they will only be used via a CI-specific UX such as `dotnet workload restore --build-only`.
```
richlander, line 147: As you know, I'm a fan of this. I wonder (in the fullness of time) if we need for of these kinds, to enable more scenarios. One thought is to model the kinds on CLI verbs (`build`, `publish`, `test`, and `run`).

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+Note that `extends` is functionally a dependency system and a way to factor out common sets of packages from workloads. By analogy to package managers such as apt-get and NuGet, workloads are metapackages that only permit unversioned dependencies and packs are packages that are only installable transitively.
```
richlander, line 149: Having spent a lot of time with Dockerfiles, I think `extends` will prove to be incredibly liberating.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Pack Definitions
+
+Pack definitions take the following form:
+
+| Key | Type | Value | Required |
+|--|--|--|--|
+| `kind` | string | Type of the pack. Valid values are `sdk`, `framework`, `library`, `template`, `tool`. | Yes |
+| `version` | string | Version of the pack. | Yes |
```
richlander, line 158: Is this different from the `version` in the manifest format, covered earlier? I notice that the workload definition doesn't carry a version. I would expect them to be the same.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+| `alias-to` | object | Alias this definition to the provided package ID | No |
+
+The `alias-to` key allows aliasing this pack definition to a different underlying pack ID on each host platform. The keys are host platform RIDs, and the values are pack IDs. The `"*"` platform ID may be used to alias a pack on all platforms.
```
richlander, line 161: I think this will be incredibly important.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+    "version": "1.0",
+    "kind": "sdk",
+    "alias-to": {
+        "osx-x64": "foo.sdk.compiler.mac",
+        "win-x64": "foo.sdk.compiler.windows"
+    }
+}
+```
+
+Workload definitions and SDK imports would refer to this as `"foo.sdk.compiler"`, but on Windows the pack that gets installed and resolved would be `"foo.sdk.compiler.windows"`. The `"foo.sdk.compiler.windows"` pack would be present on disk and its ID would be visible in resolved paths, but otherwise transparent and not able to be used directly.
```
richlander, line 176: Have you given any thought to the cross-compiler scenario? For example, Windows x64 compiler targeted Windows x64 vs Windows x64 compiler targeting Linux x64 (and all the other possible combinations).

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+Another use of aliases is to install multiple version bands of the same pack by synthesizing new versioned pack IDs:
+
+```json
+"foo.framework": {
```
richlander, line 181: Is there a reason why this first one needs to use an alias since the name matches?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+## Example Platform SDK Manifest
+
+Here is a *hypothetical* example manifest. It's not prescriptive but demonstrates concepts an patterns that can be used by platform implementations.
+
+```json5
+{
+    "version": 5,
+    "workloads": {
+        // this is a dev workload that would typically be installed by a developer getting
+        // started with this platform. it's composed of several smaller build workloads
```
richlander, line 208: This surprised me. Would this workload not be used by someone more experienced?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+                "Xamarin.Android.Sdk",
+                "Xamarin.Android.BuildTools",
+                "Xamarin.Android.Framework",
+                "Xamarin.Android.Runtime",
+                "Mono.Android.Sdk"
+            ]
+        },
+        // on CI machines, this will only be installed if
+        // the app actually targets the armv7a architecture
+        "xamarin-android-build-armv7a": {
```
richlander, line 236: There is clearly a lot of convention at play. That's good, however, it would be good to see that more defined.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
+# .NET SDK Workload Manifests
+
+# Overview
```
richlander, line 3: This is a little off-topic, and I don't have a better place for these feedback, so choosing here.

We've been having some conversations about adopting how to declare workloads that are not explicitly tied to a TFM. ASP.NET Core and Windows Forms are good examples. We've been discussing three models to declare a dependence on such a workload in a project file (using ASP.NET Core as the example):

* `UserAspnetCore`
* `<Workloads>AspnetCore</Workloads>`
* `Sdk="Microsoft.NET.Sdk.Web"`

It is worth clarifying if these are all equivalent from the perspective of tying a project file to a workload, or if there are semantic or technical differences.

If anyone has a preference for the scheme they would like to see used, independent from my question, please feel free to weigh in.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+        }
+    }
+}
+```
+
+# Side-by-Side Platform Updates
+
+In .NET 5, the [TFM is used to specify the .NET version](https://github.com/dotnet/designs/blob/master/accepted/2020/net5/net5.md) (e.g. .NET 5, .NET 6), and may also be used to specify a target platform (e.g. iOS, Android, WinUI) to access platform-specific API. If the TFM specifies a platform, it also implicitly or explicitly specifies a *Platform API Version*. Developers are generally expected to use the latest platform API version, specify a minimum OS version, and [use runtime checks to guard use of APIs from newer OS versions](minimum-os-version/minimum-os-version.md). Templates burn in the most recent OS API version so that new projects start out in the recommended configuration, warning-free and with access to the latest APIs.
+
+It is strongly recommended that updates *within* an SDK feature band only contain low-risk servicing updates, not new features. This means that developers can feel safe installing servicing updates. They know that it's not going to allow them to accidentally make the project depend on APIs that will break it for other members on their team who are not as up to date.
```
richlander, line 339: We should strengthen this. We already have a very strong policy on minimal churn low-risk servicing updates, and it's not up for debate. However, there are some new, very specific, scenarios where we may need to relax this policy. Obviously, those need to be considered with an understanding of risk, too, but I don't think that's a key issue.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+However, there is a major exception to this: OS releases do not coincide with .NET releases, so we need to be able to release platform API updates between .NET releases.
+
+The solution is to ship two versions of the platform APIs, runtime, build logic, and tools, and have the build logic determine which to use [based on the platform API version](workload-resolvers.md#side-by-side-sdk-support-for-new-os-releases). This essentially makes these platform API updates opt-in on the project level: projects can express their dependency on the updated SDK or workload, and build system can then provide better errors when the SDK has not been updated appropriately.
+
+> ***NOTE:*** Projects depending on any new SDK functionality (e.g. those that might appear in 3.1.200 series compared to 3.1.100 series) would have similar issues, and those dependencies can be handled (in a limited fashion) with global.json. However, platform updates are expected to be far more impactful than other changes that might be shipped in SDK updates, may happen between SDK series releases, and can be expressed in a more clear and comprehensible way using the OS platform versions, so we address them specifically.
```
richlander, line 345: It would be good to clarify if you can use g.j to pin to 3.1.105 as a conscious way of avoiding the OS platform update that comes in 3.1.106.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+        "version": "14.2.3",
+        "kind": "framework",
+        "alias-to": {
+            "*": "Xamarin.iOS.Framework"
+        }
+    }
+}
+```
+
+The only purpose of these side-by-side versions is to make it so that servicing updates to an SDK series do not unexpectedly bring in API changes. They are not intended for long term use. **When a new SDK series is released, the versioned workloads and their packs should be removed from the manifest**.
```
richlander, line 397: > SDK series

We should use "series" or "band" consistently.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+> ***NOTE:*** Projects depending on any new SDK functionality (e.g. those that might appear in 3.1.200 series compared to 3.1.100 series) would have similar issues, and those dependencies can be handled (in a limited fashion) with global.json. However, platform updates are expected to be far more impactful than other changes that might be shipped in SDK updates, may happen between SDK series releases, and can be expressed in a more clear and comprehensible way using the OS platform versions, so we address them specifically.
+
+## Side-by-Side Manifest Pattern
+
+The pattern is to duplicate the build workloads and the packs they use, and add versioned suffixes to one of the copies of each, for example `ios-14` and `ios`. Note that duplicate packs may be created using aliases to synthesize packs that don't correspond to real NuGet IDs.
```
richlander, line 349: I thought our planned policy was to introduce 15.0 as opt-in until we moved to a band, at which point 15.0 would be the only option. This seems like 15.0 being the default right away.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+- Project specifies 13.0 API: use 14.0 API, and emit a warning that newer API is being used
+- Project specifies 14.0 API: use 14.0 API, and emit a warning that newer API is available
+- Project specifies 15.0 API: use 15.0 API
+- Project specifies 16.0 API: emit an error that the SDK must be updated
+
+We will need to develop a standard pattern or helper logic for this to make sure it’s implemented consistently.
+
+# Appendix: SDK Feature Releases for Platform Updates
+
+An alternate option that was considered was to make SDK feature releases when platform API updates are needed. For example, if a new version of iOS was released after 5.0.100, we could make a 5.0.200 SDK release with the updated platform API.
```
richlander, line 422: > was to make SDK feature releases

To make this a little clearer, it could read "was to release a new SDK in higher feature band when ..."

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+## SDK Packs
+
+SDK packs are MSBuild SDK NuGet packages, but they are installed to `dotnet/packs/{pack-id}/{pack-version}`. They are expected to be automatically referenced by the targets.
+
+## Library Packs
+
+Library packs are normal NuGet packages. When a library pack is installed, the nupkg package file is placed in `dotnet/library-packs/` but are *not* extracted. This location is used as a directory feed by NuGet.
+
+If ones of these packages gets used, it will be extracted into the global packages folder. If the SDK is updated and the library pack is removed or replaced with a newer version, projects that have already been created and restored will continue to be able to use the extracted copy.
```
wli3, line 73: This is just the behavior of using local feed.

accepted/2020/workloads/workload-manifest.md
wli3: It is similar. However NuGetFallback folder will not be copied to global package folder

richlander:
I added a link to this PR in: https://github.com/dotnet/designs/blob/master/accepted/2020/workloads/workloads.md. Would be good to update that content in this PR to the final home of the doc.

accepted/2020/workloads/workload-manifest.md
KevinCathcart: Since you asked for people to weigh in, here are a bunch of my thoughts:

As a user, I would strongly prefer that Sdk="Microsoft.NET.Sdk.Web" continue to work in .NET 5, so as to have one less thing to need to change. It obviously needs to be able to work for projects targeting 3.x, since if installing .NET 5 SDK broke building 3.x apps unless you use global.json, people would be upset.

Furthermore, I'm not certain that we want every SDK the users might need to specify to become a unique workload. For example, consider `Microsoft.NET.Sdk.Worker`. It sets some capabilities, some properties, and enables copying `*.config` and `*.json` to the output (since M.E.Hosting applications are likely to want to use appsetting.json). Should that become a separate workload? Realistically it could and probably should just be bundled with whatever lets you build simple console apps, which would surely be included in the default .NET 5 SDK install.

Now perhaps that particular SDK is killed off in favor of incorporating that logic into the core SDK, activated by a property or something. But similar things could obviously exist. Not to mention that third party frameworks will still exist that will have their own `Sdk=` needed, unless there is a plan to support for third parties creating their own workloads. (Well, presumably they could, but without a pre-installed manifest, they would not get the desired experience.)

Per the resolver spec, the missing workloads are already discovered by a new SDK Resolver when an SDK being imported is found to not be present. I'd expect that would work just as well being triggered by the project file itself. 

Granted, that this does complicate resolving. Knowing which SDK pack is needed would not always uniquely resolve to a single workload. How `dotnet bundle restore` would pick between them is unclear. One advantage of `<Workloads>AspnetCore</Workloads>` is the ability to explicitly specify the workload by name, which could be used to override the the default workload selection logic if you really wanted something different than it decided. Presumably, something like "dotnet bundle restore --build-only" would be smart enough to see the specified pack is a "dev" one, and look though its extends, find only the build workloads, and install just the packs those need.

richlander: Great feedback. Thanks for taking the time.

First, we don't plan to break the existing syntax. That will continue to work. This conversation is a question of what the new scheme should be.

It's subtle but the topic at hand doesn't really relate to workloads exactly. For example, we already have and plan to continue to use `UseWinforms` and `UseWPF` properties. We're still planning to include these two components together in the Windows desktop workload, which will be declared via the `net5.0-windows` TFM.

It's more complicated with ASP.NET Core since it doesn't have its own TFM. ASP.NET Core will be available within `net5.0`, however, we don't want it to be required to run any .NET 5.0 app. As a result, we need a specific opt-in gesture. In short, it either can work the new way (with the `UseFoo` style syntax with the benefit that ASP.NET and WPF, for example, will have a symmetrical experience, or ASP.NET can be different and work the old way.

We talked to the ASP.NET team. They didn't like either option, for opposite reasons. I suspect they will end up going with the `UseFoo` style syntax so that the platform looks like same for all app types.

Make sense?

KevinCathcart: Admittedly your first post in this comment series did not distinguish that WPF would be part of a windows workload, and not a separate one. I’ve not actually used the WindowsDesktop sdk, so had not noticed that it used only properties instead of SDKs to distinguish Winforms and WPF. (Admittedly, since using both in the same project is possible, properties are more ergonomic than using sdks, since sdks would require adding a stand alone sdk element to the file to specify the second one if both were used). 

My main objection to `UseFoo` for scenarios like ASP.NET Core, is that SDKs are already a user exposed concept, so adding a magic property to enable one merely for consistency, just seems to be adding more user exposed concepts. After all, third party SDKs will realistically not be able to take advantage of magic properties to load themselves, but could use Sdk= to do so. (Yeah, technically, there are ways a third party sdk could use a magic property to do this, but only if already installed via an installer/package-manager. It could not benefit from automatic discovery that the sdk is needed but not installed. And third party SDKs hooking themselves into every single build on the system is probably not something you would want to encourage, given how easily doing so could cause weird breakage.)

accepted/2020/workloads/workload-manifest.md
dsplaisted: > There will be a mechanism that allows the targets to omit the pack version from the framework reference.

The pack version is already not normally specified on the `FrameworkReference` item.  It is specified in the `KnownFrameworkReference` which gets matched up with the `FrameworkReference` in the ProcessFrameworkReferences task.  The `KnownFrameworkReference` items are defined in Microsoft.NETCoreSdk.BundledVersions.props.

I was imagining that each workload would have it's own bundled versions file with the targeting and runtime pack version information.  This would be a duplicate of the data in the manifest.  It sounds like you're suggesting that the information should be read from the manifest during the build.  I think either way would work, and it's a trade off between where you have complexity.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+## Targeting and Runtime Packs
+
+The [targeting and runtime packs](https://github.com/dotnet/designs/pull/50) introduced in .NET Core 3.0 are permitted in workloads. Targeting packs contain the framework assemblies that a project compiles against at build time, and runtime packs contain the framework and runtime binaries used to execute the project. They install into `dotnet/packs/{pack-id}/{pack-version}` and are resolved at build time using `FrameworkReference` MSBuild items.
+
+Framework references are expected to be added to a project automatically by the manifest’s MSBuild targets based on the TFM. There will be a mechanism that allows the targets to omit the pack version from the framework reference. Targets may add multiple targeting or runtime pack references, for example when building a `net5.0-ios` project they could add a reference to a pack containing the `net5.0` reference assemblies and a pack containing the reference assemblies for the iOS platform bindings.
+
+## SDK Packs
+
+SDK packs are MSBuild SDK NuGet packages, but they are installed to `dotnet/packs/{pack-id}/{pack-version}`. They are expected to be automatically referenced by the targets.
```
dsplaisted, line 67: Details for how this works are here: https://github.com/dotnet/designs/blob/master/accepted/2020/workloads/workload-resolvers.md

accepted/2020/workloads/workload-manifest.md
dsplaisted: Library packs don't need to be partitioned into separate folders for each version band, but I'm not sure whether this is the case for template packs.

accepted/2020/workloads/workload-manifest.md
dsplaisted: > Multiple workloads might be installed for a given SDK band, which will result in a multiple manifests being installed.

The manifests will be there whether or not the workloads are installed.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+At any given point in time, the latest available version of all the manifests for an SDK series must always be able to be combined into a single consistent whole (in memory, not persisted to disk).
+
+Each manifest may have a `WorkloadManifest.targets` MSBuild file beside it that is unconditionally imported into all projects. This targets file contains conditioned imports to add automatic referenced to SDK packs based on project properties, and its usage and recommended patterns are described in more detail in the [workload resolvers spec](/workload-resolvers.md).
+
+## Packaging
+
+Manifests are packaged and distributed as NuGet packages. They will use the existing package type called `DotnetPlatform` so they cannot be accidentally used as a `PackageReference`. The manifest’s `sdk-manifest.json` and `WorkloadTargets.targets`  are in the root folder of the NuGet. The ID of the NuGet is `{manifest-id}-{sdk-band}` and the version of the NuGet is the manifest version.
```
dsplaisted, line 102: `WorkloadManifest.targets` is the filename used in the previous paragraph and in the workload resolver spec.

```suggestion
Manifests are packaged and distributed as NuGet packages. They will use the existing package type called `DotnetPlatform` so they cannot be accidentally used as a `PackageReference`. The manifest’s `sdk-manifest.json` and `WorkloadManifest.targets`  are in the root folder of the NuGet. The ID of the NuGet is `{manifest-id}-{sdk-band}` and the version of the NuGet is the manifest version.
```

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+Workload definitions take the following form:
+
+| Key | Type | Value | Required |
+|--|--|--|--|
+| `abstract` | bool | If `true`, this workload can only be extended, and is never exposed directly as an installable workload. Default is `false`. | No |
+| `kind` | string | Either `build` or `dev`. Default is `dev`. | No |
+| `description` | int | User-visible description for the workload. | Yes if dev and non-abstract |
+| `packs` | string array | IDs of the packs that are included in the workload. | No |
+| `extends` | string array | IDs of workloads whose packs should be included in this workload. | No |
+| `platforms` | string array | Limits the workload and workloads that extend it to only be shown and installed on these host platforms. The strings are RIDs. | False |
```
dsplaisted, line 143: I'm not sure if we want to use RIDs for this or not.  They may be more complex than we need, and I think we'll need the RID graph when processing the workload manifest.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+Note that `extends` is functionally a dependency system and a way to factor out common sets of packages from workloads. By analogy to package managers such as apt-get and NuGet, workloads are metapackages that only permit unversioned dependencies and packs are packages that are only installable transitively.
+
+## Pack Definitions
+
+Pack definitions take the following form:
+
+| Key | Type | Value | Required |
+|--|--|--|--|
+| `kind` | string | Type of the pack. Valid values are `sdk`, `framework`, `library`, `template`, `tool`. | Yes |
```
dsplaisted, line 157: I guess `library` is the type for targeting and runtime packs, but it would be good to spell that out if it's not already.

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+The `alias-to` key allows aliasing this pack definition to a different underlying pack ID on each host platform. The keys are host platform RIDs, and the values are pack IDs. The `"*"` platform ID may be used to alias a pack on all platforms.
+
+For example, an SDK pack might contain a compiler that's a native executable and hence depends on the host platform:
+
+```json
+"foo.sdk.compiler": {
+    "version": "1.0",
+    "kind": "sdk",
+    "alias-to": {
+        "osx-x64": "foo.sdk.compiler.mac",
```
dsplaisted, line 170: Is it possible to have different version numbers for the mac and windows compiler packages?  What would that look like?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+# Format
+
+Manifests are json files. Comments are supported, both `//` and `/* */` styles.
+
+The toplevel is a JSON object, containing the following keys:
+
+| Key | Type | Value | Required |
+|--|--|--|--|
+| `version` | int | The version of the manifest. Must match the version of the NuGet package that contains the manifest. | Yes |
```
dsplaisted, line 126: If this has to match the NuGet package version, then the value needs to be a string in order to hold the semantic version, right?

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+
+A workload or framework reference that referenced `"foo.framework"` would get `"foo.framework"` version `2.0.4`, while a workload or framework reference that referenced `"foo.framework.1"` would get `"foo.framework"` version `1.3.2`,
+
+## Example Platform SDK Manifest
+
+Here is a *hypothetical* example manifest. It's not prescriptive but demonstrates concepts an patterns that can be used by platform implementations.
+
+```json5
+{
+    "version": 5,
```
dsplaisted, line 205:
```suggestion
    "version": "5.0.0",
```

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+**If any version of the pack in the manifest is higher than the version bound**, emit a warning telling the developer that a newer OS API version is available, and explaining how to update the MSBuild property. This will guide developers to keep the value current.
+
+**If no version of the pack in the manifest complies with the version bound**, emit an error telling the developer that their .NET SDK is too old.
+
+For example, if the SDK contains 14.0 and 15.0 APIs:
+
+- Project specifies 13.0 API: use 14.0 API, and emit a warning that newer API is being used
+- Project specifies 14.0 API: use 14.0 API, and emit a warning that newer API is available
+- Project specifies 15.0 API: use 15.0 API
+- Project specifies 16.0 API: emit an error that the SDK must be updated
```
dsplaisted, line 416: I think these rules are specific to the workload.  I don't think Windows is ever going to force you to a newer API version, and may not have side-by-side SDK versions.

accepted/2020/workloads/workload-manifest.md
mhutch: Probably? I assume most packs would fall under `System.*` or `Microsoft.*` which are already reserved prefixes. Added a note of clarification.

accepted/2020/workloads/workload-manifest.md
mhutch: Rewrote to avoid introducing the term "manifest band"

accepted/2020/workloads/workload-manifest.md
mhutch: this is handled by the aliasing system defined later in the document, added a note to that effect

accepted/2020/workloads/workload-manifest.md
mhutch: Tried to explain this a bit better and added a link to the example.

accepted/2020/workloads/workload-manifest.md
mhutch: Yeah, I'm not a fan of the duplication as it allows things to get out of sync and means the manifest is no longer a means of composition.

accepted/2020/workloads/workload-manifest.md
mhutch: explicitly clarified this

accepted/2020/workloads/workload-manifest.md
mhutch: As in, if you try to invoke a tool that's defined in the manifest but not installed?

mhutch: Added a note about printing an error message. we can extend it later to kick off workload installation.

accepted/2020/workloads/workload-manifest.md
mhutch: All packs are resolved via the band manifest; the pack storage location is pretty much an implementation detail. The pack locations are split by pack kind but that isn't strictly necessary, it just makes the disk structure a bit more comprehensible.

In practice, the same pack id+version may be included in multiple bands, so the storage can be deduplicated by using the same location for all bands.

mhutch: As for extraction on demand - there are two kinds of packs and they're stored as nupkgs for different reasons.

1) Templates are stored as nupkgs is because that's what the template engine currently consumes. They could be unpacked on installation if the template engine were changed to support consuming unpacked templates in place, and I've added a note to this effect.

2) Library packs are stored as nupkgs because it's actually just a download cache in the form of the directory feed. They're handed off to NuGet and it does what it wants - currently, unpacking them into the global package cache, which is unaffected by future workload operations. This means that updating workloads will not break existing projects that reference the packs with an exact version.

accepted/2020/workloads/workload-manifest.md
mhutch: A manifest may contain multiple workloads though 🤷‍♀️

accepted/2020/workloads/workload-manifest.md
mhutch: Fair, it's pretty redundant

accepted/2020/workloads/workload-manifest.md
mhutch: I meant that you could install previews of a workload into a stable SDK band by subscribing to a preview feed that contained manifest updates for that band that contained preview packs. Clarified.

accepted/2020/workloads/workload-manifest.md
mhutch: That's a good point. I listed the conditions under which the baseline manifest would not need to be updated.

accepted/2020/workloads/workload-manifest.md
mhutch: This is conformant but not very useful except as an ID registration for downloading an updated manifest:
```json
{
    "version": 0,
    "workloads": {},
    "packs": {}
}
```

accepted/2020/workloads/workload-manifest.md
mhutch: Yeah, I'd like to dig into this some more sometime. I think we'll need some real world examples of workloads before we can start trying to divide them into scenario-specific pieces like this. Right now it's just a "do you want it listed or not" flag.

accepted/2020/workloads/workload-manifest.md
mhutch: Clarified this.

accepted/2020/workloads/workload-manifest.md
mhutch: Yes, that's exactly the kind of scenario this covers. Added a specific case.

accepted/2020/workloads/workload-manifest.md
mhutch: Good catch, it doesn't need the alias

accepted/2020/workloads/workload-manifest.md
```diff
@@ -0,0 +1,430 @@
...
+        "*": "foo.framework",
+    }
+},
+```
+
+A workload or framework reference that referenced `"foo.framework"` would get `"foo.framework"` version `2.0.4`, while a workload or framework reference that referenced `"foo.framework.1"` would get `"foo.framework"` version `1.3.2`,
+
+## Example Platform SDK Manifest
+
+Here is a *hypothetical* example manifest. It's not prescriptive but demonstrates concepts an patterns that can be used by platform implementations.
```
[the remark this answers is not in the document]
mhutch: That's fair. Feels similar to active voice vs passive voice.

accepted/2020/workloads/workload-manifest.md
mhutch: They might restore a project they already had or install a more complete workload knowing the optional features they'd need? Added a clarification.

accepted/2020/workloads/workload-manifest.md
mhutch: Yeah. This is maybe getting out of the scope of this doc though as it's more about how to structure the packs that make up a workload. Maybe something for an authoring guide?

accepted/2020/workloads/workload-manifest.md
mhutch: You can't - workloads are completely independent of SDK servicing version. 3.1.105 would get the platform update too. Added a note to the 'Manifest Versioning' section.

accepted/2020/workloads/workload-manifest.md
mhutch: Fleshed out the example a bunch

accepted/2020/workloads/workload-manifest.md
mhutch: yes, it is, that's what I meant  by "directory feed". changed to "local feed".

I thought it was good to be explicit about the behavior and why it was desirable in this case.

accepted/2020/workloads/workload-manifest.md
mhutch: It's actually `framework`. Added a note.

accepted/2020/workloads/workload-manifest.md
mhutch: Ironically I didn't use RIDs in my implementation and you fixed it? Thanks :D

accepted/2020/workloads/workload-manifest.md
mhutch: Not if we require that the manifests use integers as their package version...

mhutch: See in "manifest versioning"

>The “version” of a manifest package can then be as simple as a monotonic integer value.

accepted/2020/workloads/workload-manifest.md
mhutch: This pattern accounts for side by side versions. If they don't have side by side versions why would they not force you to a newer API version? It's literally the only option.

mhutch:
Given we shipped work that implements this spec, I'm going to merge if there isn't any further feedback.

PR #131, opened 2020-06-04
Update feature switch spec

sbomer:


sbomer:
Another argument for opt-in: it prevents the confusing behavior with opt-out where we take care to filter down `RuntimeHostConfigurationOption` in the SDK, but the filtered list still has items which don't correspond to feature substitutions.

With opt-in, `<Link>true</Link>` would exactly match the options with feature substitutions.

vitek-karas:
I'm not against making this opt-in, maybe @marek-safar will have a different opinion.

Counter-argument: It is already sort of opt-in since the author has to use the name of the runtime config option in the substitution.xml - so adding an explicit opt-in to the targets file is basically duplicating that intent. So if the author introduces both ends (the typical case), it feels a bit like busy work. On the other hand if the two sides are introduced independently, then it makes more sense.

MichalStrehovsky:
Btw: could we keep `Trim` instead of `Link` name in options that are user-facing for consistency with `PublishTrimmed`?

IL Linker doesn't fit the definition of the word [linker in computing](https://en.wikipedia.org/wiki/Linker_(computing)). It does an optimization that corresponds to `/opt:ref` or `--remove` optimization linker can do, but it can't do the single primary purpose of a linker (link objects). I always found that name confusing.

marek-safar:
I agree with @vitek-karas that the opt-in is kind of redundant and it's odd to have opt-in for linker only and not for other RuntimeHostConfigurationOption targets.

Another aspect to also consider it that RuntimeHostConfigurationOption should not be required (with or without opt-in) for the feature definitions. I can imagine scenarios where the feature switch is applicable to specific SDK (e.g. aspnetcore) only.

eerhardt:
> Another aspect to also consider it that RuntimeHostConfigurationOption should not be required (with or without opt-in) for the feature definitions. I can imagine scenarios where the feature switch is applicable to specific SDK (e.g. aspnetcore) only.

The `@(RuntimeHostConfigurationOption)` MSBuild item is how you get information put into the `runtimeconfig.json` file, which is then flown into the `AppContext`. Even if aspnetcore, Xamarin, or Blazor were defining the feature switch, the intention outlined in the spec is that the switch is configurable using `AppContext`, so you wouldn't need to ILLink the app in order to get the behavior. See the 2nd Goal in the doc:

> Consistent behavior across different build configurations - the app should get the exact same behavior whether the linker was used on it or not

_____

My basic reason for thinking `@(RuntimeHostConfigurationOption)` should declare that it is ILLinker aware is that not all configuration settings that go into the `runtimeconfig.json` are applicable as ILLink features. As of today, 8 out of 9 are not applicable. And 3 of the ones we have defined so far are non-boolean so they explicitly CAN'T be flowing into the ILLinker or else it will fail.

marek-safar:
OK, let's go with explicit metadata for an option which will be passed to the linker

accepted/2020/feature-switch.md
```diff
@@ -99,13 +99,13 @@ Example of a feature flag without support for linker optimizations:
     ...
 </ItemGroup>
 ```
-Note that even features with `Link == true` will still only be optimized by the linker if there is a corresponding substitution definition.
+Note that even features with `Trim == true` will still only be optimized by the linker if there is a corresponding substitution definition. Generally, `Trim` should only beb set to `true` for items which do have a corresponding subtitution definition.
```
vitek-karas, line 102:
```suggestion
Note that even features with `Trim == true` will still only be optimized by the linker if there is a corresponding substitution definition. Generally, `Trim` should only be set to `true` for items which do have a corresponding substitution definition.
```

accepted/2020/feature-switch.md
```diff
@@ -69,7 +69,7 @@ The property's value will be burned in by the linker when a corresponding featur
...
+This requires a mapping between the MSBuild property name, the full name of the feature switch for runtime configuration (which would show up in `.runtimeconfig.json`) and the read-only static property in the managed code which is used to branch the behavior. The mappings will be defined by [`RuntimeHostConfigurationOption`](https://github.com/dotnet/sdk/blob/36ef8b2aa8e5d579c921704bdab69a7407936889/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L347) which includes an AppContext configuration name and its value based on the values of user-facing MSBuild properties, and by `ILLink.Substitutions.xml` files embedded in the assembly defining the features. `RuntimeHostConfigurationOption` items will have optional boolean `Trim` metadata which indicates that the feature setting should be optimized by the linker. This metadata can be set either in the SDK (for features which have defined linker optimizations) or by developers (if they want to disable linking of a particular feature). Only items in `RuntimeHostConfigurationOption` with `Trim == true` will flow to the linker.
```
vitek-karas, line 73: We should add a line somewhere which says that `RuntimeHostConfigurationOption` item with `Trim == true` must have a boolean value (`true` or `false`) otherwise linker will fail on it.

PR #134, opened 2020-06-11
Improve Activity API usability and OpenTelemetry integration (Part 2)

tarekgh:


tarekgh:
https://github.com/dotnet/designs/blob/master/accepted/2020/diagnostics/activity-improvements.md

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,156 @@
...
+    public partial class Activity : IDisposable
+    {
+        // The setter was private and now we are making it public
+        public ActivityKind Kind { get; set; } = ActivityKind.Internal;
+    }
+```
+
+## ActivityTagsCollection
+
+OT has the concept of [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes) which is a list of key-value pairs mapping a string to some value. the value can be numeric, bool, string types. Also, the value can be an array of one of these basic types.
```
reyang, line 47:
```suggestion
OT has the concept of [Attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes) which is a list of key-value pairs mapping a string to some value. The value can be numeric, bool, string types. Also, the value can be an array of one of these basic types. 
```

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,156 @@
...
+        public ActivityEvent(string name, IEnumerable<KeyValuePair<string, object>>? attributes)
+        public ActivityEvent(string name, DateTimeOffset timestamp, IEnumerable<KeyValuePair<string, object>>? attributes)
+        public IEnumerable<KeyValuePair<string, object>> Attributes { get; }
+    }
+
+    // change it to the following:
+    public readonly struct ActivityEvent
+    {
+        public ActivityEvent(string name, ActivityTagsCollection? attributes)
+        public ActivityEvent(string name, DateTimeOffset timestamp, ActivityTagsCollection? attributes)
```
reyang, line 135: Just curious (probably not related to this PR), would `attributes` get deep-copied or a shallow copy? (e.g. if the tags collection has reference to another tags collection)

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,156 @@
...
+        public void AddTag(string key, double value) { }
+        public void AddTag(string key, bool value) { }
+        public void AddTag(string key, int [] value) { }
+        public void AddTag(string key, double [] value) { }
+        public void AddTag(string key, bool [] value) { } // how interesting this one?
+        public void AddTag(string key, string [] value) { }
+        public void AddTag(string key, ActivityTagsCollection value) { }
+
+        // We cannot change Tags property so we are introducing a new property with a new name
+        public ActivityTagsCollection TypedTags { get; }
```
reyang, line 112: Clarification question - if we enumerate through the `TypedTags`, do we get the string tag added via `AddTag(string, string)` or we have to do another enumeration through `Tags`?

tarekgh: TypedTags should return all tags including the tags added with `AddTag(string, string)`

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: This is needed to be decided which behavior is the correct one. I'll try to look at OT specs to know if it mentions the expected behavior.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,156 @@
...
+    {
+        public ActivityLink(ActivityContext context, IEnumerable<KeyValuePair<string, object>>? attributes)
+        public IEnumerable<KeyValuePair<string, object>>? Attributes { get; }
+    }
+
+    // change it to the following:
+    public readonly partial struct ActivityLink  : IEquatable<ActivityLink>
+    {
+        public ActivityLink(ActivityContext context, ActivityTagsCollection? attributes)
+        public ActivityTagsCollection? Attributes { get; }
```
reyang, line 154: I'm not good at naming things and I'm on the fence here (naming is always the most challenging discussion since everyone has opinion).

It seems we prefer to call it `Attributes` here since `ActivityLink` is a brand new thing which mirrors `SpanLink`, however we decided to go with `Activity.TypedTags` instead of `Activity.Attributes` due to the alignment with existing `ActivityTags`.

Looks like we have three options:
1. Things newly added in .NET 5 will have a consistent name `Attributes` which aligns with OpenTelemetry spec. - downside is that Tag is becoming the weird legacy thing (we can probably have a method `SetAttribute(string, string)` which mirrors what `AddTag` is doing).
2. We will use `Tags` and `TypedTags` consistently across the newly added APIs - downside is that it looks different from OpenTelemetry, I'm guessing it is probably fine since Activity -> Span is already different and the name `Tag` is already established in .NET for years (unless OpenTelemetry later will add another concept called `tags`, which I don't think will ever happen 🤣).
3. The current approach proposed by this PR.

I don't have a strong opinion, personally I would prefer option 1.

tarekgh: This is actually a good point. I am inclining to the consistency inside .NET APIs. I mean we should use Tags all the way instead of Attributes. I believe this will be less confusing to the APIs users. even for OT developers will be easy to map .NET Tags to OT Attributes and avoid the confusion mapping 2 names instead of one. I'll modify the proposal with that and let's discuss it more as part of the process.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,165 @@
...
+
+We already exposed the ActivityContext type as described by the OT specs. There is a property inside this type called `IsRemote` which we didn't include in the first place because it was under discussion with the OT spec committee. It has been confirmed this property is needed and will stay as part of the OT specs.
+
+[ActivityContext.IsRemote](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spancontext) is a Boolean flag which returns true if the ActivityContext was propagated from a remote parent. When creating children from remote spans, their IsRemote flag MUST be set to false.
+
+```c#
+    public readonly partial struct ActivityContext : IEquatable<ActivityContext>
+    {
+        // This constructor already exist but we are adding extra defaulted parameter for isRemote.
+        public ActivityContext(ActivityTraceId traceId, ActivitySpanId spanId, ActivityTraceFlags traceFlags, string? traceState = null, isRomte = false) {...}
```
pjanotti, line 17:
```suggestion
        public ActivityContext(ActivityTraceId traceId, ActivitySpanId spanId, ActivityTraceFlags traceFlags, string? traceState = null, isRemote = false) {...}
```

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,165 @@
...
+        public bool IsRemote { get; }
+    }
+```
+
+Activity class will need to expose such property too to be able to set it as part of the included context. This is needed by the OT asp.net and Http client OT adapters.
+
+```c#
+    public partial class Activity : IDisposable
+    {
+        public bool IsRemote { get; set; }
```
pjanotti, line 29: It will be good to describe the case that it needs to be set after creation usually speeds up the discussion, same for the Kind setter below.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+***<u>OpenQuestion</u>***
+
+We need to decide about the behavior of the old property `Activity.Tags` when adding tags of non-string types values. The options are
+
+- Ignore all non-string typed values and return only the string type value tags.
+- Convert the non-string type values to string before returning it.
+
+I am inclining to the first option to avoid handling ToString with more complex types (e.g. arrays or ActivityTagsCollection).
```
pjanotti, line 123: 👍

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: good idea.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+There are a couple of proposals how to address this issue in the .NET runtime. The idea is the runtime need to generate the trace Id and send it to the samplers. If the sampler using the passed trace id and decide to sample in this case, the Activity which will get created because of that sampler decision will be using the generated trace Id in its context.
+
+Solution options:
+
+- Before calling `ActivityListener.GetRequestedDataUsingContext`, there will be a check if we don't have a parent, then we generate a random trace Id and use it in the context sent to the callback. If decided to sample in, we all use this trace Id during the Activity object creation. This option doesn't need any API changes and will be fully handled inside the implementation.
+- Add a new property to the `ActivityCreationOptions` struct (e.g. `NullParentTraceId`) which will carry the trace id that can be used by the sampler when detecting no parent context. obviously this will need API change.
+
+I am inclining to the first solution for the following reasons:
```
noahfalk, line 181: Sounds fine to me, lets make sure to document carefully : ) The current property name says that the ActivityContext represents the parent and its a somewhat creative interpretation to use the TraceId portion of it as a newly generated random value instead.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+
+This document is listing the proposed changes and the rationale for the proposal. **As we still receiving more feedback and there are some issues under discussion, it is expected this documents will get updated before we do the full design review for the proposed changes.**
+
+## ActivityContext.IsRemote
+
+We already exposed the ActivityContext type as described by the OT specs. There is a property inside this type called `IsRemote` which we didn't include in the first place because it was under discussion with the OT spec committee. It has been confirmed this property is needed and will stay as part of the OT specs.
+
+[ActivityContext.IsRemote](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spancontext) is a Boolean flag which returns true if the ActivityContext was propagated from a remote parent. When creating children from remote spans, their IsRemote flag MUST be set to false.
```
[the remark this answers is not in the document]
tarekgh: That s actually an excellent question. Actually I was interpreting it differently in the part you said `If we then get the Activity.Context for the Activity that is returned from this call, we expect IsRemote = FALSE.`. But it looks your interpretation looks more correct. @reyang @cijothomas does @noahfalk interpretation is matching your understanding? in addition to @noahfalk question, I am wondering Activity.IsRemote will never (or shouldn't) be true too

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+
+## Activity.Tags
+
+As mentioned earlier, Activity has Tags property which returns the list of the activity tags. Tags are a list of `KeyValuePair<string, string>`. Tags are proven to be not enough for OT scenarios as we need to support more value types (numeric, bool, numeric array, and nested tags). Activity today has the method `public Activity AddTag(string key, string? value)`. The proposal here is to add more `AddTag` overload accepting more types and add a new getter property return the list of all typed tags.
+
+```C#
+    public partial class Activity : IDisposable
+    {
+        public void AddTag(string key, int value) { }
+        public void AddTag(string key, double value) { }
```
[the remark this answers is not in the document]
tarekgh: This design doc is written before such a discussion. but I have some clarification questions to ensure the expected behavior: 
- Should we allow all types with no restrictions? 
    - if yes, did we think what OT will do when encountering types not specified in the OT specs? 
    - if no, I'll have a concern about the usage as users can pass some invalid types and we'll throw at that time. 
-  Having AddTag(string, string) and AddTag(string, object) can be a little confusing to the users. it is like `Equal()` pattern. but in general, I am less concerned about this here.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: Sounds good. This change will not need API change. if all agree, I can start implementing this part from now.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+        public ActivityEvent(string name, DateTimeOffset timestamp, IEnumerable<KeyValuePair<string, object>>? attributes)
+        public IEnumerable<KeyValuePair<string, object>> Attributes { get; }
+    }
+
+    // change it to the following:
+    public readonly struct ActivityEvent
+    {
+        public ActivityEvent(string name, ActivityTagsCollection? tags)
+        public ActivityEvent(string name, DateTimeOffset timestamp, ActivityTagsCollection? tags)
+        public ActivityTagsCollection? Tags { get; }
```
cijothomas, line 149: Should we call this Tags or TypedTags ?

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+    {
+        public ActivityLink(ActivityContext context, IEnumerable<KeyValuePair<string, object>>? attributes)
+        public IEnumerable<KeyValuePair<string, object>>? Attributes { get; }
+    }
+
+    // change it to the following:
+    public readonly partial struct ActivityLink  : IEquatable<ActivityLink>
+    {
+        public ActivityLink(ActivityContext context, ActivityTagsCollection? tags)
+        public ActivityTagsCollection? Tags { get; }
```
cijothomas, line 167: TypedTags instead of Tags

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: I prefer Tags here. we used TypedTags in Activity because it had Tags. Also, here, the return type is `ActivityTagsCollection` which simplicity telling the expected  types.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+        // We cannot change Tags property so we are introducing a new property with a new name
+        public ActivityTagsCollection TypedTags { get; }
+    }
+```
+
+***<u>OpenQuestion</u>***
+
+We need to decide about the behavior of the old property `Activity.Tags` when adding tags of non-string types values. The options are
+
+- Ignore all non-string typed values and return only the string type value tags.
```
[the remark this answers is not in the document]
cijothomas: Agree.

accepted/2020/diagnostics/activity-improvements-2.md
cijothomas: So that we can consistently say - TypedTags in Activity is same as Attributes in Span

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: ditto. Please let me know if you disagree and we can talk more about it.

accepted/2020/diagnostics/activity-improvements-2.md
cijothomas: Yes. this must be documented clearly to avoid confusion.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+There are a couple of proposals how to address this issue in the .NET runtime. The idea is the runtime need to generate the trace Id and send it to the samplers. If the sampler using the passed trace id and decide to sample in this case, the Activity which will get created because of that sampler decision will be using the generated trace Id in its context.
+
+Solution options:
+
+- Before calling `ActivityListener.GetRequestedDataUsingContext`, there will be a check if we don't have a parent, then we generate a random trace Id and use it in the context sent to the callback. If decided to sample in, we all use this trace Id during the Activity object creation. This option doesn't need any API changes and will be fully handled inside the implementation.
```
cijothomas, line 178: just for my understanding - if the sampler returns none, then we wont end up creating the activity. In that case the compute power spend in generating the random id is wasted. Is it an expensive thing?

I was thinking of an alternate approach where sampler (if cares about traceid) instead generates the traceid and sends it back somehow. Only if sampler depends on traceid, then we pay compute cost for generatetraceid.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: I am not worried about that. In general, moving forward we should message 'Tags' in BCL are the 'Attributes'. Otherwise, we'll stick forever describing the differences between `tags`, `typed tags` and `attributes`.  Note, the API is called AddTag and I don't think we need to rename it to AddTypedTag.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: > I was thinking of an alternate approach where sampler (if cares about traceid) instead generates the traceid and sends it back somehow. Only if sampler depends on traceid, then we pay compute cost for generatetraceid.

What happen when you have multiple listeners and each giving you different trace id? 

Would it be reasonable to assume the trace if for null parent is constant? I mean we generate it once and use it all over for all null parents?

cijothomas: ya i didnt think about the multiple listener case. 
This proposed approach is fine.

accepted/2020/diagnostics/activity-improvements-2.md
cijothomas: so span.AddAttribute -> activity.AddTag. - this is good. Both support string->int/bool/etc. Same for ActivityEvent.AddTag, ActivityLink.AddTag.

For retrieving full attributes the api is: Activity.TypedTags, ActivityEvent.Tags, ActivityLink.Tags -- this is bit inconsistent. In Span API this would be consistent across Span.Attributes, Link.Attributes, Event.Attributes.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,185 @@
...
+    public readonly partial struct ActivityLink  : IEquatable<ActivityLink>
+    {
+        public ActivityLink(ActivityContext context, ActivityTagsCollection? tags)
+        public ActivityTagsCollection? Tags { get; }
+    }
+```
+
+## Automatic Trace Id  generation in case of null parent
+
+Sampler like probabilistic sampler can depend on the trace id for the sampling decision. Usually before creating any Activity using ActivitySource, the `ActivityListener.GetRequestedDataUsingContext` callback will get called which usually get implemented by the samplers. The parent context is sent as a parameter to this callback and the sampler gets the trace id from this parent context to perform the sampling action. When there is no parent, we send the default context which is nulling the trace id. The samplers depending on trace id are not going to be able to do the sampling at that time. The samplers even cannot generate and use a trace id because it cannot inform the Activity creation process to use the generated trace Id when creating the Activity object.
```
cijothomas, line 173: Can you add about the proposal @noahfalk said yesterday about `ActivityListener.GetRequestedDataUsingParentId` - This becoming an optional thing, and .NET will provide a default implementation where it'll try to generate `ActivityContext` from the string parentid, and invoke the other callback `ActivityListener.GetRequestedDataUsingContext`.
if ActivityContext cannot be extracted from parentid, then the default implementation could return .None.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: let me ask, does `Activity.Tags` still considered Attributes or not? my point here is, I don't like using TypedTags, we used it because of the reason we previously exposed have Tags. That means we shouldn't use it in other places. the other idea we can do if you seeing this a concern, instead of having `Activity.TypedTags` we can have `Activity.GetTags()` instead. doing that we'll be consistent all the way.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: >  This becoming an optional thing, and .NET will provide a default implementation where it'll try to generate ActivityContext from the string parentid, and invoke the other callback ActivityListener.GetRequestedDataUsingContext.

I am not sure that what @noahfalk said. I understood that will be the listener's job to do I think and not the .NET. I may be mistaken but @noahfalk can clarify what he said.

accepted/2020/diagnostics/activity-improvements-2.md
cijothomas: I think Activity.Tags cannot be considered Attributes as its only a sub-set of the full attributes.

My personal feeling is to have consistency, hence .GetTags() everywhere - Activity, Link, Event would give me consistency. Would love to hear from others as well.

tarekgh: > I think Activity.Tags cannot be considered Attributes as its only a sub-set of the full attributes.

If it is subset then it is qualified to be `attributes`. The other way around is wrong. `Tags` is `attributes`. attributes not necessarily be `Activity.Tags`. 

> My personal feeling is to have consistency, hence .GetTags() everywhere - Activity, Link, Event would give me consistency. Would love to hear from others as well.

I am open to all suggestions. but I didn't mean to use GetTags everywhere. GetTags will be in Activity only and Tags will be a property in other places. I am not sure why this will be confusing or inconsistent. anyway, I am thinking to schedule a meeting to go through all feedback here and discuss it in person. I appreciate all feedback I am getting from you and others.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: I am wondering, do we need to do a copy at all? I understand the specs saying Event is immutable but does that need to copy (either shallow or deep) the attributes? what python or other implementations do in such cases? I know nested attributes are not approved/implemented yet but I am wondering in general about the current state.

accepted/2020/diagnostics/activity-improvements-2.md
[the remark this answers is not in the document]
cijothomas: Thanks @noahfalk and @tarekgh I had some misunderstanding here as well, and looks like you clarified a lot.

Based on this, an ActivityContext with IsRemote=false only lives temporarily during the ActivityCreation stage, where Sampler can access it and make decision based on it. There would never be an Activity which has IsRemote=true.
@reyang can you share if your understanding is the same?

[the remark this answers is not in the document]
cijothomas: This also means there is no need of making IsRemote settable.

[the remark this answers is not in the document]
tarekgh: great! based on that, we don't need `Activity.IsRemote` at all then.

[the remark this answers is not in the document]
reyang: I'm a bit confused, is this about `Activity.IsRemote` or `ActivityContext.IsRemote`?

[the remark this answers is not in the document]
cijothomas: Activity.IsRemote was proposed here to do wrap ActivityContext.IsRemote. You can ignore Activity.IsRemote every existed.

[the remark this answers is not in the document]
cijothomas: > great! based on that, we don't need `Activity.IsRemote` at all then.

Yes.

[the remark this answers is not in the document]
tarekgh: I'll update the proposal accordingly.

accepted/2020/diagnostics/activity-improvements-2.md
cijothomas: @noahfalk Can you confirm this? I believe my comment above is what you said, but please do confirm/correct me :)

accepted/2020/diagnostics/activity-improvements-2.md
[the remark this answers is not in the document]
noahfalk: > Should we allow all types with no restrictions?

I'd propose yes, but we should make sure we have conventions that work well with OpenTelemetry and provide that to people as guidance. If customers don't follow the guidance and use other types I'd suggest that OpenTelemetry should ignore those tags.

> The question here, if someone calls Activity.AddTag(string, (object) string), I assume this should be equivalent to Activity.AddTag(string, string) and Activity.Tags should include it in its return.

Seems reasonable to me : )

> Having AddTag(string, string) and AddTag(string, object) can be a little confusing to the users. it is like Equal() pattern. but in general, I am less concerned about this here.

Anything in particular we should be looking out for? I am not aware of past feedback/confusion that the Equals() pattern caused.

accepted/2020/diagnostics/activity-improvements-2.md
noahfalk: I do find the inconsistency annoying between Tags and TypedTags though none of the alternatives have seemed that good either. One suggestion that maybe you guys would like:
- Add a concept of activity Attributes, define Tags as the subset of Attributes that are strings, and via docs encourage new code to use Attributes
- In Activity:
  AddTag(string name, string val) { return AddAttribute(name, val); }
  [new API] AddAttribute(string name, object val)
  IEnumerable Tags { return Attributes.Where(a => a is string); }
  [new API] IEnumerable Attributes
- ActivityTagCollection becomes ActivityAttributeCollection
- ActivityEvent.Tags renamed ActivityEvent.Attributes
- ActivityLink.Tags renamed ActivityLink.Attributes
- Any parameter/variable names in the new APIs that reference 'tags' are replaced with 'attributes'


 I take a little solace that all the Tags/TypedTags properties are intended for listeners and the audience of developers who will ever write an ActivityListener is probably small. So at least if things are a little messy not many people will ever have to deal with it.

accepted/2020/diagnostics/activity-improvements-2.md
noahfalk: Yep! @cijothomas captured my suggestion accurately : )

accepted/2020/diagnostics/activity-improvements-2.md
noahfalk: > Is it an expensive thing?

You were right to be concerned. I knew there was cost to it but I didn't remember quite how bad it was (benchmark below). Our default implementation of creating a default random id is almost 200ns on my machine. I am hoping a simple/fast sampling algorithm can decide whether or not to create a span in 20-100 ns so making all samplers pay 200ns for something they didn't want to look at it is pretty bad.

I think we need a design proposal where this cost is opt-in or we need to substantially improve the implementation. A few impromptu ideas:
- A boolean property on the listener that turns on seeding the parent context with a pre-generated random id
- A ref parameter on the GetDataRequestDelegate allowing a listener to populate the id (listeners could be expected not to regenerate the value if it is already non-zero)
- Another variation of the GetDataRequestDelegate that has an extra trace id parameter

You guys might have other ideas as you look into this and I can dig into it more with you if that would be helpful. There are also a few underlying implementation performance issues you may want to look at in the runtime to understand what is fundamentally slow vs. what happens to be slow because of current implementation:
- ActivityTraceId is using a string as its internal representation of the id, this incurs byte -> character conversions and allocation of a string
- ActivityTraceId.CreateRandom() uses Guid.NewGuid() which:
   - Isn't uniformly random
   - invokes down to a p/invoke and empirically was noticably slower than using Random.GetNext() 4 times


### Benchmark

|               Method |       Mean |     Error |    StdDev |
|--------------------- |-----------:|----------:|----------:|
| RandomId_DefaultImpl | 192.483 ns | 0.7703 ns | 0.7206 ns |
|      RandomId_Faster | 145.251 ns | 1.0275 ns | 0.9611 ns |
|     FixedIdFromBytes |  46.855 ns | 0.1189 ns | 0.1054 ns |
|      FixedIdFromInts |  47.236 ns | 0.5055 ns | 0.4728 ns |
|       FourRandomInts |  28.746 ns | 0.5567 ns | 0.5207 ns |
|         OneRandomInt |   6.508 ns | 0.0540 ns | 0.0451 ns |

````
using BenchmarkDotNet.Attributes;
using BenchmarkDotNet.Jobs;
using BenchmarkDotNet.Running;
using Microsoft.CodeAnalysis.CSharp.Syntax;
using System;
using System.Diagnostics;
using System.Runtime.InteropServices;


public class Program
{
    public static void Main(string[] args)
    {
        var summary = BenchmarkRunner.Run<RandomPerf>();
    }
}


public class RandomPerf
{
    private Random _random;
    private byte[] _fixedIdBytes;
    private int[] _fixedIdInts;

    [GlobalSetup]
    public void Setup()
    {
        _random = new Random();
        _fixedIdBytes = new byte[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16 };
        _fixedIdInts = new int[] { 0x04030201, 0x08070605, 0x0c0b0a09, 0x100f0e0d };
    }

    [Benchmark]
    public ActivityTraceId RandomId_DefaultImpl()
    {
        return ActivityTraceId.CreateRandom();
    }

    [Benchmark]
    public ActivityTraceId RandomId_Faster()
    {
        Span<int> ints = stackalloc int[4];
        ints[0] = _random.Next();
        ints[1] = _random.Next();
        ints[2] = _random.Next();
        ints[3] = _random.Next();
        return ActivityTraceId.CreateFromBytes(MemoryMarshal.Cast<int,byte>((ReadOnlySpan<int>)ints));
    }

    [Benchmark]
    public ActivityTraceId FixedIdFromBytes()
    {
        return ActivityTraceId.CreateFromBytes(_fixedIdBytes);
    }

    [Benchmark]
    public ActivityTraceId FixedIdFromInts()
    {
        return ActivityTraceId.CreateFromBytes(MemoryMarshal.Cast<int, byte>(_fixedIdInts));
    }

    [Benchmark]
    public int FourRandomInts()
    {
        Span<int> bytes = stackalloc int[4];
        bytes[0] = _random.Next();
        bytes[1] = _random.Next();
        bytes[2] = _random.Next();
        bytes[3] = _random.Next();
        return bytes[3];
    }

    [Benchmark]
    public int OneRandomInt()
    {
        return _random.Next();
    }
}
````
````

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 11 above)
Comment (cijothomas) [reply accepted/2020/diagnostics/activity-improvements-2.md:149]: Agree with Attributes - it is simpler name than TypedTags.

If we have an option to introduce a new property for attributes, could we consider having `Set` as well as `Add`?
Everything you said plus:
activity.SetAttribute() -overriding existing key behavior. (This is what is needed for [OpenTelemetry](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes))

The only notable difference now between Span and Activity is Add vs Set behavior for attributes/tags.

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 14 above)
Comment (cijothomas) [reply accepted/2020/diagnostics/activity-improvements-2.md:178]: The issue of generating random traceid, and we end up not sampling it in (and hence wasting cpu cycles for generating id) only occurs for the very root span of a distributed trace. I am assuming that most production distributed traces would contain many spans (probably 10s), and only 1 of them will have be rootspan, and hence pays the penalty. (that too if sampling returns NotSample at the root)

I liked the idea of Samplers decide if they are dependent on TraceID, and if a null one is passed, it generates one itself. In case of multiple listeners, only one end up generating traceid, as other listeners gets this generated one. This may put additional complication on samplers, and want to see if this is worth it based on my understanding above. (that this is only for rootspan)
+ @reyang please share your thoughts. We chatted earlier that Python is doing this already - i.e generate random TraceId() and waste the compute power if Span doesn't end up getting created.

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 11 above)
Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:149]: Although the name `Attributes` was suggested and dismissed before for the sake of consistency with Tags, I don't mind having it back as @noahfalk suggested. 

I don't think we need to have Add and Set for the attributes. The OT defines the behavior of adding Attributes and we'll just follow that. we'll have only SetAttribute. https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#set-attributes

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 14 above)
Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:178]: I am not sure what is wrong to have a constant trace id for null parent which will be generated once? why do we need to make this very complicated by plumping and defining a behavior which will not be easy for listener developers to understand or care about? I am feeling we are overengineering this part. I am really concerned about multiple listeners scenarios which may have one listener override the other listener's desire. having one listener take a decision either generating a trace id or asking for generating trace id is kind of unfair and will depend on the order of the registered listeners.

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 15 above)
Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:173]: > if ActivityContext cannot be extracted from parentid, then the default implementation could return .None.

What do you mean by default implementation here?

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 11 above)
Comment (cijothomas) [reply accepted/2020/diagnostics/activity-improvements-2.md:149]: We need AddAttribute() internally to handle AddTag() right? Asuming AddTag will be modified to simply do AddAttribute internally.

But anyway - OT only need SetAttribute, and we are good: :)

Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:149]: > We need AddAttribute() internally to handle AddTag() right? Asuming AddTag will be modified to simply do AddAttribute internally.

These are implementation details. I don't think we need to add a new API just to handle the implementation.

Comment (cijothomas) [reply accepted/2020/diagnostics/activity-improvements-2.md:149]: Yup.

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 15 above)
Comment (cijothomas) [reply accepted/2020/diagnostics/activity-improvements-2.md:173]: I meant .NET ActivitySource/Listener will provide a default implementation for `GetRequestedDataUsingParentId`. OpenTelemetry (or others) may optionally override it, if they are not happy with the default.

The default implementation (part of DiagnoticSource package) is something like this:

```
GetRequestedDataUsingParentId = (ref ActivityCreationOptions<string> options) => 
{
var parentid = options.Parent;
if (IsW3C(parentID))
  {
    var activityContext = Extract(parentId);
    call the GetRequestedDataUsingContext()
  }
else
{
   ActivityDataRequest.None;
}
},
```

Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:173]: Shouldn't this be handled on the level of the `ActivitySource.StartActivity(string name, ActivityKind kind, string parentId, IEnumerable<KeyValuePair<string, string?>>? tags = null, IEnumerable<ActivityLink>? links = null, DateTimeOffset startTime = default)` instead? 

I mean the implementation of this API will try to extract the context from the parent Id and will call `GetRequestedDataUsingContext`. If it couldn't extract the context, will continue calling `GetRequestedDataUsingParentId`. If the listener cares about `GetRequestedDataUsingParentId`, can implement it. If the listener doesn't care, it should leave it null in the listener (which effectively is the same as returning None). Doing your suggestion will make GetRequestedDataUsingParentId is active for all listeners which all of them will pay the cost of this extra call while it is not really needed.

Comment (cijothomas) [reply accepted/2020/diagnostics/activity-improvements-2.md:173]: You are right. I focused too much on the GetRequestedDataUsingParentId callback!

Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:173]: Thanks, @cijothomas. It is always good to hear other ideas and compare and choose the best. your feedback is really valuable.

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 9 above)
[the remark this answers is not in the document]
Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:9]: I have updated the doc according to this discussion.

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 15 above)
Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:133]: I have updated the doc according to this discussion.

File: accepted/2020/diagnostics/activity-improvements-2.md (hunk 14 above)
Comment (tarekgh) [reply accepted/2020/diagnostics/activity-improvements-2.md:138]: I am keeping this as an open issue. will try to discuss it offline with @noahfalk and decide what can do here.

File: accepted/2020/diagnostics/activity-improvements-2.md [hunk 16]
```diff
@@ -0,0 +1,145 @@
...
+        
+        public ActivityAttributesCollection Attributes { get; }
+    }
+```
+
+`Activity.Tags` behavior is not going to change and will always return the list of tags/attributes which has string typed values only. while the new API `Activity.Attributes' is going to return all attributes with all different type values.
+
+## Proposal Changing already exposed APIs in previous preview releases
+
+As we are introducing the ActivityAttributesCollection, we can take advantage of that now and modify some of the exposed APIs to use this collection instead. Also, we are renaming Attributes to Tags for consistency inside BCL APIs.
```
cijothomas, line 83: renaming Attributes to Tags - did you mean the other way around?

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,145 @@
...
+        public ActivityEvent(string name, IEnumerable<KeyValuePair<string, object>>? attributes)
+        public ActivityEvent(string name, DateTimeOffset timestamp, IEnumerable<KeyValuePair<string, object>>? attributes)
+        public IEnumerable<KeyValuePair<string, object>> Attributes { get; }
+    }
+
+    // change it to the following:
+    public readonly struct ActivityEvent
+    {
+        public ActivityEvent(string name, ActivityAttributesCollection? tags)
+        public ActivityEvent(string name, DateTimeOffset timestamp, ActivityAttributesCollection? tags)
```
cijothomas, line 100: why no attributes instead of tags

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,145 @@
...
+
+The proposed change here is we'll have `ActivitySource.StartActivity(..., string parentId,....)` try to extract the context from the input parent Id. If it succeed to do that, it will call the listener callback `ActivityListener.GetRequestedDataUsingContex` instead and `ActivityListener.GetRequestedDataUsingParentId` will not get called. If it fail to construct the context, it will continue calling `ActivityListener.GetRequestedDataUsingParentId` if the listener implemented it. Listener implementers now have the choice to implement `ActivityListener.GetRequestedDataUsingParentId` only if they cares about parent Ids that don't include context information.
```
cijothomas, line 127: note: based on the actual implementation, this text needs correction. (assuming this design doc will serve as doc for some period of time)
https://github.com/dotnet/runtime/pull/38312#discussion_r445046515

accepted/2020/diagnostics/activity-improvements-2.md
cijothomas: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#shouldsample

OT spec for sampling makes this separate - i.e parentcontext and traceid are passed separately. Would it make sense to do that here? So that there would be clear expectation what the traceid would be - it'd be the same traceid as parentcontext.traceid. if parent is null, then traceid is the id of the "to be created" activity.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: it should be `attributes`. I'll fix it.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: Yes, I am going to update the text here to reflect latest discussions.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: The spec is saying `TraceId of the Span to be created. It can be different from the TraceId in the SpanContext. Typically in situations when the Span to be created starts a new Trace.`. do we have any real scenario the samplers will need to get 2 trace Ids (one from the parent and one for the span to be created with)? how this will be useful to the samplers? I am expecting whoever calling `ActivitySource.StartActivity` is going to ensure passing the trace id in the parent context to have the Activity object created with this trace Id. 

I am expecting the text in this part of the document is going to change to accommodate the points that @noahfalk raised. I'll expect there will be some API change here.

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,144 @@
...
+
+```c#
+    public class ActivityAttributesCollection : IReadOnlyCollection<KeyValuePair<string, object>>
+    {
+        public ActivityAttributesCollection() { }
+        public ActivityAttributesCollection(IEnumerable<KeyValuePair<string, object>>) { }
+
+        // If the key was not set before, it will get added to the collections with the input value.
+        // If the key was set before, the mapped value of this key will be updated with the new input value.
+        public void Set(string key, int value) { }
```
cijothomas, line 55: Will we expose more overloads for bool, string etc for values?

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,144 @@
...
+As mentioned earlier, Activity has Tags property which returns the list of the activity tags. Tags are a list of `KeyValuePair<string, string>`. Tags are proven to be not enough for OT scenarios as we need to support more value types (numeric, bool, arrays...etc.). Activity today has the method `public Activity AddTag(string key, string? value)`. The proposal here is to add `SetAttribute()` which can set attributes value for a string key. 
+
+In general, Attributes will be considered a superset of Tags. 
+
+```C#
+    public partial class Activity : IDisposable
+    {
+        public void SetAttribute(string key, object value) { }
+        
+        public ActivityAttributesCollection Attributes { get; }
```
cijothomas, line 75: Can you confirm which of the following would be the recommended usage pattern?

`activity.Attributes.Set("key", "value")`

or

`activity.SetAttribute("key", "value");`

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: No. this intended to be object and not int. I'll fix that now.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: This is raising a good point. Usaers should use 
```C#
activity.SetAttribute("key", "value");
```

and `Activity.Attributes` property is always going to return a new Collection instance (like what the Tags do today). If we think this can be confusing, we can return IEnumerable<...> instead of the collection. This will be less confusing. 

@cijothomas @noahfalk what you think?

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: I fixed it. Thanks for the catch.

accepted/2020/diagnostics/activity-improvements-2.md
cijothomas: `activity.Tags` return `IEnumerable` only. I think making it same for `activity.Attributes` too is the best option. So the only way to set attribute is by using `activity.SetAttribute`.

tarekgh: I am fine with returning IEnumerable (in Activity, ActivityEvent, and ActivitLink)

tarekgh:
I have updated the doc reflecting the final edition. thanks all for your feedback!

accepted/2020/diagnostics/activity-improvements-2.md
```diff
@@ -0,0 +1,152 @@
...
+There are a couple of proposals how to address this issue in the .NET runtime. The idea is the runtime need to generate the trace Id and send it to the samplers. If the sampler using the passed trace id and decide to sample in this case, the Activity which will get created because of that sampler decision will be using the generated trace Id in its context.
+
+The proposal here is to expose the `ActivityListener` property
+
+```c#
+    public sealed class ActivityListener : IDisposable // already existing type
+    {
+        public bool PregenerateNewRootId { get; set;} // other name suggestion: GenerateTraceIdForRootContext
```
cijothomas, line 137: RootId - may get confused with Activity.RootId. 

`GenerateTraceIdForNullParent` - does this sound better? `GenerateTraceIdForRootActivity` is also okay. The term "root" span is used throughout official OT specs.

tarekgh: I added your suggestion in the doc and we discuss that in the design review.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: This is decided and I reflected the design on the doc.

accepted/2020/diagnostics/activity-improvements-2.md
tarekgh: This is decided to not doing any copies and instead the created event object will just hold on the passed input collection.

PR #138, opened 2020-07-10
Mark Windows-specific APIs

terrajobst:


accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+[MinimumOSPlatform("windows7.0")]
+```
+
+## Entire assemblies
+
+There is a set of assemblies that we can mark wholesale as being
+Windows-specific:
+
+> **OPEN**: Should `System.Threading.Overlapped.dll` be removed?
+> **OPEN**: I was told that part of `DirectoryServices` is now cross-platform. Is that true?
```
yaakov-h, line 33: Seems to be.
https://github.com/dotnet/runtime/pull/35380
https://github.com/dotnet/runtime/pull/36669

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **System.Diagnostics.EventLog.dll**
+* **System.Diagnostics.PerformanceCounter.dll**
+* **System.DirectoryServices.dll**
+* **System.DirectoryServices.AccountManagement.dll**
+* **System.DirectoryServices.Protocols.dll**
+* **System.IO.FileSystem.AccessControl.dll**
+* **System.IO.FileSystem.DriveInfo.dll**
+* **System.IO.Pipes.AccessControl.dll**
+* **System.Management.dll**
+* **System.Net.Http.WinHttpHandler.dll**
```
adamsitnik, line 47: It looks like we don't ship `net5.0` version of `System.Net.Http.WinHttpHandler.dll`  (missing `NetCoreAppCurrent` in targets):

https://github.com/dotnet/runtime/blob/a9944fa8a721af673999d074caf1dd74617f6ba6/src/libraries/System.Net.Http.WinHttpHandler/src/System.Net.Http.WinHttpHandler.csproj#L5

does it mean that this assembly should not be annotated with the new attribute?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **System.Runtime.InteropServices.WindowsRuntime.dll**
```
adamsitnik, line 48: we don't ship it anymore? https://www.nuget.org/packages/System.Runtime.InteropServices.WindowsRuntime/

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **System.Security.AccessControl.dll**
+* **System.Security.Cryptography.Cng.dll**
+* **System.Security.Cryptography.ProtectedData.dll**
```
adamsitnik, line 51: we don't ship a 5.0 package, does it mean that this assembly should not be annotated with the new attribute?

https://github.com/dotnet/runtime/blob/f9fc9fc2433a35c2dc5dc2ba1d62136e7e682be4/src/libraries/System.Security.Cryptography.ProtectedData/src/System.Security.Cryptography.ProtectedData.csproj#L4

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **System.Security.Principal.Windows.dll**
+* **System.ServiceProcess.ServiceController.dll**
+* **System.Threading.AccessControl.dll**
```
adamsitnik, line 54: we don't ship a 5.0 package, does it mean that this assembly should not be annotated with the new attribute?

https://github.com/dotnet/runtime/blob/f9fc9fc2433a35c2dc5dc2ba1d62136e7e682be4/src/libraries/System.Threading.AccessControl/src/System.Threading.AccessControl.csproj#L3

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
+# Marking Windows-only APIs
```
akoeplinger, line 1: was the addition of the 14MB platform-comapt-data.xlsx intentional? (commenting here since I can't comment on the binary file)

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **Microsoft.Win32.Registry.dll**
+* **Microsoft.Win32.Registry.AccessControl.dll**
+* **Microsoft.Win32.SystemEvents.dll**
+* **System.Diagnostics.EventLog.dll**
+* **System.Diagnostics.PerformanceCounter.dll**
+* **System.DirectoryServices.dll**
+* **System.DirectoryServices.AccountManagement.dll**
+* **System.DirectoryServices.Protocols.dll**
+* **System.IO.FileSystem.AccessControl.dll**
+* **System.IO.FileSystem.DriveInfo.dll**
```
akoeplinger, line 44: this works fine on Unix, e.g. here's the output of `foreach (var d in DriveInfo.GetDrives()) Console.WriteLine(d.RootDirectory);` on my Mac:

```
/
/dev
/System/Volumes/Data
/private/var/vm
/System/Volumes/Data/home
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `RunAsClient(PipeStreamImpersonationWorker)`
+* **PipeStream**
+    - `set_ReadMode(PipeTransmissionMode)`
+    - `WaitForPipeDrain()`
+* **PipeTransmissionMode**
+    - `Message`
+
+### System.IO.Ports
+
+> **OPEN** Should we mark the entire type/assembly?
```
akoeplinger, line 157: the implementation on dotnet/runtime has multiple .FreeBSD.cs, .Linux.cs etc files so it presumably works cross-platform?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `.ctor(Int32, Int32, String)`
+    - `.ctor(Int32, Int32, String, Boolean)`
+    - `OpenExisting(String)`
+    - `TryOpenExisting(String, Semaphore)`
+* **Thread**
+    - `SetApartmentState(ApartmentState)`
+* **WaitHandle**
+    - `SignalAndWait(WaitHandle, WaitHandle)`
+    - `SignalAndWait(WaitHandle, WaitHandle, Int32, Boolean)`
+    - `SignalAndWait(WaitHandle, WaitHandle, TimeSpan, Boolean)`
```
akoeplinger, line 475: why are these Windows only?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `Create(SocketAddress)`
+* **SocketAddress**
+    - `.ctor(AddressFamily)`
+    - `.ctor(AddressFamily, Int32)`
+    - `get_Family()`
+    - `ToString()`
+
+### System.Net.Sockets
+
+* **Socket**
```
akoeplinger, line 250: most of these APIs don't sound like they are Windows-specific?

accepted/2020/windows-specific-apis/windows-specific-apis.md
stephentoub: System.DirectoryServices.Protocols is supported on Linux now.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+> **OPEN**: I was told that part of `DirectoryServices` is now cross-platform. Is that true?
+
+* **Microsoft.Win32.Registry.dll**
+* **Microsoft.Win32.Registry.AccessControl.dll**
+* **Microsoft.Win32.SystemEvents.dll**
+* **System.Diagnostics.EventLog.dll**
+* **System.Diagnostics.PerformanceCounter.dll**
+* **System.DirectoryServices.dll**
+* **System.DirectoryServices.AccountManagement.dll**
+* **System.DirectoryServices.Protocols.dll**
```
stephentoub, line 42: System.DirectoryServices.Protocols is supported on Linux now.

accepted/2020/windows-specific-apis/windows-specific-apis.md
stephentoub: Yes, DriveInfo is fully available on unix.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **System.IO.FileSystem.AccessControl.dll**
+* **System.IO.FileSystem.DriveInfo.dll**
+* **System.IO.Pipes.AccessControl.dll**
+* **System.Management.dll**
+* **System.Net.Http.WinHttpHandler.dll**
+* **System.Runtime.InteropServices.WindowsRuntime.dll**
+* **System.Security.AccessControl.dll**
+* **System.Security.Cryptography.Cng.dll**
```
stephentoub, line 50: What about marking System.Security.Cryptography.OpenSSL as being Linux-only?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+> **OPEN** Should we mark the entire type?
+
+* **DpapiProtectedConfigurationProvider**
+    - `Decrypt(XmlNode)`
+    - `Encrypt(XmlNode)`
+
+### System.Diagnostics
+
+* **Process**
+    - `GetProcessById(Int32, String)`
```
stephentoub, line 94: This is supported as long as the string refers to the local machine.  This is an example of where the currently proposed mechanism is insufficient.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `GetProcesses(String)`
```
stephentoub, line 95: Ditto

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `GetProcessesByName(String, String)`
```
stephentoub, line 96: Ditto

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+
+### System.Drawing
+
+> **OPEN** Should we mark the entire type?
+
+* **Graphics**
+    - `CopyFromScreen(Int32, Int32, Int32, Int32, Size)`
+    - `CopyFromScreen(Int32, Int32, Int32, Int32, Size, CopyPixelOperation)`
+    - `CopyFromScreen(Point, Point, Size)`
+    - `CopyFromScreen(Point, Point, Size, CopyPixelOperation)`
```
stephentoub, line 122: Don't all of the above work on Linux (and on macOS if the "SYSTEM_DRAWING_COMMON_FORCE_X11" environment variable is set)?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **MemoryMappedFile**
+    - `CreateFromFile(FileStream, String, Int64, MemoryMappedFileAccess, HandleInheritability, Boolean)`
+    - `CreateFromFile(String)`
+    - `CreateFromFile(String, FileMode)`
+    - `CreateFromFile(String, FileMode, String)`
+    - `CreateFromFile(String, FileMode, String, Int64)`
+    - `CreateFromFile(String, FileMode, String, Int64, MemoryMappedFileAccess)`
+    - `CreateNew(String, Int64)`
+    - `CreateNew(String, Int64, MemoryMappedFileAccess)`
+    - `CreateNew(String, Int64, MemoryMappedFileAccess, MemoryMappedFileOptions, HandleInheritability)`
```
stephentoub, line 135: All of the above Create* methods are fully supported on unix as long as the supplied string is null.  This is an example of where the currently proposed mechanism is insufficient.  These should not be marked as Windows-only.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `OpenExisting(String)`
+    - `OpenExisting(String, MemoryMappedFileRights)`
+    - `OpenExisting(String, MemoryMappedFileRights, HandleInheritability)`
+
+### System.IO.Pipes
+
+* **NamedPipeClientStream**
+    - `get_NumberOfServerInstances()`
+* **NamedPipeServerStream**
+    - `RunAsClient(PipeStreamImpersonationWorker)`
```
stephentoub, line 148: This works on unix.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **PipeStream**
+    - `set_ReadMode(PipeTransmissionMode)`
```
stephentoub, line 150: This works on unix as long as it's set to PipeTransmissionMode.Byte.  We shouldn't just blanketly say it's not supported on unix; a more nuanced mechanism is needed.

accepted/2020/windows-specific-apis/windows-specific-apis.md
stephentoub: It does. @krwq added that support.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `set_HeaderWait(TimeSpan)`
+    - `set_MinSendBytesPerSecond(Int64)`
+    - `set_RequestQueue(TimeSpan)`
+* **IPEndPoint**
+    - `Create(SocketAddress)`
+* **SocketAddress**
+    - `.ctor(AddressFamily)`
+    - `.ctor(AddressFamily, Int32)`
+    - `get_Family()`
+    - `ToString()`
```
stephentoub, line 246: None of the above IPEndPoint or SocketAddress APIs are Windows-specific.

accepted/2020/windows-specific-apis/windows-specific-apis.md
stephentoub: They're not.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `IOControl(Int32, Byte[], Byte[])`
+    - `IOControl(IOControlCode, Byte[], Byte[])`
+    - `ReceiveFrom(Byte[], EndPoint)`
+    - `ReceiveFrom(Byte[], Int32, Int32, SocketFlags, EndPoint)`
+    - `ReceiveFrom(Byte[], Int32, SocketFlags, EndPoint)`
+    - `ReceiveFrom(Byte[], SocketFlags, EndPoint)`
+    - `ReceiveFromAsync(SocketAsyncEventArgs)`
+    - `ReceiveMessageFrom(Byte[], Int32, Int32, SocketFlags, EndPoint, IPPacketInformation)`
+    - `ReceiveMessageFromAsync(SocketAsyncEventArgs)`
+    - `SetIPProtectionLevel(IPProtectionLevel)`
```
stephentoub, line 265: All of the above Socket APIs work on unix.

The only one that's more nuanced is IOControl, where it only works for some control codes.  For that one, since it looks like we have the ability to annotate enum members as being Windows-only, we should do so for the ones that aren't supported.  The three that are supported as of now are IOControlCode.NonBlockingIO, IOControlCode.DataToRead, and IOControlCode.OobDataRead.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **SocketTaskExtensions**
+    - `AcceptAsync(Socket)`
+    - `AcceptAsync(Socket, Socket)`
+    - `ReceiveFromAsync(Socket, ArraySegment<Byte>, SocketFlags, EndPoint)`
+    - `ReceiveMessageFromAsync(Socket, ArraySegment<Byte>, SocketFlags, EndPoint)`
```
stephentoub, line 270: All of the above are supported on unix.

AcceptAsync(Socket, Socket) is more nuanced, in that it works fine on Unix as long as the second Socket argument is null.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **TcpListener**
+    - `AllowNatTraversal(Boolean)`
+* **TransmitFileOptions**
+    - `Disconnect`
+    - `ReuseSocket`
```
stephentoub, line 275: Both of the above options work on unix.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `.ctor(Int32, CspParameters)`
+    - `get_CspKeyContainerInfo()`
+    - `ImportCspBlob(Byte[])`
+
+### System.Security.Cryptography.X509Certificates
+
+* **X509Chain**
+    - `.ctor(IntPtr)`
+
+### System.ServiceModel
```
stephentoub, line 346: Someone familiar with WCF will need to comment on this section.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+* **SecurityNegotiationException**
+    - `.ctor(SerializationInfo, StreamingContext)`
+* **X509ServiceCertificateAuthentication**
+    - `set_CertificateValidationMode(X509CertificateValidationMode)`
+
+### System.Threading
+
+* **EventWaitHandle**
+    - `.ctor(Boolean, EventResetMode, String)`
+    - `.ctor(Boolean, EventResetMode, String, Boolean)`
```
stephentoub, line 462: The above constructors work fine on unix, as long as the string is null.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `OpenExisting(String)`
+    - `TryOpenExisting(String, EventWaitHandle)`
+* **Semaphore**
+    - `.ctor(Int32, Int32, String)`
+    - `.ctor(Int32, Int32, String, Boolean)`
```
stephentoub, line 467: The above constructors work fine on unix, as long as the string is null.

accepted/2020/windows-specific-apis/windows-specific-apis.md
stephentoub: @kouvel, do these not work on unix?

kouvel: They should work on Unix too, I think they were fixed to work since 2.1 with https://github.com/dotnet/coreclr/pull/16383

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: @terrajobst noted during the API review meeting that this was intentional so that the analysis that went into this could be captured. The thought was that since this is the `designs` repo, it's OK to include the binaries in it.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: We still want to apply the attribute. Since we cannot reference the attribute from `net5.0` though, we will need to create an internal implementation of the attribute within this assembly using the same namespace and type name. This is the same approach that was taken for nullable annotations.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: We can remove `WindowsRuntime` from the list

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: As was discussed during the review meeting, we need to design a means for a _Platform Exclusion_ attribute and use that for `OpenSSL`. While we don't expect to ship such a feature in .NET 5, we need to make sure the design is baked before .NET 5 ships to be certain we have a path forward.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: We can use an internal attribute for this one too.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: We can use an internal attribute for this one too.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: We will remove this from the list. The general guidance is that for cases where some argument values or combinations are supported cross-platform, we will not mark the API as platform-specific.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: We will remove this from the list

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: We will remove this from the list

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: I noted down that we would remove the `Create` methods but leave the `Open` methods in the list here. I am also removing the `CreateOrOpen` methods since they also look to unconditionally throw on Unix. Please let me know if I've misunderstood those though.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: Removing all of `System.IO.Ports` from the list.

accepted/2020/windows-specific-apis/windows-specific-apis.md
jeffhandley: I'm removing those Socket APIs from the list (except for `SetIPProtectionLevel`) and I'm adding the `IOControlCode` enum members other than `NonBlockingIO`, `DataToRead`, and `OobDataRead`.

accepted/2020/windows-specific-apis/windows-specific-apis.md
stephentoub: > I am also removing the CreateOrOpen methods since they also look to unconditionally throw on Unix.

CreateFromFile and CreateNew are not Windows-only. CreateOrOpen and OpenExisting are Windows-only.

jeffhandley: OK got it; adding `CreateOrOpen` back into the list. Thanks.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+### System.Diagnostics
+
+* **Process**
+    - `GetProcessById(Int32, String)`
+    - `GetProcesses(String)`
+    - `GetProcessesByName(String, String)`
+    - `set_MaxWorkingSet(IntPtr)`
+    - `set_MinWorkingSet(IntPtr)`
+    - `Start(String, String, SecureString, String)`
+    - `Start(String, String, String, SecureString, String)`
```
adamsitnik, line 100: what about `EnterDebugMode()` and `LeaveDebugMode()` which are nops on non-Windows?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `get_LoadUserProfile()`
+    - `get_Password()`
+    - `get_PasswordInClearText()`
+    - `set_Domain(String)`
+    - `set_LoadUserProfile(Boolean)`
+    - `set_Password(SecureString)`
+    - `set_PasswordInClearText(String)`
+* **ProcessThread**
+    - `set_PriorityLevel(ThreadPriorityLevel)`
+    - `set_ProcessorAffinity(IntPtr)`
```
adamsitnik, line 112: `IdealProcessor`, `ResetIdealProcessor` are nops on Unix. Should they be marked too?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `GetComInterfaceForObject<T, TInterface>(T)`
+    - `GetIUnknownForObject(Object)`
+    - `GetNativeVariantForObject(Object, IntPtr)`
+    - `GetNativeVariantForObject<T>(T, IntPtr)`
+    - `GetObjectForIUnknown(IntPtr)`
+    - `GetObjectForNativeVariant(IntPtr)`
+    - `GetObjectForNativeVariant<T>(IntPtr)`
+    - `GetObjectsForNativeVariants(IntPtr, Int32)`
+    - `GetObjectsForNativeVariants<T>(IntPtr, Int32)`
+    - `GetStartComSlot(Type)`
```
adamsitnik, line 304: what about `GetEndComSlot`?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,475 @@
...
+    - `GetTypedObjectForIUnknown(IntPtr, Type)`
+    - `GetTypeFromCLSID(Guid)`
+    - `GetTypeInfoName(ITypeInfo)`
+    - `GetUniqueObjectForIUnknown(IntPtr)`
+    - `QueryInterface(IntPtr, Guid, IntPtr)`
+    - `Release(IntPtr)`
+    - `ReleaseComObject(Object)`
```
adamsitnik, line 204: there are few more methods in `Marshal` that throw `PlatformNotSupportedException`:

* `CreateWrapperOfType(object, Type)`
* `CreateWrapperOfType<T, TWrapper>(T)`
* `GetComObjectData(object,object)`
* `SetComObjectData(object,object,object)`
* `GetIDispatchForObject(object)`

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/Runtime/InteropServices/Marshal.NoCom.cs

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,360 @@
...
+
+### System.Security.Cryptography
+
+* **CspKeyContainerInfo**
+    - `get_Accessible()`
```
adamsitnik, line 209: `ctor` is missing: https://github.com/dotnet/runtime/blob/ecdf9de474459f17efb90fd64609d164bcf454d4/src/libraries/System.Security.Cryptography.Csp/src/System/Security/Cryptography/CspKeyContainerInfo.Unix.cs#L8

```suggestion
    - `ctor()`
    - `get_Accessible()`
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,360 @@
...
+    - `get_HardwareDevice()`
+    - `get_KeyContainerName()`
+    - `get_KeyNumber()`
+    - `get_MachineKeyStore()`
+    - `get_Protected()`
+    - `get_ProviderName()`
+    - `get_ProviderType()`
+    - `get_RandomlyGenerated()`
+    - `get_Removable()`
+    - `get_UniqueKeyContainerName()`
```
adamsitnik, line 220: enitre type is Windows specifc, it's impossible to create it on Unix: https://github.com/dotnet/runtime/blob/ecdf9de474459f17efb90fd64609d164bcf454d4/src/libraries/System.Security.Cryptography.Csp/src/System/Security/Cryptography/CspKeyContainerInfo.Unix.cs#L6-L25

```suggestion
* **CspKeyContainerInfo** 
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,360 @@
...
+* **DSACryptoServiceProvider**
+    - `get_CspKeyContainerInfo()`
+    - `ImportCspBlob(Byte[])`
```
adamsitnik, line 223: the `ImportCspBlob` is supported on Unix, the `ctors` are not:

https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Security.Cryptography.Csp/src/System/Security/Cryptography/DSACryptoServiceProvider.Unix.cs#L105

https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Security.Cryptography.Csp/src/System/Security/Cryptography/DSACryptoServiceProvider.Unix.cs#L40-L48

```suggestion
    - `ctor(CspParameters)`
    - `ctor(int, CspParameters)`
    - `get_CspKeyContainerInfo()`
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,360 @@
...
+* **PasswordDeriveBytes**
+    - `CryptDeriveKey(String, String, Int32, Byte[])`
+* **RC2CryptoServiceProvider**
+    - `set_UseSalt(Boolean)`
+* **RSACryptoServiceProvider**
+    - `.ctor(CspParameters)`
+    - `.ctor(Int32, CspParameters)`
+    - `get_CspKeyContainerInfo()`
+    - `ImportCspBlob(Byte[])`
```
adamsitnik, line 232: looks like it is supported: https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Security.Cryptography.Csp/src/System/Security/Cryptography/RSACryptoServiceProvider.Unix.cs#L135-L139

```suggestion
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: I've added a note on partially supported APIs

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: Agreed, that's what @bartonjs also said

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: Added

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: Removed

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,378 @@
...
+* **SecurityNegotiationException**
+    - `.ctor(SerializationInfo, StreamingContext)`
+* **X509ServiceCertificateAuthentication**
+    - `set_CertificateValidationMode(X509CertificateValidationMode)`
+
+### System.Threading
+
+* **EventWaitHandle**
+    - `OpenExisting(String)`
+    - `TryOpenExisting(String, EventWaitHandle)`
```
adamsitnik, line 373: I am not sure how to deal with the ctor:

it calls `CreateEventCore`:

https://github.com/dotnet/runtime/blob/ecdf9de474459f17efb90fd64609d164bcf454d4/src/libraries/System.Private.CoreLib/src/System/Threading/EventWaitHandle.cs#L22-L27

which throws on Unix `if name != null`

https://github.com/dotnet/runtime/blob/6072e4d3a7a2a1493f514cdf4be75a3d56580e84/src/libraries/System.Private.CoreLib/src/System/Threading/EventWaitHandle.Windows.cs#L19-L23

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: Added

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: Added

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: I'm on the fence for those. On the one hand, if the APIs don't throw it might be overkill. On the other hand, if the APIs don't do what I want, it's also not great. I'd say for now let's err on the side of reducing false positive and not mark those.

/cc @stephentoub

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: Fixed

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: See above. We decided to mark those because these APIs work as long as name is null.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,378 @@
...
+* **Semaphore**
+    - `OpenExisting(String)`
+    - `TryOpenExisting(String, Semaphore)`
+* **Thread**
+    - `SetApartmentState(ApartmentState)`
```
adamsitnik, line 399: what about the `ApartmentState` (it's obsolete) and `TrySetApartmentState`? the seem to do the same thing

accepted/2020/windows-specific-apis/windows-specific-apis.md
stephentoub: Yeah, let's leave them unannotated.  Lots of things end up being best-effort, and the cleanest bar right now is whether the API is unusable because it throws an exception unconditionally.

accepted/2020/windows-specific-apis/windows-specific-apis.md
terrajobst: @stephentoub looked it up. `TrySetApartmentState` always returns `false` on non-Windows and `SetAparmentState` throws if `TrySetApartmentState` returns `false`. Hence, `SetAparmentState` should be marked by `TrySetApartmentState` shouldn't.

adamsitnik: @terrajobst thanks!

terrajobst:
FYI. I've done a run for .NET 5 and the latest `Microsoft.Windows.Compatibility` pack. Once I've merged the delta, I'll flip this back to api-ready-for-review.

terrajobst:
@akoeplinger I've rebased out the Excel file 🙂

accepted/2020/windows-specific-apis/windows-specific-apis.md
mconnew: We have zero api's which are Windows specific.

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,402 @@
...
+of arguments will not be marked as platform-specific. In other words, we'll
+optimize for fewer false positives.
+
+In some cases, we get away with marking enum members as platform-specific. If
+the API is popular enough, we may want to provide a custom analyzer that checks
+for certain arguments.
+
+## Open Issues
+
+* We should run the tool again .NET 5
```
jeffhandley, line 47:
```suggestion
* We should run the tool against .NET 5
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,402 @@
...
+
+* **MessageSecurityException**
+    - `.ctor(SerializationInfo, StreamingContext)`
+* **SecurityAccessDeniedException**
+    - `.ctor(SerializationInfo, StreamingContext)`
+* **SecurityNegotiationException**
+    - `.ctor(SerializationInfo, StreamingContext)`
+* **X509ServiceCertificateAuthentication**
+    - `set_CertificateValidationMode(X509CertificateValidationMode)`
+
```
jeffhandley, line 392: Per @mconnew's comment, this suggestion will remove the `ServiceModel` types.

```suggestion
```

stephentoub: Do we know what caused them to show up here in the first place?

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,402 @@
...
+### System.Configuration
+
+* **DpapiProtectedConfigurationProvider**
+
+### System.Diagnostics
+
+* **Process**
+    - `EnterDebugMode()`
+    - `LeaveDebugMode()`
```
adamsitnik, line 122: these two are `nops` and to be consistent with other nops, we should remove them from the list

context: https://github.com/dotnet/designs/pull/138/files#r454448405

```suggestion
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,402 @@
...
+    - `CreateOrOpen(String, Int64, MemoryMappedFileAccess)`
+    - `CreateOrOpen(String, Int64, MemoryMappedFileAccess, MemoryMappedFileOptions, HandleInheritability)`
+    - `OpenExisting(String)`
+    - `OpenExisting(String, MemoryMappedFileRights)`
+    - `OpenExisting(String, MemoryMappedFileRights, HandleInheritability)`
+
+### System.IO.Pipes
+
+* **NamedPipeClientStream**
+    - `Connect(Int32)`
```
adamsitnik, line 158: I can't see a code path where `Connect(Int32)` throws on Unix.

https://github.com/dotnet/runtime/blob/08643df14054d1273002254c909fa099649ebfcf/src/libraries/System.IO.Pipes/src/System/IO/Pipes/NamedPipeClientStream.cs#L130-L173

https://github.com/dotnet/runtime/blob/08643df14054d1273002254c909fa099649ebfcf/src/libraries/System.IO.Pipes/src/System/IO/Pipes/NamedPipeClientStream.Unix.cs#L20-L67

```suggestion
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,402 @@
...
+    - `get_NumberOfServerInstances()`
+* **NamedPipeServerStream**
+    - `WaitForConnection()`
```
adamsitnik, line 161: it's supported on Unix: https://github.com/dotnet/runtime/blob/08643df14054d1273002254c909fa099649ebfcf/src/libraries/System.IO.Pipes/src/System/IO/Pipes/NamedPipeServerStream.Unix.cs#L54-L66

```suggestion
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,402 @@
...
+    - `KeepAliveValues`
+    - `AbsorbRouterAlert`
+    - `UnicastInterface`
+    - `LimitBroadcasts`
+    - `BindToInterface`
+    - `MulticastInterface`
+    - `AddMulticastGroupOnInterface`
+    - `DeleteMulticastGroupFromInterface`
+* **Socket**
+    - `SetIPProtectionLevel(IPProtectionLevel)`
```
adamsitnik, line 210: https://github.com/dotnet/runtime/blob/08643df14054d1273002254c909fa099649ebfcf/src/libraries/System.Net.Sockets/src/System/Net/Sockets/Socket.Unix.cs#L13-L17

```suggestion
* **Socket**
    - `.ctor(SocketInformation)`
    - `SetIPProtectionLevel(IPProtectionLevel)`
```

accepted/2020/windows-specific-apis/windows-specific-apis.md
```diff
@@ -0,0 +1,402 @@
...
+    - `set_PasswordInClearText(String)`
+* **ProcessThread**
+    - `set_PriorityLevel(ThreadPriorityLevel)`
+    - `set_ProcessorAffinity(IntPtr)`
+
+### System.IO
+
+* **DriveInfo**
+    - `set_VolumeLabel(String)`
+
```
adamsitnik, line 144: https://github.com/dotnet/runtime/blob/1bdcd669e70d687bb0c56fe20aafa7c104b6aa0e/src/libraries/System.IO.FileSystem/src/System/IO/FileSystem.Unix.cs#L50-L58

```suggestion
### System.IO.FileSystem

* **File**
    - `Decrypt(string)`
    - `Encrypt(string)`
```

adamsitnik:
FWIW I've done some simple manual search of keyword "PlatformNotSupportedException"  in `*Unix.cs` files and found only 2 things that were not on the list and provided a suggestion that adds them. IMO the list is complete now.

My PR is ready https://github.com/dotnet/runtime/pull/39265

terrajobst:
The [proposal](https://github.com/dotnet/designs/pull/138) looks good.

PR #139, opened 2020-07-13
Add spec for obsoletions in .NET 5

jeffhandley:


Happypig375:
https://aka.ms/dotnet-warnings/MSLIB0004 doesn't work yet?

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,141 @@
...
+        * `System.Net.AuthenticationManager.PreAuthenticate`
+8. These Remoting APIs are no longer supported and throw PlatformNotSupportedException
+    * DiagnosticId: `MSLIB0012`
+    * APIs:
+        * `System.MarshalByRefObject.GetLifetimeService`
+        * `System.MarshalByRefObject.InitializeLifetimeService`
+9. Code Access Security (CAS) APIs are no longer supported
+    * DiagnosticId: `MSLIB0013`
+    * APIs: (all from the `System.Security.Permissions` namespace)
+        * Classes: (everything *except* `PrincipalPermission` (still implemented) and `PrincipalPermissionAttribute` (already obsolete))
```
jkotas, line 68: Is the fact that it happens to be implemented as compat quirk a good reason for not obsoleting it? I think it would be better to obsolete it too.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,141 @@
...
+    * APIs:
+        * `System.Runtime.ConstrainedExecution.PrePrepareMethodAttribute`
+        * `System.Runtime.ConstrainedExecution.ReliabilityContractAttribute`
+        * `System.Runtime.ConstrainedExecution.Cer`
+        * `System.Runtime.ConstrainedExecution.Consistency`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.ExecuteCodeWithGuaranteedCleanup`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.PrepareConstrainedRegions`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.PrepareConstrainedRegionsNoOP`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.ProbeForSufficientStack`
+2. This App Domain method is no longer implemented
```
stephentoub, line 27:
```suggestion
2. This AppDomain method is no longer implemented
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,141 @@
...
+    * DiagnosticId: `MSLIB0005`
+    * APIs:
+        * `PrepareContractedDelegate`
+2. The Global Assembly Cache is no longer implemented
+    * DiagnosticId: `MSLIB0006`
+    * APIs:
+        * `System.Reflection.Assembly.GlobalAssemblyCache`
+3. `System.Net.WebSockets.WebSocket.IsApplicationTargeting45` is no longer applicable
```
stephentoub, line 35: This is already marked Obsolete.  Are we saying we're just going to give it a diagnostic id / message?

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,141 @@
...
+        * `System.Threading.Thread.Abort()`
+        * `System.Threading.Thread.Abort(Object)`
+5. The default implementations of these cryptography algorithms are no longer supported and throw PlatformNotSupportedException
+    * DiagnosticId: `MSLIB0009`
+    * APIs:
+        * `System.Security.Cryptography.SymmetricAlgorithm.Create()`
+        * `System.Security.Cryptography.AssymetricAlgorithm.Create()`
+        * `System.Security.Cryptography.HMAC.Create()`
+        * `System.Security.Cryptography.KeyedHashAlgorithm.Create()`
+6. The CreatePbdGenerator API is no longer supported and throws PlatformNotSupportedException
```
stephentoub, line 51: I don't see this method anywhere in dotnet/runtime.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
jeffhandley: @GrabYourPitchforks - what do you think?

accepted/2020/better-obsoletion/obsoletions-in-net5.md
jeffhandley: Ah, I missed that. It's probably not worth it to add the id/url for this one (as we'd then need to have a page for the URL too).  I'll remove this one.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
jeffhandley: There was a typo (it should be Create**Pdb**Generator). Sorry about that.

* The ref is in `System.Linq.Expressions\ref\System.Linq.Expressions.cs` in the `DebugInfoGenerator` class
* The implementation is in `System.Linq.Expressions\src\System\Runtime\CompilerServices\DebugInfoGenerator.cs`

jeffhandley:
> https://aka.ms/dotnet-warnings/MSLIB0004 doesn't work yet?

Thanks for checking that. We'll get the URLs working before shipping the GA release in November.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
GrabYourPitchforks: I'm fine with obsoleting it (as warning). Just didn't want to push the issue since it wasn't a "slam dunk" like the other APIs, which all either no-op or throw PNSE.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,137 @@
...
+        * `System.Runtime.ConstrainedExecution.Cer`
+        * `System.Runtime.ConstrainedExecution.Consistency`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.ExecuteCodeWithGuaranteedCleanup`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.PrepareConstrainedRegions`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.PrepareConstrainedRegionsNoOP`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.ProbeForSufficientStack`
+1. This AppDomain method is no longer implemented
+    * DiagnosticId: `MSLIB0005`
+    * APIs:
+        * `PrepareContractedDelegate`
```
GrabYourPitchforks, line 30: This should be in the __CER__ bucket above.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
jeffhandley: @GrabYourPitchforks Two questions:

1. Should we group all of the CAS stuff into the same DiagnosticId already in use for `PrincipalPermissionAttribute` instead of creating a new one?
1. Should we use a different DiagnosticId for `PrincipalPermission` since it's less of a slam dunk?

GrabYourPitchforks: For (1), yes. Honestly there wasn't much of a point to creating a standalone diagnostic code just for `PrincipalPermissionAttribute` since it's marked _as error_ and you can't suppress it. So may as well reuse that same code for this bucket.

For (2), we plan on having a fixer at some point. IMO we should only use a different code if doing so would make the fixer easier to write / apply. I'm not really familiar with that feature, though, so somebody like @bartonjs or @stephentoub might know more.

jeffhandley: Thanks. For (2), that is enough for me to want to use a separate DiagnosticId for `PrincipalPermission`. It can be suppressed separately then too.

jeffhandley: Leaving this unresolved to give @bartonjs and @stephentoub a chance to weigh in. But I've now assigned all of the CAS items into the existing `MSLIB0002` DiagnosticId and then used the new `MSLIB0012` for `PrincipalPermission` and `PermissionState`.

bartonjs: Fixers are activated by diagnostic IDs, merely having the right ID makes the lightbulb show up (I'm given to understand).  So anything with a fixer gets a different code than things without a fixer.  (If one fixer can fix more than one sub-problem but have a sensible display name, one code is fine.  Or it can activate over multiple codes.)

bartonjs: This also effectively means anything bucketed into MSLIB0012 now can never gain a fixer.

jkotas: > This also effectively means anything bucketed into MSLIB0012 now can never gain a fixer.

I am not following. Can you expand on this?

AustinWise:
For CAS related items, the documented says "all [classes] from the `System.Security.Permissions` namespace" will be marked obsolete. What about other permissions like `System.Net.SocketPermission`?

It looks like the whole of the  `System.Security.Permissions` are shims that either do nothing or throw PNSE. Examples include:

* `Evidence`
* `EvidenceBase` and subclasses
* `HostSecurityManager`
* `SecurityContext`
* A few members of `SecurityManager` have not already been obsoleted.

There are also some CAS types in CorLib like `SecurityTransparentAttribute`. Some, like `SecurityTreatAsSafeAttribute`, are already obsolete but the message does not point the user in the right direction: "SecurityTreatAsSafe is only used for .NET 2.0 transparency compatibility.  Please use the SecuritySafeCriticalAttribute instead.".

If not all CAS types are obsoleted, it may be worth giving a rationale in the document.

GrabYourPitchforks:
> What about other permissions like System.Net.SocketPermission?

That's a good point. The obsoletions should probably include everything that subclasses `CodeAccessSecurityAttribute` or which implements `IPermission`, regardless of what namespace it's in. I suspect we'll discover some additional types not emcompassed by the above list once we mark the base types / interfaces obsolete and see what compiler failures fall out of the libraries code.

jeffhandley:
> The obsoletions should probably include everything that subclasses `CodeAccessSecurityAttribute` or which implements `IPermission`

* I'll add all derivatives of `CodeAccessSecurityAttribute` and add them to the CAS category
* I can also find the implementations of `IPermission`; should `IPermission` itself be in the list?

GrabYourPitchforks:
@jeffhandley Yeah, I'd also add `IPermission` and `IStackWalk` to the list, along with everything implementing them or exposing them. Not sure how much creep that will add though.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,236 @@
...
+            * `SecurityPermissionFlag`
+            * `StorePermissionFlags`
+            * `TypeDescriptorPermissionFlags`
+            * `UIPermissionClipboard`
+            * `UIPermissionWindow`
+            * `WebBrowserPermissionLevel`
+1. PrincipalPermission is no longer supported
+    * DiagnosticId: `MSLIB0012`
+    * APIs:
+        * `System.Security.Permissions.PrincipalPermission`
```
jkotas, line 235: @bartonjs Could you please comment on https://github.com/dotnet/designs/pull/139#discussion_r453846088 ?

I do not see why `PrincipalPermission` should be its own bucket separate from all other permissions.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
jeffhandley: @mavasani Could you help us make sure we understand this fully? If multiple conditions share a DiagnosticId, and then we later want to offer a fixer for _one_ of those conditions, would we be able to?  As @bartonjs and I understand it, DiagnosticId values need to be unique for diagnostics with fixers available.

I'm also thinking about suppression for this too.  Since `PrincipalPermission` might have use cases remaining when the other CAS-related APIs don't, I wouldn't want someone to suppress `PrincipalPermission` and inadvertently suppress all other API obsoletions from the CAS category.

GrabYourPitchforks:
Another addition for the CAS bucket: the enum value `SecurityAction.PermitOnly` should be obsolete.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
mavasani: > If multiple conditions share a DiagnosticId, and then we later want to offer a fixer for one of those conditions, would we be able to

Diagnostics can attach a custom set of key-value pair `properties` to itself, and fixer can use these properties to determine whether or not to fix a specific diagnostic. This enables reporting multiple diagnostics with same ID, when fix is desired to be offered only for a subset of these.

https://github.com/dotnet/roslyn/blob/7023b0bbd4ca161ff6005b1219da1bc60b0fa58e/src/Compilers/Core/Portable/Diagnostic/Diagnostic.cs#L98-L102

jeffhandley: Thanks, @mavasani. The `ObsoleteAttribute` diagnostics probably can't do this though, since Roslyn is generating the diagnostics from the `ObsoleteAttribute` for us. We'd have no way of attaching the key-value pairs (that I can think of).

mavasani: > I'm also thinking about suppression for this too. Since PrincipalPermission might have use cases remaining when the other CAS-related APIs don't, I wouldn't want someone to suppress PrincipalPermission and inadvertently suppress all other API obsoletions from the CAS category.

Suppression/Disabling an ID is a valid scenario of concern - currently, Roslyn compiler only supports disabling diagnostic set by IDs, a finer granularity configuration where a specific sub-bucket of diagnostics with shared ID can be turned off is not yet supported. @sharwell has proposed this feature here: https://github.com/dotnet/roslyn/issues/27712, but this needs design approval from Roslyn compiler team first.

jkotas: > `PrincipalPermission` might have use cases remaining 

Those use look pretty weak. I would not feel bad about people having to add warning disable around those if they want to keep them.

jeffhandley:
@GrabYourPitchforks I'm considering keeping `SecurityAction.PermitOnly` in the list but noting that we should use the same format of the `[Obsolete]` attribute that several other members of `SecurityAction` already have:

```
[Obsolete("Assembly level declarative security is obsolete and is no longer enforced by the CLR by default. See https://go.microsoft.com/fwlink/?LinkID=155570 for more information.")]
```

That is currently applied to `Deny`, `RequestMinimum`, `RequestOptional`, and `RequestRefuse`.

What do you think?  Should we obsolete it that way, assign it into the CAS DiagnosticId, or omit it from the obsoletions list?

jkotas:
> See https://go.microsoft.com/fwlink/?LinkID=155570

This link has information specific to .NET Framework 2.0 -> .NET Framework 4.0 specific CAS breaking change that is not  relevant to .NET Core. We should delete it in .NET Core and supersede it with bulk CAS obsoletion.

I think we should:
- Mark all types related to CAS obsolete. Keep it simply and mark everything. I do not think it is worth trying to reason about one-off cases that somebody somewhere may use. It just makes the story more complicated.
- The auto-fixer should be conservative and not remove code that might have been doing something important. https://github.com/dotnet/runtime/issues/39235 is a good start.

jeffhandley:
@jkotas / @GrabYourPitchforks - I went ahead and combined the PrincipalPermission and PermissionState into the CAS category.

Joe4evr:
I've also made a list of candidate Obsolete APIs a few months back: dotnet/runtime#33360
I think it would be good to cross-reference with that list and the discussions pointed to, since there's a few on there that I don't see here.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+
+With `dotnet/runtime` PR [#33248](https://github.com/dotnet/runtime/pull/33248),
+the `ObsoleteAttribute` gained new `DiagnosticId` and `UrlFormat` properties. With
+these properties, obsoletions can be grouped together when multiple APIs are
+marked as obsolete for the same reason. This capability is reflected in the list
+of obsoletions below. All of these obsoletions will use the `UrlFormat` of
+`https://aka.ms/dotnet-warnings/{0}`.
+
+1. The Constrained Execution Region (CER) feature is no longer implemented
+    * DiagnosticId: `MSLIB0004`
```
danmosemsft, line 17: Per offline conversation I would much prefer not using any MS/Microsoft prefix for efforts like this relating to the base libraries, which are a dotnet foundation project.

jeffhandley: We're getting consensus around `SYSLIB`. I'll move forward with that.

jeffhandley, line 17:
```suggestion
    * DiagnosticId: `SYSLIB0004`
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+        * `System.Runtime.ConstrainedExecution.ReliabilityContractAttribute`
+        * `System.Runtime.ConstrainedExecution.Cer`
+        * `System.Runtime.ConstrainedExecution.Consistency`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.ExecuteCodeWithGuaranteedCleanup`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.PrepareConstrainedRegions`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.PrepareConstrainedRegionsNoOP`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.PrepareContractedDelegate`
+        * `System.Runtime.CompilerServices.RuntimeHelpers.ProbeForSufficientStack`
+1. The Global Assembly Cache is no longer implemented
+    * DiagnosticId: `MSLIB0005`
```
jeffhandley, line 29:
```suggestion
    * DiagnosticId: `SYSLIB0005`
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+    * APIs:
+        * `System.Reflection.Assembly.GlobalAssemblyCache`
+1. `Thread.Abort` is no longer supported and throws PlatformNotSupportedException
+    * DiagnosticId: `MSLIB0006`
```
jeffhandley, line 33:
```suggestion
    * DiagnosticId: `SYSLIB0006`
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+    * APIs:
+        * `System.Threading.Thread.Abort()`
+        * `System.Threading.Thread.Abort(Object)`
+1. The default implementations of these cryptography algorithms are no longer supported and throw PlatformNotSupportedException
+    * DiagnosticId: `MSLIB0007`
```
jeffhandley, line 38:
```suggestion
    * DiagnosticId: `SYSLIB0007`
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+    * APIs:
+        * `System.Security.Cryptography.SymmetricAlgorithm.Create()`
+        * `System.Security.Cryptography.AssymetricAlgorithm.Create()`
+        * `System.Security.Cryptography.HMAC.Create()`
+        * `System.Security.Cryptography.KeyedHashAlgorithm.Create()`
+1. The CreatePdbGenerator API is no longer supported and throws PlatformNotSupportedException
+    * DiagnosticId: `MSLIB0008`
```
jeffhandley, line 45:
```suggestion
    * DiagnosticId: `SYSLIB0008`
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+    * APIs:
+        * `System.Runtime.CompilerServices.DebugInfoGenerator.CreatePdbGenerator`
+1. The AuthenticationManager Authenticate and PreAuthenticate APIs are no longer supported and throw PlatformNotSupportedException
+    * DiagnosticId: `MSLIB0009`
```
jeffhandley, line 49:
```suggestion
    * DiagnosticId: `SYSLIB0009`
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+    * APIs:
+        * `System.Net.AuthenticationManager.Authenticate`
+        * `System.Net.AuthenticationManager.PreAuthenticate`
+1. These Remoting APIs are no longer supported and throw PlatformNotSupportedException
+    * DiagnosticId: `MSLIB0010`
```
jeffhandley, line 54:
```suggestion
    * DiagnosticId: `SYSLIB0010`
```

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,213 @@
...
+    * APIs:
+        * `System.MarshalByRefObject.GetLifetimeService`
+        * `System.MarshalByRefObject.InitializeLifetimeService`
+1. Code Access Security (CAS) APIs are no longer supported
+    * DiagnosticId: `MSLIB0002` (reusing the existing DiagnosticId from `PrincipalPermissionAttribute`)
```
jeffhandley, line 59:
```suggestion
    * DiagnosticId: `SYSLIB0002` (reusing the existing DiagnosticId from `PrincipalPermissionAttribute`)
```

jeffhandley:
@danmosemsft @Joe4evr - I just added an "Other Considerations" section that talks to why we chose the APIs we did for .NET 5.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
```diff
@@ -0,0 +1,234 @@
...
+    * APIs:
+        * `System.Threading.Thread.Abort()`
+        * `System.Threading.Thread.Abort(Object)`
+1. The default implementations of these cryptography algorithms are no longer supported and throw PlatformNotSupportedException
+    * DiagnosticId: `SYSLIB0007`
+    * APIs:
+        * `System.Security.Cryptography.SymmetricAlgorithm.Create()`
+        * `System.Security.Cryptography.AssymetricAlgorithm.Create()`
+        * `System.Security.Cryptography.HMAC.Create()`
+        * `System.Security.Cryptography.KeyedHashAlgorithm.Create()`
```
jeffhandley, line 43: @bartonjs `HashAlgorithm.Create()` also throws (and there's a unit test including it in the same list as these others). Should we obsolete it as well?

terrajobst:
Review notes:

* Can we turn off all obsoletions? I'll follow up with the I'll file an issue.
* We shouldn't reuse `PrincipalPermissionAttribute` b/c it's an error (we shouldn't be using the same ID for different severities)
* We should avoid using the same if for where some of the APIs will have fixers and some of them don't

jeffhandley:
Thanks, @terrajobst. 

> * We shouldn't reuse `PrincipalPermissionAttribute` b/c it's an error (we shouldn't be using the same ID for different severities)

I will keep `PrincipalPermissionAttribute` as `SYSLIB0002` (as an error) and then I will add the rest of CAS into `SYSLIB0003` (as a warning).

Sorry @GrabYourPitchforks -- this will steal `SYSLIB0003` out from under you.

jeffhandley:
The way it's playing out with `PrincipalPermissionAttribute` is:

1. I'll leave the existing diagnostic on its constructor (as an error)
2. I'll apply the new diagnostic to the type (as a warning)

We cannot put errors on types because of our type forwarding infrastructure--the type forward files would then fail to build because they are referencing types that have obsoletions as errors, which cannot be suppressed.

We cannot put all of the new diagnostics on constructors since we're obsoleting the interfaces; the types themselves need to then be obsoleted as well.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
jeffhandley: I had combined it in but then ended up having to keep the `PrincipalPermissionAttribute` constructor as its own ID since it's an **error** level and the others are **warnings**. But `PrincipalPermission` falls into the general CAS category.

bartonjs:
> I will keep PrincipalPermissionAttribute as SYSLIB0002 (as an error) and then I will add the rest of CAS into SYSLIB0003 (as a warning).

They don't need to be adjacent.  In fact, being not adjacent early on will show that we don't care about such trivialities.

accepted/2020/better-obsoletion/obsoletions-in-net5.md
bartonjs: > HashAlgorithm.Create() also throws (and there's a unit test including it in the same list as these others). Should we obsolete it as well?

Yep.

terrajobst:
Yeah, let's not do unnatural acts for aesthetics. We can't keep it up anyways. We should think of SYSLIBXXXX as an opaque ID.

Gnbrkm41:
Thread.Suspend seems to be missing (it's unsupported along with `Thread.Abort` I believe) from the list.

jeffhandley:
> Thread.Suspend seems to be missing (it's unsupported along with `Thread.Abort` I believe) from the list.

`Thread.Suspend` and `Thread.Resume` have already been obsoleted. Although we should consider unifying all of the obsoleted `Thread` APIs into a single diagnostic ID and message (`SYSLIB0006`).  @terrajobst -- any objections to that?

terrajobst:
Depends. If we're changing diagnostic IDs for already shipped obsolete messages, we probably want to consider this separately. I'm not opposed to that, but we shouldn't do one-offs without considering the larger picture.

If we only shipped the obsoletions in a preview of .NET 5, feel feee to change it.

PR #143, opened 2020-07-18
Add spec for platform exclusion

terrajobst:


accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+    {
+        public UnsupportedOSPlatformAttribute(string platformName);
+    }
+```
+
+The semantics of this attribute are as follows:
+
+* Marks an assembly, module, or API as unsupported on a given OS platform.
+* The `platformName` can contain no version number, a single version number, or
+  a range:
```
jeffhandley, line 99: I'm not a fan of supporting ranges in strings like this. It starts to get hard to read/maintain IMHO, and it introduces more opportunity for brittleness.

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+    - Developer can modify the project file to explicitly indicate that they
+      want to see issues with APIs that are unsupported in Blazor.
+
+## Requirements
+
+### Goals
+
+* Be able to mark APIs that are unsupported on a particular OS
+* Allows platforms to be considered supported by default with the ability to add
+  or remove platforms from the project file and SDK.
```
eerhardt, line 60: What does `and SDK` mean in this sentence?

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+* Blazor Web Assembly is only considered supported when building a Blazor Web
+  Assembly app. For regular class libraries (`netstandard` or `net5.0`) the
+  default assumes that it's not supported and thus won't generate any warnings.
+  However, the developer must still be able to manually include Blazor Web
+  Assembly manually.
+* The annotations for unsupported APIs can express that an API was only
+  unsupported for a specific version.
```
eerhardt, line 67: A `specific version`? or a range of versions?

accepted/2020/platform-exclusion/platform-exclusion.md
eerhardt: Agreed. Having `min` and `max` version properties is a lot easier to reason about IMO. Especially since the parameter is named `platformName`, which says nothing about `version`.

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+### Hardcoded APIs in the analyzer
+
+### Build configuration for platforms
+
+In order to indicate which platforms the analyzer should by warn about, we're
+adding some metadata to MSBuild:
+
+```XML
+<ItemGroup>
+    <SupportedPlatform Include="android" />
```
eerhardt, line 175: Looking at https://github.com/dotnet/sdk/pull/12547, we already have an MSBuild item named `SupportedTargetPlatform`. Are we intending this to be the same thing as that? If not, these names are too similar to both be used.

cc @dsplaisted @sfoslund

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+        public UnsupportedOSPlatformAttribute(string platformName);
+    }
+```
+
+The semantics of this attribute are as follows:
+
+* Marks an assembly, module, or API as unsupported on a given OS platform.
+* The `platformName` can contain no version number, a single version number, or
+  a range:
+    - `browser`. Marks the API as unsupported on all version of `browser`. This
```
eerhardt, line 100: Note that versioning in `browser` is a bit awkward today. It is currently hard-coded to be `1.0.0.0`:

https://github.com/dotnet/runtime/blob/562763b5498bf53b68517b68221a89ad80ef6d16/src/libraries/System.Private.CoreLib/src/System/Environment.Browser.cs#L23

I'm not sure how we would version the "browser" OS.

cc @marek-safar @lewing @steveisok

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+OS-specific API. Good examples include the Windows Registry or Apple-specific
+graphic APIs. This is very different from browser support for specific HTML and
+CSS features: those generally aren't browser specific, which is why browser
+checks never worked well for the web. But for OS specific APIs, asking whether
+you're actually running on that OS is very workable. The only fragile aspect
+here is getting the version number right, but the analyzer enforces that.
+
+However, this approach doesn't work well for APIs that represent concepts that
+aren't OS specific, such as threading or I/O. This is a problem for OS platforms
+that are more constrained, such as Blazor Web Assembly, which runs in a browser
```
akoeplinger, line 28: use WebAssembly without space (applies throughout the doc).

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+In this spec, we're proposing a mechanism by which we can mark APIs as
+unsupported.
+
+## Scenarios and User Experience
+
+* Building a Blazor Web Assembly app
+    - Using an API that is unsupported will be flagged by the analyzer.
+* Building a class library
+    - Building a regular class library (targeting `netstandard` or `net5.0`)
+      will not complain about APIs that are unsupported in Blazor.
```
akoeplinger, line 49: I'd switch Blazor with WebAssembly here and in the next paragraph. There's nothing specific to Blazor here, it's a limitation of the browser/wasm.

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+      earlier. This will be used for APIs that were unsupported before but can
+      be supported starting with a later version of the OS. This is different
+      from applying `[MinimumOSPlatform("windows10.0.19041")]` in that the
+      presence of this attribute doesn't cause the API to be considered
+      unsupported on other platforms.
+
+### Analyzer behavior
+
+The existing analyzer that handles `[MinimumOSPlatform]` will be modified to
+handle `[UnsupportedOSPlatform]` (previously named `[RemovedInOSPlatform]`)
```
akoeplinger, line 114: [platform-checks.md](https://github.com/dotnet/designs/blob/0c0581895fc59ac03c02088b4407cf73d3562943/accepted/2020/platform-checks/platform-checks.md) should be updated to mention the renaming of this attribute.

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+are that would prevent the analyzer from flagging usage of the API.
+
+```C#
+// The API is supported everywhere except when running in a browser
+[UnsupportedOSPlatform("browser")]
+public extern void Api();
+
+public void Api_Usage()
+{
+    if (!RuntimeInformation.IsOSPlatform("browser"))
```
akoeplinger, line 127: are we supporting a string overload for `IsOSPlatform` now? or is this a typo and it should be `OSPlatform.Browser`?

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+        Api();
+    }
+}
+```
+
+### Hardcoded APIs in the analyzer
+
+### Build configuration for platforms
+
+In order to indicate which platforms the analyzer should by warn about, we're
```
akoeplinger, line 170:
```suggestion
In order to indicate which platforms the analyzer should warn about, we're
```

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+The targets of the Blazor Web Assembly SDK would initialize this as follows:
+
+```XML
+<ItemGroup>
+    <SupportedPlatform Remove="@(SupportedPlatform)" />
+    <SupportedPlatform Include="browser" />
+</ItemGroup>
+```
+
+These items needs to be passed to invocation of CSC the analyzer configuration.
```
akoeplinger, line 192: not sure what this sentence means :)

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+AFAIK they only supported properties today, but we can have a target that
```
akoeplinger, line 193:
```suggestion
AFAIK they only support properties today, but we can have a target that
```

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+converts these items into a semicolon separated list of platforms.
+
+Using items instead of a property makes it easier for the developer to add or
+remove from the set:
+
+When building a class library that is also supposed to also work in Blazor Web Assembly,
```
akoeplinger, line 199:
```suggestion
When building a class library that is supposed to also work in the browser,
```

accepted/2020/platform-exclusion/platform-exclusion.md
sfoslund: Right now we're using `SupportedTargetPlatform` as an itemgroup containing the supported `TargetPlatformVersion`s that correspond to the specified `TargetPlatformIdentifier`.

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+
+public void Api_Usage()
+{
+    if (RuntimeInformation.IsOSPlatformOrLater("iOS12.0") &&
+        RuntimeInformation.IsOSPlatformOrEarlier("ios14.0"))
+    {
+        Api();
+    }
+}
+```
```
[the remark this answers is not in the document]
terrajobst: I'm not in love with the string based approach either, but it seemed simple enough to give it a shot. FWIW, I'm not concerned about syntax issues because the entire design is around having an analyzer. We can trivially verify that the user entered sensible information.

I'm not fan of the enum and the default attribute. It makes the simple cases extremely verbose and error prone. If I want to mark an API as unsupported on a particular OS, I have to do this:

```
[OSPlatformSupport("ios", OSPlatformSupportLevel.Unupported)]
```

Then, when we make it supported we all of the sudden have to change this to:

```
[OSPlatformSupportDefault(OSPlatformSupportLevel.Supported)]
[OSPlatformSupport("ios12.0")]
```

That's not very intuitive. And we'd have another way to obsolete an API:

```
[OSPlatformSupportDefault(OSPlatformSupportLevel.Obsolete)]
```

I'm not saying my proposal is free of problems, but this particular spin looks worse to me than strings.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: > Especially since the parameter is named `platformName`, which says nothing about `version`.

To be fair, the approved API shape for `[MinimumOSVersion("ios12.0")]` is also named `plaformName` and allows for version numbers. Granted, the range is new.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: By "project" I mean "in the project file", which is usually expressed by the project template or by manual editing. By "SDK" I meant via the .props and .targets provided that is linked via SDK (which is expressed by the `Sdk` attribute).

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: Range, I'll reword that

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: So no then. Your notion of supported is "what can this SDK produce" while mine is "what platforms is the author of this project file intending to support". We probably need a different name then.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: I don't think we would; it's an synthetic platform that basically means "the subset of .NET APIs that can run in the browser sandbox". The set and behaviors might be a moving target between .NET versions, but the semantic definition of the sandbox would remain fixed.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: Yessir

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: Right, but we won't have a generic UX for targeting browser/wasm in the .NET 5 timeframe. For simplicity, this doc treats "Blazor WebAssembly" and "browser/wasm" as synonyms.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: Will do if this is the direction we're going with.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: We do support a string overload, per https://github.com/dotnet/runtime/issues/33331#issuecomment-650326500

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: Only god knows what I mean by my typos

accepted/2020/platform-exclusion/platform-exclusion.md
jeffhandley: If the browser sandbox evolves to support some of the concepts, would we just remove the attributes then?

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+
+```C#
+// On Windows, the API was unsupported up to and including version 10.0.19041.
+// The API is considered supported everywhere else without constraints.
+[UnsupportedOSPlatform("windows0.0-10.0.19041")]
```
jeffhandley, line 137: Is it correct to interpret this as equivalent to `[MinimumOSPlatform("windows10.0.19042")]`, except that it doesn't imply that all other platforms are unsupported?

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+public extern void Api();
+
+public void Api_Usage()
+{
+    if (!RuntimeInformation.IsOSPlatformOrEarlier("windows10.0.19041"))
```
jeffhandley, line 142: I liked the rationale we used for `IsOSPlatformEarlierThan` where the upper bound was exclusive.

accepted/2020/platform-exclusion/platform-exclusion.md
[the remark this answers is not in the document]
jeffhandley: > I'm not fan of the default attribute.

I'm honestly not thrilled with `OSPlatformDefault` either, but that concept seems to be what we're missing:
**When a platform is marked to be supported starting on a version, what does that mean for unspecified platforms? Are they _also_ supported (on all versions), or are they _unsupported_ entirely?**

> And we'd have another way to obsolete an API

We've already done that with `ObsoletedInOSPlatform`, right?  I was just trying to retain that concept here.  Should we revisit that?

accepted/2020/platform-exclusion/platform-exclusion.md
eerhardt: > Your notion of supported is "what can this SDK produce"

Should we rename the SDK properties then? I think the name `SupportedTargetPlatform` sounds like it is what target platforms the current project supports, not necessarily what the SDK supports.

accepted/2020/platform-exclusion/platform-exclusion.md
eerhardt: So are we saying that we think the "browser" OS is un-versionable?

accepted/2020/platform-exclusion/platform-exclusion.md
akoeplinger: @terrajobst I don't see this in the API excerpt you linked, it only mentions a new string overload for `IsOSPlatformOrLater` and `IsOSPlatformEarlierThan`, but not for the plain `IsOSPlatform`

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+The semantics of this attribute are as follows:
+
+* Marks an assembly, module, or API as unsupported on a given OS platform.
+* The `platformName` can contain no version number, a single version number, or
+  a range:
+    - `browser`. Marks the API as unsupported on all version of `browser`. This
+      will be used for APIs that can't be supported on a particular OS platform.
+    - `ios14.0`. Marks the API as unsupported starting with iOS 14.0. This will
+      be used for OS platform APIs that were removed/disabled.
+    - `windows0.0-10.0.19041`. Marks the API as unsupported for 10.0.19041 and
```
bartonjs, line 104: If ranges are to be supported, I recommend git-style .. instead of a hyphen-minus.  Hyphens seem like they're more likely to appear in an OS name than a .. would be.

accepted/2020/platform-exclusion/platform-exclusion.md
bartonjs, line 142:
```suggestion
    if (!RuntimeInformation.IsOSPlatformEarlierThan("windows10.0.19041"))
```

(OK, not sure about the version for your boundary condition here, but there is no IsOSPlatformOrEarlier; it's `<`, not `<=`)

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -0,0 +1,217 @@
...
+// The API is only supported on iOS. There, it started to be supported in
+// version 12.0 and stopped being supported in version 14.
+[MinimumOSPlatform("iOS12.0")]
+[UnsupportedOSPlatform("ios14.0")]
+public extern void Api();
+
+public void Api_Usage()
+{
+    if (RuntimeInformation.IsOSPlatformOrLater("iOS12.0") &&
+        RuntimeInformation.IsOSPlatformOrEarlier("ios14.0"))
```
bartonjs, line 159:
```suggestion
        RuntimeInformation.IsOSPlatformEarlierThan("ios14.0"))
```

Fixes the logic bug, and uses a method that actually exists.  Win/win :smile:.

jeffhandley:
I keep thinking about how the crux of the challenge we're having is that marking a specific OSPlatform as supported _implies_ that no other platforms are supported, but we don't _always_ want that behavior.

```csharp
[MinimumOSPlatform("windows7.0")]
public void WindowsOnlyAPI() {}

[MinimumOSPlatform("ios14")]
public void WorksEverywhereButOnIosVersion14IsRequired() {}
```

If we set that problem aside for a moment, and we focus purely on _excluding_ an OSPlatform, then I think we're otherwise OK.  The following would _just work_, wouldn't it?

```csharp
[UnsupportedOSPlatform("browser")]
public void WorksEverywhereExceptBrowser() {}
```

I believe the ranges were being introduced so that we could use `[UnsupportedOSPlatform]` to denote a combination of _supported_ and _unsupported_, specifying which version ranges are unsupported, leaving anything outside the range as supported **without implying exclusive platform support**.  Using `[UnsupportedOSPlatform]` in that way feels awkward to me, and less expressive than using `[MinimumOSPlatform]` to indicate _support_.

I'm really concerned about the range support, regardless of the syntax. It makes the data harder to reason over as a human code reader, and potentially more brittle when reasoning over in the analyzer implementation. I think we'd introduce a higher likelihood that the data will be incomprehensible and then the analyzer will need to introduce more diagnostics for when that's the case.

So what if we did the following?

1. Rename `RemovedInOSPlatformAttribute` to `UnsupportedOSPlatformAttribute`, but otherwise leave it as-is
    * This specifies the _lower-bound version_ in which an API became unsupported
2. Add a property to `MinimumOSPlatformAttribute` that indicates whether or not the platform should be treated as exclusive

I think we could then satisfy our scenarios:

```csharp
[MinimumOSPlatform("windows7.0")]
public void WindowsOnlyAPI() {}

[MinimumOSPlatform("ios14.0", IsExclusive = false)]
public void WorksEverywhereButOnIosVersion14IsRequired() {}

[UnsupportedOSPlatform("browser")]
public void WorksEverywhereExceptBrowser() {}
```

The evolution of an API going from unsupported to having a minimum version required for support would involve changing `[UnsupportedOSPlatform("ios")]` to `[MinimumOSPlatform("ios14", IsExclusive = false)]`.  If there are multiple platforms that add support over time, then they would each need to specify `IsExclusive = false`.

To show an edge case where something comes in and out of support for a platform repeatedly:

```csharp
[MinimumOSPlatform("ios12.0", IsExclusive = false)]
[UnsupportedOSPlatform("ios14.0")]
[MinimumOSPlatform("ios12.0", IsExclusive = false)]
[MinimumOSPlatform("ipados16.0", IsExclusive = false)]
public void Worked_In_12_And_13_But_Not_14_Or_15_But_Works_Again_In_16() {}
```

If this approach has merit, then I'd suggest that we also rename `MinimumOSPlatformAttribute` to `SupportedOSPlatformAttribute`.  Both `SupportedOSPlatform` and `UnsupportedOSPlatform` would accept a platform name with an optional version such that the version is always treated as the lower-bound.

By default, a `[SupportedOSPlatform]` attribute denotes exclusive support on that platform. But specifying `IsExclusive = false` on the attribute allows support to be turned on for the platform without affecting support on other platforms.  An `[UnsupportedOSPlatform]` attribute never implies anything about any other platforms.

To address the [concern of introducing another means for obsoleting an API]([MinimumOSPlatform("ios12.0", IsExclusive = false)]), I suggest we remove the `ObsoletedInOSPlatformAttribute` that we introduced.

To summarize, I propose that we:

1. Rename `RemovedInOSPlatformAttribute` to `UnsupportedOSPlatformAttribute`
2. Rename `MinimumOSPlatformAttribute` to `SupportedOSPlatformAttribute`
3. Remove `ObsoletedInOSPlatformAttribute`
4. Add an `IsExclusive` Boolean property to `SupportedOSPlatformAttribute`, that defaults to `true`

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -58,61 +88,169 @@ unsupported.
...
+
+Similar to [platform-checks] this model still allows for OS vendors to obsolete
+and remove APIs:
+
+```C#
+[UnsupportedOSPlatform("windows")]
+[SupportedOSPlatform("windows10.0.1903")]
+[ObsoletedInOSPlatform("windows10.0.1909")]
+[SupportedOSPlatform("windows10.0.2004")]
+public void DoesNotWorkOnWindows();
```
jeffhandley, line 209:
```suggestion
[UnsupportedOSPlatform("windows")]
[SupportedOSPlatform("windows10.0.1903")]
[ObsoletedInOSPlatform("windows10.0.1909")]
[UnsupportedOSPlatform("windows10.0.2004")]
public void DoesNotWorkOnWindows();
```

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -58,61 +88,169 @@ unsupported.
...
+```
+
+This describes an API that is supported by all platforms, except for Windows.
+And on Windows, it's supported for Windows 10 1903 and 1909.
```
jeffhandley, line 213:
```suggestion
This describes an API that is supported by all platforms, except for Windows.
And on Windows, it's supported for Windows 10 1903 and 1909, but it was
obsoleted in 1909 and then became unsupported again in 2004.
```

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -58,61 +88,169 @@ unsupported.
...
+
+Platform-specific APIs like the iOS and Android bindings are still expressible
+the same way:
+
+```C#
+[SupportedOSPlatform("ios12.0")]
+[ObsoletedInOSPlatform("ios13.0")]
+[UnsupportedOSPlatform("ios14.0")]
+[SupportedOSPlatform("ipados13.0")]
+public void OnlyWorksOniOS();
```
jeffhandley, line 223: Should we use a `params string[] platformNames`-based constructor for the attributes?  If so, we could then have the following:

```suggestion
[SupportedOSPlatform("ios12.0", "ipados13.0")]
[ObsoletedInOSPlatform("ios13.0")]
[UnsupportedOSPlatform("ios14.0")]
public void OnlyWorksOniOS();
```

With the iOS family of platforms, this might become a common need.

accepted/2020/platform-exclusion/platform-exclusion.md
```diff
@@ -132,14 +270,32 @@ public void Api_Usage()
...
 
 public void Api_Usage()
 {
-    if (!RuntimeInformation.IsOSPlatformOrEarlier("windows10.0.19041"))
+    if (!RuntimeInformation.IsOSPlatform("windows") ||
+         RuntimeInformation.IsOSPlatformOrLater("windows10.0.19041"))
+    {
+        Api();
+    }
+}
```
jeffhandley, line 286: I think we ran into trouble with `||` checks, right @buyaa-n?  Would this work or would we this need to be expressed differently?

accepted/2020/platform-exclusion/platform-exclusion.md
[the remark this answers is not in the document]
jeffhandley: The new heuristic for sorting the platforms by version and identifying whether `SupportedOSPlatform` or `UnsupportedOSPlatform` comes first accomplishes the job of determining the default, without requiring any gymnastics that would be error-prone.  Resolving this thread.

accepted/2020/platform-exclusion/platform-exclusion.md
jeffhandley: I think this is what you meant

accepted/2020/platform-exclusion/platform-exclusion.md
jeffhandley: In the latest discussions, we're moving away from the ranges concept since it was basically a workaround for marking support on a minimum version without implying lack of support on other (unspecified) platforms.

accepted/2020/platform-exclusion/platform-exclusion.md
buyaa-n: Array might work if we know all supported platforms when creating the attribute. But for adding a new platform supported we will need to add a new element to the array? isn't it mean we will need update the existing attribute? By my understanding, we want to avoid updating existing attributes

accepted/2020/platform-exclusion/platform-exclusion.md
buyaa-n: Right, it is not supported directly, but with the given API proposal with a combination of attributes its good to have/support this i guess, we can try to support this.

buyaa-n:
> 2. I think we had problems with the `||` based guards; I'll need @buyaa-n to weigh in there.

This is a more meaningful scenario than the other `||` example we discussed before, for this one we should try to add support

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: Fair point, I'll update the spec.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: I've changed this to this:

```C#
    if (RuntimeInformation.IsOSPlatformOrLater("ios12.0") &&
        !RuntimeInformation.IsOSPlatformOrLater("ios14.0"))
    {
        Api();
    }
```

@buyaa-n can we track state like this in the analyzer? You need to compute a lower bound and an upper bound.

* `RuntimeInformation.IsOSPlatformOrLater("iOS12.0")`. You know platform is `isOS` and the minimum is `12.0`
* `!RuntimeInformation.IsOSPlatformOrLater("os14.0")`. You know the platform is either not `iOS` or the version is higher than 14.
* `RuntimeInformation.IsOSPlatformOrLater("iOS12.0") && !RuntimeInformation.IsOSPlatformOrLater("os14.0")`. You know the platform is `iOS` and the version range is `[12.0,14.0)`.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: See [notes here](https://github.com/dotnet/runtime/issues/39989#issuecomment-665197220). We decided against `params.

accepted/2020/platform-exclusion/platform-exclusion.md
terrajobst: @dsplaisted were should I file a bug to discuss renaming the SDK's `SupportedTargetPlatform` item group>

accepted/2020/platform-exclusion/platform-exclusion.md
buyaa-n: It's not supported with the current implementation, but i think we can make that evaluation based on flow analysis result looks `&&` logic is more doable than `||`

accepted/2020/platform-exclusion/platform-exclusion.md
dsplaisted: In dotnet/sdk.

Note that there was already a `SupportedTargetFramework` item which we use to populate the TargetFramework dropdown on the VS property page, so we went with the same pattern for `SupportedTargetPlatform`.

terrajobst: Makes sense. In that sense I think we should come up with a new name for the analyzer. Let me chew on that.

PR #148, opened 2020-07-22
Provide default activity name value from [CallerMemberName]

kzu:
Add proposal for new StartActivity overload and default value with `CallerMemberName`


kzu:
Closing, the flow should be the other way around: Issue first, potential document later :). 

Created https://github.com/dotnet/designs/issues/149 to replace this.

PR #155, opened 2020-09-03
Initial draft of spec for System.BinaryData

terrajobst:


accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,195 @@
...
+        public override int GetHashCode();
+        public override string ToString();
+    }
+}
+namespace System.Text.Json
+{
+    public static class BinaryDataJsonExtensions
+    {
+        public static Task SetObjectAsJsonAsync<T>(this BinaryData data, T jsonSerializable, CancellationToken cancellationToken = default);
+        public static void SetObjectAsJson<T>(this BinaryData data, T jsonSerializable, CancellationToken cancellationToken = default);
```
bartonjs, line 154: I think once we've gone the route that S.T.Json is providing the extension methods we should add overloads that take the options type.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,195 @@
...
+        public BinaryData(ReadOnlySpan<byte> data);
+        public BinaryData(byte[] data);
+        public BinaryData(object jsonSerializable, Type? type = null);
+        public BinaryData(ReadOnlyMemory<byte> data);
+        public BinaryData(string data);
+        public static BinaryData FromBytes(ReadOnlyMemory<byte> data);
+        public static BinaryData FromBytes(ReadOnlySpan<byte> data);
+        public static BinaryData FromBytes(byte[] data);
+        public static BinaryData FromObject<T>(T jsonSerializable, CancellationToken cancellationToken = default);
+        public static Task<BinaryData> FromObjectAsync<T>(T jsonSerializable, CancellationToken cancellationToken = default);
```
bartonjs, line 177: In the ctors/factory version we probably still want "Json" in the name of the serialize/deserialize methods.  If nothing else, it sets up the pattern for the next data format type.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,195 @@
...
+        public void SetStream(Stream stream);
+        public static Task SetStreamAsync(Stream stream, CancellationToken cancellationToken = default);
+        public void SetString(string data);
+        public static implicit operator ReadOnlySpan<byte>(BinaryData data);
+        public static implicit operator ReadOnlyMemory<byte>(BinaryData data);
+        public ReadOnlyMemory<byte> ToBytes();
+        public Stream ToStream();
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override bool Equals(object? obj);
+        [EditorBrowsable(EditorBrowsableState.Never)]
```
bartonjs, line 144: These EB(Never)s are part of the Azure SDK style, but we don't do them in the BCL (AFAIK), probably want to remove them from the spec.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,195 @@
...
+
+```C#
+namespace System
+{
+    public readonly struct BinaryData
+    {
+        public void SetBytes(ReadOnlyMemory<byte> data);
+        public void SetBytes(ReadOnlySpan<byte> data);
+        public void SetBytes(byte[] data);
+        public void SetStream(Stream stream);
```
akoeplinger, line 135: who owns the stream, or does this essentially loop over the stream and read all the bytes and store it internally?

bartonjs: All of the setters are data sinks (or the factory methods/ctors in the other variant).  The Streams aren't held as references, they're drained into a buffer.

This ends up being important, so that things like ToBytes don't need Async variants.

eiriktsarpalis: Should we be concerned about the potential for misuse of this particular method? E.g. passing NetworkStream's whose length doesn't fit in a single buffer.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,195 @@
...
+quite appealing we need to verify that it realizes these goal. If we can't, we
+should go back to the design of using constructors and factory methods, even if
+that reduces our flexibility.
+
+### API design with extension methods
+
+```C#
+namespace System
+{
+    public readonly struct BinaryData
```
GrabYourPitchforks, line 130: Remove _readonly_ modifier.

accepted/2020/binary-data/binary-data.md
GrabYourPitchforks, line 154: The first parameter should be annotated _ref this_, otherwise you'll end up mutating a copy of the struct, which is a glorified (and expensive) no-op. We may want to create an analyzer for people writing their own extension methods for this type.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,195 @@
...
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public override int GetHashCode();
+        public override string ToString();
+    }
+}
+namespace System.Text.Json
+{
+    public static class BinaryDataJsonExtensions
+    {
+        public static Task SetObjectAsJsonAsync<T>(this BinaryData data, T jsonSerializable, CancellationToken cancellationToken = default);
```
GrabYourPitchforks, line 153: We cannot have async setters as extension methods. If `BinaryFormatter` is a struct it'll have copy-by-value semantics, which means we can't hold a reference back to the original copy (since refs can't survive across async points). For async methods, the only viable option is to have a truly static factory method which returns `Task<BinaryData>`.

bartonjs: Presumably you mean `BinaryData`, not `BinaryFormatter`? :)

GrabYourPitchforks: My brain's own autocomplete makes me sad. :)

accepted/2020/binary-data/binary-data.md
KrzysztofCwalina: why?

KrzysztofCwalina (commented):
I think the option with Set methods is just too ugly. I think extensibility should be provided either by "extensions everything" or by a serializer abstraction (or both)

accepted/2020/binary-data/binary-data.md
GrabYourPitchforks: > passing NetworkStream's whose length doesn't fit in a single buffer.

This API is akin to the various `ReadToEnd` methods we have throughout the framework. I don't think those methods have historically been misused.

AustinWise:
Excluding the JSON-related methods, I think this API makes sense. There are many ways in .NET to express binary blobs, and this API unifies them when designing an API.

I don't understand the JSON use case. If a user encounters an API that takes a `BinaryData` object as parameter, they will have to check the documentation for what sort of JSON to put into the object. The type system will not ensure they put the right messages types into the `BinaryData`. To use the `QueueClient` example, why would an API designer use `BinaryData` in their API when they could instead do one of the following:

* Define a method on `QueueClient` for each message type or message. This is a fully statically-type client.
* Define a abstract `MessageBase` class or `IMessage` interface that all message types inherit from. The `SendMessageAsync` would accept this type are a parameter.

accepted/2020/binary-data/binary-data.md
jnm2: > The Streams aren't held as references, they're drained into a buffer.

This seemed the most likely, but it would be nice to spell this out in the XML documentation for the method.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,177 @@
...
+2. But we need to balance implementation flexibility concerns with usability.
+   The primary goal of this type is to provide usability via conveniences. If we
+   don't make the type super usable, the feature will fail. If we feel that the
+   factories aren't discoverable enough, we should combine it with constructors,
+   even if that reduces our flexibility.
+
+### Shipping vehicle
+
+**Assembly**: System.Binary.Data.dll
+**NuGet Package**: System.Binary.Data (stable)
```
Nirmal4G, line 119:
```suggestion
**Assembly**: System.Memory.Data.dll
**NuGet Package**: System.Memory.Data (stable)
```

**OR**

```suggestion
**Assembly**: System.Data.Binary.dll
**NuGet Package**: System.Data.Binary (stable)
```
I personally prefer the latter as it is more direct in meaning to the API's usage.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,177 @@
...
+  usability, not perf. Performance critical scenarios should be enabled by
+  providing overloads with the existing low-level concepts, such as span.
+
+## Design
+
+We're concerned with hard coding the type to `System.Text.Json`, but we
+acknowledge that defaults are very valuable. If we need to support other
+serializers we can provide additional arguments or APIs.
+
+This type can't live in corlib because it would depend on `System.Text.Json`
```
Nirmal4G, line 92: If it can't be a part of foundational APIs, then why are we placing it in `System.*` namespace?

accepted/2020/binary-data/binary-data.md
terrajobst: Oh, that's bug. I mean to write `BinaryDay` because that's the name of the type.

accepted/2020/binary-data/binary-data.md
terrajobst: Being in corlib isn't the same as being foundational. `System.Text.Json` isn't in corlib either.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,177 @@
...
+        public static BinaryData FromBytes(ReadOnlyMemory<byte> data);
+        public static BinaryData FromBytes(ReadOnlySpan<byte> data);
+        public static BinaryData FromBytes(byte[] data);
+        public static BinaryData FromObjectAsJson<T>(T jsonSerializable, CancellationToken cancellationToken = default);
+        public static Task<BinaryData> FromObjectAsJsonAsync<T>(T jsonSerializable, CancellationToken cancellationToken = default);
+        public static BinaryData FromStream(Stream stream);
+        public static Task<BinaryData> FromStreamAsync(Stream stream, CancellationToken cancellationToken = default);
+        public static BinaryData FromString(string data);
+        public static implicit operator ReadOnlyMemory<byte>(BinaryData data);
+        public ReadOnlyMemory<byte> ToBytes();
```
MiYanni, line 137: Please make this `AsBytes` since it is non-allocating. It is a C# community convention. It is in framework APIs too. https://stackoverflow.com/a/13867955/294804

terrajobst: That's a good point, we missed that. @KrzysztofCwalina @bartonjs, I assume we agree with that?

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,144 @@
...
+        public BinaryData(byte[] data);
+        public BinaryData(object jsonSerializable, JsonSerializerOptions options = default, Type? type = null);
+        public BinaryData(ReadOnlyMemory<byte> data);
+        public BinaryData(string data);
+        public static BinaryData FromBytes(ReadOnlyMemory<byte> data);
+        public static BinaryData FromBytes(byte[] data);
+        public static BinaryData FromObjectAsJson<T>(T jsonSerializable, JsonSerializerOptions options = default, CancellationToken cancellationToken = default);
+        public static BinaryData FromStream(Stream stream);
+        public static Task<BinaryData> FromStreamAsync(Stream stream, CancellationToken cancellationToken = default);
+        public static BinaryData FromString(string data);
```
danmosemsft, line 131: > String encoding is assumed to be UTF8

What is string in context here -- eg., this string is regular UTF-16. Are you referring to eg., `byte[] data` ? Should UTF-8 appear in any method names?

accepted/2020/binary-data/binary-data.md
bartonjs: `AsBytes` does seem more idiomatic than `ToBytes`.  ImmutableArray mixes an As(Memory/Span) and a To(Builder), so it's not unheard of for a single type to use both.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,150 @@
...
+### Goals
+
+* Stable API
+* Make it easy for people to construct bytes from common data types (byte
+  arrays, spans, memories, string, JSON serialized objects)
+* Make it easy for people to convert bytes to common data types of the same
+  representations
+* Provide opinionated defaults in order to reduce complexity
+    - String encoding is assumed to be UTF8
+    - Object serialization is done with `System.Text.Json`, with specific options
```
JoshLove-msft, line 76: I realize this may need a larger discussion, but have we considered porting over something like `ObjectSerializerBase` as a serialization abstraction into .NET as well? We could then have the Azure.Core [ObjectSerializer](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Serialization/ObjectSerializer.cs) inherit from it, and we would be able to include the `ObjectSerializer` APIs on `BinaryData` as we have currently on the [type ](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core.Experimental/src/Primitives/BinaryData.cs#L251) that is in preview in Azure.

accepted/2020/binary-data/binary-data.md
JoshLove-msft: I think this was considered but we felt it wasn't necessarily worth deviating from the To/From pattern used throughout the type.

NinoFloris: For F# having the implicit conversions available as normal methods is helpful, quite painful having to invoke the op_Implicit method.

pakrym: I think we've done this after the UX study to help customers find all related `To*` overloads because they are grouped together.

accepted/2020/binary-data/binary-data.md
GrabYourPitchforks: "encoding assumed to be UTF-8" means that any string data provided here will be converted to UTF-8, and the resulting `BinaryData` instance will contain the UTF-8 representation of that string data. I don't think it's necessary to state UTF-8 in any API name.

KrzysztofCwalina (approved):
Looks good to me! Thanks!

stephentoub:
@terrajobst, this is purely OOB, right? This shouldn't be in netcoreapp. I continue to think this isn't worth its weight, and I don't think introducing yet another exchange here is a good thing,  but if it's just another nuget package folks that do like it can depend on,  c'est la vie.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,150 @@
...
+The type is for "user data", that is, the contract for the API is "give me a
+blob of binary data". It's not for protocol-like APIs where the underlying
+encoding is part of the API contract, e.g. a REST API.
+
+We don't see the need to, for example, provide new virtual methods on Stream or
+`HttpClient`.
+
+Please note: Neither the constructors nor the factories copy data -- they wrap
+the input data, so long it's already binary. Anything that requires
+transcoding/serializtion (e.g. string, object) isn't wrapped. While the goal
```
jkotas, line 111:
```suggestion
transcoding/serialization (e.g. string, object) isn't wrapped. While the goal
```

accepted/2020/binary-data/binary-data.md
terrajobst: Yes, we talked about this and this seems like a good idea. I just filed https://github.com/dotnet/runtime/issues/43669 to track this. We only excluded this from this feature for expediency.

terrajobst:
> @terrajobst, this is purely OOB, right? This shouldn't be in netcoreapp.

I've called out that this API won't be added to `Microsoft.NetCore.App`

accepted/2020/binary-data/binary-data.md
terrajobst: OK, based on conversations and too keep things consistent, I've spec'ed it as `AsBytes()`. @JoshLove-msft and @pakrym please tell me if you feel strongly that this is the wrong thing.

JoshLove-msft:
To help speed up implementation:
BinaryData - https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core.Experimental/src/Primitives/BinaryData.cs

ReadOnlyMemoryStream - https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core.Experimental/src/Primitives/ReadOnlyMemoryStream.cs

Tests - https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core.Experimental/tests/BinaryDataTests.cs
https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core.Experimental/tests/ReadOnlyMemoryStreamTests.cs

PR that was going to move it to Azure.Core that has some perf suggestions from @pakrym - https://github.com/Azure/azure-sdk-for-net/pull/16050

accepted/2020/binary-data/binary-data.md
JoshLove-msft: That's great news! 

One point to clarify - say we end up adding in `ObjectSerializer` (or maybe `ObjectSerializerBase` since the shipped version in Azure.Core is called `ObjectSerializer`) after shipping `BinaryData`. If we then want to expose the `ObjectSerializer` within `BinaryData` in `ToObject/FromObject` methods, would this change how we think about the `FromObjectAsJson/ToObjectAsJson` methods?  `FromObject/ToObject` would be able to be used with Json as well if you pass in an ObjectSerializer instance that uses Json, e.g. the [JsonObjectSerializer](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core/src/Serialization/JsonObjectSerializer.cs), so it might make more sense to name all methods as FromObject/ToObject and indicate that the default behavior is Json through the parameter name as we [currently do](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/core/Azure.Core.Experimental/src/Primitives/BinaryData.cs#L235). Is this something we would consider changing after the initial release of BinaryData?

accepted/2020/binary-data/binary-data.md
KrzysztofCwalina: I honestly think the distinction between As and To is something only a few understand and for the rest it will be just an annoyance, as users will have to continuously think "is the conversion I want a To or an As conversion". When everything starts with To, you just type "To" and select from intellisense list.

accepted/2020/binary-data/binary-data.md
bartonjs: > Is [renaming ToObjectAsJson] something we would consider changing after the initial release of BinaryData?

No, once it's stable (especially as a System.* namespaced type) it's that way basically forever.

> If we then want to expose the ObjectSerializer within BinaryData in ToObject/FromObject methods, would this change how we think about the FromObjectAsJson/ToObjectAsJson methods?

I don't think so.  The ones that take the configured serializer do whatever the provided serializer instance does. The ones that don't take it provide a set of ease-of-use defaults.  E.g. if both JSON and protobuf were top-level ease-of-use constructs you can't have

```C#
public byte[] ToObject<T>(object jsonSerializable, JsonSerializerOptions options = default);
public byte[] ToObject<T>(object protobufSerializable, ProtobufSerializerOptions options = default);
```

Well, you can, but then it's a compile error for any to call `ToObject(obj)` since there are two valid targets (so if they're done across two different versions then the Protobuf method introduced a breaking change; if they're in the same version it's as if neither had defaults).  Providing the data format name in the method allows for defaulted options to be used across all formats.  (This is also true for the constructors, a protobuf-based ctor can't have defaulted options, so that format might be method-only... but since it's in a V2+ of the type there is already a body of people to spread the awareness of the factory methods).

> or maybe `ObjectSerializerBase` since the shipped version in Azure.Core is called `ObjectSerializer`

There's a rule for this :smile::
* AVOID naming base classes with a “Base” suffix if the class is intended for use in public APIs.

But the naming and scope of that type/feature is a different issue.

accepted/2020/binary-data/binary-data.md
MiYanni: It is a common convention for those using LINQ or Entity Framework regularly. Personally, I understood it within the first 2 years of writing C# professionally. It is necessary to distinguish between `As` and `To` because in LINQ, `To` resolves the query while `As` does not. For this use case, our usage of `BinaryData` looks bizarre currently as it is littered with `ToBytes().ToArray()`, which infers we're allocating a new object to then allocate that as an array, which is not the case. Intellisense shouldn't trump common patterns that are reinforced by frameworks and community code, IMO.

terrajobst: I feel myself agreeing with @MiYanni. @bartonjs?

bartonjs: Yeah, I'm with the "ToBytes().ToArray() looks 'two' expensive", so my preference is for `AsBytes` (or AsMemory so we can also easily add AsSpan)

KrzysztofCwalina: To in LINQ is very expensive in comparison to As. Also, LINQ's As allocates. In other words, the As convention does not mean "no allocation" and the To convention does not mean allocation either. For example Memory<char>.ToString might not actually allocate a string.

tg-msft: In our UX study for this type, we initially had a `Bytes` property instead of `ToBytes()` and participants were struggling.  One of them even turned it into a `Stream` in an attempt to get a byte array out of it:

> "I looked at the ToStream directly because it looked like something natural to me, probably missed some properties on the way. Maybe it will be more intuitive if there were a method ToBytes() instead of Bytes property."

We switched to `ToBytes()` half way through and no one else struggled.  Having all the static `From*()` and instance `To*()` options align in intellisense makes it very easy for customers to discover all their options for transforming binary data.  We'd see customers scrolling back and forth over just the `From*()`s and `To*()`s in intellisense when attempting each task.  Moving one of the options to the top of the list with `AsBytes()` would make that worse.

bartonjs:
```C#
public ReadOnlyMemory<byte> AsMemory();
public ReadOnlySpan<byte> AsSpan();
public byte[] ToBytes();
```

?

KrzysztofCwalina: @bartonjs, I think the reasons we have As and To in these APIs are:

1. The API is low level API focusing on perf. We wanted to be extra explicit about the difference in perf characteristics.
2. The difference in perf characteristic between AsSpan and ToBytes is huge.

In the case of BinaryData the perf difference between ToStream and AsBytes is pretty much irrelevant, and the type is focusing on usability, not on writing low level perf to the metal code.

accepted/2020/binary-data/binary-data.md
```diff
@@ -0,0 +1,151 @@
...
+        public static BinaryData FromBytes(ReadOnlyMemory<byte> data);
+        public static BinaryData FromBytes(byte[] data);
+        public static BinaryData FromObjectAsJson<T>(T jsonSerializable, JsonSerializerOptions options = default, CancellationToken cancellationToken = default);
+        public static BinaryData FromStream(Stream stream);
+        public static Task<BinaryData> FromStreamAsync(Stream stream, CancellationToken cancellationToken = default);
+        public static BinaryData FromString(string data);
+        public static implicit operator ReadOnlyMemory<byte>(BinaryData data);
+        public static implicit operator ReadOnlySpan<byte>(BinaryData data);
+        public ReadOnlyMemory<byte> AsBytes();
+        public T ToObjectFromJson<T>(JsonSerializerOptions options = default, CancellationToken cancellationToken = default);
```
JoshLove-msft, line 142: Is it correct that this (and FromObjectAsJson) should take cancellationToken?
The sync version of JsonSerializer.Serialize and JsonSerializer.Deserialize do not take cancellationTokens.

SOHODeveloper:
I just upgraded to preview 9 of the Azure.Messaging.ServiceBus package and that brought along System.Memory.Data 1.0.0.0
While I appreciate the discussions above and the removal of the pattern ToBytes().ToArray(), I really think that ToBytes() (or AsBytes() as mentioned) above is better.  ToBytes tells us we are getting a byte[] where as ToArray() is at best ambiguous.  What type of array am I getting?

JohnNilsson:
https://github.com/dotnet/runtime/issues/41686 is closed and this is merged so I suppose the window for design suggestions is closed. But jus in case:

The static `FromObjectAsJson<T>` seems problematic as it ties a particular serializer to the type without an obvious way to add similar overloads for other serializers. To that end, might it be better to have a generic "builder" type for that use case, allowing extension methods to hook into particular builder types?

The invocation would then be `From<T>().AsJson()` where extension mehods like `AsJson()` would be defined as `(this BinaryData.Builder<T> builder, JsonSerializerOptions? options = default)` allowing further constraints on `T` if desirable.

PR #157, opened 2020-09-10
Rename TargetPlatformMinVersion to SupportedOSPlatformVersion

terrajobst:


dsplaisted (approved):

dsplaisted:
FYI @wli3

nkolev92 (approved):
Given that `SupportedOSPlatformVersion` is not used by NuGet in the framework representation right now, there are no concerns with this change from NuGet.Client side.

fyi @zkat

Nirmal4G:
TPMV and TPV are the right terminologies already. But, if we're renaming them, then, What else are those used for?

_If ain't broke, right?_

PR #164, opened 2020-11-06
OR_GREATER preprocessor symbols for TFMs

terrajobst:


accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,193 @@
...
+    - Applies to `net1.0`-`net4x` and the `NETFRAMEWORK`/`NET` symbols
+    - For example, `net4.8` will define `NETFRAMEWORK`, `NET48`,
+      `NET48_OR_GREATER` `NET472_OR_GREATER`, `NET471_OR_GREATER`, etc.
+* .NET Core
+    - Applies to `netcoreappX.Y` and the `NETCOREAPP` symbol
+    - For example, `netcoreapp3.1` will define `NETCOREAPP`, `NETCOREAPP3_1`,
+      `NETCOREAPP3_1_OR_GREATER`, `NETCOREAPP3_0_OR_GREATER` etc.
+* .NET 5 and later
+    - Applies to `netX.Y` and the `NET` symbol
+    - For example, `net6.0` will define `NET`, `NET6_0`, `NET6_0`, `NET5_0` etc.
```
Tratcher, line 163: I'm not sure what happened here? It makes sense for net6.0 to define NET5_0_OR_GREATER, but not NET5_0.
```suggestion
    - For example, `net6.0` will define `NET`, `NET6_0` etc.
```

richlander:
What would the current [docs table](https://docs.microsoft.com/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if#remarks) look like? I currently rely on it heavily for this scenario. I believe it is a key aspect of the design.

image

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,193 @@
...
+    - This will also include the corresponding defines a .NET Core X.Y would
+      have gotten. For example, `net6.0` will also define `NETCOREAPP`,
+      `NETCOREAPP6_0`, `NETCOREAPP5_0_OR_GREATER`, `NETCOREAPP3_1_OR_GREATER`
```
Tratcher, line 166: So NETCOREAPP6_0 is intentionally redundant with NET6_0?

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: That was multi-caret editing mishap. Fixed.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Interesting. I assumed the SDK also defines `NETCOREAPP5_0` when you target `net5.0`. Turns out, it only defines `NETCOREAPP`. I'll update the spec to remove `NETCOREAPP6_0` and `NETCOREAPP5_0_OR_GREATER`.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,194 @@
...
+
+This also makes the code more resilient for cases where you add new frameworks
+with lower versions later. For example, if you start targeting .NET Core 3.0,
+the former code would go down the path for .NET 6 while the latter will fail
+with "Unhandled TFM", which is a lot more logical for the person having to
+maintain the code.
+
+## Scenarios and User Experience
+
+### Multi-targeting for devices to provide a cross-platform API
```
dougbu, line 77: @terrajobst it's not clear how this section relates to the rest of the document. Does `WINDOWS7_0` actually mean `WINDOWS7_0_OR_GREATER`❔ If not and there's no plan to add `WINDOWS7_0_OR_GREATER`, suggest removing this part.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,194 @@
...
+        #if NET5_0_OR_HIGHER
+            FastBlurUsingHardwareIntrinsics();
+        #else
+            SlowBlurUsingSimpleMath();
+        #endif
+    }
+}
+```
+
+## Requirements
```
dougbu, line 131: Do the requirements include out-of-support TFMs❔ If not, suggest adding that as a non-goal.

dougbu (commented):
👍

terrajobst:
@richlander 

> What would the current [docs table](https://docs.microsoft.com/dotnet/csharp/language-reference/preprocessor-directives/preprocessor-if#remarks) look like? I currently rely on it heavily for this scenario. I believe it is a key aspect of the design.

Good point, I've added a section to clarify that.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: The requirements section includes the list of frameworks for which we will do the defines, so I think that's called out.

marcpopMSFT:
Should we exclude ones for runtimes that are no longer in support (eg NET2_0 or NETCOREAPP1_0)?

marcpopMSFT:
Are you imagining this for the 6.0.100 sdk or an earlier 5.0.N00 SDK?

terrajobst:
> Are you imagining this for the 6.0.100 sdk or an earlier 5.0.N00 SDK?

That's a great question. I didn't consider this, I'll have to think about that. Would feature work like this make the bar?

terrajobst:
> Should we exclude ones for runtimes that are no longer in support (eg NET2_0 or NETCOREAPP1_0)?

Oh, I think I now understand @dougbu's [question](https://github.com/dotnet/designs/pull/164#discussion_r518992759). I think it's easier to understand if we support all version that the SDK supports building for. IOW, I would only exclude defines for TFMs that one cannot build for.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
+# OR_GREATER preprocessor symbols for TFMs
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+Today, projects automatically define `#if` symbols based on the friendly TFM
+representation. This includes both versionless- as well as version-specific
+symbols. For .NET 6, we plan to add `XXX_OR_GREATER` variants that are
+accumulative in nature:
+
+TFM                 | Existing | New
```
jaredpar, line 10:
```suggestion
TFM                 | Existing | Additional
```

The word "New" here is ambiguous between "this is the new defines we generate that replace the old ones" and "these are the new defines in addition to the older ones". Rest of the doc clears that up but think it could help to be more explicit here.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+(`NET`). However, the code still needs to be revised when you add lower versions
+of existing TFMs.
+
+With this feature, you'll be able to express it like this, which is a lot more
+intuitive:
+
+```C#
+void M()
+{
+    #if NET5_0_OR_GREATER
```
jaredpar, line 55: VS standard formatting is going to put the `#if` in column 0 but here they are at column 4. Do we want this to look more like the VS experience here?

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+    - For example, `netcoreapp3.1` will define `NETCOREAPP`, `NETCOREAPP3_1`,
+      `NETCOREAPP3_1_OR_GREATER`, `NETCOREAPP3_0_OR_GREATER` etc.
+* .NET 5 and later
+    - Applies to `netX.Y` and the `NET` symbol
+    - For example, `net6.0` will define `NET`, `NET6_0`, `NET6_0_OR_GREATER`,
+      `NET5_0_OR_GREATER` etc.
+    - This will also include the corresponding defines a .NET Core 3.1 successor
+      would have gotten. For example, `net5.0` will also define `NETCOREAPP`,
+      `NETCOREAPP3_1_OR_GREATER`, `NETCOREAPP3_0_OR_GREATER` etc. But it will
+      *neither* define a `NETCOREAPP5_0` or `NETCOREAPP5_0_OR_GREATER`.
```
jaredpar, line 171: Does this go all the way back to `NETCOREAPP1_1_OR_GREATER`?

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+
+### Multi-targeting for devices to provide a cross-platform API
+
+Joletta is building a cross-platform library for geolocation. Since there is no
+built-in API for accessing geolocation, she needs to call different APIs for
+different platforms, so she uses `#if` to write different code, depending on the
+OS she is building for, plus a .NET Standard implementation that just throws
+`PlatformNotSupportedException`.
+
+Since she needs to provide custom implementations per OS, she wants to fail
```
CoffeeFlux, line 85: I don't understand what this has to do with `OR_GREATER`. It's great to have platform ifdefs like that, but they seem unrelated to the rest of the document. A better example in this vein would be making use of some platform-specific APIs only available on a certain OS version.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+fallback logic for the slow version in the `#else` branch, with no call to
+`#error`.
+
+```C#
+public class Picture
+{
+    public void Blur(float radius)
+    {
+        #if NET5_0_OR_HIGHER
+            FastBlurUsingHardwareIntrinsics();
```
CoffeeFlux, line 126: I'm not sure I really like the intrinsics example, mostly because this is architecture-dependent as well and so the checks are always more complicated than this.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+API, adding a new version wouldn't change this code, so he simply puts the
+fallback logic for the slow version in the `#else` branch, with no call to
+`#error`.
+
+```C#
+public class Picture
+{
+    public void Blur(float radius)
+    {
+        #if NET5_0_OR_HIGHER
```
CoffeeFlux, line 125:
```suggestion
        #if NET5_0_OR_GREATER
```

accepted/2020/or-greater-defines/or-greater-defines.md
CoffeeFlux: Agreed here. We also use that style for the BCL, and I think we should be consistent.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+of existing TFMs.
+
+With this feature, you'll be able to express it like this, which is a lot more
+intuitive:
+
+```C#
+void M()
+{
+    #if NET5_0_OR_GREATER
+        Net5OrHigherBehavior();
```
CoffeeFlux, line 56: Should you be using `Net5OrGreaterBehavior` here, and just generally throughout the document? There appear to be a bunch of mentions of 'Higher' still lingering.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+        #else
+            #error Provide implementation
+        #endif
+    }
+}
+```
+
+### When running on a later version of .NET, use a newer API
+
+Zachariah is building a .NET imaging library. On .NET 5, we have added enough
```
dsplaisted, line 110: Nit

```suggestion
Zachariah is building a .NET imaging library. In .NET 5, we have added enough
```

accepted/2020/or-greater-defines/or-greater-defines.md
dsplaisted, line 171:
```suggestion
      *neither* define a `NETCOREAPP5_0` nor `NETCOREAPP5_0_OR_GREATER`.
```

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+    - This will also include the corresponding defines a .NET Core 3.1 successor
+      would have gotten. For example, `net5.0` will also define `NETCOREAPP`,
+      `NETCOREAPP3_1_OR_GREATER`, `NETCOREAPP3_0_OR_GREATER` etc. But it will
+      *neither* define a `NETCOREAPP5_0` or `NETCOREAPP5_0_OR_GREATER`.
+* .NET 5 and later with operating systems
+    - The OS flavors will also gain the `XXX_OR_GREATER` variants
+    - For example, `net5.0-windows10.0.19222.0` will also define `WINDOWS`,
+      `WINDOWS10_0_19041_0`, `WINDOWS10_0_19041_0_OR_GREATER`, etc.
+* Xamarin
+    - This covers the existing Xamarin offerings, the new .NET 6-based Xamarin
```
dsplaisted, line 177:
```suggestion
    - This covers the existing Xamarin offerings, the new .NET 6-based iOS and Android support
```

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+      is handled by the previous sections.
+    - The existing Xamarin platforms don't have versioned preprocessor symbols
```
dsplaisted, line 179: I think there is some level of preprocessor symbols defined for Android API levels.  @mhutch @jonathanpeppers, @Redth

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,218 @@
...
+    - Thus, we won't be adding those
+
+### Doc changes
+
+The [documentation table][docs-table] in the [#if (C# reference)][docs] will
+need to change as follows:
```
dsplaisted, line 185: The updated table feels too verbose to me.  Instead of having an exhaustive table, perhaps the documentation should have a short description of what preprocessor symbols will be defined as well as examples.

dsplaisted (approved):

dsplaisted:
> > Are you imagining this for the 6.0.100 sdk or an earlier 5.0.N00 SDK?
> 
> That's a great question. I didn't consider this, I'll have to think about that. Would feature work like this make the bar?

I'd love to get this into 5.0.2xx (ie VS 16.9).

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: I find your lack of taste disturbing, but I succumb to the larger principle of consistency :-)

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Reworded to `Proposed Additions`

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Great catch, fixed.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Fixed

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Fixed

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Yes and no. Hardward intrinsics have a uniform API, regardless of architecture. The expectation is that the code uses runtime checks via `IsSupported`. The problem here is that if you're using .NET Standard 2.0, you don't have the APIs that allow you write that code in the first place, that's what the `#if` is really solving.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Excellent. Fixed.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: I played with that; the benefit of an exhaustive list is that it leaves no room for interpretation. For example, if someone wonders whether there is `NETCOREAPP6_0_OR_HIGHER` then can easily check that.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: > Does this go all the way back to NETCOREAPP1_1_OR_GREATER?

Yes, that's what the `etc.` was supposed to indicate. I've replaced that with an explicit range.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: > I don't understand what this has to do with OR_GREATER. It's great to have platform ifdefs like that, but they seem unrelated to the rest of the document.

Right. One pushback we got for promoting the `OR_GREATER` variants is that some think that it's good that existing code fails when new frameworks/platforms are being targeted. The point of this scenario is to outline how this line of thinkin fits in with this design.

> A better example in this vein would be making use of some platform-specific APIs only available on a certain OS version.

That's what the other scenario covers. `When running on a later version of .NET, use a newer API`. Whether that's OS API or a .NET API doesn't matter.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: I find your lack of taste disturbing, but I succumb to the larger principle of consistency :-)

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: See my response [here](https://github.com/dotnet/designs/pull/164#issuecomment-723326659). Unless there is a significant cost increase, I think it's easier if do this for all TFMs that can be targeted from the .NET 5 SDK.

accepted/2020/or-greater-defines/or-greater-defines.md
CoffeeFlux: Ah gotcha, makes sense. Thanks for clarifying!

accepted/2020/or-greater-defines/or-greater-defines.md
CoffeeFlux: Oh I see, so it's about promoting that pattern rather than the specific defines being added here. I'd still consider putting this as the second scenario in the document; it just seems off to me that the first scenario listed doesn't even use the new defines. Though perhaps this is just a 'me problem' and no one else was confused here :smile:

accepted/2020/or-greater-defines/or-greater-defines.md
jonathanpeppers: Here is where we generate preprocessor symbols on Android:

https://github.com/xamarin/xamarin-android/blob/5fc88a1a7cf67f2045e790b30e43dd7a5a28cd4f/src/Xamarin.Android.Build.Tasks/Tasks/GetAndroidDefineConstants.cs#L28-L38

`__ANDROID_16__` will be present for 16 or higher, for example.

We were planning on keeping these for .NET 6 in addition to any new ones, so we don't break existing code.

terrajobst: I'll add this to the sepc

terrajobst: @Jonathanpeppers @mhutch 

I've spec'ed this as follows. Please let me know if you have any concerns:

> * Xamarin
>    - This covers the existing Xamarin offerings, the new .NET 6-based iOS and
>      Android support is handled by the previous sections.
>    - The existing Xamarin Apple platforms don't have versioned preprocessor
>      symbols, but Android does: `__ANDROID_16__`
>    - We don't plan on adding `OR_GREATER` flavors like
>      `__ANDROID_16_OR_GREATER__`. Instead, we're only doing those for .NET 6.

dotMorten:
I like this and could use this. What I don't like however is that this entire thing feels like a hack. I'd rather be able to say `#if NETVERSION>=5.0` like C++ allows. The number of preprocessors you have to define over time will otherwise just get insane.

terrajobst:
> I like this and could use this. What I don't like however is that this entire thing feels like a hack. I'd rather be able to say `#if NETVERSION>=5.0` like C++ allows. The number of preprocessors you have to define over time will otherwise just get insane.

I'll let @jaredpar comment on this. My personal opinion is that while there is some appeal it feels like the scenario where this is useful is limiting and waiting for a language change to do this feels wrong to me.

dotMorten:
> and waiting for a language change to do this feels wrong to me.

Oh no I agree. As I said I could use this, but I'm worried about this in the longer term as it'll get unwieldy.

jaredpar:
The C# pre-processor has always had a minimalist design and that included not having logical comparisons like this. I think it would be reasonable to add the logical operators as it still keeps the design pretty minimalistic and doesn't creep into full macro territory. At the same time though the usual question is whether it's worth cost of doing it here. 

One point to consider here is the broader ecosystem. Consider if we push to change the C# compiler to allow `NETVERSION >= 5.0` and as a consequence stop generating `OR_GREATER` defines. At that point we have effectively raised the minimum bar for language participation in the .NET ecosystem to include supporting conditional defines with logical operators. That potentially pushes new costs to say F# to participate here. I say *potentially* because I don't know off hand if they support this or not.

terrajobst:
Also, the question is what the values are. For example, would three and four digit version numbers be supported? There is no intrinsic C# type for those.

drieseng:
I'm ok with having such default symbols when you select a target framework, but - last time I checked - these are "magic" symbols that are added implicitly.

I'd like us to have a way to override (remove) such symbols. Personally, I tend to use feature symbols (eg. FEATURE_DICTIONARY_TRYADD) instead and would't want to be bothered by such TFM symbols (by intellisense).

Wouldn't it be possible to explicitly set these symbols in the project file?
For example:

```xml
  <PropertyGroup>
    <TargetFrameworks>net5.0;net45</TargetFrameworks>
  </PropertyGroup>
  <PropertyGroup Condition="'$(TargetFramework)=='net5.0'">
    <TargetFrameworkConstants>NET5_0;...</TargetFrameworkConstants>
  </PropertyGroup>
  <PropertyGroup Condition="'$(TargetFramework)=='net45'">
    <TargetFrameworkConstants>NET45;...</TargetFrameworkConstants>
  </PropertyGroup>
```  

This would make these symbols easily discoverable for developers, and would allow for a setting on the Build configuration panel of a project that can be used to enable or disable these TMS symbols.

Here's a (sample) project file with these TFM symbols enabled in the **Debug** and disabled in the **Release** build configuration (this doesn't make sense, it's just to show how this could be handled):

```xml
  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFrameworks>net5.0;net45</TargetFrameworks>
  </PropertyGroup>
  <PropertyGroup Condition="'$(TargetFramework)=='net5.0'">
    <TargetFrameworkConstants>NET5_0;...</TargetFrameworkConstants>
  </PropertyGroup>
  <PropertyGroup Condition="'$(TargetFramework)=='net45'">
    <TargetFrameworkConstants>NET45;...</TargetFrameworkConstants>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
    <DefineConstants>TRACE;$(TargetFrameworkConstants)</DefineConstants>
  </PropertyGroup>
  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|AnyCPU'">
    <DefineConstants>TRACE</DefineConstants>
  </PropertyGroup>
</Project>
```
I'm aware that my comment will probably be considered off topic. This issue just reminded me of my concerns with such implicit symbols.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,221 @@
...
+    - For example, `net6.0` will define `NET`, `NET6_0`, `NET6_0_OR_GREATER` and
+      `NET5_0_OR_GREATER`.
+    - This will also include the corresponding defines a .NET Core 3.1 successor
+      would have gotten. For example, `net5.0` will also define `NETCOREAPP`,
+      `NETCOREAPP3_1_OR_GREATER`..`NETCOREAPP1_0_OR_GREATER`.
+    - This will *neither* define a `NETCOREAPP5_0` nor
+      `NETCOREAPP5_0_OR_GREATER`.
+* .NET 5 and later with operating systems
+    - The OS flavors will also gain the `XXX_OR_GREATER` variants
+    - For example, `net5.0-windows10.0.19222.0` will also define `WINDOWS`,
```
akoeplinger, line 175:
```suggestion
    - For example, `net5.0-windows10.0.19041.0` will also define `WINDOWS`,
```

so it matches with the preprocessor symbol examples

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,221 @@
...
+      `WINDOWS10_0_19041_0`, `WINDOWS10_0_19041_0_OR_GREATER`, etc.
+* Xamarin
+    - This covers the existing Xamarin offerings, the new .NET 6-based iOS and
+      Android support is handled by the previous sections.
+    - The existing Xamarin Apple platforms don't have versioned preprocessor
+      symbols, but Android does: `__ANDROID_16__`
+    - We don't plan on adding `OR_GREATER` flavors like
+      `__ANDROID_16_OR_GREATER__`. Instead, we're only doing those for .NET 6.
```
akoeplinger, line 183: Make it clearer that there will be new symbols for Android/iOS:

```suggestion
      `__ANDROID_16_OR_GREATER__`. Instead, we're only adding the symbols described in the previous section for .NET 6.
```

This effectively means we'll be adding `ANDROID_21_OR_GREATER` since the TFM will be `net6.0-android21` right?

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,221 @@
...
+> | .NET Standard     | `NETSTANDARD`, `NETSTANDARD2_1`, `NETSTANDARD2_0`, `NETSTANDARD1_6`, `NETSTANDARD1_5`, `NETSTANDARD1_4`, `NETSTANDARD1_3`, `NETSTANDARD1_2`, `NETSTANDARD1_1`, `NETSTANDARD1_0`, `NETSTANDARD2_1_OR_GREATER`, `NETSTANDARD2_0_OR_GREATER`, `NETSTANDARD1_6_OR_GREATER`, `NETSTANDARD1_5_OR_GREATER`, `NETSTANDARD1_4_OR_GREATER`, `NETSTANDARD1_3_OR_GREATER`, `NETSTANDARD1_2_OR_GREATER`, `NETSTANDARD1_1_OR_GREATER`, `NETSTANDARD1_0_OR_GREATER` |
+> | .NET 6 (and .NET Core) | `NET5_0`, `NETCOREAPP`, `NETCOREAPP3_1`, `NETCOREAPP3_0`, `NETCOREAPP2_2`, `NETCOREAPP2_1`, `NETCOREAPP2_0`, `NETCOREAPP1_1`, `NETCOREAPP1_0`, `NET6_0_OR_GREATER`, `NET5_0_OR_GREATER`, `NETCOREAPP3_1_OR_GREATER`, `NETCOREAPP3_0_OR_GREATER`, `NETCOREAPP2_2_OR_GREATER`, `NETCOREAPP2_1_OR_GREATER`, `NETCOREAPP2_0_OR_GREATER`, `NETCOREAPP1_1_OR_GREATER`, `NETCOREAPP1_0_OR_GREATER` |
```
akoeplinger, line 194: I wonder if we can leave off the `_0` from .NET5 and later, since we don't plan on releasing minor versions?

It'd save a few characters to type: `NET5_OR_GREATER`.

accepted/2020/or-greater-defines/or-greater-defines.md
```diff
@@ -0,0 +1,221 @@
+# OR_GREATER preprocessor symbols for TFMs
```
akoeplinger, line 1: Was there a discussion around the name, e.g. I'd have intuitively called this `OR_LATER` instead.

accepted/2020/or-greater-defines/or-greater-defines.md
poke: That probably goes hand in hand with [a discussion on Twitter](https://twitter.com/jaredpar/status/1324004651948150784) that happened a few days ago. Basically, `NET40_OR_GREATER` would probably conflict with what `NET40` refers to right now.

HurricanKai:
I'd be interested why this is preferred over feature specific defines, ie `SPAN_SUPPORT` or such, in my experience it's rare to check for versions, usually it's just a way to check for features, like span, Intrinsics, etc.

accepted/2020/or-greater-defines/or-greater-defines.md
akoeplinger: ah, right, I didn't see that .NET Framework symbols use the shortened name so they'd conflict.

bgrainger, line 194:
```suggestion
> | .NET 6 (and .NET Core) | `NET6_0`, `NETCOREAPP`, `NETCOREAPP3_1`, `NETCOREAPP3_0`, `NETCOREAPP2_2`, `NETCOREAPP2_1`, `NETCOREAPP2_0`, `NETCOREAPP1_1`, `NETCOREAPP1_0`, `NET6_0_OR_GREATER`, `NET5_0_OR_GREATER`, `NETCOREAPP3_1_OR_GREATER`, `NETCOREAPP3_0_OR_GREATER`, `NETCOREAPP2_2_OR_GREATER`, `NETCOREAPP2_1_OR_GREATER`, `NETCOREAPP2_0_OR_GREATER`, `NETCOREAPP1_1_OR_GREATER`, `NETCOREAPP1_0_OR_GREATER` |
```

Additionally, why does this define all of `NETCOREAPP3_1`, `NETCOREAPP3_0`, `NETCOREAPP3_1_OR_GREATER`, `NETCOREAPP3_0_OR_GREATER`? That seems incompatible with the note immediately below:

> Version-specific symbols are only defined for the version you're targeting.

It's also incompatible with the current behaviour, where targeting `net5.0` _doesn't_ define `NETCOREAPP3_1`.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: I proposed OR_LATER before and I was told that OR_GREATER is more consistent with other existing prior art.

accepted/2020/or-greater-defines/or-greater-defines.md
terrajobst: Syntax-wise, yes. But I'm not sure where we landed for the TFM versions of Android. AFAIK we weren't using API version but product versions, but that's unrelated to this. The defines would match whatever the TFM version is.

terrajobst:
@drieseng

> I'm ok with having such default symbols when you select a target framework, but - last time I checked - these are "magic" symbols that are added implicitly.
>
> I'd like us to have a way to override (remove) such symbols. Personally, I tend to use feature symbols (eg. FEATURE_DICTIONARY_TRYADD) instead and would't want to be bothered by such TFM symbols (by intellisense).

They are magic in that they are automatically defined by the SDK, but from the C# standpoint they are all just regular defines. In other words, the compiler doesn't know of any of them. It's worth pointing out that TFMs aren't the only implicit defines. We also have one for configuration. However, you can them off already:

* `DisableImplicitFrameworkDefines`
* `DisableImplicitConfigurationDefines`

Just set the ones to `true` you don't want. I'll clarify this behavior for the spec as well.

terrajobst:
I believe at this point all feedback is addressed, so I'm merging.

terrajobst:
@HurricanKai 

> I'd also be interested why this is preferred over feature specific defines, ie `SPAN_SUPPORT` or such, in my experience it's rare to check for versions, usually it's just a way to check for features, like span, Intrinsics, etc.

Sorry, I missed your comment earlier. I don't think we're going to add defines at that level of granularity because we don't want to be in the business of deciding how features will be ported between .NET implementations as this has been proven to be very difficult to predict. However, if you want to work this, you totally can. You can disable the implicit TFM defines and define them as @drieseng outlined [above](https://github.com/dotnet/designs/pull/164#issuecomment-724552466).

dsplaisted:
@marcpopMSFT @sfoslund It looks like this is ready for us to start implementation

accepted/2020/or-greater-defines/or-greater-defines.md
akoeplinger: We did land on using the API versions, I'll need to update the design doc :)

akoeplinger:
Copying here so it doesn't get lost in the resolved code reviews:

>akoeplinger 2 days ago
>Was there a discussion around the name, e.g. I'd have intuitively called this OR_LATER instead.

>terrajobst 10 hours ago
>I proposed OR_LATER before and I was told that OR_GREATER is more consistent with other existing prior art.

@terrajobst not trying to start a bikeshedding discussion about the name here but can you point me to some of that existing prior art?

marcpopMSFT:
[Here](https://docs.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindows7orgreater) is some windows prior art.  Not sure what else is out there.

terrajobst:
> [Here](https://docs.microsoft.com/en-us/windows/win32/api/versionhelpers/nf-versionhelpers-iswindows7orgreater) is some windows prior art. Not sure what else is out there.

Yep, I believe those are the ones that were pointed out to me

jkotas:
There is a lot of prior art in commonly spoken language where "later" is way more common

Search for ["Windows 8 or later"](https://www.bing.com/search?q=%22Windows%208%20or%20later%22): 28,000,000 hits
Search for ["Windows 8 or greater"](https://www.bing.com/search?q=%22Windows%208%20or%20greater%22): 11 hits

terrajobst:
> Search for ["Windows 8 or greater"](https://www.bing.com/search?q=%22Windows%208%20or%20greater%22): 11 hits

windows 7 or greater:  586,000,000 hits



😄 

But in fairness:

windows 7 or later:  1,630,000,000 hits

terrajobst:
I'm not attached to the name. Unless someone has strong feelings I'm going to rename `OR_GREATER` to `OR_LATER` then.

terrajobst:
https://github.com/dotnet/designs/pull/165/files

dotMorten:
OR_GREATER makes more sense IMHO. It's a version check, not a date check

bgrainger:
Worst case, .NET 5.1 comes out chronologically _after_ .NET 6.0 and we all get into a flamewar about whether it's "LATER" or not. 😀

terrajobst:
@dotMorten and @bgrainger, please comment on the active PR: #165.

PR #165, opened 2020-11-13
Rename OR_GREATER to OR_LATER

terrajobst:
Based on [feedback](https://github.com/dotnet/designs/pull/164#issuecomment-726021864)

### DIFF
--- a/accepted/2020/or-later-defines/or-later-defines.md
+++ b/accepted/2020/or-later-defines/or-later-defines.md
@@ -1,16 +1,16 @@
-# OR_GREATER preprocessor symbols for TFMs
+# OR_LATER preprocessor symbols for TFMs
 
 **PM** [Immo Landwerth](https://github.com/terrajobst)
 
 Today, projects automatically define `#if` symbols based on the friendly TFM
 representation. This includes both versionless- as well as version-specific
-symbols. For .NET 6, we plan to add `XXX_OR_GREATER` variants that are
+symbols. For .NET 6, we plan to add `XXX_OR_LATER` variants that are
 accumulative in nature:
 
 TFM                 | Existing | Proposed Additions
 --------------------|----------|--------------------------------------------
-`net5.0`            | `NET5_0` | `NET5_0_OR_GREATER`
-`net6.0`            | `NET6_0` | `NET6_0_OR_GREATER`, `NET5_0_OR_GREATER`
+`net5.0`            | `NET5_0` | `NET5_0_OR_LATER`
+`net6.0`            | `NET6_0` | `NET6_0_OR_LATER`, `NET5_0_OR_LATER`
 
 The goal is to align the semantics between package folder names and `#if`
 without breaking existing code.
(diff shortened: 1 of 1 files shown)


sebastienros:
Can we just count the legal votes?

akoeplinger (approved):

dotMorten:
OR_GREATER makes more sense IMHO. It's a version check, not a date check

It also makes sense as part of the discussion we had about potentially having range check support in the future. Ie `#if NETVERSION>=5.0`.

davkean:
Windows does OR_GREATER: https://docs.microsoft.com/en-us/windows/win32/sysinfo/getting-the-system-version.

ericsampson:
Will a N-1 minor release ever happen after a N major? If so, @dotMorten has a point...

dotMorten:
@ericsampson My understanding is it isn't the _plan_, but also never say never...

JamesNK:
Docs tend to use "or later" when talking about version numbers:

https://www.google.com/search?q=site%3Adocs.microsoft.com+%22or+later%22

Nirmal4G:
Can't we have both?! 😎

Tyrrrz:
How long until full pattern matching support in compilation directives?

Nirmal4G:
@dotMorten If you guys are sure about range check then do that instead of this.

I think having two similar things might make it worse than intented.

I vote for range check. It's simple.

dotMorten:
@Nirmal4G range check is a _potential_ future long-term feature. Nothing for sure.

mehrandvd:
> OR_GREATER makes more sense IMHO. It's a version check, not a date check. 5.1 could potentially be released after 6.0.
> 
> It also makes sense as part of the discussion we had about potentially having range check support in the future. Ie `#if NETVERSION>=5.0`.

What about OR_NEWER?

devlead (changes requested):
Greater makes more sense IMHO, there might be things released later that aren't greater.

umbarov:
GREATER goes well with numbers. LATER goes well with code names.
```
NET5_0_OR_GREATER
```
```
NET_PUMPKIN_OR_LATER
```

PathogenDavid:
If we’re going to have a bikeshed party, I’d like to complicate things by proposing `_OR_NEWER`, [which is what Unity uses](https://docs.unity3d.com/Manual/PlatformDependentCompilation.html).

(I’m half joking, but I do think there’s some merit to remaining consistent with a segment of the .NET ecosystem that has actively encouraged/enabled platform-specific code through preprocessor for quite some time.)

theo-albers:
I suggest OR_HIGHER

alefranz:
> OR_GREATER makes more sense IMHO. It's a version check, not a date check. 5.1 could potentially be released after 6.0.
> 
Or even 3.1.11 would be a source of confusion in a month time, as released later then 5.0

jaredpar:
Do we need to take this now `OR_LATER`?

dotMorten:
@jaredpar I think it needs to be `AND_NEVER`

terrajobst:
OK, considering the voting distribution I think we leave this as `OR_GREATER`.

PR #166, opened 2020-11-21
Fixed example links in proposal template

PathogenDavid:
I also noticed the link to the ASP.NET Designs repo in the readme (https://github.com/aspnet/specs) is dead, but I'm not sure where  it was moved to. (Or maybe it's just private?)

### DIFF
--- a/meta/template.md
+++ b/meta/template.md
@@ -12,8 +12,8 @@ check out [this document](proposals.md).
 
 ## Examples
 
-* [Windows Compatibility Pack](https://github.com/dotnet/designs/blob/master/accepted/compat-pack/compat-pack.md)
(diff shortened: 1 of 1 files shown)


akoeplinger (approved):

davidfowl:
It's not dead., it's just private.

PathogenDavid:
@davidfowl Thanks for confirming!

@akoeplinger And thanks for the merge!

akoeplinger:
Ok, we should remove it from the readme then :) @PathogenDavid mind sending a PR?

PathogenDavid:
> Ok, we should remove it from the readme then :) @PathogenDavid mind sending a PR?

Beep boop: https://github.com/dotnet/designs/pull/168

Part of me wonders if maybe we should just mark it as internal, but I suppose anyone who needs to know about that repo already does.

PR #168, opened 2020-11-25
Removed internal-only link to the ASP.NET specs repo

PathogenDavid:
As requested in https://github.com/dotnet/designs/pull/166#issuecomment-733954027

Part of me wonders if maybe we should just mark it as internal, but I suppose anyone who needs to know about that repo already does.


akoeplinger:
> Part of me wonders if maybe we should just mark it as internal, but I suppose anyone who needs to know about that repo already does.

I agree.

akoeplinger (approved):

PR #169, opened 2020-11-30
Update the "linking libraries" document to match current plans

vitek-karas:
This change updates the document to:
* Match the behavior as implemented in .NET 5
* Describe the goals and how we plan to achieve them as of .NET 6 beginning
* Provide a more detailed description of the various techniques used to solve trim warnings
* Make the description more generic to apply to all libraries, not just the core framework
* Include description of wider user experiences
* Add links to issues and proposals as of .NET 6 planning

### DIFF
--- a/accepted/2020/linking-libraries.md
+++ b/accepted/2020/linking-libraries.md
@@ -517,7 +517,8 @@ See [Default values for feature switches in trimmed console apps](https://github
 for a discussion of such defaults for console applications.
 
 Important note is that these changes will be made based on the intent
-to trim the app (`PublishTrimmed=true` in project file as described above)
+to trim the app (`PublishTrimmed=true` in project file as described in
+[Analysis and feedback](#Analysis-and-feedback))
 and thus will take effect in every stage of SDK.
 So `dotnet run` on such app will still have these features disabled,
 not just `dotnet publish`. As already mentioned, the goal is to have a
--- a/accepted/2020/linking-libraries.md
+++ b/accepted/2020/linking-libraries.md
@@ -22,14 +22,15 @@ an application and trim the classes and methods that are not used by the
 application. In many scenarios, this results in a significantly smaller packaged
 application.
 
-This document describes how to modify libraries to work well with the ILLink.
+This document describes how to modify libraries to work well with ILLink.
 
-*Note: For .NET Core, our goal is to use the .NET libraries implemented in dotnet/runtime for all
-.NET applications. In previous releases, Xamarin and Blazor applications
-have been using the Mono libraries from the mono/mono repo, which have been written with size in mind.
-However, the libraries implemented in dotnet/runtime have traditionally been concerned with throughput
-and not size. As such, it may take multiple releases to meet all size goals
-with the .NET libraries.*
+*Note: For .NET Core, our goal is to use the .NET libraries implemented in
+`dotnet/runtime` for all .NET applications. In previous releases,
+Xamarin and Blazor applications have been using the Mono libraries from the
+`mono/mono` repo, which have been written with size in mind.
+However, the libraries implemented in dotnet/runtime have traditionally
+been concerned with throughput and not size. As such, it may take multiple
+releases to meet all size goals with the .NET libraries.*
 
 In the context of trimming applications, there are two main concerns:
 
@@ -43,7 +44,7 @@ Therefore it is critical that any necessary code must be preserved in the applic
 
 ## Scenarios and User Experience
 
-The ILLink can be executed on a .NET application and the developer will be
+ILLink can be executed on a .NET application and the developer will be
 confident that the trimmed application works correctly, while reducing its
 final size.
 
@@ -63,14 +64,15 @@ There are a number of things the author of a library needs to undertake in order
 to make it work well with trimmed applications. In general they fall into
 the two buckets outlined above:
 
-1. Trimming compatibility
-1. Smaller Size
+1. Ensuring code required for the app the work is preserved by the linker
+1. Ensuring as much of the library as possible is trimmable if not needed 
+by the app
 
 ### Trimming compatibility
 
 Ensuring required code is preserved by the linker is not easy. .NET has a very
 rich Reflection capability that enables dynamic invocation of code. There are
-many cases where a static analysis tool (like the ILLink) is not able to know
+many cases where a static analysis tool (like ILLink) is not able to know
 which code is required and which isn't. For example, a .NET application could
 read a Type name from a file and create a new instance of that Type through
 Reflection. When the linker analyzes the application statically, it doesn't see
@@ -115,8 +117,8 @@ in the context of an application. The goal is for the library code itself to
 not generate any warnings (as those are not actionable) and instead either
 resolve them (by providing annotations which help the trimming tool correctly
 recognize all dependencies) or to annotate public APIs such that the problem
-can be resolved in the call site by the tool automatically or by the user following the guidance
-provided by the warning and the affected API it calls out.
+can be resolved in the call site by the tool automatically or by the user
+following the guidance provided by the warning and the affected API it calls out.
 
 Depending on the specific problematic pattern, different solutions can be employed.
 The list below is not complete list, it's just the most common patterns we've seen
@@ -126,22 +128,29 @@ be made for any affected feature.
 The order of these is essentially the recommended order in which to try to
 apply the solutions as they go roughly in increasing order of complexity.
 
-#### Refactor hardcoded reflection
+#### Hardcoded reflection
 
 The trim tool can already recognize a lot of reflection patterns if they are hard
 coded (type and member names specified as literals in the code). But it doesn't work
 on all such patterns. If the affected code is using what can be described as
 hard coded reflection it is often possible to rewrite it in such a way that the
 trim tool can recognize all of its patterns and avoid any issues that way. If that
-is not possible or practical, a reasonable attempt should be made to improve the trimming
-tool and teach it about such pattern.
+is not possible or practical, a reasonable attempt should be made to improve
+the trimming tool and teach it about such pattern.
 
 Example of a refactoring: [Using statically known types instead of types from reflection.](https://github.com/dotnet/runtime/commit/29c4895d13ed8768ee5e8307ccef01dff38c1289#diff-58f0fbbb8b2d1c2d89fdae163e981f01697cb4c2de1fe67e6e171504bc615586L1176-L1178)
 
-#### Annotate with DynamicallyAccessedMembers
+#### Requirements on `Type` values
 
-If the code passes around values of the `System.Type` type, these can be annotated
-with [`DynamicallyAccessedMembersAttribute`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute.-ctor?view=net-5.0)
+Frequently a certain method uses `Type` value to perform some reflection
+(for example getting a value of a property from the type), but the type
+in question is not known by the method, it's passed in as a parameter for example.
+Another part of the code uses such method and passes in a known type.
+Such pattern is technically possible to analyze statically as both the type
+and the requirements on it are known. To help the trim tool analyze such cases
+the `Type` values can be annotated with the requirements.
+
+This is done by adding the [`DynamicallyAccessedMembersAttribute`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicallyaccessedmembersattribute.-ctor?view=net-5.0)
 to declare requirements on the type stored in the value. The trim tool uses
 this information to both enforce the fulfillment of the requirements
 as well as validate that these are sufficient for the code using the values.
@@ -158,7 +167,7 @@ Example of using these annotations:
 [Annotate type parameter to include constructor if the method body asks for it
 via `GetConstructor`.](https://github.com/dotnet/runtime/pull/36532/files#diff-58f0fbbb8b2d1c2d89fdae163e981f01697cb4c2de1fe67e6e171504bc615586L1116-L1131)
 
-#### Suppress warnings
+#### False positives
 
 If the warning while technically correct does not affect an app's functionality,
 consider suppressing it. For this add [`UnconditionalSuppressMessageAttribute`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.unconditionalsuppressmessageattribute?view=net-5.0)
@@ -168,7 +177,7 @@ but it will be kept in the code for all configurations).
 Example of suppressing warning on code which doesn't change behavior when trimmed:
 [Trimming fields on attributes doesn’t affect equality](https://github.com/dotnet/runtime/commit/0f97c7af662c68862995c0b2ab59e87242eff459#diff-a911305f19a82ebd3a64f9293f9e5347ff15c74d5440f6bc56af65f1fa56e9b8L17-L18)
 
-#### Declare explicit dynamic dependency
+#### Dependencies which are static but not analyzable by the tool
 
 In cases where the trimming tool is not able to infer the dependency or
 the developer wants to be explicit, the [`DynamicDependencyAttribute`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicdependencyattribute?view=net-5.0)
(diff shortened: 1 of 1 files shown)


accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
-- `Assembly.Load/LoadFrom/LoadFile`
-    - Dynamically loading assemblies into a trimmed application is problematic because if the linker isn't aware of those assemblies, it won't know to preserve the methods required by those assemblies. If the core application doesn't need `string.Split`, the linker will trim the `Split` method away. When an assembly that the linker didn't know about is loaded dynamically, it may need `string.Split`, and fail to load.
-- `System.Reflection.Emit`
-    - A method or property that is only called by dynamically generated code may be trimmed if the linker isn't aware of it.
+There are a number of things the author of a library needs to undertake in order
+to make it work well with trimmed applications. In general they fall into
+the two buckets outlined above:
+
+1. Trimming compatibility
+1. Smaller Size
```
stephentoub, line 67:
```suggestion
1. Ensuring code required for correctness is preserved by the linker
1. Ensuring as much of the library as is possible is trimmable if not directly used
```
or something like that.  As written, these aren't really things that can be "undertaken".

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+it's possible the app will not work correctly after trimming.
+
+See the [Linker Reflection Flow Design](https://github.com/mono/linker/blob/master/docs/design/reflection-flow.md)
+for more information on how the process works.
+
+#### Annotating code
+
+Some library code can use reflection or other problematic patterns and thus
+running the trimming tool on it can break its functionality and it generates
+trim analysis warnings.
```
stephentoub, line 105:
```suggestion
Some library code can use reflection or other problematic patterns such that
running the trimming tool on it can break its functionality and generate
trim analysis warnings.
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+
+Warnings originating in the library code itself are not actionable by the end
+user as they frequently point to internal implementation details and it's
```
stephentoub, line 108:
```suggestion
user as they frequently point to internal implementation details, and it's
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+the affected functionality.
+
+For this reason, affected libraries need to be annotated to help trimming tools
+to better understand the code and to generate warnings which are actionable
+in the context of an application. The goal is for the library code itself to
+not generate any warnings (as those are not actionable) and instead either
+resolve them (by providing annotations which help the trimming tool correctly
+recognize all dependencies) or to annotate public APIs such that the problem
+can be resolved in the call site by the tool automatically or the warning
+is generated pointing at the user code which uses the affected API.
```
stephentoub, line 119:
```suggestion
can be resolved in the call site by the tool automatically or by the user following the guidance
provided by the warning and the affected API it calls out.
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+
+Depending on the specific problematic pattern different solutions can be employed.
```
stephentoub, line 121:
```suggestion
Depending on the specific problematic pattern, different solutions can be employed.
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+The list below is not complete list, it's just the most common patterns we've seen
```
stephentoub, line 122: The headers of each section below describe the solution rather than the pattern/problem.

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+so far. The list below describes the progression of improvements which should
+be made for any affected feature.
+
+The order of these is essentially the recommended order in which to try to
+apply the solutions as they go roughly in increasing order of complexity.
+
+#### Refactor hardcoded reflection
+
+The trim tool can already recognize lot of reflection patterns if they are hard
```
stephentoub, line 131:
```suggestion
The trim tool can already recognize a lot of reflection patterns if they are hard
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+coded (type and member names specified as literals in the code). But it doesn't work
+on all such patterns. If the affected code is using what can be described as
+hard coded reflection it is often possible to rewrite it in such a way that the
+trim tool can recognize all of its patterns and avoid any issues that way. If that
+is not possible or practical a reasonable attempt should be made to improve the trimming
```
stephentoub, line 136:
```suggestion
is not possible or practical, a reasonable attempt should be made to improve the trimming
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+tool and teach it about such pattern.
+
+Example of a refactoring: [Using statically known types instead of types from reflection.](https://github.com/dotnet/runtime/commit/29c4895d13ed8768ee5e8307ccef01dff38c1289#diff-58f0fbbb8b2d1c2d89fdae163e981f01697cb4c2de1fe67e6e171504bc615586L1176-L1178)
+
+#### Annotate with DynamicallyAccessedMembers
+
+If the code passes around values of `System.Type` type, these can be annotated
```
stephentoub, line 143:
```suggestion
If the code passes around values of the `System.Type` type, these can be annotated
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+[Cross method annotation](https://github.com/mono/linker/blob/master/docs/design/reflection-flow.md#cross-method-annotations)
+has more details on the design of this trimming feature.
+
+Example of using these annotations:
+[Annotate type parameter to include constructor if the method body asks for it
+via `GetConstructor`.](https://github.com/dotnet/runtime/pull/36532/files#diff-58f0fbbb8b2d1c2d89fdae163e981f01697cb4c2de1fe67e6e171504bc615586L1116-L1131)
+
+#### Suppress warnings
+
+If the warning while technically correct does not affect app's functionality,
```
stephentoub, line 163:
```suggestion
If the warning while technically correct does not affect an app's functionality,
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+* `Assembly.Load/LoadFrom/LoadFile`
+    - Dynamically loading assemblies into a trimmed application is problematic
+because if the trim tool isn't aware of those assemblies, it won't know to preserve
+the methods required by those assemblies. If the core application doesn't need
+`string.Split`, trimming will remove the `Split` method. When an assembly that
+the trim tool didn't know about is loaded dynamically, it may need `string.Split`,
+and fail to run.
+* `System.Reflection.Emit`
+    * A method or property that is only called by dynamically generated code may
+be trimmed if the linker isn't aware of it.
```
stephentoub, line 266: This seems like too broad a category to include here.  There are plenty of S.R.Emit solutions that can be appropriately annotated to teach the linker what APIs are actually used, and in many cases, follow patterns with hardcoded reflection the linker already understands.

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
-See https://github.com/dotnet/runtime/issues/31712 and https://github.com/dotnet/runtime/issues/35199 for more information on this issue.
+### Smaller Size
 
-To solve this, we will follow the approaches outlined in https://github.com/dotnet/runtime/issues/31712 to not rely on the Embedded Descriptor Files in order to make the Reflection usages linker safe. Instead, we will add the appropriate linker annotations and not embed these descriptor files into our libraries. This will allow applications to safely call, or not call, the APIs and the linker will either preserve or trim the code correctly.
+To make applications smaller, we have to let the trim tool know what code is not
+necessary and can be removed. A lot of times it is obvious that if a method
+isn't called, it can be trimmed and the trim tool will recognize this automatically.
+But there are places where unnecessary code looks like it is being called.
+
+#### Removing features
```
stephentoub, line 306: As with the earlier bulleted list, the previous paragraph reads to me as if it's going to be followed by a list of places where unnecessary code doesn't actually get trimmed, but then these headers aren't about such categories of problems but rather about the solutions to such problems.  Consider changing these headers to describe the category of problem that all share the same solution, rather than naming it based on the solution.

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
-Since these 2 scenarios require the use of Embedded Descriptor xml files, we will need to teach the linker to respect feature switches inside an Embedded Descriptor xml file.
+To solve these, we can define a [feature switch](#Feature-Switches) for each of
+these scenarios. An application (or SDK) can default these to on or off, as appropriate.
+For example, a Xamarin iOS application can safely turn COM support off.
 
 #### Analyzing Canonical Applications
 
-Application size is similar to throughput performance in that you can't guess at what could effect an application size. You have to measure it. In order to measure it, we will need to define a set of canonical applications that we want to measure. Once we have these applications, we can link them and then analyze the output to see what pieces are remaining, why, and investigate how to remove the unnecessary pieces.
+Application size is similar to throughput performance in that you can't guess
+at what could effect an application size. You have to measure it. In order to
```
stephentoub, line 377:
```suggestion
at what could affect an application size. You have to measure it. In order to
```

accepted/2020/linking-libraries.md
vitek-karas: I removed the RefEmit item as it's not a good example. As you mention in lot of cases it is actually perfectly compatible with trimming, or can be made that way without to much trouble.

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+application.
+
+This document describes how to modify libraries to work well with the ILLink.
+
+*Note:For .NET Core frameworks our goal is to use the .NET Libraries for all
+.NET applications. In previous releases, Xamarin and Blazor applications
+have been using the Mono libraries, which have been written with size in mind.
+However, the .NET Libraries have traditionally been concerned with throughput
+and not size. As such, it may take multiple releases to meet all size goals
+with the .NET Libraries.*
```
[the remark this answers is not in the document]
vitek-karas: It is the case for template applications for Blazor and Xamarin, where we're basically on par with mono. It's not the case for some other features - for example our serialization stacks are much larger than the relatively simplistic implementations in mono.

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,583 @@
...
+* `EventSource`
+    * On some devices the `EventSource` logic is unnecessary because it has no
+place to write the events. We can introduce a feature switch to remove
+`EventSource` code/calls. [Issue tracking this work](https://github.com/dotnet/runtime/issues/37414)
+
+#### Usage of descriptor files
+
+Trimming .NET applications has been around for a while in the Xamarin scenarios.
+To add the ability to overcome limitations of the trim tool,
+a descriptor files where introduces. These are instructions to the trim tool
```
sbomer, line 348:
```suggestion
descriptor files were introduced. These are instructions to the trim tool
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,583 @@
...
+is capable of displaying, developers could use PerfView to do the size analysis.
+
+### Analysis and feedback
+
+The goal is to make the experience predictable, as such the plan is to employ
+multiple tools to provide feedback to developers about the application's compatibility
+with trimming (and other modes of deploying the app). These tools will run in
+different stages of the SDK pipeline. Roslyn analyzers will be added to provide
+immediate feedback about the source code during code editing and in inner build loops.
+There’s already ILLink tool which performs trimming but also analyses the app
```
sbomer, line 435:
```suggestion
There’s already ILLink which performs trimming but also analyses the app
```

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,583 @@
...
+effect on for example building the application, that is setting `PublishSingleFile=true`
+will not produce a single-file from `dotnet build`.
+
+More details provided in [Analyzers for publish only functionality](https://github.com/dotnet/sdk/issues/14562)
+
+### Feature Switches
+
+Feature switches were introduced to make it possible to disable functionality
+which is typically included by default with a given .NET Core application,
+but it's either incompatible with trimming or it makes the resulting application
```
sbomer, line 460:
```suggestion
but either is incompatible with trimming or makes the resulting application
```

accepted/2020/linking-libraries.md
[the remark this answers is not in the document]
stephentoub: > for example our serialization stacks are much larger than the relatively simplistic implementations in mono

Which serialization stacks? DCS? XML? BinaryFormatter?

eerhardt (approved):
This looks great Vitek. I have mostly wording and punctuation suggestions.

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,572 @@
...
+
+Example of suppressing warning on code which doesn't change behavior when trimmed:
+[Trimming fields on attributes doesn’t affect equality](https://github.com/dotnet/runtime/commit/0f97c7af662c68862995c0b2ab59e87242eff459#diff-a911305f19a82ebd3a64f9293f9e5347ff15c74d5440f6bc56af65f1fa56e9b8L17-L18)
+
+#### Declare explicit dynamic dependency
+
+In cases where the trimming tool is not able to infer the dependency or
+the developer wants to be explicit, the [`DynamicDependencyAttribute`](https://docs.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.dynamicdependencyattribute?view=net-5.0)
+can be used to state an explicit dependency from a method to a type/method.
+The trimming tool doesn’t reason about this dependency. If the annotated method
```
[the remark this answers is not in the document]
vitek-karas: Agreed - removed the sentence.

accepted/2020/linking-libraries.md
```diff
@@ -1,170 +1,583 @@
...
+either make the hard coded code recognizable by the trim tool directly or use
+`DynamicDependencyAttribute` to explicitly express a dependency. Libraries should
+ideally not have any descriptor files in them, as those unconditionally add code
+into the application potentially making it unnecessarily large.
+[Example of a discussion on solutions of this problem](https://github.com/dotnet/runtime/issues/31712)
+
+There are 2 existing cases which are somewhat special here worth mentioning:
 
-There are 2 existing cases here worth mentioning:
 1. Types only used for Debugging
```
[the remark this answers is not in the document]
vitek-karas: I just reworded it here a bit. I think it's valuable to show this in the descriptor file - we actually implement both via descriptor files even though they're behind a feature switch.
We had a discussion if these should move to some random "Root" method or if we should keep them in the descriptor. I think the descriptor usage in these cases is OK - so I'll leave it here.

eerhardt (approved):
:shipit:

accepted/2020/linking-libraries.md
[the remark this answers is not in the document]
vitek-karas: I know that Marek mentioned XML specifically as being very large in .NET Core while the mono implementation is tiny (doesn't have the same feature-set but it's obviously enough for mono). I haven't measured this myself.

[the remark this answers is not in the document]
marek-safar: > I'd heard that we expect we're already at approximate parity. Is that not the case?

That's approximately correct for "template" like apps where we have done initial work to get the size under control. However, that's not the ultimate goal.

> Which serialization stacks?

All of them

> XML specifically as being very large in .NET Core 

XML as a whole thing is very trimming unfriendly (no work has done there and the code was not written with minimal dependencies in mind). For now, we managed to drop the dependencies to XML altogether but this might not work for end-user apps.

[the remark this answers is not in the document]
stephentoub: > All of them

Mono had its own implementation of JsonSerializer?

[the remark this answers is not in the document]
stephentoub: No. a) That's a completely different API. b) It's not a serializer. c) Its use is discouraged. d) That existed here, too: https://github.com/dotnet/corefx/tree/release/3.0/src/System.Json/src/System/Json.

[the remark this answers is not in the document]
stephentoub: And, DCS... mono had its own independent implementation?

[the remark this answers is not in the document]
marek-safar: > That's a completely different API

You asked and I'm only showing up what developers had available and not proposing touching that in any form or shape. This was JSON API with serialization support over DataContract.

> DCS... mono had its own independent implementation?

We had tweaked .NET FX implementation with static generators https://github.com/mono/mono/tree/master/mcs/class/System.Runtime.Serialization/ReferenceSources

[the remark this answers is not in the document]
stephentoub: > You asked

Right, I'm trying to understand the comment that was made.  For any given serialization stack, if it's the case that mono has a functionally-equivalent implementation that's much more trimmable, that's very interesting and should give us leads on how to proceed to improve the ones in dotnet/runtime.  If it's the case that mono used the same code and just tweaked it, then we should be able to copy those same tweaks rather than starting an investigation from scratch.  But if it's the case that we're talking about a completely different API rather than different implementations of the same API, that's apples-to-oranges and isn't interesting for this conversation / document.

vitek-karas:
I'm merging this as all feedback has been reacted to. Please let me know if there's anything else I'll be happy to fix it in a separate PR.

PR #170, opened 2020-12-04
platform-exclusion.md: platform attributes should use Build not Version with Win10

antonfirsov:
According to our [conceptual documentation](https://docs.microsoft.com/en-us/dotnet/standard/analyzers/platform-compat-analyzer) we should use Build not Version, eg. `windows10.0.18362` instead of `windows10.0.1903`, otherwise it would be inconsistent with TFM-s and capability checking API-s. Also note that Windows 10 versions are switching to new a scheme (`20H2`, `21H1` ...), which will make it impossible to use them with the attribute.

Despite the 3 reviewers (!), I almost made a mistake in dotnet/runtime#45494, which is probably the first PR in our libs using a specific version with `SupportedOSPlatformAttribute`. It was @filipnavara who spotted the mistake.

@buyaa-n @jeffhandley @terrajobst please confirm if this is correct, I need to be 100% sure before merging that PR.

### DIFF
--- a/accepted/2020/platform-exclusion/platform-exclusion.md
+++ b/accepted/2020/platform-exclusion/platform-exclusion.md
@@ -147,11 +147,11 @@ public void DoesNotWorkOnWindows();
 ```
 
 Now let's say that in .NET 6 we made the API work in Windows, but only on
-Windows `10.0.1903`. We'd update the API as follows:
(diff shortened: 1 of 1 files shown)


buyaa-n:
> please confirm if this is correct, I need to be 100% sure before merging that PR.

That is correct, it should be same what SDK uses

terrajobst (approved):

PR #174, opened 2021-02-02
.NET 6.0 TFMs

terrajobst:


accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,108 @@
...
+
+> ***OPEN QUESTION**: Should `net6.0-maccatalyst` also have a compatibility
+> relationship with any of the **existing** Xamarin TFMs, such as `xamarin.mac`
+> or `xamarin.tvos`?*
+>
+> ***OPEN QUESTION**: Should `net6.0-maccatalyst` also have a compatibility
+> relationship with any of the **new** Xamarin TFMs, such as `net6.0-mac`?*
+>
+> ***OPEN QUESTION**: Should `net6.0-maccatalyst` generate a warning when assets
+> from `xamarin.ios` are being used?
```
terrajobst, line 84: @Redth @rolfbjarne @marek-safar please help me answer these.

Redth: I think that warnings would be useful any time assets are consumed in the `net6.0-maccatalyst` TFM from `net6.0-ios`, `xamarin.ios10`, and if we add support: `net6.0-mac` and `xamarin.mac20`. 

A couple of reasons:
1. There may of course be code that throws at runtime (some iOS API's do not work in Catalyst and some AppKit/Cocoa API's do not work as well).
2. Bindings to native frameworks/libraries generally will not work (the native assets themselves would need to be rebuilt for Catalyst).
3. MSBuild logic, especially in the older, non-net6 TFM's may not work correctly.

Redth: It's also probably worth explicitly stating that `net6.0-maccatalyst` does not have compatibility with `net6.0-ios` or `net6.0-mac` assets.  I had originally assumed we might add this, but I see you did not include that and realize that was likely quite intentional, and a better choice the more I think about it.

Redth: > ***OPEN QUESTION**: Should `net6.0-maccatalyst` also have a compatibility
> relationship with any of the **existing** Xamarin TFMs, such as `xamarin.mac`
> or `xamarin.tvos`?*

`xamarin.ios` - yes (as you have in the table in the spec)

`xamarin.mac/watchOS/tvos` - no:

I would guess most NuGets in the ecosystem that would have `xamarin.mac` assets would already have `xamarin.ios` assets as well.  Also, Catalyst supports a large subset of the iOS API's, but a smaller subset off the Mac (AppKit/Cocoa) API's (at least officially).  I don't think there's enough value in adding compatibility for `Xamarin.mac` assets vs the potential incompatibilities in that scenario.  Same for other Xamarin apple TFM's.

Redth: > ***OPEN QUESTION**: Should `net6.0-maccatalyst` also have a compatibility
> relationship with any of the **new** Xamarin TFMs, such as `net6.0-mac`?*

No.

I originally thought yes for `net6.0-ios` however I'm not convinced this is the correct answer.

1. If you are building for the new `net6.0-ios` target, it doesn't seem unreasonable to expect to include the `net6.0-catalyst` target if that's the real intent.  I expect library authors will see PR's to add the TFM rather quickly if the code is actually able to be shared with no or minimal changes.
2. Native frameworks/libs need to be rebuilt for `net6.0-maccatalyst` and are slightly different than the native assets you would build for `net6.0-ios` bindings.  These would be incompatible.

No for `net6.0-mac`, `net6.0-tvos`, etc as well.  These would be potentially even less compatible with catalyst as they support a more narrow set of iOS API's.

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,110 @@
...
+|--------------------|------------------------------------------|
+| net6.0             | (subsequent version of net5.0)           |
+| net6.0-windows     | (subsequent version of net5.0-windows)   |
+| net6.0-android     | xamarin.android                          |
+|                    | (+everything else inherited from net6.0) |
+| net6.0-ios         | xamarin.ios                              |
+|                    | (+everything else inherited from net6.0) |
+| net6.0-macos       | xamarin.mac                              |
+|                    | (+everything else inherited from net6.0) |
+| net6.0-maccatalyst | xamarin.ios                              |
```
Redth, line 43: Can this compatibility mapping be _removed_ in future major .NET version?  Can `net7.0-maccatalyst` be no longer compatible with `xamarin.ios` while `net6.0-maccatalyst` is?  

I'm not completely sure if this is desirable, just trying to get a sense of if this can go away at some point in the future easily enough.

Can we warn today and err tomorrow?

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,110 @@
+# .NET 6.0 Target Frameworks
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+In [.NET 5.0][net5.0] we defined a new syntax for target frameworks (TFM) that
+allows us to avoids tools and humans to use a decoder ring, such as the .NET
```
danmosemsft, line 6:
```suggestion
makes it unnecessary to use a decoder ring, such as the .NET
```

richlander:
With the context I have, LGTM.

accepted/2021/net6.0-tfms/net6.0-tfms.md
rolfbjarne: I agree with @Redth on all three questions.

accepted/2021/net6.0-tfms/net6.0-tfms.md
rolfbjarne, line 43: A related question is whether it's possible to add compatibility for other TFMs in the future or not: can `net7.0-ios` be made compatible with `net7.0-maccatalyst` if we want to at that point, even if `net6.0-ios` is not compatible with `net6.0-maccatalyst`?

marek-safar: Is the expectation for catalyst and ios to be so highly compatible that the users won't need the special table again to know which ios version is compatible with catalyst version?

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,110 @@
...
+TFM automatically compatible with:
+
+1. Itself
+2. Earlier versions of itself
+3. The corresponding `netX.Y` TFM
+4. Earlier versions of the corresponding `netX.Y` TFM
+
+The `net5.0` TFM has the additional twist that it's compatible with
+`netcoreapp3.2` (and earlier), `netstandard2.1` (and earlier) as well as .NET
+Framework (`net4.x` and earlier, but under a warning). The .NET Framework
```
yaakov-h, line 62: .NET Framework TFMs don't have decimal points, e.g. `net472`.

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,110 @@
...
+|                    | (+everything else inherited from net6.0) |
+| net6.0-ios         | xamarin.ios                              |
+|                    | (+everything else inherited from net6.0) |
+| net6.0-macos       | xamarin.mac                              |
+|                    | (+everything else inherited from net6.0) |
+| net6.0-maccatalyst | xamarin.ios                              |
+|                    | (+everything else inherited from net6.0) |
+| net6.0-tvos        | xamarin.tvos                             |
+|                    | (+everything else inherited from net6.0) |
+| net6.0-watchos     | xamarin.watchos                          |
```
spouliot, line 47: if we do not plan on supporting watchOS then we should not document a TFM for it
c.c. @davidortinau

accepted/2021/net6.0-tfms/net6.0-tfms.md
marek-safar: @jeffhandley will this need a work in platform compact analyzer?

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: Good catch

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: > Can this compatibility mapping be _removed_ in future major .NET version? Can `net7.0-maccatalyst` be no longer compatible with `xamarin.ios` while `net6.0-maccatalyst` is?

Keeping the compat relationship rules doesn't cost us, but removing would (because it requires logic to enforce an upper boundary). It would also mess with the developer's intuition on versioning rules. I would say that (1) the old assets are naturally used less because people tend to upgrade and (2) whatever makes us comfortable with people using old assets with issues on .NET 6 should also hold for future versions. I don't think going out of our way to make this no longer work causes more trouble than it is worth.

terrajobst: @marek-safar 

> @jeffhandley will this need a work in platform compact analyzer?

Not sure what you're asking. My understanding is that starting with .NET 6, developers either target Catalyst (via `net6.0-maccatalyst`) or iOS (via `net6.0-maccatalyst`) but there is no tooling for building a single binary that can on both (well, ignoring `net6.0`, because that one wouldn't offer any Catalyst/iOS bindings). So in that sense, it doesn't really matter how Catalyst and iOS bindings work. You can have disjoint bindings or a single set of bindings with annotations -- at any point in time, the compiler/analyzer only sees a single set of bindings with version annotations.

And when referencing `xamarin.ios` from Catalyst the platform compat analyzer doesn't do anything because it only looks a the code that is being compiled; it doesn't consider indirect dependencies on platform-specific APIs.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: Awesome, I have addressed all issues now.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: Makes sense, will remove watchOS.

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,159 @@
...
+    - `maccatalyst-arm64`
+    - `maccatalyst-arm64-13`
+    - `maccatalyst-arm64-14`
+    - `maccatalyst-x64`
+    - `maccatalyst-x64-13`
+    - `maccatalyst-x64-14`
+* `net6.0-macos`
+    - TBD
+* `net6.0-tvos`
+    - TBD
```
terrajobst, line 126: @marek-safar What are the RIDs we need for Android, iOS, macOS, and tvOS?

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: You can add new compat mappings in new versions of the TFM. You can only add compa mappings to existing TFM versions if those take effect when nothing is compatible (otherwise it would be a breaking change).

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,159 @@
...
+## RIDs
+
+We need to add RIDs for all these platforms:
+
+* `net6.0-android`
+    - TBD
+* `net6.0-ios`
+    - TBD
+* `net6.0-maccatalyst`
+    - `maccatalyst-arm64`
```
yaakov-h, line 117: Apple refer to the Apple Silicon Macs as `arm64e` currently. Is .NET 6 going to use any of the extended features or is it going to target a plain `arm64` instruction set?

accepted/2021/net6.0-tfms/net6.0-tfms.md
marek-safar: There is a lot of them. If you want to list them all you can start with listing everything in

Android-arch: x64;x86;arm;arm64
Android-versions: 21 to 29 range
ios/tv-arch: arm64;x64
ios/tv-versions: 11 to 14 range
ios-arch: arm
ios-vesion: 10 only

accepted/2021/net6.0-tfms/net6.0-tfms.md
marek-safar: It matters for runtime libraries annotations. For example, will `UnsupportedOSPlatform ("ios")` be enough for maccatalyst build? Similarly, are we expecting publicly annotate unsupported API with `UnsupportedOSPlatform ("ios")` to not work also on Catalyst and not to confuse developers?

rolfbjarne: > It matters for runtime libraries annotations. For example, will `UnsupportedOSPlatform ("ios")` be enough for maccatalyst build? Similarly, are we expecting publicly annotate unsupported API with `UnsupportedOSPlatform ("ios")` to not work also on Catalyst and not to confuse developers?

It would become messy if we mixed iOS and Mac Catalyst this way, because there is Mac Catalyst-only API (as there is iOS-only API), and we'd end up with rules like "`UnsupportedOSPlatform ("ios")` applies to Mac Catalyst unless there's a Mac-Catalyst-specific availability attribute, in which case those take precedence."

I think it's best to try to avoid this complication, and just say that availability attributes for iOS do not apply to Mac Catalyst.

rolfbjarne (approved):

accepted/2021/net6.0-tfms/net6.0-tfms.md
marek-safar: I believe that's how the tooling works today so we'd need to explicitly block this configuration.

dsplaisted (approved):

accepted/2021/net6.0-tfms/net6.0-tfms.md
akoeplinger: Right now we're targeting `arm64` but support for `arm64e` is tracked with https://github.com/dotnet/runtime/issues/42738 and https://github.com/dotnet/runtime/issues/34121

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,159 @@
...
+    - `maccatalyst-arm64-13`
+    - `maccatalyst-arm64-14`
```
akoeplinger, line 119: the version comes before the arch in the RID and with a dot:

```suggestion
    - `maccatalyst.13-arm64`
    - `maccatalyst.14-arm64`
```

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,159 @@
...
+    - `maccatalyst-x64`
+    - `maccatalyst-x64-13`
+    - `maccatalyst-x64-14`
```
akoeplinger, line 122:
```suggestion
    - `maccatalyst.13-x64`
    - `maccatalyst.14-x64`
```

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,159 @@
...
+compatibility rules with more levels, using a dash for this new TFM would make
+it harder to add support for this later.
+
+### Why didn't we name it `net6.0-catalyst`?
+
+Looks like [Apple refers to it as "Mac Catalyst"](https://developer.apple.com/mac-catalyst/).
+
+### Why did we make `net6.0-maccatalyst` compatible with the existing Xamarin TFMs?
+
+It's true that for native code existing binaries largely don't work in Mac
```
akoeplinger, line 143: not just largely, existing native binaries don't work at all:

```suggestion
It's true that for native code existing binaries don't work in Mac
```

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: I would expect the Xamarin bindings to no use `[UnsupportedOSPlatform]` at all -- I think APIs that only work in iOS would be `[SupportedOSPlatform("ios<version>"]`, APIs that only work in Catalyst be `[SupportedOSPlatform("maccatalyst<version>"]` and APIs that work in both would be marked as `[SupportedOSPlatform("ios<version>"][SupportedOSPlatform("ios<version>"]`.

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,159 @@
...
+The [built-in compatibility rules][net5.0] of the `netX.Y-platform` TFMs make a
+TFM automatically compatible with:
+
+1. Itself
+2. Earlier versions of itself
+3. The corresponding `netX.Y` TFM
+4. Earlier versions of the corresponding `netX.Y` TFM
+
+The `net5.0` TFM has the additional twist that it's compatible with
+`netcoreapp3.2` (and earlier), `netstandard2.1` (and earlier) as well as .NET
```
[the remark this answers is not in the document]
terrajobst: Ah yes, my bad.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: @marek-safar, I've added as section to the spec about platform annotations which should answer your question.

marek-safar: @terrajobst I think that should work for Xamarin binding. We'll see if that will be sufficient also for runtime (e.g. different Console support).

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,329 @@
...
+example of this. The same rational can be applied to `netcoreapp` where this
+would logically be the server implementation (or maybe a .NET Core 3.x desktop
+implementation).
+
+A `net6.0-maccatalyst` project referencing a NuGet package should behave as
+follows:
+
+* It should prefer `xamarin.ios` assets over `netcoreapp` and `netstandard`
+  assets.
+    - However, it should still prefer `net5.0`/`net6.0` assets over
```
praeclarum, line 96: This seems wrong to me but it could be because I don't understand net5/6. Why would you ever want a less platform specific implementation over a more specific implementation? I would be very surprised to reference a library and find out it can't present UI because the net5/6 was chosen over ios version.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: Good point; cross-platform APIs would follow the Blazor WebAssembly pattern of using `UnsupportedOSPlatform`.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: `xamarin.ios` is the old Xamarin TFM. The assumption here is that if a package is updated, it should target iOS/Catalyst using the new TFMs (`net6.0-ios` and `net6.0-maccatalyst` respectively).

praeclarum: OK, so the requirement is that if someone updates to net5/6 then they should update their iOS TFMs too? Logical but no obvious. Gonna have to spread that tip.

terrajobst: > OK, so the requirement is that if someone updates to net5/6 then they should update their iOS TFMs too? Logical but no obvious.

My assumption, and I could totally be wrong here, is that we probably want support producing the old Xamarin TFM from the new tooling. So I'd think once you upgrade the tools to .NET 6, you're naturally already on the new TFMs anyway.

praeclarum: I feel like this migration is going much slower than you think. I anticipate a lot of my apps will be mixed for years and I would prefer nugets to keep working. I'm thinking there will be a lot of nugets out there with ALL these TFMs. That said, as long as the tooling doesn't regress on the old TFMs, that old code will keep working.

terrajobst: > @terrajobst I hand author all my nuspecs. 😅



> I feel like this migration is going to be much slower than you think. I anticipate a lot of my apps will be mixed for years 

That's fair

> and I would prefer nugets to keep working.

Also fair; let me think about the compatibility relationship. While I'm all for making more stuff work I'm also very concerned with having rules that are simple enough to ambiguities and surprises. Let me chew on that.

Redth: I tend to also think that if you're adding `net6.0-ios` support, it's not a stretch to expect you to add `net6.0-maccatalyst` as another target as well.  The tooling will not be super helpful in allowing your to construct a nupkg with a mix of old `Xamarin.iOS` and newer `net6.0-ios` TFM's to begin with unless you're producing artisanal, hand-crafted nupkg's like Frank, so I would expect many package authors to drop support for the old TFM at this point.  Even if you were going to mix `net6.0-ios` support with `Xamarin.iOS` support in the same package, it's not too crazy to throw `net6.0-maccatalyst` in there as well, right?

I think if you had all 3 TFM's (`net6.0-ios`, `net6.0-maccatalyst`, and `Xamarin.iOS`) in your nupkg, you could still support both old Xamarin and new .NET6 without issue.  

I think the only correction that might be needed here is to say `net6.0 assets are preferred' instead of net6.0 _and_ net5.0 since Xamarin was never supported on net5.0.  

The rule would be, make sure you have `net6.0-ios` and `net6.0-maccatalyst` in your packages if you want to support both iOS and macOS in a .NET 6 app, but you could still include `Xamarin.iOS` assets if you want to support apps targeting the old TFM.

jonathanpeppers: This is experimental (but maybe collection of hacks is a better description): https://github.com/xamarin/Xamarin.Legacy.Sdk

What this thing does is allow `dotnet build` running under .NET 6 to load the legacy Xamarin MSBuild targets in a multi-targeted project. So starting with something like a `net6.0-android;net6.0-ios` class library, you can add `monoandroid11.0;xamarin.ios10`.

The question is: how many packages will need to even use Xamarin.Legacy.Sdk? We've been thinking only components Xamarin maintains would need to do this. Things like AndroidX or Google Play Services.

praeclarum: I’m pretty satisfied with the arguments here. It’s logical and @Redth seems to have the bases covered.

nkolev92: I get that there are arguments for preferring platform specific over generational compatibility, but another perspective is why is a fallback, something that *might* work, being preferred over something that *will work*.

Redth: > I get that there are arguments for preferring platform specific over generational compatibility, but another perspective is why is a fallback, something that _might_ work, being preferred over something that _will work_.

In this case the "something that will work" while technically correct from an API availability perspective, is not really true in practice since the `net5.0`  assets would not contain the platform specific 'switch' implementation that is required to be useful for these platforms.  So, yes, maybe things compile, but at runtime they will throw or otherwise be completely unusable.  In many cases in the Xamarin ecosystem, if there are `xamarinios` assets in a package, it is not just adding value on top of the `netstandard` assets, but rather it's a full bait and switch where the `netstandard` assets are really just considered a reference assembly with the public API contracts and stub implementations which throw at runtime.

My recommendations:
 - `net6.0-maccatalyst` prefers fallback path (in order) of: `net6.0`, `xamarinios`, `netstandard` 
 - `net6.0-ios` prefers the fallback path (in order) of `net6.0`, `xamarinios`, `netstandard`
 - The fallback paths explicitly exclude `net5.0` assets
 
Since Xamarin was never compatible in the net5 generation of TFM's and we're already seeing some packages (such as ReactiveUI) which ship assets for both `net5.0` and `xamarinios` today, it's important that we don't prefer `net5.0` assets over `xamarinios`.  If you have a `xamarinios` project today you'll get `xamarinios` assets instead of `net5.0` assets, and changing this for projects upgraded to `net6.0-ios` would break/change the expected behaviour.

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,347 @@
...
+      `xamarin.ios`.
+    - It also shouldn't be compatible with any other existing Xamarin TFM (such
+      as `xamarin.mac`)
+* Generate NuGet warning [NU1701] when a `xamarin.ios` asset is being used
```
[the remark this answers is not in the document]
terrajobst: Agreed. As I said on email, I'm not particularly fond of this either. But I can get behind this because this doesn't introduce new crazy rules between the new TFMs but between a new TFM and an old TFM. I can justify that with our obligation to existing customers.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: > * `net6.0-maccatalyst` prefers fallback path (in order) of: `net6.0`, `xamarinios`, `netstandard`

This makes sense. I have updated the spec accordingly.

> * `net6.0-ios` prefers the fallback path (in order) of `net6.0`, `xamarinios`, `netstandard`

That doesn't make sense to me. Why would you prefer `net6.0` over `xamarinios`? I thought we'd want to treat `xamarinios` as the previous version of `net6.0-ios`, which is basically how `net5.0` is treated to `netcoreapp`.

nkolev92: I'm probably a bit pedantic here, but we should avoid using the term fallback here. 
Fallback is usually correlated with AssetTargetFallback, and this isn't that. Sounds like the ask is to have explicit compatibility between these frameworks.

accepted/2021/net6.0-tfms/net6.0-tfms.md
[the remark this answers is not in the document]
nkolev92: If xamarin.ios is preferred over netcoreapp, I think that the warning is weird as well. Same reasoning as above applies, it's not really a fallback, it's just how compatibility is in this case.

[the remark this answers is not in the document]
terrajobst: The problem I see is this:

1. Silently using the .NET Core assets is unlikely to result in a workable application due to bait & switch
2. Slenetly using xamarin.ios, while having a higher chance of working, might not work either because Mac Catalyst isn't 100% compatible with iOS.

So it seems to me the precedence rules + warning is best user experience we can deliver.

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,396 @@
...
+1. `net6.0`
+1. `xamarin.ios` (no warning)
+1. `net5.0`
+1. `netcoreapp3.1` – `1.0`
+1. `netstandard2.1` – `1.0`
+1. `net4.x` – `1.0` (warning)
+
+**net6.0-maccatalyst**
+
+1. `net6.0-ios`
```
Redth, line 124: This should be `net6.0-maccatalyst`, yes ?

Redth:
```suggestion
1. `net6.0-maccatalyst`
```

accepted/2021/net6.0-tfms/net6.0-tfms.md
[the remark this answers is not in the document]
nkolev92: Given 1, why not warn in the case of .NET Core assets older than 3.1 as well? 

The thing that stands out to me is the following (actual names abstracted)

Given A: 

A - works
A' - works
B - works (with a warning)
C - not likely to work (no warning). 

I know it's an oversimplification, but just curious :)

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,396 @@
...
+| net6.0-tvos        | xamarin.tvos                             |
+|                    | (+everything else inherited from net6.0) |
+
+### Compatibility rules
+
+The [built-in compatibility rules][net5.0] of the `netX.Y-platform` TFMs make a
+TFM automatically compatible with:
+
+1. Itself
+2. Earlier versions of itself
```
mhutch, line 71: Maybe we should make it clear this list order does not correspond to precedence?

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,396 @@
...
+
+1. `net6.0-ios`
+1. `net6.0`
+1. `xamarin.ios` ([NU1701] warning)
+1. `net5.0`
+1. `netcoreapp3.1` – `1.0`
+1. `netstandard2.1` – `1.0`
+1. `net4.x` – `1.0 ([NU1701] warning)
+
+**net6.0-macos**
```
mhutch, line 132: Why doesn't this support xamarin.mac assets?

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,396 @@
...
+
+1. `net6.0-macos`
+1. `net6.0` - `5.0`
+1. `netcoreapp3.1` – `1.0`
+1. `netstandard2.1` – `1.0`
+1. `net4.x` – `1.0` ([NU1701] warning)
+
+**net6.0-tvos**
```
mhutch, line 140: Why doesn't this support `xamarintvos` assets?

accepted/2021/net6.0-tfms/net6.0-tfms.md
```diff
@@ -0,0 +1,396 @@
...
+    {
+        public static bool IsMacCatalyst();
+        public static bool IsMacCatalystVersionAtLeast(int major, int minor = 0, int build = 0);
+    }
+}
+```
+
+## Platform support annotations
+
+It seems the Xamarin team wants to have a single assembly with bindings for the
```
mhutch, line 165: Can we clarify the purpose of having a single assembly?

accepted/2021/net6.0-tfms/net6.0-tfms.md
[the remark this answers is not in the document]
terrajobst: It's not that netcoreapp/netstandard assets wouldn't work in Mac Catalyst; it's that if a package offers both xamarin.ios and netcoreapp/netstandard then the likelyhood of netcoreapp/netstandard being the right ones is low. IOW, it's probabilty isn't just a function of the project TFM and the a given offered TFM, it's about which TFM to pick given a set of TFMs. And that's what the precedence rules try to accomodate.

Does that make more sense?

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: I believe it is the default precedence for a regular `netX.Y-platform`, no?

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: A good point, will add.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: A good point, will add.

accepted/2021/net6.0-tfms/net6.0-tfms.md
terrajobst: The point I was trying to make was less about a single assembly vs multiple but a having a combined set of bindings for Mac-based platforms. Is that not accurate?

mw-developers (commented):
Duplicate of #174

PR #178, opened 2021-02-16
Proposal to disable multi-level lookup by default

vitek-karas:
This document describes the reasoning and provides detailed description of the proposed changes in behavior.


proposed/disable-multi-level-lookup-by-default.md
```diff
@@ -0,0 +1,204 @@
...
+1. Use the version of the specified "dotnet root" - this would be more in line
+   with what the "dotnet root" parameter tries to convey. This might be technically
+   a bit challenging in some cases.
+
+I think it should be OK to go with option 1 - rely on the version of `hostfxr` -
+but this will need more detailed investigation.
+
+## Alternative solutions
+
+### Off by default everywhere
```
jkotas, line 154: We will effectively end up here eventually once .NET 5 and lower are out of support.

We may want to say that we will delete support multi-level-lookup one year after .NET 5 and lower are out of support, so that we do not have to carry it forever.

proposed/disable-multi-level-lookup-by-default.md
```diff
@@ -0,0 +1,204 @@
...
+This is a proposal to change the default to disable multi-level lookup.
+
+* Applications will change behavior based on the target .NET version.
+This version will be determined by the `tfm` property in `.runtimeconfig.json`.
+    * Applications targeting .NET 6 and above will act as if `DOTNET_MULTILEVEL_LOOKUP`
+    is set to `0` by default.
+    * Applications targeting .NET 5 and below (.NET Core 3.\*, .NET Core 2.\*, ...)
+    will keep their current behavior which is as if `DOTNET_MULTILEVEL_LOOKUP`
+    is set to `1`.
+    * All applications will continue to obey an explicit setting of
```
jkotas, line 51: I think it would be better to always ignore `DOTNET_MULTILEVEL_LOOKUP` for .NET 6+ apps, with no option to override.

proposed/disable-multi-level-lookup-by-default.md
jkotas: https://github.com/dotnet/designs/pull/178/files#diff-1ae0e0e27bffb4488839ad85db6c7024d71b809120c6a4cf46fcae385ba3f745R51 would have to be changed to this to work.

proposed/disable-multi-level-lookup-by-default.md
brianrob: I like this idea.  Are there any scenarios that we're aware of where MLL behavior is desired?

proposed/disable-multi-level-lookup-by-default.md
```diff
@@ -0,0 +1,204 @@
...
+
+### New dotnet CLI actions for install/update
+
+This is effectively an extension to the discrepancy between app behavior and
+`dotnet --info` behavior. What is the desired behavior for the new CLI actions -
+this is yet to be determined.
+
+### Way to dismiss informational message about the change
+
+There should be some way to dismiss the information messages printed by
```
brianrob, line 118: We could also consider only showing the message for a period of time, so that we don't have to have a file-based scheme.  Imagine something like "Show it for 1 week past installation."

proposed/disable-multi-level-lookup-by-default.md
```diff
@@ -0,0 +1,204 @@
...
+typically a very explicit action, and so it's reasonable to assume users
+will keep these consistent.
+
+Doing the same for applications is a tougher proposition because they're not
+governed by one `global.json` (typically repos or solutions have only one `global.json`),
+instead they're governed by their own `.runtimeconfig.json` which may
+(and usually are) targeting multiple versions even from one project/repo.
+
+### Off by default everywhere and auto-roll-forward
+
```
brianrob, line 183: I'm afraid that this would cause the same kind of problems that we run into now with MLL.  Users won't know that they've been upgraded, and if something doesn't work quite right, there won't be an immediately obvious way to show them that they've been upgraded.

proposed/disable-multi-level-lookup-by-default.md
brianrob: If we go with something like this, we could also show the warning message unconditionally in places like `dotnet --info`, and then just remove the message when we remove MLL in a future patch.

proposed/disable-multi-level-lookup-by-default.md
richlander: I'd be happy with showing a message during previews only.

proposed/disable-multi-level-lookup-by-default.md
richlander: Yup. I think this is the best plan.

proposed/disable-multi-level-lookup-by-default.md
richlander: I think a variant of this is the winning ticket:

- .NET 7+ doesn't honor MLL.
- There are good controls to enable manual roll-forward of apps (including global controls), with nothing automatic.

brianrob: Sounds great.  Thanks @richlander.  Do we need to make any changes to this doc, or do you think this is ready for merge?

proposed/disable-multi-level-lookup-by-default.md
vitek-karas: > Are there any scenarios that we're aware of where MLL behavior is desired?

Haven't really seen one.

vitek-karas: Thanks everybody - changed the doc to effectively remove MLL for TFM 7+.

proposed/disable-multi-level-lookup-by-default.md
vitek-karas: Added both options to the list.

proposed/disable-multi-level-lookup-by-default.md
vitek-karas: Changed the proposal to remove MLL for .NET 7+ apps. Also added a section about the plan to completely remove MLL for everything once all downlevel runtimes are out of support.

proposed/disable-multi-level-lookup-by-default.md
vitek-karas: I agree with @richlander  on this. For applications we're almost there (missing exact version override, but roll forward works everywhere). For SDK we should add commandline/env options.

I think that if we pay close attention to the experience when things fail - so our error messages/docs, there should be much less need to make things work magically as well as up-front messages.

vitek-karas:
I updated the doc for .NET 7 and applied feedback. Most notable changes:
* Target .NET 7 as the release where this will happen
* For .NET 7+ apps, multi-level lookup is completely disabled with no override

@richlander @jkotas @brianrob @MarcoRossignoli please take another look.

Once we agree on this I'll move the file to accepted and update the index and so on to merge it.

MarcoRossignoli:
LGTM 
`dotnet test` relies on DOTNET_ROOT(SDK one) for the apphost otherwise `dotnet exec` for the correct architecture(should not have any impact on fx resolution), also we can recognize current sdk version and the target tfm before starting the host to output the warning.

cc: @nohwnd

proposed/disable-multi-level-lookup-by-default.md
```diff
@@ -0,0 +1,222 @@
...
+The message should only be shown when there's observable difference in behavior.
+For CLI commands that means when running the command with multi-level lookup on
+would end up using different version of SDK than when running it with
+multi-level lookup off.
+
+`dotnet --info` should print the message in all cases where it's running
+from a private install location (so the global one doesn't match the current).
+
+We should look into UX around `dotnet run` and `dotnet test`, if these should
```
MarcoRossignoli, line 110: in case of fx failure for 7+(or temporary) are we changing the UX or do we keep the standard one?

proposed/disable-multi-level-lookup-by-default.md
```diff
@@ -0,0 +1,222 @@
...
+also print out a message when targeting downlevel runtimes.
+
+We should also consider expanding the error message printed out by the host
```
MarcoRossignoli, line 113: strongly agree with this one, usually on test plat we rely a lot on failure resolution message to understand what's happened.

brianrob: Agreed - this is a good way to educate folks if they happen to encounter behavior they don't expect.

proposed/disable-multi-level-lookup-by-default.md
```diff
@@ -115,20 +129,24 @@ this is yet to be determined.
...
+   .NET 7 previews, but remove them for final release. 
 1. Require the user to set `DOTNET_MULTILEVEL_LOOKUP=0` - which will disable
-   the message.
+   the message. Would go against our plans to completely remove the feature
+   in the future as we would train users to set this anyway.
 1. Only show the message once (would require a way to remember that it was shown,
-   similar to first run CLI experience).
+   similar to first run CLI experience). The timing is problematic, the first
+   invocation could be from a script and thus invisible to the user.
+   Alternative is to show it for a time period after install.
```
brianrob, line 147: Something else to consider is a way to suppress the warnings that writes into a file that already gets loaded.  That way, the cost of loading a new file isn't added into the mix, but the user has a way to make them disappear once they no longer want to see them.

brianrob:
Overall, LGTM.  Thanks @vitek-karas!

proposed/disable-multi-level-lookup-by-default.md
vitek-karas: I think we could try to do a good job at showing the message only when it actually matters. In the error case, we could even go and do the lookup "with MLL enabled" and compare the results, and only show the message if they differ. Not saying we need to do that, but it's possible.

For the SDK lookup part - I wonder if we should add a message always, or just when it fails. The message should be shown only when using private runtime anyway, which is already a relatively small subset of overall uses, and I want to see how much more we can reduce it without taking a noticable perf hit (running the result "with" MLL might be too expensive, but could measure that).

brianrob: Sounds good.  I agree that running "with" is likely too expensive.  I remember digging into the cost of MLL at one point, and thinking that it would not be ideal to have to run it to decide whether or not to expose error messages.  Hopefully, we can get to a point where we only run MLL when explicitly requested for those versions of .NET that support require it.

PR #183, opened 2021-02-25
Align minimum-os-version with platform-checks

mhutch:
Update the Minimum OS Version spec to align with the later designs in the Platform Checks speck.

### DIFF
--- a/accepted/2020/minimum-os-version/minimum-os-version.md
+++ b/accepted/2020/minimum-os-version/minimum-os-version.md
@@ -42,7 +42,7 @@ Availability annotations are intended to be used both in third-party assemblies 
 
 Earlier versions of this section proposed a form of these annotation attributes based on the Xamarin.iOS/Xamarin.Mac attributes, but it has been superseded by a further refined and generalized form defined in the [attributes section of the Platform Checks spec](../platform-checks/platform-checks.md#attributes).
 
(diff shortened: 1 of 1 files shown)


accepted/2020/minimum-os-version/minimum-os-version.md
```diff
@@ -40,124 +40,53 @@ To enable tooling to provide a good experience, assemblies should be annotated w
...
-[Introduced(PlatformIdentifier.macOS, 10, 2)]
-[Deprecated(PlatformIdentifier.macOS, 10, 9)]
-[Removed(PlatformIdentifier.macOS, 10, 12, 1)]
-public class Foo : NSObject { ... }
-```
-
-The purpose of keeping ‘removed’ APIs and annotating them with `RemovedAttribute` is to allow apps and libraries to continue to use these APIs in fallback paths for older OS versions. This is a pattern that exists on Apple platforms.
-
-If and when C# support target typing for enums, we could add convenience overloads that would improve readability:
+These availability attributes omit the `Deprecated` availability kind from the Xamarin attibutes as it’s redundant, and rename `Introduced` to `Supported` to better reflect their alternate purpose of annotating multi-platform assemblies. They also omit the ability to differentiate availability based on architecture for reduction of scope, thought this could be added if needed in future.
```
rolfbjarne, line 45:
```suggestion
These availability attributes omit the `Deprecated` availability kind from the Xamarin attributes as it’s redundant, and rename `Introduced` to `Supported` to better reflect their alternate purpose of annotating multi-platform assemblies. They also omit the ability to differentiate availability based on architecture for reduction of scope, though this could be added if needed in future.
```

mhutch:
@terrajobst ping

terrajobst (approved):

PR #186, opened 2021-03-11
Add UTF-8 design doc stub

krwq:
**Note: This document is just a stub for the structure of the documentation.**

We have some scenarios we'll be adding but we will be adding each of them as a separate PR so that they can have separate discussion as needed.


accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,21 @@
+# Improve UTF-8 support
+
+**DRAFT**
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+UTF-8 is currently the most popular encoding on the web with [more than 95% of all websites using it](https://w3techs.com/technologies/cross/character_encoding/ranking). Additionally it's used as a default encoding on Linux. When .NET was created there was many character encodings used widely and UTF-8 was not dominant at the time - .NET has chose to use UTF-16. As UTF-8 became more popular it become more obvious that UTF-8 has many memory advantages over other encodings including UTF-16 and therefore its great adoption.
```
GrabYourPitchforks, line 7: In addition to Linux, UTF-8 is also going to become the preferred code page on Windows in a future release.

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+## Benefits of using UTF-8 over UTF-16
```
GrabYourPitchforks, line 9: For completeness, we should also add a drawbacks section. The primary drawbacks are:

* If you perform heavy text processing or utilize globalization-related features heavily (e.g., a GUI application localized to your native language), UTF-16 is often faster than UTF-8 due to technical details of its implementation.
* For string which consist primarily of CJK characters, UTF-8 is less memory-efficient than UTF-16.

These conditions aren't expected to occur frequently in web applications or other application types where TTFB and memory utilization are critical metrics.

GrabYourPitchforks (approved):
Thank you for getting the skeleton structure up! Much appreciated! :)

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+- Reduces heap allocations - in many apps the number of memory needed for storing text is expected to be reduced twice
+- Improves quality, readability, and maintainability of code - performance oriented apps frequently store UTF-8 strings as byte arrays or workaround the platform limitations in other ways
+- Improves integration with subsystems that use UTF-8 natively - for example on Linux virtually all APIs take UTF-8 strings - they currently need to transcode from/to UTF-16 - sometimes multiple times - on every API call
```
jkotas, line 13: > on Linux virtually all APIs take UTF-8 strings

Technically, Linux APIs take C strings (ie zero terminated sequences of bytes). UTF-8 encoding for these strings is a convention, but not a requirement. For example, it is easy to create a file with a filename that is invalid UTF-8.

accepted/2021/utf8/requirements/README.md
```diff
@@ -0,0 +1,26 @@
+# Requirements
+
+**DRAFT**
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+## [UTF-8 Strings should be pre-validated](validation.md)
+
+- allow us to do better optimizations, for example when enumerating runes
```
jkotas, line 9: As we discussed offline, I do not think this is the right choice for .NET. I think we should be more like golang and the validity of the encoding should be a soft requirement.

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,21 @@
...
+
+**DRAFT**
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+UTF-8 is currently the most popular encoding on the web with [more than 95% of all websites using it](https://w3techs.com/technologies/cross/character_encoding/ranking). Additionally it's used as a default encoding on Linux. When .NET was created there was many character encodings used widely and UTF-8 was not dominant at the time - .NET has chose to use UTF-16. As UTF-8 became more popular it become more obvious that UTF-8 has many memory advantages over other encodings including UTF-16 and therefore its great adoption.
+
+## Benefits of using UTF-8 over UTF-16
+
+- Reduces heap allocations - in many apps the number of memory needed for storing text is expected to be reduced twice
```
jkotas, line 11:
```suggestion
- Reduces heap allocations - the amount of memory needed for storing text may be reduced up 2x compared to UTF-8 encoding
```

accepted/2021/utf8/requirements/README.md
```diff
@@ -0,0 +1,26 @@
...
+- it's consistent with other ecosystems
+- better safety and reliability
+
+## String/UTF-8 String literals should be in read-only memory (read-only section of PE file)
+
+- allows to share strings between processes
+
+## UTF-8 Strings should be zero-terminated
```
jkotas, line 17: This is implementation detail, not a requirement. The requirement should be "Performant managed/native interop for UTF-8 strings".

accepted/2021/utf8/requirements/README.md
```diff
@@ -0,0 +1,26 @@
...
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+## [UTF-8 Strings should be pre-validated](validation.md)
+
+- allow us to do better optimizations, for example when enumerating runes
+- it's consistent with other ecosystems
+- better safety and reliability
+
+## String/UTF-8 String literals should be in read-only memory (read-only section of PE file)
```
jkotas, line 13: This is implementation detail, not a requirement.

accepted/2021/utf8/requirements/README.md
```diff
@@ -0,0 +1,26 @@
+# Requirements
```
jkotas, line 1: Are these requirements for UTF-8 support in general, or something more specific?

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,21 @@
...
+- Improves quality, readability, and maintainability of code - performance oriented apps frequently store UTF-8 strings as byte arrays or workaround the platform limitations in other ways
+- Improves integration with subsystems that use UTF-8 natively - for example on Linux virtually all APIs take UTF-8 strings - they currently need to transcode from/to UTF-16 - sometimes multiple times - on every API call
+- Competitivenes with other platforms that offer UTF-8 primitives - most of the popular languages currently offer some sort of UTF-8 string support while .NET only allows transcoding to/from byte array
+- Reduced CPU time for transcoding between UTF-16 and UTF-8 - currently many apps get data as UTF-8 and/or need to produce UTF-8 document output - while transcoding is very fast and in a typical complex app it will not show up very high on the perf metrics it can still add up to couple of percent of CPU cycles
```
jkotas, line 15: It would be useful to have "Data" section with more detailed hard data.

accepted/2021/utf8/requirements/README.md
stephentoub: Presumably the requirement would be more like "Reduced memory consumption on a densely-packed machine"?

accepted/2021/utf8/requirements/README.md
GrabYourPitchforks: Krzysztof, would it make sense to call these opinionated behaviors / areas of investigation rather than requirements? That more accurately describes that these are open for discussion and modification.

accepted/2021/utf8/README.md
sylveon: I don't believe that's much of an all-in: the -A APIs internally convert to UTF-16 and then call the appropriate -W API. This is just increasing compatibility with many programs which blindly assumed that -A APIs had UTF-8. I don't expect the native UTF-16 encoding to ever flip.

jkotas: I agree that the Windows OS as a whole is always going to be UTF-16 native. I guess that there may be clusters of new APIs that are UTF-8 native.

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,21 @@
...
+**DRAFT**
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+UTF-8 is currently the most popular encoding on the web with [more than 95% of all websites using it](https://w3techs.com/technologies/cross/character_encoding/ranking). Additionally it's used as a default encoding on Linux. When .NET was created there was many character encodings used widely and UTF-8 was not dominant at the time - .NET has chose to use UTF-16. As UTF-8 became more popular it become more obvious that UTF-8 has many memory advantages over other encodings including UTF-16 and therefore its great adoption.
+
+## Benefits of using UTF-8 over UTF-16
+
+- Reduces heap allocations - in many apps the number of memory needed for storing text is expected to be reduced twice
+- Improves quality, readability, and maintainability of code - performance oriented apps frequently store UTF-8 strings as byte arrays or workaround the platform limitations in other ways
```
sylveon, line 12: At the inconvenience of UTF-16 code now dropping in quality, readability and maintainability because of the need to do workarounds being transferred to them.

accepted/2021/utf8/README.md
sylveon, line 13: WinRT APIs only take UTF-16 strings, so this makes integration with them as well as producing them much harder and need transcoding.

tannergooding: There are likewise certain interop structs and APIs on Windows that only take ~~UTF16~~ `WCHAR_T`. The vast majority do have both `A` and `W` overloads, however.

accepted/2021/utf8/requirements/README.md
jeffhandley: > Krzysztof, would it make sense to call these opinionated behaviors / areas of investigation rather than requirements?

I want to get the skeleton/stub merged in so that we can then iterate on these specific docs individually and have focused discussions on them. So this approach sounds good to me.

accepted/2021/utf8/requirements/README.md
jeffhandley: For UTF-8 support in general. We expect this will become the index page for the collection of requirements docs.

accepted/2021/utf8/README.md
jeffhandley: @jkotas What do you think about including that in each of the scenarios docs that will be produced so that the performance data section could accompany the specific characteristics of each scenario separately?

jkotas: Sounds good to me.

accepted/2021/utf8/README.md
GrabYourPitchforks: > I guess that there may be clusters of new APIs that are UTF-8 native.

This matches my understanding.

krwq: I'll leave this as is since I'm not seeing any definite source of truth for what Windows is going to do. The only thing we can say for sure is that they recommend UTF-8 usage

accepted/2021/utf8/requirements/README.md
krwq: I agree, I will change this whole section to `Areas of investigation`

accepted/2021/utf8/requirements/README.md
krwq: Since I'm changing the whole doc into `Areas of investigation` I'll leave the title as is (in form of question) but use this sentence instead of first bullet point since it captures this better

accepted/2021/utf8/README.md
krwq: Adding this to the drawbacks section

accepted/2021/utf8/README.md
krwq: Adding this to the drawbacks section

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+UTF-8 is currently the most popular encoding on the web with [more than 95% of all websites using it](https://w3techs.com/technologies/cross/character_encoding/ranking). Additionally it's used as a default encoding on Linux. When .NET was created there was many character encodings used widely and UTF-8 was not dominant at the time - .NET has chose to use UTF-16. As UTF-8 became more popular it become more obvious that UTF-8 has many memory advantages over other encodings including UTF-16 and therefore its great adoption.
+
+## Benefits of using UTF-8 over UTF-16
+
+- Reduces heap allocations - the amount of memory needed for storing text may be reduced up to 2x compared to UTF-16 encoding
+- Improves quality, readability, and maintainability of code - performance oriented apps frequently store UTF-8 strings as byte arrays or workaround the platform limitations in other ways
+- Improves integration with subsystems that are UTF-8 or mostly UTF-8 - for example on Linux most of the APIs are UTF-8 strings are by convention - they currently need to transcode from/to UTF-16 - sometimes multiple times - on every API call
```
jkotas, line 13:
```suggestion
- Improves integration with subsystems that are UTF-8 or mostly UTF-8 - for example on Linux most of the APIs are UTF-8 strings by convention - they currently need to transcode from/to UTF-16 - sometimes multiple times - on every API call
```

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,30 @@
+# Improve UTF-8 support
+
+**DRAFT**
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+UTF-8 is currently the most popular encoding on the web with [more than 95% of all websites using it](https://w3techs.com/technologies/cross/character_encoding/ranking). Additionally it's used as a default encoding on Linux. When .NET was created there was many character encodings used widely and UTF-8 was not dominant at the time - .NET has chose to use UTF-16. As UTF-8 became more popular it become more obvious that UTF-8 has many memory advantages over other encodings including UTF-16 and therefore its great adoption.
```
jkotas, line 7:
```suggestion
UTF-8 is currently the most popular encoding on the web with [more than 95% of all websites using it](https://w3techs.com/technologies/cross/character_encoding/ranking). Additionally it's used as a default encoding on Linux. When .NET was created there was many character encodings used widely and UTF-8 was not dominant at the time - .NET has chosen to use UTF-16. As UTF-8 became more popular it become more obvious that UTF-8 has many memory advantages over other encodings including UTF-16 and therefore its great adoption.
```

krwq:
@jkotas can this be merged? We can do more corrections in the next PRs here as we add scenarios

jkotas:
Fine with me

jeffhandley (approved):

accepted/2021/utf8/README.md
```diff
@@ -0,0 +1,30 @@
+# Improve UTF-8 support
+
+**DRAFT**
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
```
terrajobst, line 5: We just removed the PM/Dev markers. Please change this, rebase on top of main, and run `./update-index`  in the root.

```suggestion
**Owner** [Immo Landwerth](https://github.com/terrajobst)
```

terrajobst:
I'm not a fan of these stub documents with that many files. I'd prefer a single document that uses our template.

jeffhandley:
> I'm not a fan of these stub documents with that many files. I'd prefer a single document that uses our template.

I thought we had an agreement on this one a while back, but perhaps we had a miscommunication @terrajobst. As we start to add in the different scenarios, this will get unwieldy as a single document. And the back-and-forth on the validation doc was a good example for why we need standalone PRs on focused topics.

terrajobst:
> I thought we had an agreement on this one a while back, but perhaps we had a miscommunication @terrajobst.

Must have missed that conversation then :-)

> As we start to add in the different scenarios, this will get unwieldy as a single document. And the back-and-forth on the validation doc was a good example for why we need standalone PRs on focused topics.

Let's see. So far, I've not a seen a single design where multiple documents would have been better.

jeffhandley:
I talked with @terrajobst offline. We are flattening the structure by removing the subfolders and consolidating the README docs into the top-level README, but leaving the ability to have child documents.

PR #187, opened 2021-03-12
Adding the skeleton for statics-in-interfaces

tannergooding:
CC. @terrajobst, @MadsTorgersen 

This is the skeleton (following the same template as for utf8) for the statics-in-interfaces/generic-math design for the libraries.


stephentoub (approved):
It looks like a skeleton :)

danmoseley (approved):
I approve this skeleton. A fine skeleton🙂

PR #189, opened 2021-03-15
.NET runtime security mitigations

jkotas:


accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+GC heap compaction can create a second copy of secrets stored in managed memory. It makes it impossible for an application to reliably clear in-memory secrets stored in managed memory after they are no longer used.
+
+We will introduce GC feature that clears unused memory after GC heap compaction. Having the GC clear memory after compaction improves the security of applications for which heap dumps and in-memory probes are within their scope of threat. [dotnet/runtime#10480](https://github.com/dotnet/runtime/issues/10480) has more details.
+
+## Dangerous Deserialization
+
+Object deserialization vulnerabilities continue to be a major issue for many type safe languages. These vulnerabilities allow the attacker to introduce unexpected object graphs into the remote process.
+
+We plan to gradually [deprecate binary formatter](https://github.com/dotnet/designs/blob/main/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md) that has been the source of the worst .NET serialization vulnerabilities. We will work on excising other similar dangerous serialization patterns from the .NET ecosystem.
```
blowdart, line 184: _We will work on excising other similar dangerous serialization patterns from the .NET ecosystem._

This, whilst laudable, has bad optics "Microsoft kills popular package for save files" is what it reads like. Do you really mean the whole ecosystem, including 3rd party code, or just our own?

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Object and Array Pooling
+
+`ArrayPool<T>` and other similar custom object pools are equivalent of the classic C malloc. Manual lifetime management traditionally suffers from use-after-free and double-free security vulnerabilities. Many years were spent on implementing diagnostics and mitigations of these vulnerabilities for malloc.
+
+We need to implement similar diagnostic and mitigation capabilities for ArrayPool\&lt;T\&gt; and friends. It is [not unusual](https://github.com/dotnet/corefx/pull/37270) to find security bugs caused by incorrect `ArrayPool<T>` use even in dotnet/runtime libraries.
+
+We are also going to be continually pushed to expose APIs like those we use internally which make potential misuse even easier, e.g. `ValueStringBuilder`. Thus far we have simply said &quot;we're not comfortable doing that&quot;, but it is unlikely going to be a satisfactory answer long-term. It is something we need to keep thinking about.
```
blowdart, line 172: It's a satisfactory answer for me :) More seriously satisfactory is a value judgement. How about 

_We are continually pushed to expose potentially dangerous APIs we use internally, e.g. `ValueStringBuilder`. Thus far we have simply said &quot;we're not comfortable doing that&quot;, and we will continue to examine the balance of security against utility". If an API can be safe by default it is more likely to be included._

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
+# .NET Runtime Security Mitigations
+
+[Jan Kotas](https://github.com/jkotas)
+
+_From Wikipedia: Mitigation is the reduction of something harmful or the reduction of its harmful effects._
+
+Security is our top concern. The .NET platform combines many layers of security measures that include design and code security reviews, threat modeling, fuzzing of sensitive components, analyzers to detect potential security issue, documentation of security best practices, shipping of regular security updates throughout the support life cycle, bug bounty programs and mandatory security trainings for engineers. This redundancy of methods is known as [defense-in-depth](https://en.wikipedia.org/wiki/Defense_in_depth_(computing)).
```
blowdart, line 7: Nit: _potential security issues_ not issue.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+The security mitigations prevent or reduce probability of each step in this sequence.
+
+The mitigations for memory safety violations must be typically applied to all code running in a process to be fully effective. The exploits often combine memory safety violations and code from multiple modules in the process. For example, memory safety violation in C/C++ module and code generated by .NET runtime JIT can be combined together to produce an exploit. Thus, missing mitigations in .NET modules can allow exploiting vulnerabilities in C/C++ modules that would be hard or impossible to exploit in isolation, and vice versa.
+
+# Executable Space Protection
+
+_Known as Data Execution Prevention (DEP) on Windows_
```
blowdart, line 52: Maybe just change the title to Executable Space Protection / Data Execution Prevention rather than have the italicized sentence?

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+## W^X
+
+[W^X](https://en.wikipedia.org/wiki/W%5EX) is one of the most fundamental mitigations. It blocks the simplest attack path by disallowing memory pages to be writeable and executable at the same time. .NET runtime has been missing this mitigation so far and the lack of this mitigation has (correctly) resulted in us not considering more advanced mitigations. Large number of pages that are both writeable and executable in a typical .NET process is a ripe target for attacks that simply inject new code.
+
+Apple has made the W^X mandatory for future versions of macOS desktop operating system as part of Apple Silicon transition. It motivated us to schedule implementation of this mitigation for .NET 6, on all supported operating systems. Our principle is to treat all supported operating systems equally with respect to security, where possible.
+
+## Protect Intermediate Language (IL) Code
+
+IL code is equivalent of executable code. The IL code might be overwritten with malicious IL. An attacker could then wait for the IL to be compiled or re-compiled via tiering and have malicious code injected that way.
```
blowdart, line 64: What is tiering? Explanation needed?

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+We do not see strong need for introducing out of process JIT for .NET runtime in foreseeable future. The primary reason is that .NET does not support sand-boxed execution of untrusted code anymore. The out-of-proc JIT has advantages for execution of untrusted (verifiable) code. The security benefits of out-of-proc JIT for trusted unverifiable IL code are less interesting since strong soundness validation is not feasible for unverifiable IL code. The secondary reasons include availability of more secure options that disallow runtime code generation altogether, the implementation cost and necessity of a second process to run alongside the primary process.
+
```
blowdart, line 91: is it work noting here that shipping a compiler _at all_ has been a problem for restricted environments? Every now and again I still get requests on how to disable the .net compiler in framework installs. Or rosyln.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+# Control-flow Integrity
+
+Once attackers are no longer able to inject arbitrary code into the process thanks to executable space protections, they start looking for creative ways to leverage existing code in the process to mount the attack. [Return-oriented Programming](https://en.wikipedia.org/wiki/Return-oriented_programming) (ROP) and [Jump-oriented Programming](https://www.csc2.ncsu.edu/faculty/xjiang4/pubs/ASIACCS11.pdf) (JOP) are generalized exploit techniques that take advantage of situations where the program uses indirection for control transfer. The affected indirection is return address for ROP and function pointer for JOP. Control-flow integrity adds validation of these indirections to eliminate or reduce chance of them being used to mount successful attacks.
+
+The control-flow integrity validations are emerging area. They are not yet standardized across operating systems or hardware vendors.
```
blowdart, line 96: NIT: "an" emerging area.

blowdart: Are you saying here we won't support them unless they're standardized? That's what the last sentence insinuates.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Shadow Stack
+
+A second stack of return addresses that is mostly invisible to the program is maintained on the side in addition to the ordinary stack. Return instruction uses the second stack to confirm that the return address on the ordinary stack has not been tampered with. This prevents attacks from leveraging stack buffer overruns to overwrite return addresses and hijack the control flow.
+
+Low-level techniques used by .NET runtime such as return address hijacking for GC thread suspension have to be updated for compatible with shadow stack.
+
+We have been working closely with Windows team on ensuring that the shadow stack support in the operating system can work well together with .NET runtime. The result of this collaboration are new Windows APIs e.g. `RtlGetReturnAddressHijackTarget` that is in private preview currently. The new APIs should help other language runtimes to enable this mitigation. Our plan is complete support for shadow stack mitigation on Windows in .NET 6.
```
blowdart, line 108: Can we talk about it publicly if it's in preview? There's no windows documentation on google searches for this, so if this document is meant to be public the API reference may need to go.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+# Language, Libraries and Runtime Features
+
+Library and language design have always played a key role in mitigating security bugs by preventing or discouraging them from being introduced in the first place.
+
+## Unsafe API and C# language constructs
+
+Number of newer APIs such as `Memory<T>` and friends come with security risks that are not clearly communicated.
+
+We need to adopt a more holistic approach to unsafe features and APIs that we expose for public use. Developers need feedback in the IDE so that they understand that they are using constructs that are unsafe and are [informed of the specific risks that come with using those APIs](https://github.com/dotnet/runtime/issues/41418). Over time, we bias more to constructs and paradigms that are safe by default (possibly inspired by Rust).
```
blowdart, line 164: Opinion: An IDE warning is not enough. Not everyone will use Visual Studio. There needs to be non-IDE feedback, even if it's just compile time warnings.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+We are also going to be continually pushed to expose APIs like those we use internally which make potential misuse even easier, e.g. `ValueStringBuilder`. Thus far we have simply said &quot;we're not comfortable doing that&quot;, but it is unlikely going to be a satisfactory answer long-term. It is something we need to keep thinking about.
+
+## Secrets in Presence of GC Heap Compaction
+
+GC heap compaction can create a second copy of secrets stored in managed memory. It makes it impossible for an application to reliably clear in-memory secrets stored in managed memory after they are no longer used.
+
+We will introduce GC feature that clears unused memory after GC heap compaction. Having the GC clear memory after compaction improves the security of applications for which heap dumps and in-memory probes are within their scope of threat. [dotnet/runtime#10480](https://github.com/dotnet/runtime/issues/10480) has more details.
```
blowdart, line 178: Nit: "We will introduce _a_ GC feature"

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+  - [Dangerous Deserialization](#dangerous-deserialization)
+- [Looking forward](#looking-forward)
+  - [Plan for .NET 6](#plan-for-net-6)
+- [Appendix](#appendix)
+  - [Timing Side Channel Attacks](#timing-side-channel-attacks)
+  - [Confidential Computing](#confidential-computing)
+
+# Anatomy of an exploit
+
+Memory security vulnerability starts with a memory safety violation. That includes bugs such as writing outside of an array's bounds or using a pointer after freeing the associated object. When an attacker can influence the pointer or data involved, they have at least some control of the program's memory (this is referred to as a write-what-where primitive). Once an attacker has a write-what-where primitive, they may be able to use that to change the program's execution to do something malicious.
```
danmoseley, line 44: I guess these violations can be caused by either our own bug (in the runtime, shims or unsafe code) or in customer app code (in native libraries or unsafe code). Is it worth comparing these in terms of (a) frequency/impact (b) susceptibility to mitigations below?
Some of these will apply to the runtime, but will not affect customer-created vulnerabilities.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Protect Intermediate Language (IL) Code
+
+IL code is equivalent of executable code. The IL code might be overwritten with malicious IL. An attacker could then wait for the IL to be compiled or re-compiled via tiering and have malicious code injected that way.
+
+.NET runtime should treat the IL code same way as directly executable code, ie make IL read-only or otherwise protected wherever possible.
+
+## Address Space Layout Randomization (ASLR)
+
+[Address space layout randomization](https://en.wikipedia.org/wiki/Address_space_layout_randomization) arranges all code and data structures in memory randomly that makes it harder for the attacker to reliably transfer control to particular exploited function or modify particular data structure. .NET runtime has been compatible with ASLR since .NET Framework 3.5 SP1 (released in 2008).
```
danmoseley, line 70: I assume if the customer app loads native libraries that aren't ASLR aware, the OS loader still supports ASLR for our runtime.

Do we verify that all our native code (beyond coreclr/jit/gc) opt into ASLR? I guess this falls into the category of outer ring checks we do.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+Also, operating systems often offer opt-in or opt-out mechanisms to disallow runtime code generation, for example [Arbitrary Code Guard](https://docs.microsoft.com/en-us/windows/security/threat-protection/microsoft-defender-atp/customize-exploit-protection) (ACG) on Windows, [Allow Execution of JIT-compiled Code Entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_allow-jit) on macOS.
+
+Mono and .NET Native for UWP have been shipping .NET runtimes that abide by the runtime code generation restriction. [Native AOT Form Factor](https://github.com/dotnet/runtimelab/tree/feature/NativeAOT) experiment explores this space further.
+
+IL code interpreters circumvent the environment restriction on runtime code generation and allow .NET libraries that depend on runtime code generation to work, with lower performance. From a security point of view, IL code interpreters are not desirable in environments with disallowed code generation since IL code is equivalent of executable code as described above.
```
danmoseley, line 84: Could an interpreter do slightly better than ref emit, eg., by refusing to interpret 'unsafe' IL?

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Stack Buffer Overflow Protection
+
+[Stack Buffer Overflow](https://en.wikipedia.org/wiki/Stack_buffer_overflow) protection is among the older security mitigations. It has been known as &quot;GS cookie&quot; on Windows. It is designed to catch writes into stack memory outside the intended bounds.
+
+The first .NET runtime that included it was .NET Framework 2.0, and the implementation has not changed much since then. The mitigation in C/C++ compilers have seen improvements in the meantime (for example, [Enhanced GS](https://msrc-blog.microsoft.com/2009/03/20/enhanced-gs-in-visual-studio-2010/) or -[fstack-protector-strong](https://lwn.net/Articles/584225/)).
```
GrabYourPitchforks, line 152: Might be worth pointing out that we put stack cookies into some managed routines as well. (Currently, any _localloc_ IL instruction causes the cookie to be emitted into the current frame.) It's not strictly a C/C++ mitigation.

accepted/2021/runtime-security-mitigations.md
danmoseley: This is a public repo.

accepted/2021/runtime-security-mitigations.md
GrabYourPitchforks: With `ValueStringBuilder` specifically, I think we said we're comfortable shipping it once we get good language guardrails for its usage. C# doesn't have a strong ownership model, but we could reconsider that. We could also introduce copy-only structs (see https://gist.github.com/GrabYourPitchforks/7ab6a440100467a82cfe5998cd1e91be for a basic proposal), or we could write a Roslyn analyzer to govern `ValueStringBuilder` usage.

So I wouldn't say we're _opposed_ to doing it, just that we're waiting for some changes in the ecosystem before we introduce it publicly.

GrabYourPitchforks:
Thanks Jan for getting this posted!

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+We will review and update stack buffer overflow protection in .NET runtime to match state of the art.
+
+# Language, Libraries and Runtime Features
+
+Library and language design have always played a key role in mitigating security bugs by preventing or discouraging them from being introduced in the first place.
```
danmoseley, line 158: I don't see mention of new analyzers. Do we have any potential analyzers in mind to help spot security issues, or existing analyzers to evangelize?
cc @jeffhandley

accepted/2021/runtime-security-mitigations.md
danmoseley: Even if it's not secure by default, if we expose an API that's potentially dangerous, we should aim to simultaneously offer an analyzer that makes it significantly harder to use insecurely.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+Number of newer APIs such as `Memory<T>` and friends come with security risks that are not clearly communicated.
+
+We need to adopt a more holistic approach to unsafe features and APIs that we expose for public use. Developers need feedback in the IDE so that they understand that they are using constructs that are unsafe and are [informed of the specific risks that come with using those APIs](https://github.com/dotnet/runtime/issues/41418). Over time, we bias more to constructs and paradigms that are safe by default (possibly inspired by Rust).
+
+## Object and Array Pooling
+
+`ArrayPool<T>` and other similar custom object pools are equivalent of the classic C malloc. Manual lifetime management traditionally suffers from use-after-free and double-free security vulnerabilities. Many years were spent on implementing diagnostics and mitigations of these vulnerabilities for malloc.
+
+We need to implement similar diagnostic and mitigation capabilities for ArrayPool\&lt;T\&gt; and friends. It is [not unusual](https://github.com/dotnet/corefx/pull/37270) to find security bugs caused by incorrect `ArrayPool<T>` use even in dotnet/runtime libraries.
```
danmoseley, line 170: https://github.com/dotnet/runtime/issues/7532 has some discussion.

accepted/2021/runtime-security-mitigations.md
danmoseley: Just curious, did we ever estimate what sort of perf hit this would imply? Would it need to be opt-in?

accepted/2021/runtime-security-mitigations.md
GrabYourPitchforks: Not sure how useful this would be in practice. There are two things that can make code "unsafe". The first - as you point out - is using unverifiable IL patterns. The second is calling methods marked `[SecurityCritical]` or otherwise CAS-annotated. Since the .NET runtime no longer has a concept of CAS, there's no way for the runtime itself to know that a method is unsafe-equivalent.

For example, assume you make a call to `Marshal.WriteByte(IntPtr, byte)`, which allows you to stomp on any address in memory. You don't need unverifiable code to make such a call. In .NET Framework, the runtime would complain if transparent (not explicitly trusted) code calls this method, but that's because the method itself is marked `[SecurityCritical]`. In .NET \[Core+\], these annotations are gone, so it looks like any other "safe" method.

accepted/2021/runtime-security-mitigations.md
danmoseley: We certainly have responsibilities to help make our ecosystem more secure wrt serializers - that could be achieved in many ways. But it's partly "our job" IMO

accepted/2021/runtime-security-mitigations.md
jkotas: These APIs are shipping in publicly available Windows preview channels. I will drop the API name from this doc since the API name alone without documentation is useless.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+Full memory encryption protects all memory transparently. AMD Secure Memory Encryption (SME) and Intel Total Memory Encryption (TME) are existing or planned implementations of this technology. We can expect it to become the baseline of all premium cloud offerings within the next decade. While full memory encryption does not require any fundamental changes of the software stack at .NET runtime level, it may require updates of designs and best practices for diagnostics and telemetry to avoid unintentional egress of confidential data.
+
+Enclaves protect a portion of memory. It is implemented by Intel Software Guard Extensions (SGX). To fit into Enclaves programming model, the software stack typically requires refactoring of sensitive computation into isolated component that runs inside the Enclave and [uses specialized APIs for communication](https://openenclave.io/sdk/). The current enclave capabilities are too limitted to fully suport a general purpose .NET runtime. .NET runtime for enclaves would be a niche [form factor](https://github.com/dotnet/designs/blob/main/accepted/2020/form-factors.md) with number of limitations.
```
danmoseley, line 213:
```suggestion
Enclaves protect a portion of memory. It is implemented by Intel Software Guard Extensions (SGX). To fit into Enclaves programming model, the software stack typically requires refactoring of sensitive computation into isolated component that runs inside the Enclave and [uses specialized APIs for communication](https://openenclave.io/sdk/). The current enclave capabilities are too limited to fully suport a general purpose .NET runtime. .NET runtime for enclaves would be a niche [form factor](https://github.com/dotnet/designs/blob/main/accepted/2020/form-factors.md) with number of limitations.
```

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+We do not see strong need for introducing out of process JIT for .NET runtime in foreseeable future. The primary reason is that .NET does not support sand-boxed execution of untrusted code anymore. The out-of-proc JIT has advantages for execution of untrusted (verifiable) code. The security benefits of out-of-proc JIT for trusted unverifiable IL code are less interesting since strong soundness validation is not feasible for unverifiable IL code. The secondary reasons include availability of more secure options that disallow runtime code generation altogether, the implementation cost and necessity of a second process to run alongside the primary process.
```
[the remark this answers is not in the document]
danmoseley: Could we imagine ever offering API to make it easier to run your own code, or a sub application, in an _OS enforced_ sandbox (eg., a low integrity process)? Is that something our ecosystem would use?

accepted/2021/runtime-security-mitigations.md
yaakov-h, line 170:
```suggestion
We need to implement similar diagnostic and mitigation capabilities for `ArrayPool<T>` and friends. It is [not unusual](https://github.com/dotnet/corefx/pull/37270) to find security bugs caused by incorrect `ArrayPool<T>` use even in dotnet/runtime libraries.
```

accepted/2021/runtime-security-mitigations.md
GrabYourPitchforks: I think I heard a 5% perf hit \[to GC\] being thrown around somewhere? Every discussion I've had on this feature assumes customers would opt-in to it.

accepted/2021/runtime-security-mitigations.md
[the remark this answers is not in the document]
GrabYourPitchforks: Unlikely that such an API would be generalizable across operating systems. The particular behavior you'd want is very scenario-specific. For example: should the sandbox be quota-restricted? Should it have access to network or filesystem resources, and if so should it have full access or restricted access?

I think it'd be possible to create a package to allow for this. But the API surface would look different across each OS, and we'd really want that package to be maintained by the appropriate OS team, _not_ the libraries team.

accepted/2021/runtime-security-mitigations.md
GrabYourPitchforks: Should we hyperlink "recompiled via tiering" to https://github.com/dotnet/runtime/blob/aa1eeb3dff7ae8b670e33e66a935d56591c2638d/docs/design/features/tiered-compilation.md?

accepted/2021/runtime-security-mitigations.md
davidwrighton: I believe this refers more to the concept that it isn't feasible to achieve mitigations with the matching security characteristics across operating systems and hardware vendors as the capabilities are not standardized enough. In particular, strong mitigations in this space typically require involvement with the OS in ways that are fundamentally different for each OS, and the hardware mitigations in the space are only available on a subset of hardware and also have tight OS integration that differs from OS to OS making finding a least common denominator api concept is difficult.

accepted/2021/runtime-security-mitigations.md
davidwrighton: We may wish to increase the amount of cases where we add protection. In particular due to the presence of unsafe arithmetic on ref parameters it has become much easier to write code that have problems.

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+Apple has made the W^X mandatory for future versions of macOS desktop operating system as part of Apple Silicon transition. It motivated us to schedule implementation of this mitigation for .NET 6, on all supported operating systems. Our principle is to treat all supported operating systems equally with respect to security, where possible.
+
+## Protect Intermediate Language (IL) Code
+
+IL code is equivalent of executable code. The IL code might be overwritten with malicious IL. An attacker could then wait for the IL to be compiled or re-compiled via tiering and have malicious code injected that way.
+
+.NET runtime should treat the IL code same way as directly executable code, ie make IL read-only or otherwise protected wherever possible.
```
stephentoub, line 66: Also seems like an area where we should avoid applying certain kinds of optimizations.  For example, DynamicMethod fits squarely in the target area of writeable managed memory for the purposes of generating code to be executed.  Today, it doesn't use ArrayPool, but there have been suggestions in the past that it do so for some of the potentially large arrays it creates.  As part of reducing avenues of attack, this is a case where we should probably avoid using pooled arrays (we probably wouldn't have done so anyway due to poor tradeoffs, but wanted to throw it out as an example).

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Intel Control-flow Enforcement Technology (CET)
+
+[CET](https://software.intel.com/content/www/us/en/develop/articles/technical-look-control-flow-enforcement-technology.html) is set of hardware features that enable efficient implementation of integrity checks for control flow indirections. CET first appeared in Tiger Lake processors launched in September 2020.
+
+### Shadow Stack
+
+A second stack of return addresses that is mostly invisible to the program is maintained on the side in addition to the ordinary stack. Return instruction uses the second stack to confirm that the return address on the ordinary stack has not been tampered with. This prevents attacks from leveraging stack buffer overruns to overwrite return addresses and hijack the control flow.
+
+Low-level techniques used by .NET runtime such as return address hijacking for GC thread suspension have to be updated for compatible with shadow stack.
```
stephentoub, line 106:
```suggestion
Low-level techniques used by .NET runtime such as return address hijacking for GC thread suspension have to be updated for compatibility with shadow stack.
```

accepted/2021/runtime-security-mitigations.md
stephentoub: I expect in .NET 6 we're going to ship something like it in support of interpolated strings, but it'll be buried in System.Runtime.CompilerServices; while someone could use it directly, it'll fit in with the rest of System.Runtime.CompilerServices and System.Runtime.InteropServices as "if you touch this stuff, it's unsafe and you need to know what you're doing".

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+`ArrayPool<T>` and other similar custom object pools are equivalent of the classic C malloc. Manual lifetime management traditionally suffers from use-after-free and double-free security vulnerabilities. Many years were spent on implementing diagnostics and mitigations of these vulnerabilities for malloc.
+
+We need to implement similar diagnostic and mitigation capabilities for `ArrayPool<T>` and friends. It is [not unusual](https://github.com/dotnet/corefx/pull/37270) to find security bugs caused by incorrect `ArrayPool<T>` use even in dotnet/runtime libraries.
+
+We are also going to be continually pushed to expose APIs like those we use internally which make potential misuse even easier, e.g. `ValueStringBuilder`. Thus far we have simply said &quot;we're not comfortable doing that&quot;, but it is unlikely going to be a satisfactory answer long-term. It is something we need to keep thinking about.
+
+## Secrets in Presence of GC Heap Compaction
+
+GC heap compaction can create a second copy of secrets stored in managed memory. It makes it impossible for an application to reliably clear in-memory secrets stored in managed memory after they are no longer used.
```
stephentoub, line 176: This is certainly true in the general case.  We and others do have code, though, that pins managed memory prior to storing secrets in it, and zeros it out prior to unpinning.  The pinned-object heap has made handling such things easier, too.

accepted/2021/runtime-security-mitigations.md
jkotas: > the customer app loads native libraries that aren't ASLR aware, the OS loader still supports ASLR for our runtime.

Yes, it does. Note that the non-ASLR enabled library can be used as trampoline to attack anything else in the process, including our runtime. It is really important for these mitigations to be enabled for all code in the process as discussed in the "Anatomy of an exploit" chapter.

> Do we verify that all our native code (beyond coreclr/jit/gc) opt into ASLR?

Yes, it is part of the binskim gate. It actually found some Linux binaries that were not opted in correctly when it was enabled for the first time.

danmoseley: Naïve question, is there anything we can do to help customers realize when they load native code into a .NET process that isn't ASLR enabled? Or if not technically at least educate through documentation.

accepted/2021/runtime-security-mitigations.md
marek-safar: Analyzers are mentioned few times in the comments. It might be worth calling out how we are going to enforce them to reduce the security risks.

accepted/2021/runtime-security-mitigations.md
jkotas: >  move-only structs

Do we have one central issue to point to that describes the problem and discusses potential designs?

accepted/2021/runtime-security-mitigations.md
jkotas: > I guess these violations can be caused by either our own bug or in customer app code

Or by a bug in OS or 3rd party library that your app happens to use, as pointed out later in the doc.

>  Is it worth comparing these in terms of (a) frequency/impact (b) susceptibility to mitigations below?

It may be a good topic for a research paper. Research papers doing this sort of analysis for specific domain are fairly common. They typically find that memory safety issues are very common source of security bugs (ie like half of all security bugs).

accepted/2021/runtime-security-mitigations.md
```diff
@@ -0,0 +1,213 @@
...
+
+We should be mindful of this mitigation and avoid designs (performance optimizations in particular) that invalidate it.
+
+## Disallow Runtime Code Generation
+
+Disallowing runtime code generation altogether is the ultimate form of executable space protection. It is often combined with verification of signatures of all code being executed.
+
+The runtime code generation has been unconditionally disallowed on most game consoles and Apple devices, to protect business models and consumer experiences. We expect that the set of environments that unconditionally disallow runtime code generation is going to grow over time. It will include cloud infrastructure and the most critical cloud services where the restriction will be enforced across the system via Hypervisor-Protected Code Integrity (HCVI) and related technologies.
```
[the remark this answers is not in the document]
jkotas: Cloud confidential compute is protecting against one class of risks. Disallowing runtime code generation is protecting against different class of risks. I see them as independent. Many customers will choose one and not the other, especially once cloud confidential compute becomes commodity.

Of course, the most critical infrastructure will aim for maximum protection, by requiring or opting-into all security measure available.

accepted/2021/runtime-security-mitigations.md
jkotas: Opened https://github.com/dotnet/runtime/issues/50389

accepted/2021/runtime-security-mitigations.md
jkotas: > Do we have any potential analyzers in mind to help spot security issues, or existing analyzers to evangelize?

I have marked a few to make them show up in this query: https://github.com/dotnet/runtime/issues?q=is%3Aopen+label%3Acode-analyzer+label%3Asecurity . I am sure we can come up with more ideas in this category.

PR #191, opened 2021-03-21
Specify that TFV includes the 'v' prefix on it's value, and also fix comparisons to use msbuild functions instead.

AraHaan:
The documentation file that I looked at was incorrect at this and when I tried to do the comparison at the bottom I noticed there was errors with it and then looked internally in the SDK and found out it actually prefixes it with an ``v`` and the one I needed was the one without the ``v`` prefix.

As such I felt this document needed some changes and also to note for future viewers what to use if they do not want the prefix from TFV.

### DIFF
--- a/accepted/2020/net5/net5.md
+++ b/accepted/2020/net5/net5.md
@@ -1042,7 +1042,7 @@ And to check for things before a specific version do a comparison like this:
 
 If we had used ``TargetFrameworkVersion`` instead of ``_TargetFrameworkVersionWithoutV`` we would see this error for both conditions:
 
-````error MSB4086: A numeric comparison was attempted on "$(TargetFrameworkVersion)" that evaluates to "v3.0" instead of a number, in condition "'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' >= '3.0'".````
+``error MSB4086: A numeric comparison was attempted on "$(TargetFrameworkVersion)" that evaluates to "v3.0" instead of a number, in condition "'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' >= '3.0'".``
 ``error MSB4086: A numeric comparison was attempted on "$(TargetFrameworkVersion)" that evaluates to "v3.0" instead of a number, in condition "'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' &lt; '3.0'".``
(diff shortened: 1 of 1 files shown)


accepted/2020/net5/net5.md
```diff
@@ -1020,13 +1027,24 @@ In MSBuild you can't easily do comparisons like:
 <ItemGroup Condition="'$(TargetFramework)' >= 'net5.0'`">
 ```
 
-because that would be a string comparison. Rather, you need to do comparisons
+because that would be a string comparison. Rather, you need to do a comparison
 like this:
 
 ```xml
-<ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(TargetFrameworkVersion)' >= '3.0'">
+<ItemGroup Condition="'$(TargetFrameworkIdentifier)' == '.NETCoreApp' AND '$(_TargetFrameworkVersionWithoutV)' >= '3.0'">
```
jkotas, line 1034: The proper way to do this is using `VersionLessThan` msbuild function, like: `$([MSBuild]::VersionLessThan($(TargetFrameworkVersion), '3.0'))`

accepted/2020/net5/net5.md
```diff
@@ -493,6 +494,12 @@ Specifically:
   the customer can override this in the project file to a lower version (using a
   higher version than `TargetPlatformVersion` should generate an error).
 
+ * **_TargetFrameworkVersionWithoutV is normally set only internally in the Microsft.NET.Sdk SDK**.
```
jkotas, line 497: We do not want to be recomending or encouraging use of internal msbuild properties. There is a better way to do it without them.

jkotas:
> The documentation file that I looked at was incorrect at this 

Can you please submit PR to fix the documentation? The specs in this repo are not meant to be a replacement for documentation.

AraHaan:
Oh I was using the file I PR'd here for documentations on this, but I could look at see if normal documentations for it needs updating as well.

AraHaan:
Should I open a pull request for the .NET SDK's that use ``_TargetFrameworkVersionWithoutV`` to use the MSBuild functions as well too in case a future version of msbuild decides to nuke the internal property version?

jkotas:
This is being discussed in https://github.com/dotnet/sdk/issues/2618#issuecomment-802024656 . You may want to ask on that issue first whether PR to clean it up would be welcomed or start with a PR that cleans it up in just a few files to validate the approach.

AraHaan:
It looks like some people call the msbuild functions with the 2nd arg either wrapped inside of ``'``'s or even without them.

PR #192, opened 2021-03-22
Preview Features

terrajobst:


accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+```C#
+[assembly: RequiresPreviewFeatures]
+```
+
+This means you can look at a binary and tell whether it was built using preview
+features. This can be used for auditing and for detecting transitive
+dependencies ([discussed later](#api-analyzer)).
+
+### runtime.json
```
AaronRobinsonMSFT, line 228: Is this supposed to be `runtimeconfig.json`?

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+Whether or not `EnablePreviewFeatures` was set to true for the application
+should be written into the `runtime.json` file such that the host/runtime can
+use it to make decisions. Alternatively, we could say this is encoded by the
+fact that the assembly the host considers the entry point has the
+`RequiresPreviewFeatures` assembly level attribute.
+
+### Compilation context
+
+When `EnablePreviewFeatures` is true, the `LangVersion` property should be set
```
AaronRobinsonMSFT, line 238: Does this mean that we can forever assume that if `EnablePreviewFeatures` is `true` then `LangVersion` was also `true`? Should everyone be able to infer that or is that an implementation detail at present?

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+## Stakeholders and Reviewers
+
+* Runtime teams
+* C#, VB, F# compiler and IDE teams
+* Library teams
+* Project system team
+* SDK team
+* MSBuild team
+* NuGet team
```
jkotas, line 182: Should Powershell be also in this list? They may need to do work to prevent powershell scripts taking accidental depedency on unstable preview surface.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
+# Preview Features
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+Starting with .NET 5, we've moved to fixed release schedule whereby we ship one
```
stephentoub, line 5:
```suggestion
Starting with .NET 5, we've moved to a fixed release schedule whereby we ship one
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+## Scenarios and User Experience
+
+### Enabling a preview runtime feature
+
+Maddy is building a Tye application. She reads in a blog post that the .NET
+runtime team is experimenting with a new GC mode, code named FancyGC.
+
+She wants to see how this improves the performance of her backend. She's glad to
+see that the new GC mode was shipped as part of .NET 6, so she doesn't need to
```
stephentoub, line 65: In this example, the new GC mode is the default, such that there's no additional XML like `<ServerGarbageCollection>true</ServerGarbageCollection>`?

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+figure out how to run a new version of .NET in her CI lab. So she creates a
+feature branch and enables the new GC mode simply by changing the service's
+project file as follows:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
+    <EnablePreviewFeatures>True</EnablePreviewFeatures>
```
jkotas, line 75: Where does this example says that it enables `FancyGC`?

The preview runtime features that are just performance optimizations should be on a different plan, and it should be possible to enable them indendently.

stephentoub, line 75: Why is this separate from langversion?  Why can't we just use the existing `<LangVersion>preview</LangVersion>` to mean "I'm opting in to preview stuff"?  I don't think many folks realistically differentiate between language, library, and runtime, nor should they need to?  From my perspective, we already have a mechanism for this... we should just use it.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+should be written into the `runtime.json` file such that the host/runtime can
+use it to make decisions. Alternatively, we could say this is encoded by the
+fact that the assembly the host considers the entry point has the
+`RequiresPreviewFeatures` assembly level attribute.
+
+### Compilation context
+
+When `EnablePreviewFeatures` is true, the `LangVersion` property should be set
+to `Preview`. This avoids customers having to turn on preview features for the
+language separately.
```
stephentoub, line 240: I'm still not understanding why the new EnablePreviewFeatures is necessary at all / why we can't just use LangVersion=preview.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+                    AttributeTargets.Struct |
+                    AttributeTargets.Enum |
+                    AttributeTargets.Constructor |
+                    AttributeTargets.Method |
+                    AttributeTargets.Property |
+                    AttributeTargets.Field |
+                    AttributeTargets.Event, AllowMultiple = true, Inherited = false)]
+    public sealed class RequiresPreviewFeaturesAttribute : Attribute
+    {
+        public PreviewAttribute();
```
stephentoub, line 269:
```suggestion
        public RequiresPreviewFeaturesAttribute();
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+}
+```
+
+We'll ship a built-in analyzer that will report diagnostics when an API is being
+used that is marked `[RequiresPreviewFeatures]` but the consumer (member, type,
+module, assembly) isn't marked. The default severity is `error`.
+
+This analyzer will also validate that if any referenced assembly has applied
+`[RequiresPreviewFeatures]`, that the consuming assembly must also have this
+attribute applied.
```
stephentoub, line 280: Why an analyzer rather than the compiler itself?

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+### Goals
+
+* Customers can discover preview features
+* Customers are made aware if a particular feature is in a preview state so that
+  they always know which part of the product is fully supported and which parts
+  aren't.
+* Customer have an easy way to audit if their applications or libraries are
+  using preview features.
+* Customers can't accidentally depend on preview features, either directly or
```
jkotas, line 156: Does this cover also depencies taken via reflection? It may be worth having a paragraph dedicated to reflection.

Consider:
```
void f(dynamic x) => x.ExperimentalMethod();
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+behavioral changes and also allows certain parts of .NET to make deliberate
+breaking changes if necessary.
+
+As a result, this makes it viable to innovate across runtime, libraries, and
+languages, allowing us to build more compelling features that would have been
+almost impossible to do on .NET Framework. Examples include static linking,
+default interface members, and ref structs (such as `Span<T>`).
+
+However, it's sometimes challenging to deliver cross-cutting feature work in a
+single release, especially when they are breaking new ground in terms of
```
jonsequitur, line 28:
```suggestion
single release, especially when it's breaking new ground in terms of
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+see that the new GC mode was shipped as part of .NET 6, so she doesn't need to
+figure out how to run a new version of .NET in her CI lab. So she creates a
+feature branch and enables the new GC mode simply by changing the service's
+project file as follows:
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
marek-safar, line 74: What will happen when different or multiple TFMs are used?

accepted/2021/preview-features/preview-features.md
YairHalberstadt: Additionally this current design makes it possible to `EnablePreviewFeatures` whilst explicitly setting `LangVersion` to e.g. 7. I think it would be somewhat confusing if that left you able to use C# 9 features.

Using just `LangVersion` would alleviate this issue.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
+# Preview Features
+
+**PM** [Immo Landwerth](https://github.com/terrajobst)
+
+Starting with .NET 5, we've moved to fixed release schedule whereby we ship one
+version of .NET once a year, at RTM quality, with every other year being a long
+term support (LTS) release.
+
+This is in stark contrast to the .NET Framework days where we had multiple years
```
danmoseley, line 9:
```suggestion
This is in contrast to the .NET Framework days where we had multiple years
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+for a given release, typically two to three years.
+
+Shipping more frequently has many advantages, such as being able to react
+quicker to trends and having an organization and customer base that is used to
+shipping & absorbing technology quicker to make the most of today's fast moving
+world.
```
danmoseley, line 15:
```suggestion
shipping & absorbing technology more quickly.
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+breaking changes if necessary.
+
+As a result, this makes it viable to innovate across runtime, libraries, and
+languages, allowing us to build more compelling features that would have been
+almost impossible to do on .NET Framework. Examples include static linking,
+default interface members, and ref structs (such as `Span<T>`).
+
+However, it's sometimes challenging to deliver cross-cutting feature work in a
+single release, especially when they are breaking new ground in terms of
+expressiveness. A good example of such a feature is the upcoming improved
```
danmoseley, line 29: is expressiveness the right word? Or, could you explain what you mean by it?

accepted/2021/preview-features/preview-features.md
danmoseley: It would also be easier to communicate.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+During lunch, Adriana discusses a design problem with one of the architects who
+mentions that the upcoming static interface members in C# might be able to solve
+this issue. Adriana does a quick web search and finds a blog post from the C#
+team that showcases how this feature might look like. She wonders how she can
+play with that. Optimistic, she copy & pastes the sample code into her project
+which immediately generates a compilation error:
+
+> The feature 'static interface members' is currently in Preview and
+> "unsupported". To use Preview features, use the 'preview' language version.
```
danmoseley, line 94: Is this sufficient warning that Adriana is entering an unsupported configuration? Maybe add something like "To return to a supported configuration, change back to a supported language version."

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+help to speed things up considerably. They decide to give it a spin, so they
+copy & paste some sample code from the blog post.
+
+The code immediately produces an error message:
+
+> error: The type 'IArithmetic\<T>' is in preview. In order to use it, you need
+> to enable preview features.
+
+After taking a quick look at the blog post, Ainsley realizes that they only
+needs to set `EnablePreviewFeatures` to `true` in the project file.
```
jnm2, line 130:
```suggestion
need to set `EnablePreviewFeatures` to `true` in the project file.
```

accepted/2021/preview-features/preview-features.md
danmoseley: BTW, this is mixing up preview language version and 'enable preview features'.

danmoseley: Or, maybe `<EnablePreviewFeatures>` is the master switch, that includes `<LangVersion>`

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+  they always know which part of the product is fully supported and which parts
+  aren't.
+* Customer have an easy way to audit if their applications or libraries are
+  using preview features.
+* Customers can't accidentally depend on preview features, either directly or
+  indirectly through some library.
+* We can deliver a reasonable experience even when a preview feature can't be
+  supported in all the workloads, for example, by failing with a sensible error
+  message.
+* We have a unified experience for enabling preview features, no matter whether
```
danmoseley, line 161: Does a unified experience mean a single switch?

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+  they are in a runtime, a compiler, a library, or span multiple parts of the
+  system.
+* We don't have to do unnatural acts to expose a preview feature, by, for
+  example, having to build a plug-in based system that allows swapping in and
+  out files.
+
+### Non-Goals
+
+* Allowing Microsoft to turn features on or off remotely, e.g. A/B testing
```
danmoseley, line 170: What about telemetry (dotnet build and from VS)

accepted/2021/preview-features/preview-features.md
danmoseley: Or make one a synonym for the other, with identical effect, and deprecate LangVersion (ie., have VS set it instead).

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+We'll ship a built-in analyzer that will report diagnostics when an API is being
+used that is marked `[RequiresPreviewFeatures]` but the consumer (member, type,
+module, assembly) isn't marked. The default severity is `error`.
+
+This analyzer will also validate that if any referenced assembly has applied
+`[RequiresPreviewFeatures]`, that the consuming assembly must also have this
+attribute applied.
+
+User code will typically only have this attribute applied at the assembly-level.
```
danmoseley, line 282: What will the compiler do when the attribute is not applied at the assembly level?
Will it require that assembly is built with `<EnablePreviewFeatures>True</EnablePreviewFeatures>` ? 

If assembly A has a method M marked with `[RequiresPreviewFeatures]`, but no assembly level attribute, and it references assembly B that has the assembly level attribute, will the compiler warn?
If assembly A has `[RequiresPreviewFeatures]` assembly level attribute, and it references assembly B that has no assembly level attribute, but does have the attribute on a method M, will the compiler/analyzer flag it? what about if A is not using M?

etc

danmoseley: I assume the answer is, the compiler errors when there is a call to another assembly that has the attribute on the called member, its type, or the assembly as  whole, and the calling method does not have the attribute on itself, its type, or the calling assembly as a whole.

I'm not sure about hte first question though.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+### Are preview features supported?
+
+No. The entire point of preview feature is to gather feedback, and make changes
+based on it. This means we can't support them:
+
+* We will generally not service previews even if they are included in LTS
+  releases. The compiler has done this successfully already. If we can't hold
+  this line, then this effort has failed.
+* This also means that people using preview feature in an LTS, they may no
```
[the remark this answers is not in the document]
danmoseley: Do you think that would happen often? I imagine that most preview features wouldn't be particularly often broken by tooling changes?

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+capability to the compiler itself.
+
+## Q & A
+
+### Are preview features supported?
+
+No. The entire point of preview feature is to gather feedback, and make changes
+based on it. This means we can't support them:
+
+* We will generally not service previews even if they are included in LTS
```
danmoseley, line 296:
```suggestion
* We will generally not service preview features even if they are included in LTS
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+  releases. The compiler has done this successfully already. If we can't hold
+  this line, then this effort has failed.
```
danmoseley, line 298: How much trouble would we go to, if any, to avoid breaking preview features when we service supported releases?

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+* This also means that people using preview feature in an LTS, they may no
+  longer be able to use the latest tooling if the later version of the tooling
+  made a breaking change for a preview feature. We consider this acceptable.
+* We don't believe previews will be around for that long, because the idea is
```
danmoseley, line 302:
```suggestion
* We don't believe preview features will be around for that long, because the idea is
```

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+* The quality bar and connotation of preview features is virtually identical to
+  preview builds of the platform. The only different is that an RTM build
+  additionally contains features that are still in the preview state.
+* The compiler has a language version `preview` that has the same connotation
+  that we use here.
+* We generally used the term *experimental* where probability of shipping is
+  <50%, e.g. in `runtimelabs`
+* Overall, the goal is draw more people in and motivate them to try the
+  features. The term *preview* seems more helpful than *experimental*.
```
danmoseley, line 418: What about preview SDK features, is that out of scope? Eg a dotnet.exe feature. It presumably could not be enabled at project level, in general.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+In this design the system has no way to prevent this and it will in fact just
+work.
+
+### Why did we use the term *preview*?
+
+Originally I used the term *experimental* but we decided to change it to
+*preview* for these reasons:
+
+* The quality bar and connotation of preview features is virtually identical to
+  preview builds of the platform. The only different is that an RTM build
```
danmoseley, line 411:
```suggestion
  preview builds of the platform. The only different is that an GA build
```
let's use GA consistently

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+### What happens when a preview feature is released?
+
+Several things will happen:
+
+1. We remove the `[RequiresPreviewFeatures]` attribution from any APIs
+2. We remove any guards from the runtime/compilers that check for the preview
+   flag (of course, compilers are likely still checking language versions, but
+   that's separate)
+
+If someone shipped a binary to nuget.org that depends on a preview feature,
```
danmoseley, line 396: Should it be an error to attempt to package a 'stable' NuGet package containing assemblies with the preview attribute?

accepted/2021/preview-features/preview-features.md
YairHalberstadt: LangVersion is needed as it can be any version number, or preview, whereas EnablePreviewFeatures is an on off switch.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+   preview, such as static interface members, which includes runtime work,
+   language work, and compiler work. Said differently, having a single preview
+   mode forces us to jointly coordinate preview features across all layers.
+
+### Why are we recording whether an assembly used preview features?
+
+For two reasons:
+
+1. It can be used for auditing
+2. It is used to enforce transitivity. An assembly that doesn't use preview
```
danmoseley, line 381: > we only have to look at the assembly level attribute.

I don't follow this -- we allow the attribute to be on members, without being on the assembly. It seems to me that the compiler needs to recognize this attribute throughout compilation.

accepted/2021/preview-features/preview-features.md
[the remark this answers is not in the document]
jkotas: > I imagine that most preview features wouldn't be particularly often broken by tooling changes?

Tooling includes C# compiler and C# language services. You open a file after VS update, there are red squiqlies everywhere and the editor is unusable (this happened to me multiple times in the past when experimenting with preview features).

accepted/2021/preview-features/preview-features.md
danmoseley: Good point. But should LangVersion=preview imply all preview features are enabled? (Possibly with EnablePreviewFeatures as another option for doing that)?

Also should EnablePreviewFeatures imply LangVersion=preview?

accepted/2021/preview-features/preview-features.md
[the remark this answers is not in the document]
danmoseley: Ah, but if the preview thing was an API, which didn't rely on language/compiler, we wouldn't expect a VS upgrade to affect it. Eg., it wouldn't remove it from intellisense.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,418 @@
...
+
+User code will typically only have this attribute applied at the assembly-level.
+However, the framework will typically only use this attribute on types and
+members without putting it on the assembly itself.
+
+***Note:*** Since F# doesn't support analyzers, we would have to add this
+capability to the compiler itself.
+
+## Q & A
+
```
jkotas, line 290: Are the preview APIs going to show up in intellisense by default, or only once the project opts in into EnablePreviewFeatures?

accepted/2021/preview-features/preview-features.md
stephentoub: > But should LangVersion=preview imply all preview features are enabled?

Yes :smile:

accepted/2021/preview-features/preview-features.md
terrajobst: I'm not opposed to that, but my understanding from the previous meeting was that we didn't want a model where individual features can be turned on/off but instead we have a single preview mode that turns on all preview features together.

terrajobst: Because LangVersion represents the language version. To me, it doesn't make sense to say that setting `LangVersion` to preview means that it also enables runtime and framework features.

I'm suggesting that `EnablePreviewFeatures` is the single setting for most users as we generally don't want users to set `LangVersion` anyway.

stephentoub: For runtime experiments that change behavior without requiring or in any way relying on additional APIs or language features, we're fine being on the same plan we've been on... a bazillion environment variables for configuring lots of things, some of which are experimental.  I would prefer if we started naming them DOTNET_ instead of COMPlus_ (https://github.com/dotnet/runtime/issues/47283), but that's orthogonal :smile:

accepted/2021/preview-features/preview-features.md
terrajobst: I used the message that the compiler currently emits if language syntax is guarded by the preview language version. I think it would be good if we could change the wording to make it more consistent with the overall preview mode.

accepted/2021/preview-features/preview-features.md
terrajobst: `EnablePreviewFeatures` should set `LangVersion` to `preview` unless the user explicitly set a value for `LangVersion`.

accepted/2021/preview-features/preview-features.md
stephentoub: > To me, it doesn't make sense to say that setting LangVersion to preview means that it also enables runtime and framework features.

It enables the full stack preview features.  I think we should continue using the thing we already have.

Sounds like we need another discussion on this point.

accepted/2021/preview-features/preview-features.md
terrajobst: Every time I suggested a compiler diagnostic I was told that we should default to an analyzer unless there is a strong reason why it can't be in an analyzer.

accepted/2021/preview-features/preview-features.md
terrajobst: Yes, they should show up by default in IntelliSense to make the discoverable.

accepted/2021/preview-features/preview-features.md
terrajobst: I thought about it but I don't see a need for that if we track this via assembly level attributes.

accepted/2021/preview-features/preview-features.md
terrajobst: I consider that out of scope because it's not clear to me that this maps well to the project level.

accepted/2021/preview-features/preview-features.md
terrajobst: I think that depends:

* If the feature only requires work in the runtime/library space, then we probably get that for free since LTS releases are in their own branch anyway. We'd have to go out of our way to change how a preview feature works there.
* However, if the preview feature requires work in the SDK (which includes the compiler) then consumers would be limited by which SDK they install. We can probably make that happen by avoiding making breaking changes in preview features in minor updates of a past major. We should probably still reserve the right to make breaking changes changes in minor updates of the current major so that we can make improvements to preview features we're currently still previewing.

accepted/2021/preview-features/preview-features.md
jkotas: > we didn't want a model where individual features can be turned on/off 

That is correct for functionality features: New APIs, new language features, etc. 

I expect that performance oriented features with no public surface are typically going to be different. Notice the shipping version of the performance oriented features are often opt-in as well. In your Fancy GC example, the shipping version of it would be most likely opt-in as well, e.g. same as ServerGC is opt-in today.

Some of the performance oriented features can be an exception from this rule and it may make sense to activate them automatically in preview mode - if we expect that the eventual shipping version of the performance oriented feature is going to be on by default as well.

Some of the performance oriented features affect functionality. A good example is IL linking. We should never auto activate features like that in preview mode.

accepted/2021/preview-features/preview-features.md
jkotas: Even when the project is opted into the preview mode?

accepted/2021/preview-features/preview-features.md
terrajobst: > What will the compiler do when the attribute is not applied at the assembly level?

I don't think the compiler/analyzer should do anything in this case. The SDK will generate the assembly level attribute based on the project setting, but this can be turned, which is what we will do to build the BCL. For obvious reason we don't System.Runtime as whole to have an assembly level attribute -- we only want this on specific types/members.

> If assembly A has a method M marked with [RequiresPreviewFeatures], but no assembly level attribute, and it references assembly B that has the assembly level attribute, will the compiler warn?

Yes. The assembly level attribute is meant to protect the user that just uses the project level setting. Advanced users (like us) can turn off the generation of the assembly level attribute and selectively apply this attribute to particular parts of the library. In this case it will only warn users that uses those APIs.

accepted/2021/preview-features/preview-features.md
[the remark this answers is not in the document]
terrajobst: @jkotas 

> This means that we can only count on receiving the feedback for preview features on RTM builds for about two months, until next VS update that has a good chance of containing some breaking change ships

I don't think that's very likely. Say we ship static interface members in .NET 6 and C# 10. The corresponding RTM of the .NET SDK is version 6.xxx. We wouldn't do any breaking changes in the tooling for that version.

However, the preview builds of .NET 7 runtime and SDK might have made breaking changes. This means users that are using the .NET 7 SDK might not be able to use preview features in .NET 6 anymore. But I think that's fine. Those users are on the fast train and they will likely use the .NET 7 runtime anyway.

Preview features are meant for users that want/need to be on the slower train but still want to play with preview features.

accepted/2021/preview-features/preview-features.md
terrajobst: Yes, because how would the user know that they need to do this in the first place?

The point is: they see an API. They start using it. The analyzer tells them it's a preview feature that they need to turn on. At that point they can decide: use the API and buy into preview behavior or not using the API but staying on the supported path.

accepted/2021/preview-features/preview-features.md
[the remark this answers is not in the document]
jkotas: > We wouldn't do any breaking changes in the tooling for that version.

I am specifically talking about VS. VS ships a new Roslyn and a new version of language services every couple of months and aggresively pushing everybody to update to it. You do not really have an option to stay on the old version.

If you just use notepad and .NET SDK from the command line, I agree that there is no problem - but that is not where the masses are.

accepted/2021/preview-features/preview-features.md
terrajobst: By "expressiveness" I mean being able to represent concepts that you couldn't do before. These are usually type system features (such as async/await, static interface members, nullability) as opposed to pure syntactic sugar (e.g. file scope namespace declarations, the using modifier for locals) or a few more types/methods in the BCL.

accepted/2021/preview-features/preview-features.md
terrajobst: "Unified experience" means we have a single concept/UX for the 90% case. The single switch proposal is a possible implementation for that.

accepted/2021/preview-features/preview-features.md
terrajobst: That's good point, my example might not have been the best. Yes, the intention was we're making a change in the GC that would be on by default moving forward, as opposed to a new mode (like server vs. workstation). I guess a better example would have been tiered-JITing.

accepted/2021/preview-features/preview-features.md
terrajobst: It sure does 🙂

FWIW, I'm not opposed to using `LangVersion` for this if we believe that's the right setting. My understanding from talking to @MadsTorgersen and @jaredpar was that we generally don't want people to set `LangVersion` in the project file.

accepted/2021/preview-features/preview-features.md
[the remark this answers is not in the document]
terrajobst: > I am specifically talking about VS

I see. I guess that means we'll have to think about that.

accepted/2021/preview-features/preview-features.md
terrajobst: I'll add them

accepted/2021/preview-features/preview-features.md
terrajobst: The property should be thought of as something like an API: the reference assembly must be marked as `RequiresPreviewFeatures` if the implementation assembly is. It's generally easiest if the customer treats this as a single setting across all TFMs as opposed to per TFM. However, just like the attribute itself, advanced users can of course selectively apply the attribute to specific APIs and thus also only specific TFMs of the project file.

accepted/2021/preview-features/preview-features.md
jcouv: Let's chat with Jared/Aleksey next week

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,420 @@
...
+                    AttributeTargets.Field |
+                    AttributeTargets.Event, AllowMultiple = true, Inherited = false)]
+    public sealed class RequiresPreviewFeaturesAttribute : Attribute
+    {
+        public RequiresPreviewFeaturesAttribute();
+    }
+}
+```
+
+We'll ship a built-in analyzer that will report diagnostics when an API is being
```
jcouv, line 276: Are we planning to use the preview mode to change the behavior of an existing API (maybe to fix a bug or to extend the range of functionality of a method)? 
Then it would not be possible to mark the API with `[RequiresPreviewFeatures]`. We'd potentially need a runtime check in the library.

accepted/2021/preview-features/preview-features.md
jcouv: Chatted with Aleksey and team (Jared is out today). An analyzer is preferred as a starting point unless there's a strong reason otherwise, as Immo had stated.

accepted/2021/preview-features/preview-features.md
mangod9: The GC example is interesting, we are in fact considering shipping the new regions functionality within the "standalone" GC, which would mean deploying the native clrgc.dll. Wonder if enabling such capability should be captured as part of this.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,420 @@
...
+
+With .NET Core, we've solved the problem of the extremely high in-place
+compatibility bar that .NET Framework has. This reduces risks for unintentional
+behavioral changes and also allows certain parts of .NET to make deliberate
+breaking changes if necessary.
+
+As a result, this makes it viable to innovate across runtime, libraries, and
+languages, allowing us to build more compelling features that would have been
+almost impossible to do on .NET Framework. Examples include static linking,
+default interface members, and ref structs (such as `Span<T>`).
```
jaredpar, line 24: Also function pointers, covariant returns, etc ... Pick what you want here though, just wanted to call out other examples :smile:

accepted/2021/preview-features/preview-features.md
jaredpar: My biggest concern is that not all of our cross cutting work will include language support. Consider hardware intrinsics as a concrete example. That is a cross cutting feature but had no language involvement. Hence seems odd to have `<LangVersion>preview</LangVersion>` as the way to opt into that. I don't have a strong objection to using `LangVersion` but that is one oddity that comes to mind. 

I do feel strongly that we should have one way of opting into cross cutting previews. By that I mean if we ended up picking say `<EnablePreviewFeatures>true</EnablePreviewFeatures>` then I would feel strongly that `<LangVersion>` should be defalut to `"preview"` when that was enabled.

stephentoub:
> Hence seems odd to have <LangVersion>preview</LangVersion> as the way to opt into that. I don't have a strong objection to using LangVersion but that is one oddity that comes to mind.

From my perspective, I think we do ourselves a disservice when we try to draw some kind of distinction between C# and the APIs that are usable from C#.

accepted/2021/preview-features/preview-features.md
jasonmalinowski: The word confused me too. The sentence can probably just end after "However, it's sometimes challenging to deliver cross-cutting feature work in a single release" -- I don't think anybody will disagree with that!

accepted/2021/preview-features/preview-features.md
terrajobst: OK, this was addressed by scoping out pure runtime work.

accepted/2021/preview-features/preview-features.md
terrajobst: For now, we landed on having the `EnablePreviewFeatures` as the primary switch which also sets `LangVersion` to `Preview` (unless the user explicitly set it).

accepted/2021/preview-features/preview-features.md
terrajobst: https://github.com/dotnet/designs/pull/192#discussion_r615043481

accepted/2021/preview-features/preview-features.md
terrajobst: We addressed this by scoping out pure runtime features from this.

accepted/2021/preview-features/preview-features.md
terrajobst: I have added an open issue in the design section.

accepted/2021/preview-features/preview-features.md
terrajobst: I have added a scenario to capture this.

accepted/2021/preview-features/preview-features.md
terrajobst: I don't think so. We'd simply make the change in a preview build of the platform and/or add a compat switch to get the old behavior back if deemed necessary.

accepted/2021/preview-features/preview-features.md
```diff
@@ -0,0 +1,572 @@
...
+marked as `[RequirePreviewFeatures]`.
+
+For instance:
+
+```C#
+void f(dynamic x) => x.PreviewMethod();
+```
+
+The problem with doing that is that it's not very pay-for-play friendly (every
+single time a method is called via reflection, we need to check for the applied
```
jkotas, line 357: In additional to not being pay-for-play, it is hard to do well. We have tried that multiple times in the past (last time with UWP) and it was always leaky.

terrajobst: Good point. I've expanded the section.

jmarolf (approved):

PR #195, opened 2021-04-01
Compile-time source generation for structured log messages

maryamariyan:
I gathered all my notes into this design doc. (this is a first draft)

/cc: 
The [Logging Generator](https://github.com/geeknoid/LoggingGenerator) Author: @geeknoid
@davidfowl @stephentoub @shirhatti @333fred @ericstj 

<details>
<summary> TODO (completed) </summary>

- [x] Add sample using exception as argument to log method: (As a general rule, the first instance of ILogger, LogLevel, and Exception are treated specially in the log method signature. Subsequent instances are treated like normal args)
```csharp
// below works
        [LoggerMessage(110, LogLevel.Debug, "M1 {ex3} {ex2}")]
        static partial void LogMethod(ILogger logger, System.Exception ex, System.Exception ex2, System.Exception ex3);

// but this warns:
        // warning SYSLIB0013: Don't include a template for ex in the logging message since it is implicitly taken care
	// DiagnosticSeverity.Warning,
        [LoggerMessage(0, LogLevel.Debug, "M1 {ex} {ex2}")]
        static partial void LogMethod(ILogger logger, System.Exception ex, System.Exception ex2);
```
- [x] Add note about ordering:
There is no constraint in the ordering at this point. So user is able to add an Exception, ILogger, LogLevel signature, or LogLevel, ILogger, Exception, or whatever. Sample: (note ILogger argument is the last argument)
```c#
        [LoggerMessage(110, LogLevel.Debug, "M1 {ex3} {ex2}")]
        static partial void LogMethod(System.Exception ex, System.Exception ex2, System.Exception ex3, ILogger logger);
```
- [x] Add a note, defining clearly the scope of uniqueness for event ids, the generator would have an event id uniqueness check as warning so it can be suppressed) + Add a sample.

> We established during reviews that unique ids per assembly wouldn't work. So, we changed the scope of uniqueness for the event IDs to be the class in which the log method is declared. It doesn't check up the inheritance hierarchy for conflicting IDs because it also wouldn't be possible cross-assembly boundaries anyways. The diagnostic message for this condition would be a warning indicating the class name.
- [x] Docs in the future would need to mention that event ID uniqueness is checked per class itself not at base classes.

- [x] Explain how the source generator detects the ILogger when using instance mode. Could we support logger APIs from base class or on interfaces?
> When generating instance-mode logging methods, the field that holds the logger is now determined by looking at available fields. You get an error if there isn't a field of the right type or if there is more than one such field. This also doesn't look up the inheritance hierarchy, but it would be possible to add this capability in the future. (show sample)

</details>


davidfowl:
I would love the 2 ideas for what gets generated in here.

proposed/logging-generator.md
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    [LoggerMessage(1, LogLevel.Trace)]
+    public partial void LogEmptyMessage();
+
+    [LoggerMessage(9, LogLevel.Trace, "Fixed message", EventName = "CustomEventName")]
+    public partial void LogWithCustomEventName();
+
+    [LoggerMessage(10, "Welcome to {city} {province}!")]
+    public partial void LogWithDynamicLogLevel(string city, LogLevel level, string province);
```
loic-sharma, line 166: My first thought when I saw this attribute was that this used some default log level (I didn't notice the parameter).  Do you think this may confuse customers? Is logging with a dynamic level common?

proposed/logging-generator.md
```diff
@@ -0,0 +1,308 @@
...
+`LoggerMessageAttribute` As shown in the examples, is used by the developer to trigger source generation:
+
+```c#
+namespace Microsoft.Extensions.Logging
+{
+    [AttributeUsage(AttributeTargets.Method)]
+    public sealed partial class LoggerMessageAttribute : Attribute
+    {
+        public LoggerMessageAttribute(int eventId, LogLevel level, string? message = null);
+        public LoggerMessageAttribute(int eventId, string? message = null);
```
loic-sharma, line 90: Just curious, why is the event ID required? Would it make sense to add a constructor without an event ID?

proposed/logging-generator.md
```diff
@@ -0,0 +1,308 @@
...
+    public void LogMethod(string name)
+    {
+        Log.LogName(_logger, name);
+    }
+
+    private static class Log
+    {
+        private static readonly Action<ILogger, string, Exception> _logName = LoggerMessage.Define<string>(LogLevel.Information, 0, @"Hello {name}");
+
+        public static void LogName(ILogger logger, string name)
```
jmezach, line 58: I would assume this to be an extension method, but it is currently missing the `this` keyword? Is that intentional?

davidfowl: It's not required to be one. I also don't think partial extension methods are supported today

proposed/logging-generator.md
davidfowl: It's required because they uniquely identify the event within a category.

loic-sharma: Sorry, I still don’t understand the requirement as the `LoggingSample1` doesn’t provide an event ID. From my perspective, this loses the uniqueness benefit but gains some ergonomics in return. If I were to use `LoggerMessage` I would likely try to use `0` as my event ID as I personally don’t mind losing uniqueness.

davidfowl: A 0 event id is wrong though, the generator will flag it and give you a compile time error. You need to specify unique event ids for your log messages. They are a contract between log producer and log consumer. It's the public API equivalent for log messages.

YairHalberstadt: Is it possible you won't know your eventId at compile time? Perhaps you store it in a config?

davidfowl: Yep, but then you won't be using this source generator because its likely your format string also comes from that configuration file. At that point, you need a runtime API to read from configuration and "register" log messages. That's what `LoggerMessage.Define` gives you.

proposed/logging-generator.md
```diff
@@ -0,0 +1,327 @@
...
+    builder.FormatCallback
+);
+```
+Through this new compiler feature, the builder would be able to recognize name holes, (here the city) or even format specifiers. However, currently the proposed builder approach is limited in its ability to get both name holes and format specifiers at once. The logger to is able to identify both of these from message templates and would be good to have this feature available.
+
+Also, an older compiler, the builder overload may not be detectable and therefore the above log APIs could end up calling the wrong overload, which would cause in logging messages to completely lose the structure provided in the input. This could be however be mitigated with an analyzer that warns user when they use string interpolated messages on older compiler versions.
+ 
+### Limitations with string interpolation builder approach:
+
+- With the interpolated string builder approach, there is currently no way to get both name holes and format specifiers, but this is supported today already by the existing logger APIs.
```
333fred, line 305: I'm not sure this is quite true, but it would depend on how creative we want to get with the format string. We could use some currently-unused character to specify a "separator" between the regular format string part and the name you want the hole to have. It would certainly take some design though.

shirhatti: @333fred I would love for that. Do we file an issue for this?

proposed/logging-generator.md
davidfowl: Who changes log level per call @geeknoid?

proposed/logging-generator.md
333fred: To be clear, I was not talking about a separate language feature here, just making the proposed builder type above smarter. I would imagine and format changes would be part of this proposal.

proposed/logging-generator.md
```diff
@@ -0,0 +1,327 @@
...
+
+- The biggest concern for using the custom struct builder for our logging approach is that we'd need to store some value for name holes eagerly in the builder. This step cannot be done lazily because the structure we generate could ultimately get serialized in different ways decided by the different consumers of `ILogger.Log` call. The end result of using the builder this way, could be less efficient than what we already have in our logging APIs, even though our current approaches already do boxing.
+
+### Benefits:
+
+- The name hole approach adds enough value that would be good to consider in the future for cases where we are given an interpolated string rather than message templates.
```
stephentoub, line 311: There are key benefits not called out here.  In particular, the interpolated string approach means callers don't need to guard the call sites if any computation is necessary to produce the arguments to the logging.  With all of the other approaches considered in this design, any such work done at the call site to produce the arguments will need to be guarded behind appropriate logging enabled checks, or else it'll end up being done even if logging is disabled.

proposed/logging-generator.md
```diff
@@ -0,0 +1,327 @@
...
+
```
stephentoub, line 312: Also, the comment made back on the original issue stated:
"A benefit there is that we can use string format syntax as a natural way which we can consider for imperative logging, but the consensus there is to keep the logging generator to enable writing more declarative logging APIs."
FWIW, I think this is a strange line being drawn between "declarative" and "imperative".  From the developer perspective, we're talking about the difference between (using one of your examples from earlier):
```C#
[LoggerMessage(0, LogLevel.Information, "Hello `{name}`")]
public static partial void LogName(this ILogger logger, string name);
```
and
```C#
public static void LogName(this ILogger logger, string name) =>
    logger.Log(0, LogLevel.Information, $"Hello `{name}`");
```
If we define "declarative" as using an attribute, sure, one does and one doesn't.  But otherwise, we're still declaring the same information, just slightly different syntax.

The valid reasons I heard that we might prefer the source generator approach were:
- Wanting to enable the exact format syntax required by https://messagetemplates.org/
- Wanting to be able to pass all of original data as-is without any complication around how it's stored prior to something being done with it other than creating a string
- Wanting to be able to supply alternative names for the holes, if the C# 10 syntax doesn't end up allowing that (but Fred is investigating)

None of the other reasons I heard were compelling to me for why we'd go the source generator route.

proposed/logging-generator.md
maryamariyan: With this issue https://github.com/dotnet/runtime/issues/45290, there were new Define overloads added that guard the call site producing the arguments with isEnabled check

proposed/logging-generator.md
maryamariyan: That's fair. I think they both offer declarative. I will update the proposal.

davidfowl: I honestly don't think these approaches are at odds but:  
- The structure *has* to be preserved (Names), that's the main thing about the string builder approach that is missing (hopefully we can make something work in C# 10).
- Not being able to lazily ToString is a problem, we don't want to allocate until a logger provider asks for the string
- The approach is declarative and toolable. We can write rules against the attribute instead of call sites that use Log.

I say we treat the builder approach as distinct from the source generator. In fact, in the later proposals we're going to generate calls to `LoggerMessage.Define<...>`. This change basically re-enforces that the source generator is a declarative approach to the model we have today.

proposed/logging-generator.md
davidfowl: @maryamariyan can you write out a sample of the latest code gen for the source generator.

proposed/logging-generator.md
stephentoub: > The structure has to be preserved (Names), that's the main thing about the string builder approach that is missing

This is my "Wanting to be able to supply alternative names for the holes" bullet.  If just `{name}` is used, it can already support that.  I suggest if specifying names separate from the expression itself is important, that the design document call that out and show it in examples: currently I don't see any examples that supply different names.

> Not being able to lazily ToString is a problem, we don't want to allocate until a logger provider asks for the string

This is my "Wanting to be able to pass all of original data as-is without any complication around how it's stored prior to something being done with it other than creating a string" bullet. But it depends on what you mean by allocating.  I imagine the builder would either format into an ArrayPool-rented char[] or serialize into an ArrayPool-rented byte[].  And of course only after checking whether logging was enabled at an appropriate logging level.

> The approach is declarative and toolable

I still question the declarative statement here.  An analyzer would search for attributes and look at attribute arguments, or an analyzer would search for Log method calls and look at method call arguments.  Is the statement really that sites are implicitly constrained by what you're allowed to write in attributes?

proposed/logging-generator.md
stephentoub: An example would be very helpful, as I'm currently missing how those APIs help here.  Just to take a concrete example, let's pick this method from ASP.NET:
https://github.com/dotnet/aspnetcore/blob/277cc564ce5a76c6613f29dd92c0168e06b41b1c/src/Shared/CertificateGeneration/CertificateManager.cs#L790-L791
Call site looks like:
https://github.com/dotnet/aspnetcore/blob/277cc564ce5a76c6613f29dd92c0168e06b41b1c/src/Shared/CertificateGeneration/CertificateManager.cs#L85-L88
Now obviously that's an event on an EventSource, but let's pretend it's not... with these logging APIs it would be translated to something like:
```C#
[LoggerMessage(2, LogLevel.Verbose, "Certificates: `{matchingCertificates}`")]
public static partial void DecribeFoundCertificates(this ILogger logger, string matchingCertificates);
```
And the call site would presumably be:
```C#
logger.DescribeFoundCertificates(ToCertificateDescription(matchingCertificates));
```
How does that caller avoid the `ToCertificateDescription(matchingCertificates)` computation without adding their own guard check?

proposed/logging-generator.md
geeknoid: The log level is not normally changed per call, it's based on config.

The specific case is that we are building a generic facility to log HTTP requests. Someone wanting to use our component gets to decide the log level our log state comes out as. The first implementation of that code needed to declare different log methods, all with the same parameters, except for the log level. And then at runtime, it needed  switch statement to decide which version of the log method to invoke. Basically, it was quite lame. Making this small extension to the logger generator to make the log level available dynamically eliminated all that crazy redundancy.

proposed/logging-generator.md
geeknoid: The logging method can be:

* A plain static method.
* An extension method.
* An instance method (where the ILogger comes from a field on the containing type).

proposed/logging-generator.md
geeknoid: From my perspective "declarative" is really about the fact it is a "guided" experience. The code generator has deep semantic understanding of what the user is doing and provides useful error messages that don't exist using completely generic mechanisms.

For example, the code generator generates warnings for duplicate event ids. It'll auto-generate a message if you don't supply one to output parameters as a JSON blob, etc.

geeknoid: In terms of supplying different names, the current implementation provides a limited form of that as it optionally supports pascal casing of names. So you can supply "foo" as a method parameter, and call it "{Foo}" in the message text (and in the downstream IReadOnlyCollection produced on the TState given to ILogger). This feature is controlled by a global option on the generator.

davidfowl: > This is my "Wanting to be able to supply alternative names for the holes" bullet. If just {name} is used, it can already support that. I suggest if specifying names separate from the expression itself is important, that the design document call that out and show it in examples: currently I don't see any examples that supply different names.

We change casing all of the time in ASP.NET Core. We also define templates separately from where we specify values so this happens *all the time*. The intent isn't to capture variables but to design log messages and their structure and pass the appropriate values after the fact.

> This is my "Wanting to be able to pass all of original data as-is without any complication around how it's stored prior to something being done with it other than creating a string" bullet. But it depends on what you mean by allocating. I imagine the builder would either format into an ArrayPool-rented char[] or serialize into an ArrayPool-rented byte[]. And of course only after checking whether logging was enabled at an appropriate logging level.

We need to eagerly stash all of the state before we ToString the overall thing. This makes it hard to avoid boxing or calling ToString on the incoming parameters. So there's an allocation forced up front to produce the builder.

> I still question the declarative statement here. An analyzer would search for attributes and look at attribute arguments, or an analyzer would search for Log method calls and look at method call arguments. Is the statement really that sites are implicitly constrained by what you're allowed to write in attributes?

Attributes force a particular pattern that we like and is easier to analyze than code. This is a style preference thing though.

stephentoub: > We need to eagerly stash all of the state before we ToString the overall thing. This makes it hard to avoid boxing or calling ToString on the incoming parameters. So there's an allocation forced up front to produce the builder.

If you don't count ArrayPool rentals as allocation (waving my hands), then there's no allocation to create the builder.  You're right there would be ToString allocations as part of formatting into that, for any non-primitive types that allocate in their ToString methods. I expect with ISpanFormattable that will go way down, though.

proposed/logging-generator.md
davidfowl: These APIs don't help there. You still need an if check here but I don't know why that's a problem...

stephentoub: > You still need an if check here but I don't know why that's a problem...

Because people forget and then we ship .NET 5 with ASP.NET paying to log data about all certificates on the box even though that logging is disabled :wink:

davidfowl: LOL, I think we should fix this code regardless.

stephentoub: https://github.com/dotnet/aspnetcore/pull/27956

proposed/logging-generator.md
```diff
@@ -0,0 +1,327 @@
+# Logging Generator for strongly-typed logging messages
+
+## Summary
+
+This is a proposal for a new approach to logging to allow for more easier setup of strongly-typed logging messages. This can be achieved by using a source generator, that gets triggered with a `LoggerMessageAttribute` on partial logging methods, and would be able to either autogenerate the implementation of such method, or produce compile-time diagnostics hinting developers into proper usage of this logging approach.
```
ericstj, line 5: Source generator is a solution, and in fact an alternative solution is a language feature.  We should be starting off with a problem statement.  

You mention
> This is a proposal for a new approach to logging to allow for more easier setup of strongly-typed logging messages

This implies the problem is *it's hard to log strongly-typed logging messages*, but it might be bigger than that.  It'd be nice to describe the problem and its characteristics.  Your goals can then talk about how to solve the problem and potentially the prioritization of the goals.

@shirhatti can you share some thoughts on this?

proposed/logging-generator.md
```diff
@@ -0,0 +1,327 @@
...
+```
+
+Here are some problems with the `LoggingSample1` sample using `LogInformation`, `LogWarning`, etc.: 
+1. We cannot provide event ID through these APIs
+2. Parameters passed are processed before LogLevel checks
+3. It requires parsing of message string on every use to find templates to substitute
+
+Because of these problems, the more efficient approach recommended today as best practices is to use [LoggerMessage.Define](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/logging/loggermessage?view=aspnetcore-5.0) APIs instead, illustrated below with `LoggingSample2`:
+
+```c#
```
ericstj, line 39: nit: this doc uses both `c#` and `csharp` for formatting.  Only the latter gets coloration in the github rendering.

proposed/logging-generator.md
```diff
@@ -0,0 +1,327 @@
+# Logging Generator for strongly-typed logging messages
+
+## Summary
+
+This is a proposal for a new approach to logging to allow for more easier setup of strongly-typed logging messages. This can be achieved by using a source generator, that gets triggered with a `LoggerMessageAttribute` on partial logging methods, and would be able to either autogenerate the implementation of such method, or produce compile-time diagnostics hinting developers into proper usage of this logging approach.
+
+## Scenarios
+
+Heads up that `LoggingSample1` and `LoggingSample2` show existing approaches to logging, and the remaining will illustrate the samples using the source generator approach with `LoggerMessageAttribute`.
+
```
ericstj, line 10: Is it a goal to support multiple languages: VB and F#?

proposed/logging-generator.md
```diff
@@ -0,0 +1,327 @@
...
+    public void LogMethod(string name)
+    {
+        _logger.LogInformation("Hello {name}", name);
+    }
+}
+```
+
+Here are some problems with the `LoggingSample1` sample using `LogInformation`, `LogWarning`, etc.: 
+1. We cannot provide event ID through these APIs
```
davidfowl, line 33: We can but they aren't required today. Leading to bad usages in real systems that want to react or detect specific event issues being logged.

proposed/logging-generator.md
davidfowl: Agreed. The samples show the progression but doesn't exactly explain the fact that the source generator is about making structured logging more convenient.

proposed/logging-generator.md
loic-sharma: > They are a contract between log producer and log consumer. It's the public API equivalent for log messages.

Could you say more about that contract? I see that [`LoggerExtensions.Log`](https://github.com/dotnet/runtime/blob/8445e4eaa3b18bfdbed1eb1507293777c590d371/src/libraries/Microsoft.Extensions.Logging.Abstractions/src/LoggerExtensions.cs#L372-L375) accepts a nullable message and defaults the event ID to 0. Does that mean that `LogInformation` and the likes break that contract if you skip the event ID parameter?

A little bit of context. I worked on SharePoint where all log statements have a unique event ID automatically generated. These logs served a dual purpose: to debug problems in production, and as signals of events for monitoring and analysis by data scientists. Unfortunately, engineers would often break monitors or data reports by changing the some logging logic. It was hard to tell which logging statements were used just for logging, and which ones were also used as signal events.

I'm now on the nuget.org team and we avoid this problem by separating logs (for debugging) and metrics (for signals of events). We avoid regressions in our metrics through testing. If we have gaps in our metrics, we can do adhoc analysis using our logs (our logs include the original message template, which we use as a proxy to an event ID).

TLDR: I'm skeptical of event IDs in logs, but maybe there are scenarios I'm unaware of :)

shirhatti: > I'm now on the nuget.org team and we avoid this problem by separating logs (for debugging) and metrics (for signals of events). We avoid regressions in our metrics through testing. If we have gaps in our metrics, we can do adhoc analysis using our logs (our logs include the original message template, which we use as a proxy to an event ID).

The contract between a log producer and log consumer is that an EventId uniquely identifies a log (and not the message template).

If in your use case, you're willing to make a even stronger contract that the message template uniquely identifies a log, then that's okay too (until you have a typo that you're afraid to fix😛) 

> TLDR: I'm skeptical of event IDs in logs. I personally prefer separating logs and metrics.

Nothing in this proposal is asking you to compute metric from logs. I wholly agree with you that it's a bad idea for a multitude of reasons.

proposed/logging-generator.md
stephentoub: > Only the latter gets coloration in the github rendering.

Not true. 
```C#
public class ThisHas
{
    public static bool Color;
}
```

proposed/logging-generator.md
shirhatti: > Thanks for the explanation! Could you say more about that contract? I see that LoggerExtensions.Log accepts a nullable message and defaults the event ID to 0. Does that mean that LogInformation and the likes break that contract if you skip the event ID parameter?

I'd recommend library authors not use that method. I suspect this was added a long time ago to placate application developers to try and reduce the concept counts in logging, but this is from before my time 🤷

proposed/logging-generator.md
ericstj: Maybe the comment colorization is different than file browser?  Here's what I see:
image
image

proposed/logging-generator.md
davidfowl: I'm sure there's more

proposed/logging-generator.md
loic-sharma: I had a chat with @shirhatti offline. What do y'all think of:

1. Clarifying in the goals of the document that this spec is trying to enforce best practices on library authors
2. Allowing uses of the `LoggerMessage` attribute without defining an event ID, but, adding a Roslyn Analyzer that warns/errors to recommend best practices

davidfowl: I think it should error by default and you can turn it off if you don't care about event ids.

proposed/logging-generator.md
maryamariyan: > @maryamariyan can you write out a sample of the latest code gen for the source generator.

here's a sample:

```csharp
    static partial class Log
    {
        [LoggerMessage(32, LogLevel.Trace, "Certificates: `{matchingCertificates}`")]
        public static partial void DescribeFoundCertificates(this ILogger logger, string matchingCertificates);
    }
```

generates:
```csharp

namespace Example
{
    partial class Log
    {
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "1.0.0.0")]
        private static readonly global::System.Action<global::Microsoft.Extensions.Logging.ILogger, string, global::System.Exception?> _DescribeFoundCertificatesCallback =
            global::Microsoft.Extensions.Logging.LoggerMessage.Define<string>(global::Microsoft.Extensions.Logging.LogLevel.Trace, new global::Microsoft.Extensions.Logging.EventId(32, nameof(DescribeFoundCertificates)), "Certificates: `{matchingCertificates}`", true);

        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "1.0.0.0")]
        public static partial void DescribeFoundCertificates(this Microsoft.Extensions.Logging.ILogger logger, string matchingCertificates)
        {
            if (logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Trace))
            {
                _DescribeFoundCertificatesCallback(logger, matchingCertificates, null);
            }
        }

    }
}
```

usage:
```csharp
logger.DescribeFoundCertificates("some matching certificates go here!");
```

stephentoub: > usage

My point is the usage with the source generator needs to be:
```C#
if (logger.IsEnabled(LogLevel.Trace))
{
    logger.DescribeFoundCertificates(Describe(matchingCertificates));
}
```

davidfowl: For this specific usage yes. We don't need to delete the feature because one of these usages requires a manual if check.

stephentoub: I'm not arguing we need to "delete the feature".  I'm arguing there's a real pit of failure here's that being swept under the rug.

davidfowl: I don't think it's being swept under the rug. Typically the way we solve that pattern is by adding a method that takes unprocessed data and manually adds the if check. It's the exact same thing you need to do today, this doesn't improve that situation. I'm not sure how to fix that other than having a lambda. There are lots of lots that don't require complex transformation and are just logging the data that you have. Maybe an analyzer can detect these use cases and recommend an IsEnabled check.

proposed/logging-generator.md
```diff
@@ -0,0 +1,488 @@
...
+    new EventId(1),
+    builder, // the CustomLoggerParamsBuilder as log state
+    exception,
+    (s, e) => s
+);
+```
+
+
+
+On C# 10 and above. `CustomLoggerParamsBuilder` would be a custom [string interpolation builder](https://github.com/333fred/csharplang/blob/2550a43b391e844faaa0a2023f66489328a41612/proposals/improved-interpolated-strings.md) defined as:
```
333fred, line 326:
```suggestion
On C# 10 and above. `CustomLoggerParamsBuilder` would be a custom [string interpolation builder](https://github.com/dotnet/csharplang/blob/d41fcd0377db1ecccfa8de9185b9aa07f894b342/proposals/improved-interpolated-strings.md) defined as:
```

proposed/logging-generator.md
maryamariyan: addressed this by adding an extra Q&A section for the source gen in the document.

proposed/logging-generator.md
```diff
@@ -0,0 +1,487 @@
...
+        builder = new CustomLoggerParamsBuilder(baseLength, formatHoleCount, logLevel);
+        return true;
+    }
+
+    private LogLevel _logLevelEnabled;
+
+    // returns the number of holes
+    public int Count => ... // depends on the type used to store key value pairs
+
+    public static readonly Func<CustomLoggerParamsBuilder, Exception?, string> FormatCallback = (builder, exception) => builder.ToString();
```
davidfowl, line 349: This doesn't need to exist right?

proposed/logging-generator.md
```diff
@@ -0,0 +1,487 @@
...
+```
+
+described in solution 1, the usage would look like:
+
+```csharp
+public static void LogName(this ILogger logger, string name, Exception exception) =>
+    logger.Log(LogLevel.Information, new EventId(0), $"Hello `{name}`", exception, (s, e) => s);
+```
+
+where `(s, e) => s` is the format callback.
```
davidfowl, line 302: `s.ToString()`?

proposed/logging-generator.md
```diff
@@ -0,0 +1,489 @@
...
+
+    // returns the number of holes
+    public int Count => ... // depends on the type used to store key value pairs
+
+    public static readonly Func<CustomLoggerParamsBuilder, Exception?, string> FormatCallback = (builder, exception) => builder.ToString();
+
+    private CustomLoggerParamsBuilder(int baseLength, int formatHoleCount, LogLevel logLevelEnabled)
+    {
+        // Initialization logic
+        _logLevelEnabled = logLevelEnabled;
```
davidfowl, line 355: Why is the level passed in here? I don't think it's needed here.

proposed/logging-generator.md
```diff
@@ -0,0 +1,489 @@
...
+    public bool TryFormatInterpolationHole<T>(T t)
+    {
+        // Store and format part as required
+        // Store name hole and value as KeyValuePair<string, object?>
+        return true;
+    }
+
+    public KeyValuePair<string, object?> this[int index]
+    {
+        get => ... // depends on the type used to store key value pairs
```
davidfowl, line 373: Its interesting what gets stored here.

proposed/logging-generator.md
```diff
@@ -0,0 +1,489 @@
...
+
+    public bool TryFormatBaseString(string s)
+    {
+        // Store and format part as required
+        return true;
+    }
+
+    public bool TryFormatInterpolationHole<T>(T t)
+    {
+        // Store and format part as required
```
davidfowl, line 366: We should call out that this would need to box T

proposed/logging-generator.md
```diff
@@ -0,0 +1,489 @@
...
+```
+
+### Q & A - The builder approach:
+
+- Question: Can I use a custom string interpolation builder in an async context?
+
+```csharp
+public async Task Method()
+{
+    // Cannot use MyStringBuilder if it's a ref struct
```
davidfowl, line 398: We can't use a ref struct anyways because we're using a generic type.

proposed/logging-generator.md
```diff
@@ -0,0 +1,489 @@
...
+    var key1 = "value1";
+    var key2 = "value2";
+    logger.LogInformation($"Some text: {key1}, {key2}");
```
davidfowl, line 401: @stephentoub are there cases where this will falls back to a string.Format call? That would be bad.

proposed/logging-generator.md
```diff
@@ -0,0 +1,489 @@
...
+1. (+1 for the C# 10 Language Feature) _With the interpolated string builder, callers don't need to guard the call sites if any computation is necessary to produce the arguments to the logging. Whereas with the source generator, the user call site would need to get wrapped around `IsEnabled` checks._
+
+For example with the source generator:
+
+```csharp
+[LoggerMessage(2, LogLevel.Trace, "Certificates: `{matchingCertificates}`")]
+public static partial void DecribeFoundCertificates(this ILogger logger, IEnumerable<X509Certificate2> matchingCertificates);
+```
+
+The user call site would need to get wrapped around `IsEnabled` check for this specific use case to skip evaluating `Describe(..)` which might be expensive to compute:
```
davidfowl, line 424: 👍🏾

proposed/logging-generator.md
stephentoub: As long as the builder isn't a ref struct, you can expose a set of TryFormatXx methods so that it won't.

proposed/logging-generator.md
```diff
@@ -0,0 +1,489 @@
...
+
+### Sample-based comparison of two approaches
+
+1. (+1 for the C# 10 Language Feature) _With the interpolated string builder, callers don't need to guard the call sites if any computation is necessary to produce the arguments to the logging. Whereas with the source generator, the user call site would need to get wrapped around `IsEnabled` checks._
+
+For example with the source generator:
+
+```csharp
+[LoggerMessage(2, LogLevel.Trace, "Certificates: `{matchingCertificates}`")]
+public static partial void DecribeFoundCertificates(this ILogger logger, IEnumerable<X509Certificate2> matchingCertificates);
```
stephentoub, line 421: Can you write this with the source generator?  What translates the IEnumerable into the resulting string?

maryamariyan: Yes, could lead to:

```csharp  
[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "1.0.0.0")]
private static readonly global::System.Action<global::Microsoft.Extensions.Logging.ILogger, System.Collections.Generic.IEnumerable<string>, global::System.Exception?> _DecribeFoundCertificatesCallback =
    global::Microsoft.Extensions.Logging.LoggerMessage.Define<System.Collections.Generic.IEnumerable<string>>(global::Microsoft.Extensions.Logging.LogLevel.Trace, new global::Microsoft.Extensions.Logging.EventId(2, nameof(DecribeFoundCertificates)), "Certificates: `{matchingCertificates}`", true); 

[global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "1.0.0.0")]
public static partial void DecribeFoundCertificates(this Microsoft.Extensions.Logging.ILogger logger, System.Collections.Generic.IEnumerable<string> matchingCertificates)
{
    if (logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Trace))
    {
        _DecribeFoundCertificatesCallback(logger, matchingCertificates, null);
    }
}
```

The underlying `LoggerMessage.Define<IEnumerable<string>>(..)` call translates `IEnumerable` to string, (using an internal `LogValues` type as TState: [link to implementation](https://source.dot.net/#Microsoft.Extensions.Logging.Abstractions/LoggerMessage.cs,241))

stephentoub: But if the dev actually wanted to control the string rather than it just concatenating ToString()s of the enumerated items, e.g.
https://github.com/dotnet/aspnetcore/blob/277cc564ce5a76c6613f29dd92c0168e06b41b1c/src/Shared/CertificateGeneration/CertificateManager.cs#L104
presumably they would instead need to write something like this:
```C#
public static void DecribeFoundCertificates(this ILogger logger, IEnumerable<X509Certificate2> matchingCertificates)
{
    if (logger.IsEnabled(LogLevel.Trace)
    {
        DecribeFoundCertificates(ToCertificateDescription(matchingCertificates));
    }
}

[LoggerMessage(2, LogLevel.Trace, "Certificates: `{matchingCertificates}`")]
private static partial void DecribeFoundCertificates(this ILogger logger, string matchingCertificates);
```
?

davidfowl: That's correct. That's the pattern we use today

maryamariyan: Yes sure, that looks like a valid sample code, when developer wants to control the string itself.

maryamariyan: I'll update the document sample using your snippet above

proposed/logging-generator.md
```diff
@@ -0,0 +1,481 @@
...
+
+Q & A - The source generator
+
+- Question: Why would we want to enforce event ID uniqueness checks if existing log APIs still allow for using default event ID values? For example with `LogInformation` we have two overloads, one taking `EventId` and the other one not. But the proposal with `LoggerMessageAttribute` only requires taking an `EventId`.
+
+Answer: The enforced restriction added via `LoggerMessage` attribute aims at providing best practices for library authors more so than it does for app developers, who in most cases do not care about event IDs.
+
+This is a great feedback. In order to make this approach less restrictive, it would make sense to also allow `LoggerMessageAttribute` skip taking event IDs in combination with an analyzer that generates an error to recommend best practices while allowing app developers to cut corners for productivity. When turned off, the event IDs would default to zero or implicit event IDs would get generated.
+
+## Solution 2: using a custom string interpolation builder
```
davidfowl, line 282: I wouldn't pose this as an either OR like I said before. Lets just list this as an optional feature we can do regardless of the source generator approach.

davidfowl: ## Supporting string interpolation builder in logging

proposed/logging-generator.md
```diff
@@ -0,0 +1,481 @@
...
+
+## Requirements
+
+### Goals
+
+Today it is difficult to declare strongly-typed code paths for ILogger messages (e.g. `LogMethod` APIs shown in both samples above) in a way that's efficient, approachable and maintainable by an entire team all at the same time. The solution we are looking for should ideally lead/guide developers into writing more declarative, less verbose, and easy to write, logging APIs with reduced boilerplate code. 
+
+Part of the goal is to enforce best practices on library authors, so they would be required to provide event IDs, add diagnostics to warn against reusing event IDs, and lead towards writing more efficient logs (all of which discourage against using `LoggingSample1`).
+
+### Non-Goals
```
davidfowl, line 68: `LoggerMessage.Define` doesn't box anything, the design of exposing the log as an `IReadOnlyList<KeyValuePair<string, object>>`  which happens for all solutions mentioned below.

proposed/logging-generator.md
```diff
@@ -0,0 +1,481 @@
...
+
+Where the implementation for `LogName` would be completed by the source generator. 
+
+### Sample with static class Log
+
+```csharp
+public static partial class Log
+{
+    [LoggerMessage(0, LogLevel.Information, "Hello `{name}`")]
+    public static partial void LogName(this ILogger logger, string name);
```
davidfowl, line 127: I'm not sure static partials work yet. 

@jaredpar this isn't in for .NET 6.

proposed/logging-generator.md
```diff
@@ -0,0 +1,481 @@
...
+public partial class LoggingSample6
+{
+    private readonly ILogger _logger;
+
+    public LoggingSample6(ILogger logger)
+    {
+        _logger = logger;
+    }
+
+    // possible when PascalCaseArguments is set to true:
```
davidfowl, line 192: This should just work by default.

proposed/logging-generator.md
```diff
@@ -0,0 +1,481 @@
...
+### Sample-based comparison of two approaches
+
+1. (+1 for the C# 10 Language Feature) _With the interpolated string builder, callers don't need to guard the call sites if any computation is necessary to produce the arguments to the logging. Whereas with the source generator, the user call site would need to get wrapped around `IsEnabled` checks._
+
+For example with the source generator, the user call site would need to get wrapped around `IsEnabled` check for this specific use case to skip evaluating `Describe(..)` which might be expensive to compute:
+
+```csharp
+public static void DecribeFoundCertificates(this ILogger logger, IEnumerable<X509Certificate2> matchingCertificates)
+{
+    if (logger.IsEnabled(LogLevel.Trace)
```
davidfowl, line 414:
```suggestion
    if (logger.IsEnabled(LogLevel.Trace))
```

proposed/logging-generator.md
```diff
@@ -0,0 +1,496 @@
...
+
+    [LoggerMessage(0, LogLevel.Information, "Hello {name}")]
+    public partial void LogName(string name);
+}
+```
+
+Where the implementation for `LogName` would be completed by the source generator. For example, the generated code for `LoggingSample3` would look like this:
+
+```csharp
+    partial class LoggingSample3
```
davidfowl, line 118: nit: Formatting, push this all the way left.

proposed/logging-generator.md
```diff
@@ -0,0 +1,496 @@
...
+}
+```
+
+
+### Sample with exception as argument to log method
+
+`ILogger.Log` API signature takes log level and optionally an exception per log call:
+
+```csharp
+    public partial interface ILogger
```
davidfowl, line 167: nit: Formatting, push this all the way left.

proposed/logging-generator.md
```diff
@@ -0,0 +1,496 @@
...
+and allow for expensive computations to be skipped when logging is not enabled.
+
+2. _The source generator provides useful error messages that don't exist using completely generic mechanisms to enforce logging best practices._
+
+For example, the generator can provide an event ID uniqueness diagnostics check with warning severity so it can be suppressed.
+
+The scope of uniqueness for the event IDs would be the class in which the log method is declared. It doesn't check up the inheritance hierarchy for conflicting IDs because it also wouldn't be possible cross-assembly boundaries. The diagnostic message for this condition would be a warning indicating the class name:
+
+```csharp
+    static partial class LogClass
```
davidfowl, line 455: nit: Formatting, push this all the way left.

proposed/logging-generator.md
```diff
@@ -0,0 +1,496 @@
...
+    public partial interface ILogger
+    {
+        void Log<TState>(Microsoft.Extensions.Logging.LogLevel logLevel, Microsoft.Extensions.Logging.EventId eventId, TState state, System.Exception? exception, System.Func<TState, System.Exception?, string> formatter);
+    }
+```
+
+Therefore, as a general rule, the first instance of ILogger, LogLevel, and Exception are treated specially in the log method signature of the source generator. Subsequent instances are treated like normal arguments to the message template:
+
+```c#
+// below works
```
davidfowl, line 176: nit: Formatting, push this all the way left.

proposed/logging-generator.md
```diff
@@ -0,0 +1,496 @@
...
+```csharp
+[LoggerMessage(0, LogLevel.Information, "Hello `{name}`")]
+public static partial void LogName(this ILogger logger, string name, Exception exception);
+```
+
+described in the previous section, the usage would look like:
+
+```csharp
+public static void LogName(this ILogger logger, string name, Exception exception) =>
+    logger.Log(LogLevel.Information, new EventId(0), $"Hello `{name}`", exception, (s, e) => s.ToString());
```
davidfowl, line 314: Not sure this works as is without adding new overloads to Logging to support the builder specifically. Essentially imagine we have a method like this:

```C#
void Log<T>(T state) => Console.WriteLine(state.ToString());
```
It has no idea what T is supposed to be when I do this:

```C#
string name = "David";
Log($"Hello {name}"); // This will call string.Format
```

I assume we'd need an overload like this:

```C#
void Log(CustomLoggerParamsBuilder loggingBuilder) => Console.WriteLine(loggingBuilder.ToString());
```

Now the compiler can perform target typing.

@stephentoub and @333fred

333fred: > `Log($"Hello {name}"); // This will call string.Format`

Well, that would presumably call the `InterpolatedStringBuilder` in the framework, but the end result is the same, it's not your custom builder type. If you want to use your own builder, you need to have an overload that takes it instead of a string.

davidfowl: Yep. The point was to call out that we'd need a new overload.

333fred: Note that if you have both, the builder-type overload should be preferred. So it's not a problem if you have an overload that does take a string.

davidfowl: It's possible to prevent fallback to string with the appropriate methods on the builder as well right?

333fred: > It's possible to prevent fallback to string with the appropriate methods on the builder as well right?

The way to do that would be to not have an overload that takes a string.

proposed/logging-generator.md
```diff
@@ -0,0 +1,621 @@
...
+  "State": {
+    "Message": "Value is 1.234568E+004",
+    "value": 12345.6789,
+    "{OriginalFormat}": "Value is {value:E}"
+  }
+}
+```
+
+
+#### `EmitDefaultMessage` : YES/NO
```
maryamariyan, line 319: I chatted with @shirhatti and @davidfowl offline about this feature and I didn't get an up vote for keeping this feature.

I'll remove it from the proposal since 
(A) It is not an already supported scenario.
(B) The behavior as seen from the console output seems a bit unexpected. 

cc: @geeknoid

maryamariyan:
@omajid as of https://github.com/dotnet/designs/pull/190 this PR is failing, what do I need to do to fix this?

omajid:
Could you run `./update-index > INDEX.md`? CI is running that command too and comparing the `INDEX.md` in this PR with the `INDEX.md` that `update-index` generates.

ericstj:
> Could you run ./update-index > INDEX.md?

That's not so friendly on Windows for all developers.  I tried to run that command with windows bash.exe and got a bunch of errors around line endings.  We should probably come up with a way to make it work regardless of developer platform.

omajid:
> That's not so friendly on Windows for all developers.

Ouch. I didn't realize that. 

> We should probably come up with a way to make it work regardless of developer platform.

I dont use Windows, but I am happy to help make this happen. Any suggestions on what a Windows-friendly workflow would look like?

> I tried to run that command with windows bash.exe and got a bunch of errors around line endings.

Should it just update the file in-place directly? Would that help with the line ending issues?

ericstj:
>> I tried to run that command with windows bash.exe and got a bunch of errors around line endings.
> Should it just update the file in-place directly? Would that help with the line ending issues?

Nevermind, I see this is a python script.  We can just tell folks to use python to run it.  I see a couple path separator issues.  I'll put up a fix.

maryamariyan:
Note to reviewers:

- The API proposal and samples got updated with the approved feedback:https://github.com/dotnet/designs/blob/ced717fc77d218b3793aa68b4ccd3ac3fe59fe50/accepted/2021/logging-generator.md#api-proposal

- Added this section in the document to explain limitations with using `Define` which source generators could handle one way or another: https://github.com/dotnet/designs/blob/ced717fc77d218b3793aa68b4ccd3ac3fe59fe50/accepted/2021/logging-generator.md#added-benefits-with-source-generators-not-supported-with-loggermessagedefine

accepted/2021/logging-generator.md
```diff
@@ -333,6 +332,162 @@ same console logs formatted using JsonConsole:
 }
 ```
 
+### Added benefits with source generators not supported with `LoggerMessage.Define`
```
davidfowl, line 332: I don't really understand this. This feels like a choice. We chose not to add those overloads. If we want to support more than 6 overloads then we need to add those overloads...

maryamariyan: even if there were overloads up to taking 16 parameter (Action limit). then the source generator would still be able to fallback to another approach to support arbitrary number of parameters

I think with this section I was trying to illustrate that source generator can deal with some of the existing limitations of `Define`, one way or another, if we wanted to allow more robust support. (arbitrary number of parameter is one of the several items of limitations explained in this section)

maryamariyan: Are recommending to switch that ability off for preview 4, to only allow up to 6 parameters until we find request for it?

davidfowl: > Are recommending to switch that ability off for preview 4, to only allow up to 6 parameters until we find request for it?

I think so. I'm not sure why the API was rejected TBH. Maybe we're waiting until customers complain 😄

davidfowl:
@maryamariyan time to merge this?

maryamariyan:
> @maryamariyan time to merge this?

It should be ready now

davidfowl (approved):

PR #200, opened 2021-04-10
Replace Python script with a .NET program

terrajobst:
While there is nothing wrong with Python, many of us don't have it installed but all of us have .NET -- for obvious reasons :-)

I've also added two shell scripts, one for Windows and one for Unix, so regardless of OS you can just regenerate the index using the same command:

    $ ./update-index

/cc @omajid


INDEX.md
```diff
@@ -39,9 +39,9 @@ Use update-index to regenerate it:
 | 2020 | [.NET Optional SDK Workloads](accepted/2020/workloads/workloads.md) | [Rich Lander](https://github.com/richlander) |
 | 2020 | [.NET Runtime Form Factors](accepted/2020/form-factors.md) | [Rich Lander](https://github.com/richlander), [Jan Kotas](https://github.com/jkotas) |
 | 2020 | [.NET SDK Workload Manifests](accepted/2020/workloads/workload-manifest.md) | [Mikayla Hutchinson](https://github.com/mhutch) |
-| 2020 | [ASN.1 BER/CER/DER Reader & Writer](accepted/2020/asnreader/asnreader.md) | [Jeremy Barton](https://github.com/bartonjs) |
 | 2020 | [Annotating APIs as unsupported on specific platforms](accepted/2020/platform-exclusion/platform-exclusion.md) | [Immo Landwerth](https://github.com/terrajobst) |
 | 2020 | [Annotating platform-specific APIs and detecting its use](accepted/2020/platform-checks/platform-checks.md) | [Immo Landwerth](https://github.com/terrajobst) |
+| 2020 | [ASN.1 BER/CER/DER Reader & Writer](accepted/2020/asnreader/asnreader.md) | [Jeremy Barton](https://github.com/bartonjs) |
```
terrajobst, line 44: These differences are because the script orders by year and then by title

INDEX.md
```diff
@@ -53,11 +53,11 @@ Use update-index to regenerate it:
...
 | 2020 | [Marking APIs that are unsupported by Blazor WebAssembly](accepted/2020/blazor-unsupported-apis/blazor-unsupported-apis.md) | [Immo Landwerth](https://github.com/terrajobst) |
 | 2020 | [Marking Windows-only APIs](accepted/2020/windows-specific-apis/windows-specific-apis.md) | [Immo Landwerth](https://github.com/terrajobst) |
-| 2020 | [OR_GREATER preprocessor symbols for TFMs](accepted/2020/or-greater-defines/or-greater-defines.md) | [Immo Landwerth](https://github.com/terrajobst) |
+| 2020 | [MSBuild SDK Resolvers and optional workloads in .NET 5](accepted/2020/workloads/workload-resolvers.md) | [Rich Lander](https://github.com/richlander), [Daniel Plaisted](https://github.com/dsplaisted) |
 | 2020 | [Obsoletions in .NET 5](accepted/2020/better-obsoletion/obsoletions-in-net5.md) | [Immo Landwerth](https://github.com/terrajobst) |
+| 2020 | [OR_GREATER preprocessor symbols for TFMs](accepted/2020/or-greater-defines/or-greater-defines.md) | [Immo Landwerth](https://github.com/terrajobst) |
```
terrajobst, line 60: Ditto

tools/update-index/Program.cs
```diff
@@ -0,0 +1,263 @@
...
+    {
+        var designs = Directory.GetFiles(directoryPath, "*.md", SearchOption.AllDirectories)
+                               .Select(p => Document.Parse(p))
+                               .Where(d => d != null)
+                               .Select(d => d!)
+                               .ToArray();
+
+        using var outputWriter = new StreamWriter(outputPath);
+
+        outputWriter.WriteLine("<!--");
```
ericstj, line 97: Not sure the current state of T4 tooling, but this would be more readable as a T4 template.

ericstj (approved):
Thanks for doing this.  While I did have some fun debugging that python script, I prefer the current approach. 😉

tools/update-index/Program.cs
terrajobst: I just tried it. Couple of things:

1. It doesn't work well without VS because the generated T4 file is only updated when you're within VS. If you're doing VS Code and CLI, then I don't even know how you can re-generate the file.
2. Suprisingly T4 seems to work just fine .NET 5
3. Razor is much nicer than T4 because you don't have to declare the transitions between code and markup. In T4 you need to use markers and you need to be very careful with line breaks in Markdown. For prosperity I've included the T4 version below. TBH, I don't find it more readable.

So yeah, I'm inclined to not use T4 here.

```C#
<#@ template language="C#" #>
<#@ assembly name="System.Core" #>
<#@ import namespace="System.Linq" #>
<#@ import namespace="System.Text" #>
<#@ import namespace="System.Collections.Generic" #>
<!--

This file is auto-generated. Direct changes to it maybe lost.

Use update-index to regenerate it:

    ./update-index

-->

# Design Index

## Meta

<#
foreach (var document in GetDocuments(DocumentKind.Meta)) { #>
* <#= GetMarkdownLink(document) #>
<#
}
#>

## Accepted

|Year|Title|Owners|
|----|-----|------|
<#
foreach (var document in GetDocuments(DocumentKind.AcceptedDesign))
{
    var year = document.Year.ToString();
    var link = GetMarkdownLink(document);
    var owners = string.Join(", ", document.Owners);
#>| <#= year #> | <#= link #> | <#= owners #> |
<#
}
#>

## Proposed

|Year|Title|Owners|
|----|-----|------|
<#
foreach (var document in GetDocuments(DocumentKind.ProposedDesign))
{
    var year = document.Year.ToString();
    var link = GetMarkdownLink(document);
    var owners = string.Join(", ", document.Owners);
#>| <#= year #> | <#= link #> | <#= owners #> |
<#
}
#>


```

PR #201, opened 2021-04-12
Use T4 for template

terrajobst:
Based on @ericstj's [suggestion](https://github.com/dotnet/designs/pull/200#discussion_r611797203) I gave T4 a spin:

Findings:

1. It doesn't work well without VS because the generated T4 file is only updated when you're within VS. If you're doing VS Code and CLI, then I don't even know how you can re-generate the file.
    - Update: There seems to be a VS Code plug-in that takes care of that.
2. Surprisingly T4 seems to work just fine in .NET 5
3. Razor is much nicer than T4 because you don't have to declare the transitions between code and markup. In T4 you need to use markers and you need to be very careful with line breaks in Markdown. TBH, I don't find it more readable.
4. Looks like it doesn't handle OS line breaks either :-)

So yeah, I'm inclined to not use T4 here.


ericstj:
I was imagining replacing the entire program with a T4 template and using @mhutch's [tool](https://github.com/mono/t4).  There are other ways to remove the boilerplate which don't take on the new dependency.  I'm fine with whatever you decide.  Thanks for humoring my suggestion.

BradBarnich:
That's strange that you have the preprocessed version in the tree and not the actual output of the template. I immensely dislike t4, but being a template system that is directly supported VS, it gets picked. This is of course a trivial example, but WriteLine doesn't exactly scale either.

Do you have razor branch as well? Might be interesting if it drops the requirement to run the template from VS.

terrajobst:
> That's strange that you have the preprocessed version in the tree and not the actual output of the template

That's by design because I'm using a runtime template (i.e. I don't generate code, I generate a Markdown file). The generation happens at runtime (not at design time). What's generated is the template here.

terrajobst:
> I was imagining replacing the entire program with a T4 template and using @mhutch's [tool](https://github.com/mono/t4). There are other ways to remove the boilerplate which don't take on the new dependency. I'm fine with whatever you decide. Thanks for humoring my suggestion.

I see. I haven't looked at that. If someone feels trigger happy, I'd be curious what this would look like. For now, it seems a good old console app does the trick and seems more maintainable to me.

PR #206, opened 2021-04-16
Clarify workload pack aliasing behavior

mhutch:
Make workload aliasing behavior explicit and enable the behavior needed by https://github.com/dotnet/aspnetcore/pull/31761. Workload reader PR to follow.

### DIFF
--- a/accepted/2020/workloads/workload-manifest.md
+++ b/accepted/2020/workloads/workload-manifest.md
@@ -205,7 +205,7 @@ The `framework` pack kind is used for runtime packs and targeting packs. The wor
 
 A pack definition with the optional `alias-to` key is an *alias pack*. An alias pack has a virtual pack ID that doesn't need to correspond to a real NuGet package. When querying the workload manifest, alias packs resolve to concrete pack IDs in a platform-dependent way. The `alias-to` value is a JSON object, where the keys are host platform RIDs, and the values are NuGet package IDs.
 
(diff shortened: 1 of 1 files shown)


lewing (commented):
This makes sense to me

accepted/2020/workloads/workload-manifest.md
```diff
@@ -218,17 +220,19 @@ For example, an SDK pack might contain a compiler that's a native executable and
 }
 ```
 
-Workload definitions and SDK imports would refer to this as `"foo.sdk.compiler"`, but on Windows the pack that gets installed and resolved would be `"foo.sdk.compiler.windows"`. The `"foo.sdk.compiler.windows"` pack would be present on disk and its ID would be visible in resolved paths, but otherwise transparent and not able to be used directly.
+Workload definitions and SDK imports would refer to this as `foo.sdk.compiler`, but on Windows the pack that gets installed and resolved would be `foo.sdk.compiler.windows`. The `foo.sdk.compiler.windows` pack would be present on disk and its ID would be visible in resolved paths, but otherwise transparent and not able to be used directly. On x64 Macs it would similarly resolve to the concrete `foo.sdk.compiler.mac` pack.
+
+On all other platforms this pack would be a no-op. If workloads that contained this pack were not functional without it, the workload author should restrict the workload itself using `"platforms": [ "osx-x64", "win-x64" ]`.
```
dsplaisted, line 225: What does it mean to be a no-op?  I think it means it would be a no-op on workload install, but what about during workload resolution?  Should the `workloadmanifest.targets` file have conditions that make sure the workload pack isn't imported on platforms where it's not available?

dsplaisted (approved):

accepted/2020/workloads/workload-manifest.md
mhutch: Clarified this in the spec

PR #207, opened 2021-04-17
Improving default string globalization experience

GrabYourPitchforks:
Draft design document for how we can flatten the learning curve for using `string`, removing globalization concepts for developers who needn't be exposed to them, and widening the pit of success for people calling string-based APIs.

__This is a draft design document and does not represent a final plan or committed work.__

See also:

- https://github.com/dotnet/runtime/issues/43956
- https://github.com/dotnet/runtime/issues/30626
- https://github.com/dotnet/runtime/issues/14065

### DIFF
--- a/accepted/2021/default_string_globalization.md
+++ b/accepted/2021/default_string_globalization.md
@@ -1,0 +1,390 @@
+# Improving the _string_ developer experience
+
+Questions? Contact: @tarekgh @safern @GrabYourPitchforks
+
+## Problem summary
+
+APIs on `string` are not consistently ordinal or linguistic, and this leads to confusion among our developer community and bugs in applications. https://github.com/dotnet/runtime/issues/43956 describes this pain in more detail, and it aggregates issues where developers experienced breaks when upgrading from .NET Core 3.1 to .NET 5 due to runtime changes exposing latent bugs in their own applications.
+
+That GitHub issue describes various options we were considering, including relying solely on Roslyn analyzers or providing runtime compatibility switches. None of these issues is very palatable because they either force the developers to think about globalization concerns or they make it difficult for library authors to reason about what behaviors their components will exhibit at runtime.
+
+This draft proposes an alternate solution which attempts to provide the best of all worlds: steering new developers toward success, letting people with existing code bases know when they have issues, and allowing existing compiled DLLs to work largely as expected.
+
+## Characteristics of an ideal solution
+
+An ideal solution to this problem should have two qualities.
+
+1. `string` and related APIs should default to ordinal behavior unless the caller makes an explicit request for a different behavior.
+
+2. Developers should not be exposed to any globalization-related concepts unless they explicitly seek globalization support.
+
+The first characteristic is fairly straightforward. APIs like `string.Compare`, `string.Contains`, `string.Equals`, `string.IndexOf`, and others should all behave in an ordinal fashion. Developers shouldn't have to rely on Roslyn analyzers or MSDN to know what behavior a given API has.
+
+The second characteristic requires a bit more explanation. Popular programming languages like Java, JavaScript, Go, Python, C/C++, and others treat string APIs as ordinal. .NET is unique in that we build globalization concepts directly into many basic `string` APIs. Unfortunately, evidence shows that .NET's behavior in this regard is largely unexpected and confusing to our developer audience.
+
+[Java 16's `String` class](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/String.html) (which is the closest to .NET's `string` class) is a good example of this principle. Per their own documentation:
+
+> Unless otherwise noted, methods for comparing Strings do not take locale into account.
+
+And in fact, the _only_ two Java `String` APIs that are culture-aware are case-conversion routines.
+
+```java
+String s = getString();
+
+// parameterless overloads use current locale by default
+String s1 = s.toUpperCase();
+String s2 = s.toLowerCase();
+
+// or you can pass a Locale instance if you don't want to use the current locale
+// Java's Locale.ROOT is roughly equivalent to .NET's CultureInfo.InvariantCulture
+String s3 = s.toUpperCase(Locale.ROOT);
+String s4 = s.toLowerCase(Locale.ROOT);
+```
+
+And even though these are the only two linguistic-by-default APIs on Java's `String` type, problematic usage of them is still so widespread that CERT's coding rules for Java [forbid calling the parameterless overloads](https://wiki.sei.cmu.edu/confluence/display/java/STR02-J.+Specify+an+appropriate+locale+when+comparing+locale-dependent+data). This CERT rule STR02-J is roughly equivalent to .NET's [CA1310 rule](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1310), which flags certain overloads of `string` APIs as problematic.
+
+Put another way, the second characteristic mentioned above might be paraphrased as "developers shouldn't be exposed to globalization-related concerns unless they have imported the _System.Globalization_ namespace into their file."
+
+## Proposed solution
+
+To solve this problem for the .NET ecosystem at large, we propose to take the following actions on `string`.
+
+1. Obsolete (as warning) all linguistic-by-default methods on `string` with a new __SYSLIBxxxx__ identifier.
+2. Mark all newly-obsoleted methods as `[EditorBrowsable(EditorBrowsableState.Never)]`.
+3. Mark all methods which take a `CultureInfo` parameter as `[EditorBrowsable(EditorBrowsableState.Never)]`, but do not obsolete them.
+4. Add simple ordinal overloads for all obsoleted methods. (These overloads _should not_ take `StringComparison` as a parameter. More on this later.)
+5. Add simple "ignore case" ordinal overloads for all methods which currently don't have them.
+6. Create fixers to migrate call sites from the now-obsoleted methods onto newer methods.
+
+And we propose to take the following actions on `string`-like APIs.
+
+1. Mark all linguistic-related APIs as `[EditorBrowsable(EditorBrowsableState.Never)]`, but do not obsolete them. (Exception: do not mark APIs if they're in the _System.Globalization_ namespace.)
+2. Create analyzers to detect `Comparer<string>.Default`, `new SortedSet<string>()`, and similar code patterns that implicitly rely on the current culture.
+
+### List of API obsoletions and additions
+
+In the diff below, a `-` indicates hiding; and a `+` indicates addition. A subset of `-` APIs are obsoleted, using the pattern described above.
+
+```diff
+ public sealed partial class String : System.Collections.Generic.IEnumerable<char>, System.Collections.IEnumerable, System.ICloneable, System.IComparable, System.IComparable<string?>, System.IConvertible, System.IEquatable<string?>
+ {
+     public static readonly string Empty;
+     [System.CLSCompliantAttribute(false)]
+     public unsafe String(char* value) { }
+     [System.CLSCompliantAttribute(false)]
+     public unsafe String(char* value, int startIndex, int length) { }
+     public String(char c, int count) { }
+     public String(char[]? value) { }
+     public String(char[] value, int startIndex, int length) { }
+     public String(System.ReadOnlySpan<char> value) { }
+     [System.CLSCompliantAttribute(false)]
+     public unsafe String(sbyte* value) { }
+     [System.CLSCompliantAttribute(false)]
+     public unsafe String(sbyte* value, int startIndex, int length) { }
+     [System.CLSCompliantAttribute(false)]
+     public unsafe String(sbyte* value, int startIndex, int length, System.Text.Encoding enc) { }
+     [System.Runtime.CompilerServices.IndexerName("Chars")]
+     public char this[int index] { get { throw null; } }
+     public int Length { get { throw null; } }
+     public object Clone() { throw null; }
+-    public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length) { throw null; }
+-    public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length, bool ignoreCase) { throw null; }
+-    public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length, bool ignoreCase, System.Globalization.CultureInfo? culture) { throw null; }
+-    public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length, System.Globalization.CultureInfo? culture, System.Globalization.CompareOptions options) { throw null; }
+     public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length, System.StringComparison comparisonType) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB, bool ignoreCase) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB, bool ignoreCase, System.Globalization.CultureInfo? culture) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB, System.Globalization.CultureInfo? culture, System.Globalization.CompareOptions options) { throw null; }
++    public static int CompareString(System.String? strA, System.String? strB) { throw null; }
++    public static int CompareStringIgnoreCase(System.String? strA, System.String? strB) { throw null; }
+     public static int Compare(System.String? strA, System.String? strB, System.StringComparison comparisonType) { throw null; }
+     public static int CompareOrdinal(System.String? strA, int indexA, System.String? strB, int indexB, int length) { throw null; }
+     public static int CompareOrdinal(System.String? strA, System.String? strB) { throw null; }
+-    public int CompareTo(object? value) { throw null; }
+-    public int CompareTo(System.String? strB) { throw null; }
++    public int CompareToString(System.String? strB) { throw null; }
++    public int CompareToStringIgnoreCase(System.String? strB) { throw null; }
+     public static System.String Concat(System.Collections.Generic.IEnumerable<string?> values) { throw null; }
+     public static System.String Concat(object? arg0) { throw null; }
+     public static System.String Concat(object? arg0, object? arg1) { throw null; }
+     public static System.String Concat(object? arg0, object? arg1, object? arg2) { throw null; }
+     public static System.String Concat(params object?[] args) { throw null; }
+     public static System.String Concat(System.ReadOnlySpan<char> str0, System.ReadOnlySpan<char> str1) { throw null; }
+     public static System.String Concat(System.ReadOnlySpan<char> str0, System.ReadOnlySpan<char> str1, System.ReadOnlySpan<char> str2) { throw null; }
+     public static System.String Concat(System.ReadOnlySpan<char> str0, System.ReadOnlySpan<char> str1, System.ReadOnlySpan<char> str2, System.ReadOnlySpan<char> str3) { throw null; }
+     public static System.String Concat(System.String? str0, System.String? str1) { throw null; }
+     public static System.String Concat(System.String? str0, System.String? str1, System.String? str2) { throw null; }
+     public static System.String Concat(System.String? str0, System.String? str1, System.String? str2, System.String? str3) { throw null; }
+     public static System.String Concat(params string?[] values) { throw null; }
+     public static System.String Concat<T>(System.Collections.Generic.IEnumerable<T> values) { throw null; }
+     public bool Contains(char value) { throw null; }
+     public bool Contains(char value, System.StringComparison comparisonType) { throw null; }
+     public bool Contains(System.String value) { throw null; }
+     public bool Contains(System.String value, System.StringComparison comparisonType) { throw null; }
+     [System.ComponentModel.EditorBrowsableAttribute(System.ComponentModel.EditorBrowsableState.Never)]
+     [System.ObsoleteAttribute("This API should not be used to create mutable strings. See https://go.microsoft.com/fwlink/?linkid=2084035 for alternatives.")]
+     public static System.String Copy(System.String str) { throw null; }
+     public void CopyTo(int sourceIndex, char[] destination, int destinationIndex, int count) { }
+     public static System.String Create<TState>(int length, TState state, System.Buffers.SpanAction<char, TState> action) { throw null; }
+     public bool EndsWith(char value) { throw null; }
++    public bool EndsWithIgnoreCase(char value) { throw null; }
+-    public bool EndsWith(System.String value) { throw null; }
+-    public bool EndsWith(System.String value, bool ignoreCase, System.Globalization.CultureInfo? culture) { throw null; }
+     public bool EndsWith(System.String value, System.StringComparison comparisonType) { throw null; }
++    public bool EndsWithString(System.String value) { throw null; }
++    public bool EndsWithStringIgnoreCase(System.String value) { throw null; }
+     public System.Text.StringRuneEnumerator EnumerateRunes() { throw null; }
+     public override bool Equals([System.Diagnostics.CodeAnalysis.NotNullWhenAttribute(true)] object? obj) { throw null; }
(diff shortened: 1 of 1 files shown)


accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+
+And even though these are the only two linguistic-by-default APIs on Java's `String` type, problematic usage of them is still so widespread that CERT's coding rules for Java [forbid calling the parameterless overloads](https://wiki.sei.cmu.edu/confluence/display/java/STR02-J.+Specify+an+appropriate+locale+when+comparing+locale-dependent+data). This CERT rule STR02-J is roughly equivalent to .NET's [CA1310 rule](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1310), which flags certain overloads of `string` APIs as problematic.
+
+Put another way, the second characteristic mentioned above might be paraphrased as "developers shouldn't be exposed to globalization-related concerns unless they have imported the _System.Globalization_ namespace into their file."
+
+## Proposed solution
+
+To solve this problem for the .NET ecosystem at large, we propose to take the following actions on `string`.
+
+1. Obsolete (as warning) all linguistic-by-default methods on `string` with a new __SYSLIBxxxx__ identifier.
```
tannergooding, line 52: I'd really like to see 1. be `Create fixers to migrate call sites` and 2. to be `Obsolete (as warning)`

Having a fixer is critical IMO and the most important part considering how widespread the existing methods are.

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+-    Ordinal = 4,
+-    OrdinalIgnoreCase = 5,
++    Normal = 4, // aliases Ordinal
++    IgnoreCase = 5, // aliases OrdinalIgnoreCase
+ }
+```
+
+## A note on `StringComparison` and `StringComparer`
+
+One notable aspect of this proposal is the complete restructuring of the `StringComparison` and `StringComparer` types. It is important to note that all members of these types will be preserved. The proposal here is to hide all existing fields / static properties and to create new "simpler" names _Normal_ and _IgnoreCase_. Both of these additions alias their ordinal siblings.
```
tarekgh, line 316: May be it is worth mentioning whoever want to create a linguistic StringComparer, can call the extension method `CompareInfo.GetStringComparer`. This will be explicit as it will be created from the CompareInfo which show the intend of using the linguistic functionality.

tarekgh (approved):

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+ public enum StringComparison
+ {
+-    CurrentCulture = 0,
+-    CurrentCultureIgnoreCase = 1,
+-    InvariantCulture = 2,
+-    InvariantCultureIgnoreCase = 3,
+-    Ordinal = 4,
+-    OrdinalIgnoreCase = 5,
++    Normal = 4, // aliases Ordinal
++    IgnoreCase = 5, // aliases OrdinalIgnoreCase
```
tannergooding, line 310: What's the replacement for `CurrentCulture` and `InvariantCulture`?

How will this impact cases like `CompareOptions.NumericOrdering`: <https://github.com/dotnet/runtime/issues/13979#issuecomment-297115553>?

How will this impact cases like `RemoveStart` and `RemoveEnd`: <https://github.com/dotnet/runtime/issues/14386>

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+
+1. `string` and related APIs should default to ordinal behavior unless the caller makes an explicit request for a different behavior.
+
+2. Developers should not be exposed to any globalization-related concepts unless they explicitly seek globalization support.
+
+The first characteristic is fairly straightforward. APIs like `string.Compare`, `string.Contains`, `string.Equals`, `string.IndexOf`, and others should all behave in an ordinal fashion. Developers shouldn't have to rely on Roslyn analyzers or MSDN to know what behavior a given API has.
+
+The second characteristic requires a bit more explanation. Popular programming languages like Java, JavaScript, Go, Python, C/C++, and others treat string APIs as ordinal. .NET is unique in that we build globalization concepts directly into many basic `string` APIs. Unfortunately, evidence shows that .NET's behavior in this regard is largely unexpected and confusing to our developer audience.
+
+[Java 16's `String` class](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/String.html) (which is the closest to .NET's `string` class) is a good example of this principle. Per their own documentation:
```
danmoseley, line 25: It's good that we're looking to learn from other ecosystems. Is there anything we can learn from the other ecosystems you mention above, beyond your note that they separate string from globalization? (Go, Rust etc)

accepted/2021/default_string_globalization.md
tarekgh: To use CompareInfo instead for the operation. e.g. `CultureInfo.CurrentCulture.CompareInfo.Compare(...)`. I think we need to revisit [https://github.com/dotnet/runtime/issues/13979](https://github.com/dotnet/runtime/issues/13979) or at least will make NumericOrdering is cultural independent. 

---
In reply to: [615181346](https://github.com/dotnet/designs/pull/207#discussion_r615181346) [](ancestors = 615181346)

accepted/2021/default_string_globalization.md
GrabYourPitchforks: Oh, absolutely. Most other ecosystems also enforce opinionated behaviors regarding UTF-8 string well-formedness. But that continues to be a point of contention for us. ;)

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+4. Add simple ordinal overloads for all obsoleted methods. (These overloads _should not_ take `StringComparison` as a parameter. More on this later.)
+5. Add simple "ignore case" ordinal overloads for all methods which currently don't have them.
+6. Create fixers to migrate call sites from the now-obsoleted methods onto newer methods.
+
+And we propose to take the following actions on `string`-like APIs.
+
+1. Mark all linguistic-related APIs as `[EditorBrowsable(EditorBrowsableState.Never)]`, but do not obsolete them. (Exception: do not mark APIs if they're in the _System.Globalization_ namespace.)
+2. Create analyzers to detect `Comparer<string>.Default`, `new SortedSet<string>()`, and similar code patterns that implicitly rely on the current culture.
+
+### List of API obsoletions and additions
```
marek-safar, line 64: Are we going to release an updated version of netstandard or are the developers targetting multiple versions going to be faced with warnings everywhere?

Happypig375:
Can't we have e.g. `CompareToNormal` as opposed to `CompareToString` to align with members in `StringComparison` and `StringComparator`?

Happypig375:
With the naming convention above, we can omit repeating `String` in the method names and have `xxxNormal` and `xxxIgnoreCase` instead of `xxx` for `char` overloads, `xxxIgnoreCase` for `char` overloads, `xxxString`, and `xxxStringIgnoreCase`.

Happypig375:
Maybe we can even provide `char` overloads for `xxxNormal` for uniformity.

Happypig375:
Also `ContainsIgnoreCase` is missing.

accepted/2021/default_string_globalization.md
tarekgh: I think the challenge here is we are going to have a new APIs which will be in .NET 6.0 and up. I don't think we can add these to the netstandard. 

CC @terrajobst

tarekgh:
> Can't we have e.g. CompareToNormal as opposed to CompareToString to align with members in StringComparison and StringComparer?

The issue here is if you are developer new to .NET, would the name `CompareToNormal` would be better than `CompareToString`? The term `Normal` is the best we can find to fit in `StringComparison` but we welcome any better names if there is any.

accepted/2021/default_string_globalization.md
GrabYourPitchforks: These wouldn't be in standard. If you're multi-targeting, I believe per the way this proposal is written, the only methods that would be marked `[Obsolete]` are methods that the code analyzers CA1307 and CA1310 would have flagged anyway as likely incorrect usage. A significant number of methods would be marked EB-never but wouldn't produce warnings.

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+
+## Characteristics of an ideal solution
+
+An ideal solution to this problem should have two qualities.
+
+1. `string` and related APIs should default to ordinal behavior unless the caller makes an explicit request for a different behavior.
+
+2. Developers should not be exposed to any globalization-related concepts unless they explicitly seek globalization support.
+
+The first characteristic is fairly straightforward. APIs like `string.Compare`, `string.Contains`, `string.Equals`, `string.IndexOf`, and others should all behave in an ordinal fashion. Developers shouldn't have to rely on Roslyn analyzers or MSDN to know what behavior a given API has.
```
akoeplinger, line 21: MSDN doesn't exist anymore :)

```suggestion
The first characteristic is fairly straightforward. APIs like `string.Compare`, `string.Contains`, `string.Equals`, `string.IndexOf`, and others should all behave in an ordinal fashion. Developers shouldn't have to rely on Roslyn analyzers or docs.microsoft.com to know what behavior a given API has.
```

accepted/2021/default_string_globalization.md
tannergooding: I think an important thing to keep in mind about scenarios like `Rust` is that they design most of their types around the principle of "zero cost abstraction": <https://doc.rust-lang.org/book/ch00-00-introduction.html#people-who-value-speed-and-stability>
> Rust is for people who crave speed and stability in a language. By speed, we mean the speed of the programs that you can create with Rust and the speed at which Rust lets you write them. The Rust compiler’s checks ensure stability through feature additions and refactoring. This is in contrast to the brittle legacy code in languages without these checks, which developers are often afraid to modify. By striving for zero-cost abstractions, higher-level features that compile to lower-level code as fast as code written manually, Rust endeavors to make safe code be fast code as well.

So while things like `str` are always sliced to valid UTF-8 (unless using one of the `unchecked` slicing methods), they ensure the relevant other components are exposed to allow working with and dealing with "raw parts", that is the `bytes` rather than `runes` (to put it into .NET terms) so power users have the ability to do things themselves when and as appropriate.

Likewise, I don't believe rust has a case insensitive comparer in `std` today. You need to pull it in through an external crate (basically `nupkg`) and specify it as the comparer to use in your `match` expressions (in this way, its more like `IComparer`, but where the implementation is a `struct` and done via generic specialization so its "zero cost" and inlining/etc happen).

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+-    public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length, bool ignoreCase, System.Globalization.CultureInfo? culture) { throw null; }
+-    public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length, System.Globalization.CultureInfo? culture, System.Globalization.CompareOptions options) { throw null; }
+     public static int Compare(System.String? strA, int indexA, System.String? strB, int indexB, int length, System.StringComparison comparisonType) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB, bool ignoreCase) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB, bool ignoreCase, System.Globalization.CultureInfo? culture) { throw null; }
+-    public static int Compare(System.String? strA, System.String? strB, System.Globalization.CultureInfo? culture, System.Globalization.CompareOptions options) { throw null; }
++    public static int CompareString(System.String? strA, System.String? strB) { throw null; }
```
safern, line 99: It feels weird that `string` will be two times in the sentence when calling these APIs (`string.CompareString`) I guess there is no better naming/options?... I definitely like this better than including `Ordinal` in the name though.

safern (approved):

accepted/2021/default_string_globalization.md
GrabYourPitchforks: Downside is that the good names were taken by the method we want to steer people away from. :(

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+
+This draft proposes an alternate solution which attempts to provide the best of all worlds: steering new developers toward success, letting people with existing code bases know when they have issues, and allowing existing compiled DLLs to work largely as expected.
+
+## Characteristics of an ideal solution
+
+An ideal solution to this problem should have two qualities.
+
+1. `string` and related APIs should default to ordinal behavior unless the caller makes an explicit request for a different behavior.
+
+2. Developers should not be exposed to any globalization-related concepts unless they explicitly seek globalization support.
```
stephentoub, line 19: Is this claim specific to APIs on String?  For example, how does this relate to formatting/parsing APIs on types like Int32, DateTime, etc., which all factor in the current culture by default?

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+1. Obsolete (as warning) all linguistic-by-default methods on `string` with a new __SYSLIBxxxx__ identifier.
+2. Mark all newly-obsoleted methods as `[EditorBrowsable(EditorBrowsableState.Never)]`.
+3. Mark all methods which take a `CultureInfo` parameter as `[EditorBrowsable(EditorBrowsableState.Never)]`, but do not obsolete them.
+4. Add simple ordinal overloads for all obsoleted methods. (These overloads _should not_ take `StringComparison` as a parameter. More on this later.)
+5. Add simple "ignore case" ordinal overloads for all methods which currently don't have them.
+6. Create fixers to migrate call sites from the now-obsoleted methods onto newer methods.
+
+And we propose to take the following actions on `string`-like APIs.
+
+1. Mark all linguistic-related APIs as `[EditorBrowsable(EditorBrowsableState.Never)]`, but do not obsolete them. (Exception: do not mark APIs if they're in the _System.Globalization_ namespace.)
```
stephentoub, line 61: Related to my previous comment, how does this relate to methods that implement interfaces or override virtuals? e.g. Int32.ToString() factoring in the current current.  The point earlier talked about not needing to understand globalization concepts... was that too broad, and this is all really just about linguistics-related globalization concepts?

accepted/2021/default_string_globalization.md
stephentoub: I don't understand why we would make this so much harder when you do want CurrentCulture{IgnoreCase}.  The concern expressed in this doc is around the defaults and the inconsistency with defaults we currently expose.  But with StringComparison/StringComparer, a developer is explicitly specifying which they want.  I don't see only downside to changing this.

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+
+So when rules like [__CA1307__](https://docs.microsoft.com/dotnet/fundamentals/code-analysis/quality-rules/ca1307) (pass an explicit `StringComparison` value) are enforced, the developer is now forced to choose between ordinal, invariant, or current-culture. It is not reasonable to expect the average developer to understand the difference between ordinal and linguistic behavior. But it _is_ reasonable for an average developer to think to themselves, "I was told _invariant_ was correct in all other situations where I encountered it. It must also be the appropriate default here.". (It would further be reasonable that this decision may be made via muscle memory, with the developer never consciously realizing that they were presented a choice.)
```
stephentoub, line 322: I don't see the argument for hiding CurrentCulture/CurrentCultureIgnoreCase.  If someone explicitly types "CurrentCulture", that's what they get; it's very clear.

accepted/2021/default_string_globalization.md
```diff
@@ -0,0 +1,390 @@
...
+
+Finally, we cannot rely on analyzers and fixers alone to detect when a linguistic, current-culture call is about to be made. If a `string` instance is passed around as `object` and some consumer has code which reads `if (obj is IComparable comparable) { comparable.CompareTo(...); }`, there's not much we can do about this. If it's too unpalatable to change the underlying behavior of `string` itself, then the best we can do is direct people to use a modern API surface and generic collections, avoiding non-generic methods like `Array.Sort(Array)`.
+
+## Future work and ecosystem implications
+
+This work has implications for GUI stacks, such as WPF and WinForms. GUI scenarios often involve performing linguistic, current-culture collation of data, such as alphabetically sorting entries in a dropdown list. It should be expected that WinForms, WPF, and similar code bases will be disproportionately hit with warnings. They have a few options available to them. In descending order of preference:
```
stephentoub, line 386: > similar code bases will be disproportionately hit with warnings

Not only hit with warnings, but there's a good chance pushing them to use ordinal behavior will actually introduce bugs.

GrabYourPitchforks: That's the weird part about all of this. The rule of thumb is: if you're sorting data for display directly in client UI, the default (_current culture_) behavior is probably appropriate. If you're on a server, a current culture comparison is probably not appropriate. So it largely depends on the context.

However, there is one universal truth: __the pattern `if (string.Compare(...) == 0)` is almost certainly a bug__, regardless of context. There are exceptions, sure, but they're very rare.

Maybe the answer is that we don't hide the original methods, but we instead query what you're doing with the result. If we see a comparison result `== 0` or `!= 0` we flag it. This will catch fewer misusages of the comparison APIs, but experience tells me that the warnings will be more actionable / of higher quality.

accepted/2021/default_string_globalization.md
GrabYourPitchforks: Perhaps I should clarify this. The concept of "ordinal" only applies to APIs which operate on or transform strings / sequences of chars. (Think: `ReadOnlySpan<char>` APIs in `MemoryExtensions`.)

For APIs which are about changing representation between string and non-string types (e.g., _string_ \<-\> _int_), I don't suggest any behavioral change. These APIs necessarily have to perform some kind of linguistic translation, so they may as well fall back to _current culture_ as they do today. That also helps solidify `Parse` and `ToString` as helpers for "please convert this data from / to a representation which is understandable by the current user."

stephentoub: > The concept of "ordinal" only applies to APIs which operate on or transform strings / sequences of chars.

Yes.  I was commenting on the second goal/requirement: "Developers should not be exposed to any globalization-related concepts unless they explicitly seek globalization support."  There are many globalization-related concepts unrelated to ordinal.

accepted/2021/default_string_globalization.md
terrajobst: I kind with @stephentoub. Having `Normal` and `IgnoreCase` present on `StringComparer` seems sufficient as it maps to the methods. Using the enum/comparer is already the advanced mode, so hiding the other concepts feels a bit like we're trying too hard to be helpful.

accepted/2021/default_string_globalization.md
terrajobst: Yeah this is the thing that urks me. Our goal is to present the developer with a simpler view of the world, but the new names look fairly misplaced and odd :-(

terrajobst:
@GrabYourPitchforks should this be marked with api-ready-for-review so we can discuss this?

danmoseley:
I expect @stephentoub would want to have a chance to be part of API review before anything's definitely decided, but no doubt helpful to start discussions.

GrabYourPitchforks:
@terrajobst I suspect if we wanted to have a public discussion, it should be a design-specific discussion with no expectation that we would finalize an API immediately. We'd want Steve and usability experts to be involved prior to finalization. This document also suggests usability testing to help test our theories.

accepted/2021/default_string_globalization.md
iSazonov: In PowerShell repo we follow simple rule - "use explicitly culture/ordinal/ignorecase everywhere".
You will find that this code is full of explicit Ordinal and OrdinalIgnorecase names
This makes the code incredibly understandable and maintainable.
Despite this, we've fixed a few case of unexpected use of culture. So the PR with new conception is great!

But the new names looks weird for me too.
As I said, having Ordinal and OrdinalIgnorecase names in PowerShell code makes it better in my opinion.
Notice, we already have CompareOrdinal(). Why not follow this pattern?
```
CompareToString -> CompareToOrdinal
CompareToStringIgnoreCase -> CompareToOrdinalIgnoreCase
```
And this is completely consistent with the new concept. 
It also makes it easier to learn and understand for people coming from other languages.

KevinCathcart:
While I appreciate the goal of having it be easy to do the correct thing and avoid accidental use of lingusitic operations, I have a lot of concerns here about ergonomics, and poor user experience. You already mention the new developer using an old tutotrial and seeing warnings. But what about experienced developers who move to a newer .NET and are confused/frustrated about overloads they know exist (and are use to using, because they are the "correct" way to do things as of right now) don't show up in Intellisense because they have been made EBNever. 

I do have to wonder if for application code, users would not be better served by have an easy way to set the Default culture of all threads to some special variant of the Invariant culture that does ordinal comparisons/equality etc instead of the unicode default algorithms. This would actually make the existing APIs generally do the right thing, and would not not be seriously regressing things for people creating GUI apps, Sadly such users will always need to decide for each case if they want a linguistic operation, or a ordinal one. This design doc says "It is not reasonable to expect the average developer to understand the difference between ordinal and linguistic behavior", but for GUI app developers kind of do need to understand. For things like parsing file formats, they often want ordinal, but for things related to user input, they often want linguistic.

Obviously libraries would still need to avoid accidentally using linguistic methods if they really want ordinal, unless the libraries will only be used by programs that use this special default culture.

mhutch:
I love the goal of this proposal but the new `*String` method names seem odd/ugly. Given the prevalence of these methods, would it be worth considering more radical methods to "fix" the current API e.g. runtime changes to enable assembly-level default comparison?

GrabYourPitchforks:
> I do have to wonder if for application code, users would not be better served by have an easy way to set the Default culture of all threads to some special variant of the Invariant culture that does ordinal comparisons/equality etc instead of the unicode default algorithms. This would actually make the existing APIs generally do the right thing, and would not not be seriously regressing things for people creating GUI apps,

We considered but rejected this proposal. The reason is that it would affect more fundamental APIs like `int.ToString`, which we generally _do_ want to match to the user's current locale by default. It would also significantly impact people writing GUI apps, who rely on things culture-aware sorting when displaying items in a list box.

mrange:
I am positive towards increasing the pit of success when it comes to string APIs. No developers expect strings API to behave differently depending on global thread static variable (current culture) until the first long-night debugging session due to a localization bug.

I am concerned about the string interpolation feature in C# and F# and how that can be made less dependent on current culture. I realize there's a `FormattableString.Invariant()` method but I doubt I get developers to type : `FormattableString.Invariant($"Hello {x}")` over `$"Hello {x}"` as large part of the charm with string interpolation is the succinctness which is then lost.

If possible, I would like to see an optional analyzer that warns for all usages string interpolation without `FormattableString.Invariant` so it's at least easy to find all the places.

huoyaoyuan:
> The reason is that it would affect more fundamental APIs like `int.ToString`, which we generally _do_ want to match to the user's current locale by default.

Being culture aware in those members are so problematic:

- `int.ToString` or `int.Parse` can't be optimized further since they have to touch current culture, or access `InvariantCulture`. Converting between `int` and `string` very likely to be used frequently. Passing `CultureInfo.InvariantCulture` or `NumberFormatInfo.InvariantInfo` is even more annoying. `double` is more impactful.
- Round-tripping of `DataTime` is a nightmare. Formatting in ddmmyy culture and parsing in mmddyy culture can easily lead to bugs, which I have already heard about.
- Usage of `IEquatable<string>` is merely unavoidable, especially for tuples and records. Manually making strings ordinal in them will lose almost all the convenience of them.

With my personal bias, as a Chinese speaker, casing and sorting are totally nothing in Chinese. *The only thing I benefit from being culture aware is displaying DateTime.*

danmoseley:
We will soon be planning for .NET 8. Is it an appropriate time to pick this up again?

drewnoakes:
Pinging @dotnet/project-system as we are considering modernising the resource editor in Visual Studio.

Smaug123:
Minor suggestion that doesn't change anything fundamental about the discussion: how about `String.HasPrefix` and `String.HasSuffix` rather than `{Starts,Ends}WithIgnoreCase`? English conveniently has (at least) two terse ways to express "B starts with A".

PR #209, opened 2021-04-26
Fix type of description field

lewing:


lewing:
I guess the trailing commas were possibly intended it just hurts my eyes.  Happy to revert that part.

PR #211, opened 2021-04-29
Metrics Design

tarekgh:
https://github.com/dotnet/runtime/issues/44445

### DIFF
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -32,7 +32,7 @@ There are two types of instruments:
 
 Meter is the factory type responsible for creating Instruments. Meter will have a name and optional version strings.
 
-### Listener
+### MeterListener
 
 The listener is the type that allows listening to the measurements reported by instruments (e.g. Counter, ObservableCounter, etc.). The listener is an important class that will be used by OpenTelemetry to implement the Metrics SDK.
 
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -216,7 +216,7 @@ namespace System.Diagnostics.Metrics
 
         protected void RecordMeasurement(
                             T measurement,
-                            KeyValuePair<string, object?> tag1) { throw null; }
+                            KeyValuePair<string, object?> tag) { throw null; }
 
         protected void RecordMeasurement(
                             T measurement,
@@ -305,7 +305,7 @@ namespace System.Diagnostics.Metrics
     {
         public void Add(T measurement) { throw null; }
         public void Add(T measurement,
-                            KeyValuePair<string, object?> tag1) { throw null; }
+                            KeyValuePair<string, object?> tag) { throw null; }
         public void Add(T measurement,
                             KeyValuePair<string, object?> tag1,
                             KeyValuePair<string, object?> tag2) { throw null; }
@@ -329,7 +329,7 @@ namespace System.Diagnostics.Metrics
         public void Record(T measurement) { throw null; }
         public void Record(
                         T measurement,
-                        KeyValuePair<string, object?> tag1) { throw null; }
+                        KeyValuePair<string, object?> tag) { throw null; }
         public void Record(
                         T measurement,
                         KeyValuePair<string, object?> tag1,
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -463,7 +463,7 @@ namespace System.Diagnostics.Metrics
     Meter meter = new Meter("io.opentelemetry.contrib.mongodb", "v1.0");
     Counter<int> counter = meter.CreateCounter<int>("Requests");
     counter.Add(1);
-    counter.Add(1, KeyValuePair<string, object>("request", "read"));
+    counter.Add(1, KeyValuePair.Create<string, object>("request", "read"));
 ```
 
 ### Listening Example
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -212,7 +212,7 @@ namespace System.Diagnostics.Metrics
         /// Record measurement overloads allowing passing different numbers of tags.
         /// </summary>
 
-        protected void RecordMeasurement(T val) { throw null; }
+        protected void RecordMeasurement(T measurement) { throw null; }
 
         protected void RecordMeasurement(
                             T measurement,
@@ -314,7 +314,7 @@ namespace System.Diagnostics.Metrics
                             KeyValuePair<string, object?> tag2,
                             KeyValuePair<string, object?> tag3) { throw null; }
         public void Add(T measurement,
-                            ReadOnlySpan<<string, object?>> tags) { throw null; }
+                            ReadOnlySpan<KeyValuePair<string, object?>> tags) { throw null; }
         public void Add(T measurement,
                             params KeyValuePair<string, object?>[] tags) { throw null; }
     }
@@ -340,7 +340,7 @@ namespace System.Diagnostics.Metrics
                         KeyValuePair<string, object?> tag2,
                         KeyValuePair<string, object?> tag3) { throw null; }
         public void Record(T measurement,
-                            ReadOnlySpan<<string, object?>> tags) { throw null; }
+                            ReadOnlySpan<KeyValuePair<string, object?>> tags) { throw null; }
         public void Record(
                         T measurement,
                         params KeyValuePair<string, object?>[] tags) { throw null; }
@@ -405,7 +405,7 @@ namespace System.Diagnostics.Metrics
     /// The listener class can be used to listen to observable and non-observable instrument
     /// recorded measurements.
     /// </summary>
-    public class MeterListener : IDisposable
+    public sealed class MeterListener : IDisposable
     {
         /// <summary>
         /// Simple constructor
@@ -432,8 +432,9 @@ namespace System.Diagnostics.Metrics
 
         /// <summary>
         /// Stop listening to a specific instrument measurement recording.
+        /// returns the associated state.
         /// </summary>
-        public void DisableMeasurementEvents(Instrument instrument) { throw null; }
+        public object? DisableMeasurementEvents(Instrument instrument) { throw null; }
 
         /// <summary>
         /// Set a callback for a specific numeric type to get the measurement recording notification
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -191,13 +191,13 @@ namespace System.Diagnostics.Metrics
 }
 ```
 
-### Non-observable Instrument Base Class
+### Instrument Base Class
 
 ```csharp
 namespace System.Diagnostics.Metrics
 {
     /// <summary>
-    /// Instrument<T> is the base class from which all non-observable instruments will inherit from.
+    /// Instrument<T> is the base class from which all instruments that report measurements in the context of the request will inherit from.
     /// Mainly it will support the CLS compliant numerical types.
     /// </summary>
     public abstract class Instrument<T> : Instrument where T : unmanaged
@@ -243,7 +243,7 @@ namespace System.Diagnostics.Metrics
 {
     /// <summary>
     /// ObservableInstrument<T> is the base class from which all observable instruments will inherit from.
-    /// It will only support the CLS compliant numerical types
+    /// It will only support the CLS compliant numerical types.
     /// </summary>
     public abstract class ObservableInstrument<T> : Instrument where T : unmanaged
     {
@@ -261,7 +261,7 @@ namespace System.Diagnostics.Metrics
         /// </summary>
         protected abstract IEnumerable<Measurement<T>> Observe();
 
-        public override bool IsObservable => throw null;
+        public override bool IsObservable => true;
     }
 }
 ```
@@ -292,13 +292,13 @@ namespace System.Diagnostics.Metrics
 }
 ```
 
-### Non-observable Instruments
+### Instruments Concrete Classes
 
 ```csharp
 namespace System.Diagnostics.Metrics
 {
     /// <summary>
-    /// The counter is a non-observable Instrument that supports non-negative increments.
+    /// The counter is an Instrument that supports non-negative increments.
     /// e.g. Number of completed requests.
     /// </summary>
     public sealed class Counter<T> : Instrument<T> where T : unmanaged
@@ -320,7 +320,7 @@ namespace System.Diagnostics.Metrics
     }
 
     /// <summary>
-    /// The histogram is a non-observable Instrument that can be used to report arbitrary values
+    /// The histogram is an Instrument that can be used to report arbitrary values
     /// that are likely to be statistically meaningful. It is intended for statistics such as the request duration.
     /// e.g. the request duration.
     /// </summary>
@@ -402,7 +402,7 @@ namespace System.Diagnostics.Metrics
 
 
     /// <summary>
-    /// The listener class can be used to listen to observable and non-observable instrument
+    /// The listener class can be used to listen to kinds of instruments.
     /// recorded measurements.
     /// </summary>
     public sealed class MeterListener : IDisposable
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -425,7 +425,7 @@ namespace System.Diagnostics.Metrics
                             Instrument instrument,
                             T measurement,
                             ReadOnlySpan<KeyValuePair<string, object?>> tags,
-                            object? cookie);
+                            object? state);
 
 
     /// <summary>
@@ -455,7 +455,7 @@ namespace System.Diagnostics.Metrics
         /// <summary>
         /// Start listening to a specific instrument measurement recording.
         /// </summary>
-        public void EnableMeasurementEvents(Instrument instrument, object? cookie = null) { throw null; }
+        public void EnableMeasurementEvents(Instrument instrument, object? state = null) { throw null; }
 
         /// <summary>
         /// Stop listening to a specific instrument measurement recording.
@@ -503,7 +503,7 @@ namespace System.Diagnostics.Metrics
             listener.EnableMeasurementEvents(instrument, null);
         }
     };
-    listener.SetMeasurementEventCallback<int>((instrument, measurement, tags, cookie) =>
+    listener.SetMeasurementEventCallback<int>((instrument, measurement, tags, state) =>
     {
         Console.WriteLine($"Instrument: {instrument.Name} has recorded the measurement {measurement}");
     });
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -422,7 +422,6 @@ namespace System.Diagnostics.Metrics
         /// this can happen when the Meter or the Listener is disposed of. Or calling Stop()
         /// on the listener.
         /// </summary>
-        // This need some clarification
         public Action<Instrument, object?>? MeasurementsCompleted { get; set; }
 
         /// <summary>
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -25,7 +25,7 @@ The Instrument is the type that will be used by the app and library authors to r
 
 There are two types of instruments:
 
-- The first type we'll call it just `Instrument` for simplicity. These instruments are called inside a request, meaning they have an associated distributed Context (with Span, Baggage, etc.). OpenTelemetry specs call this type of instrument a synchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here proposes three instrument classes of that type: `Counter` and `Histogram`.
+- The first type we'll call it just `Instrument` for simplicity. These instruments are called inside a request, meaning they have an associated distributed Context (with Span, Baggage, etc.). OpenTelemetry specs call this type of instrument a synchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here proposes two instrument classes of that type: `Counter` and `Histogram`.
 - The second type is called `ObservableInstrument` which reports measurements by a callback, once per collection interval, and lacks Context. OpenTelemetry specs call this type of instrument an asynchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here is proposing three instrument classes of that type: `ObservableCounter`, `ObservableGauge`, and `ObservableUpDownCounter`.
 
 ### Meter
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -220,22 +220,22 @@ namespace System.Diagnostics.Metrics
         protected void RecordMeasurement(T val) { throw null; }
 
         protected void RecordMeasurement(
-                            T val,
+                            T measurement,
                             KeyValuePair<string, object?> tag1) { throw null; }
 
         protected void RecordMeasurement(
-                            T val,
+                            T measurement,
                             KeyValuePair<string, object?> tag1,
                             KeyValuePair<string, object?> tag2) { throw null; }
 
         protected void RecordMeasurement(
-                            T val,
+                            T measurement,
                             KeyValuePair<string, object?> tag1,
                             KeyValuePair<string, object?> tag2,
                             KeyValuePair<string, object?> tag3) { throw null; }
 
         protected void RecordMeasurement(
-                            T val,
+                            T measurement,
                             ReadOnlySpan<KeyValuePair<string, object?>> tags) { throw null; }
     }
 }
@@ -278,7 +278,7 @@ namespace System.Diagnostics.Metrics
 namespace System.Diagnostics.Metrics
 {
     /// <summary>
-    /// A measurement stores one observed value and its associated tags. This type is used by Observable instruments' Observe() method when reporting current measurements. 
+    /// A measurement stores one observed value and its associated tags. This type is used by Observable instruments' Observe() method when reporting current measurements.
     /// with the associated tags.
     /// </summary>
     public struct Measurement<T> where T : unmanaged
@@ -306,7 +306,7 @@ namespace System.Diagnostics.Metrics
     /// The counter is a non-observable Instrument that supports non-negative increments.
     /// e.g. Number of completed requests.
     /// </summary>
-    public class Counter<T> : Instrument<T> where T : unmanaged
+    public sealed class Counter<T> : Instrument<T> where T : unmanaged
     {
         public void Add(T measurement) { throw null; }
         public void Add(T measurement,
@@ -328,7 +328,7 @@ namespace System.Diagnostics.Metrics
     /// UpDownCounter is a non-observable Instrument that supports increments and decrements.
     /// e.g. Number of items in a queue.
     /// </summary>
-    public class UpDownCounter<T> : Instrument<T> where T : unmanaged
+    public sealed class UpDownCounter<T> : Instrument<T> where T : unmanaged
     {
         public void Add(T measurement) { throw null; }
         public void Add(T measurement,
@@ -351,7 +351,7 @@ namespace System.Diagnostics.Metrics
     /// that are likely to be statistically meaningful. It is intended for statistics such
     /// e.g. the request duration.
     /// </summary>
-    public class Histogram<T> : Instrument<T> where T : unmanaged
+    public sealed class Histogram<T> : Instrument<T> where T : unmanaged
     {
         public void Record(T measurement) { throw null; }
         public void Record(
@@ -386,7 +386,7 @@ namespace System.Diagnostics.Metrics
     /// when the instrument is being observed.
     /// e.g. CPU time (for different processes, threads, user mode or kernel mode).
     /// </summary>
-    public class ObservableCounter<T> : ObservableInstrument<T> where T : unmanaged
+    public sealed class ObservableCounter<T> : ObservableInstrument<T> where T : unmanaged
     {
         protected override IEnumerable<Measurement<T>> Observe() { throw null; }
     }
@@ -396,7 +396,7 @@ namespace System.Diagnostics.Metrics
     /// when the instrument is being observed.
     /// e.g. the process heap size
     /// </summary>
-    public class ObservableUpDownCounter<T> : ObservableInstrument<T> where T : unmanaged
+    public sealed class ObservableUpDownCounter<T> : ObservableInstrument<T> where T : unmanaged
     {
         protected override IEnumerable<Measurement<T>> Observe() { throw null; }
     }
@@ -406,7 +406,7 @@ namespace System.Diagnostics.Metrics
     /// when the instrument is being observed.
     /// e.g. the current room temperature
     /// </summary>
-    public class ObservableGauge<T> : ObservableInstrument<T> where T : unmanaged
+    public sealed class ObservableGauge<T> : ObservableInstrument<T> where T : unmanaged
     {
         protected override IEnumerable<Measurement<T>> Observe() { throw null; }
     }
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -248,7 +248,7 @@ namespace System.Diagnostics.Metrics
 {
     /// <summary>
     /// ObservableInstrument<T> is the base class from which all observable instruments will inherit from.
-    /// Mainly It'll support the CLS compliant numerical types
+    /// It will only support the CLS compliant numerical types
     /// </summary>
     public abstract class ObservableInstrument<T> : Instrument where T : unmanaged
     {
@@ -262,7 +262,7 @@ namespace System.Diagnostics.Metrics
                     string? unit) : base(meter, name, description, unit) { throw null; }
 
         /// <summary>
-        /// Observe will report the non-additive values of the measurements.
+        /// Observe() fetches the current measurements being tracked by this instrument.
         /// </summary>
         protected abstract IEnumerable<Measurement<T>> Observe();
 
@@ -278,7 +278,7 @@ namespace System.Diagnostics.Metrics
 namespace System.Diagnostics.Metrics
 {
     /// <summary>
-    /// A helper class used by the Observable instruments Observe method to report the measurement values
+    /// A measurement stores one observed value and its associated tags. This type is used by Observable instruments' Observe() method when reporting current measurements. 
     /// with the associated tags.
     /// </summary>
     public struct Measurement<T> where T : unmanaged
@@ -465,7 +465,7 @@ namespace System.Diagnostics.Metrics
         /// <summary>
         /// Set a callback for a specific numeric type to get the measurement recording notification
         /// from all instruments which enabled listened to and was created with the same specified
-        /// numeric type.
+        /// numeric type. If a measurement of type T is recorded and a callback of type T is registered, that callback is used. If there is no callback for type T but there is a callback for type object, the measured value is boxed and reported via the object typed callback. If there is neither type T callback nor object callback then the measurement will not be reported.
         /// </summary>
         public void SetMeasurementEventCallback<T>(MeasurementCallback<T>? measurementCallback) { throw null; }
 
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -25,7 +25,7 @@ The Instrument is the type that will be used by the app and library authors to r
 
 There are two types of instruments:
 
-- The first type we'll call it just `Instrument` for simplicity. These instruments are called inside a request, meaning they have an associated distributed Context (with Span, Baggage, etc.). OpenTelemetry specs call this type of instrument a synchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here proposes three instrument classes of that type: `Counter`, `UpDownCounter`, and `Histogram`.
+- The first type we'll call it just `Instrument` for simplicity. These instruments are called inside a request, meaning they have an associated distributed Context (with Span, Baggage, etc.). OpenTelemetry specs call this type of instrument a synchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here proposes three instrument classes of that type: `Counter` and `Histogram`.
 - The second type is called `ObservableInstrument` which reports measurements by a callback, once per collection interval, and lacks Context. OpenTelemetry specs call this type of instrument an asynchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here is proposing three instrument classes of that type: `ObservableCounter`, `ObservableGauge`, and `ObservableUpDownCounter`.
 
 ### Meter
@@ -65,14 +65,9 @@ namespace System.Diagnostics.Metrics
         public string? Version { get; }
 
         /// <summary>
-        /// Factory Methods to create Counter, UpDownCounter, and Histogram instruments.
+        /// Factory Methods to create Counter and Histogram instruments.
         /// </summary>
         public Counter<T> CreateCounter<T>(
-                            string name,
-                            string? description = null,
-                            string? unit = null) where T : unmanaged { throw null; }
-
-        public UpDownCounter<T> CreateUpDownCounter<T>(
                             string name,
                             string? description = null,
                             string? unit = null) where T : unmanaged { throw null; }
@@ -307,28 +302,6 @@ namespace System.Diagnostics.Metrics
     /// e.g. Number of completed requests.
     /// </summary>
     public sealed class Counter<T> : Instrument<T> where T : unmanaged
-    {
-        public void Add(T measurement) { throw null; }
-        public void Add(T measurement,
-                            KeyValuePair<string, object?> tag1) { throw null; }
-        public void Add(T measurement,
-                            KeyValuePair<string, object?> tag1,
-                            KeyValuePair<string, object?> tag2) { throw null; }
-        public void Add(T measurement,
-                            KeyValuePair<string, object?> tag1,
-                            KeyValuePair<string, object?> tag2,
-                            KeyValuePair<string, object?> tag3) { throw null; }
-        public void Add(T measurement,
-                            ReadOnlySpan<<string, object?>> tags) { throw null; }
-        public void Add(T measurement,
-                            params KeyValuePair<string, object?>[] tags) { throw null; }
-    }
-
-    /// <summary>
-    /// UpDownCounter is a non-observable Instrument that supports increments and decrements.
-    /// e.g. Number of items in a queue.
-    /// </summary>
-    public sealed class UpDownCounter<T> : Instrument<T> where T : unmanaged
     {
         public void Add(T measurement) { throw null; }
         public void Add(T measurement,
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -259,7 +259,7 @@ namespace System.Diagnostics.Metrics
         /// <summary>
         /// Observe() fetches the current measurements being tracked by this instrument.
         /// </summary>
-        protected abstract IEnumerable<Measurement<T>> Observe();
+        public abstract IEnumerable<Measurement<T>> Observe();
 
         public override bool IsObservable => throw null;
     }
@@ -361,7 +361,7 @@ namespace System.Diagnostics.Metrics
     /// </summary>
     public sealed class ObservableCounter<T> : ObservableInstrument<T> where T : unmanaged
     {
-        protected override IEnumerable<Measurement<T>> Observe() { throw null; }
+        public override IEnumerable<Measurement<T>> Observe() { throw null; }
     }
 
     /// <summary>
@@ -371,7 +371,7 @@ namespace System.Diagnostics.Metrics
     /// </summary>
     public sealed class ObservableUpDownCounter<T> : ObservableInstrument<T> where T : unmanaged
     {
-        protected override IEnumerable<Measurement<T>> Observe() { throw null; }
+        public override IEnumerable<Measurement<T>> Observe() { throw null; }
     }
 
     /// <summary>
@@ -381,7 +381,7 @@ namespace System.Diagnostics.Metrics
     /// </summary>
     public sealed class ObservableGauge<T> : ObservableInstrument<T> where T : unmanaged
     {
-        protected override IEnumerable<Measurement<T>> Observe() { throw null; }
+        public override IEnumerable<Measurement<T>> Observe() { throw null; }
     }
 }
 ```
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -14,7 +14,7 @@ The proposed APIs are intended to be supported on Full Framework 4.6 and up, .NE
 
 ```MD040
 Naming in this proposal is picked up from the OpenTelemetry specification to avoid having .NET APIs
-deviating from the standard even if the standard names are not the best to use.
+deviating from the standard even if the standard names are not the best option for the .NET ecosystem.
 ```
 
 ## Metric Terminologies
--- a/accepted/2021/System.Diagnostics/Metrics-Design.md
+++ b/accepted/2021/System.Diagnostics/Metrics-Design.md
@@ -415,7 +415,7 @@ namespace System.Diagnostics.Metrics
         /// <summary>
         /// Callbacks to get notification when an instrument is published
         /// </summary>
-        public Action<Instrument>? InstrumentPublished { get; set; }
+        public Action<Instrument, MeterListener>? InstrumentPublished { get; set; }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


tarekgh:
CC @noahfalk

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+
+The OpenTelemetry architecture is splitting the measurement reporting part to a component called `APIs` and that is the part we are discussing here and proposing APIs for it. The other part is called `SDK` which will monitor, aggregate the Metrics measurements, and export the aggregated measurement to some backend. The SDK part will be implemented in the OpenTelemetry [dotnet repo](https://github.com/open-telemetry/opentelemetry-dotnet) and will not part of the .NET. The `SDK` will depend on the APIs we are going to expose in .NET.
+
+The proposed APIs will be used by the application or Library authoress to record metrics measurements. The APIs also include a listener which will allow listening to the recorded measurements events which can be used by the OpenTelemetry SDK to aggregate the measurements and export them.
+
+The proposed APIs are intended to be supported on Full Framework 4.6 and up, .NET Core all versions, and .NET 5.0 and up.
```
cijothomas, line 13: .NET Core 2.1/3.1 and above?

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+
+The OpenTelemetry Metrics APIs support reporting measurements about the execution of a computer program at run time. The Metrics APIs are designed explicitly for processing raw measurements, generally with the intent to produce continuous summaries of those measurements, efficiently and simultaneously.
+
+The OpenTelemetry architecture is splitting the measurement reporting part to a component called `APIs` and that is the part we are discussing here and proposing APIs for it. The other part is called `SDK` which will monitor, aggregate the Metrics measurements, and export the aggregated measurement to some backend. The SDK part will be implemented in the OpenTelemetry [dotnet repo](https://github.com/open-telemetry/opentelemetry-dotnet) and will not part of the .NET. The `SDK` will depend on the APIs we are going to expose in .NET.
+
+The proposed APIs will be used by the application or Library authoress to record metrics measurements. The APIs also include a listener which will allow listening to the recorded measurements events which can be used by the OpenTelemetry SDK to aggregate the measurements and export them.
```
cijothomas, line 11:
```suggestion
The proposed APIs will be used by the application or Library authors to record metrics measurements. The APIs also include a listener which will allow listening to the recorded measurements events which can be used by the OpenTelemetry SDK to aggregate the measurements and export them.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+        /// </summary>
+        public ObservableGauge<T> CreateObservableGauge<T>(
+                            string name,
+                            Func<T> observeValue,
+                            string? description = null,
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        public ObservableGauge<T> CreateObservableGauge<T>(
+                            string name,
+                            Func<Measurement<T>> observeValues,
```
cijothomas, line 118:
```suggestion
                            Func<Measurement<T>> observeValue,
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+        /// </summary>
+        public ObservableUpDownCounter<T> CreateObservableUpDownCounter<T>(
+                            string name,
+                            Func<T> observeValue,
+                            string? description = null,
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        public ObservableUpDownCounter<T> CreateObservableUpDownCounter<T>(
+                            string name,
+                            Func<Measurement<T>> observeValues,
```
cijothomas, line 139:
```suggestion
                            Func<Measurement<T>> observeValue,
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+        protected void RecordMeasurement(
+                            T val,
+                            KeyValuePair<string, object> tag1,
+                            KeyValuePair<string, object> tag2) { throw null; }
+
+        protected void RecordMeasurement(
+                            T val,
+                            KeyValuePair<string, object> tag1,
+                            KeyValuePair<string, object> tag2,
+                            KeyValuePair<string, object> tag3) { throw null; }
```
cijothomas, line 235: I hope we can add overloads for upto 5, if we see most common cases requires upto 5.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+                    Meter meter,
+                    string name,
+                    string? description,
+                    string? unit) : base(meter, name, description, unit) { throw null; }
+
+        /// <summary>
+        /// Observe will report the non-additive values of the measurements.
+        /// </summary>
+        protected abstract IEnumerable<Measurement<T>> Observe();
+    }
```
cijothomas, line 268: I assume this will override IsObservable => return true..

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+### Non-observable Instrument Base Class
+
+```csharp
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// Instrument<T> is the base class from which all non-observable instruments will inherit from.
+    /// Mainly It'll support the CLS compliant numerical types
+    /// </summary>
+    public abstract class Instrument<T> : Instrument where T : unmanaged
```
cijothomas, line 208: Where does the validation for T occurs? Will api validate that it'll be one of the numerical types? Or is that left to the SDK?

tarekgh: the validation will be done in the API level and not the SDK.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: that is right.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: do you have any data suggesting how many tags usually used in the common scenarios? I am asking because I want to avoid APIs which not going to be used. We still have the option in the future to add those if proven it is really needed. Also, we have the overload which take ReadOnlySpan<> which can help when need to use more labels.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: I think we can reword it as all supported .NET Core versions instead.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
+# Metrics APIs Design
+
+## Overview
+
+This document is discussing the .NET Metrics APIs design which implements the [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/new_api.md).
+
+The OpenTelemetry Metrics APIs support reporting measurements about the execution of a computer program at run time. The Metrics APIs are designed explicitly for processing raw measurements, generally with the intent to produce continuous summaries of those measurements, efficiently and simultaneously.
+
+The OpenTelemetry architecture is splitting the measurement reporting part to a component called `APIs` and that is the part we are discussing here and proposing APIs for it. The other part is called `SDK` which will monitor, aggregate the Metrics measurements, and export the aggregated measurement to some backend. The SDK part will be implemented in the OpenTelemetry [dotnet repo](https://github.com/open-telemetry/opentelemetry-dotnet) and will not part of the .NET. The `SDK` will depend on the APIs we are going to expose in .NET.
```
reyang, line 9:
```suggestion
The OpenTelemetry architecture is splitting the measurement reporting part to a component called `APIs` and that is the part we are discussing here and proposing APIs for it. The other part is called `SDK` which will monitor, aggregate the Metrics measurements, and export the aggregated measurement to some backend. The SDK part will be implemented in the [OpenTelemetry .NET repo](https://github.com/open-telemetry/opentelemetry-dotnet) and will not be part of the .NET. The `SDK` will depend on the APIs we are going to expose in .NET.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
cijothomas: yea. I want to specifically ask about .NET Core 2.1 which will be out-of-support in August, before .NET 6.
So the support would be:
.NET Framework 4.6 and above
.NET Core 3.1
.NET 5 and above

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+
+The proposed APIs will be used by the application or Library authors to record metrics measurements. The APIs also include a listener which will allow listening to the recorded measurements events which can be used by the OpenTelemetry SDK to aggregate the measurements and export them.
```
reyang, line 11:
```suggestion
The proposed APIs will be used by the application or library authors to record metrics measurements. The APIs also include a listener which will allow listening to the recorded measurements events which can be used by the OpenTelemetry SDK (or other potential consumers) to aggregate the measurements and export them.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+
+The proposed APIs are intended to be supported on Full Framework 4.6 and up, .NET Core all versions, and .NET 5.0 and up.
+
+```MD040
+Naming in this proposal is picked up from the OpenTelemetry specification to avoid having .NET APIs
+deviating from the standard even if the standard names are not the best to use.
```
reyang, line 17:
```suggestion
deviating from the standard even if the standard names are not the best option for the .NET ecosystem.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+- The second type is called `ObservableInstrument` which reports measurements by a callback, once per collection interval, and lacks Context. OpenTelemetry specs call this type of instrument an asynchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here is proposing three instrument classes of that type: `ObservableCounter`, `ObservableGauge`, and `ObservableUpDownCounter`.
+
+### Meter
+
+Meter is the factory type responsible for creating Instruments. Meter will have a name and optional version strings.
+
+### Listener
+
+The listener is the type that allows listening to the recorded measurements occurred by both the non-observable instrument and the observable instrument. The listener is an important class that will be used by OpenTelemetry to implement the Metrics SDK.
```
reyang, line 37: The "non-observable instrument" sounds a bit weird 🧙

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: Officially we'll support the versions that are not out of support yet. But, Systsem.Diagnostics.Diagnostics source support netstandard2.0 which will make it work with net core 2.1.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+
+### Tags
+
+Tag is the term used to refer to a key-value attribute associated with a metric event. Each tag categorizes the metric event, allowing events to be filtered and grouped for analysis.
+Current OpenTelemetry specs call it Labels but this is going to change to Attributes. We have used the tags name in the tracing APIs and we'll stick with this name for the sake of consistency with tracing.
```
reyang, line 42: It is already changed to attributes. Please refer to [this issue](https://github.com/open-telemetry/opentelemetry-specification/issues/1113) which is already closed as resolved.

accepted/2021/System.Diagnostics/Metrics-Design.md
cijothomas: No I dont have any data. 
I was just asking, if we do find data (or feedback from otel community) that more is required, would we be flexible to add more.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: Thanks, I'll update the doc accordingly.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: agree. do you have a better name I can use here? I am trying to avoid using the `sync/async` terms.

accepted/2021/System.Diagnostics/Metrics-Design.md
cijothomas: That remind - could you add in the intro that this new API will be part of DiagnosticSource package itself.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: Yes, if we see a demand on that for sure we consider adding them.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+```
+
+### Measurement class
+
+```csharp
+
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// A helper class used by the Observable instruments Observe method to repro the measurement values
```
reyang, line 279: repro -> report?

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+- The second type is called `ObservableInstrument` which reports measurements by a callback, once per collection interval, and lacks Context. OpenTelemetry specs call this type of instrument an asynchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here is proposing three instrument classes of that type: `ObservableCounter`, `ObservableGauge`, and `ObservableUpDownCounter`.
+
+### Meter
+
+Meter is the factory type responsible for creating Instruments. Meter will have a name and optional version strings.
+
+### Listener
+
+The listener is the type that allows listening to the recorded measurements occurred by both the non-observable instrument and the observable instrument. The listener is an important class that will be used by OpenTelemetry to implement the Metrics SDK.
```
cijothomas, line 37:
```suggestion
The listener is the type that allows listening to the measurements recorded by instruments. The listener is an important class that will be used by OpenTelemetry to implement the Metrics SDK.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+    /// <summary>
+    /// Instrument<T> is the base class from which all non-observable instruments will inherit from.
+    /// Mainly It'll support the CLS compliant numerical types
+    /// </summary>
+    public abstract class Instrument<T> : Instrument where T : unmanaged
+    {
+        /// <summary>
+        /// Protected constructor to create the instrument with the common properties.
+        /// </summary>
+        protected Instrument(Meter meter, string name, string? description, string? unit) :
```
cijothomas, line 213: It'd be possible for anyone to create new instruments by inheriting Instrument<T>? (like MyTimingGauge : Instrument<T>
{
RecordTime() --> internally calling RecordMeasurement().
}

accepted/2021/System.Diagnostics/Metrics-Design.md
reyang: Perhaps "The listener is the type that allows listening to the measurements reported by the instruments (e.g. `Counter`, `ObservableCounter`, etc.)."

reyang (approved):
Although the PR is marked as a "draft", it looks good to me.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// A delegate to represent the callbacks signatures used in the listener.
+    /// </summary>
+    public delegate void MeasurementCallback<T>(
+                            Instrument instrument,
+                            T measurement,
+                            ReadOnlySpan<(string TagName, object TagValue)> tags,
+                            object? cookie);
```
cijothomas, line 419: whats cookie?

reyang: state object

cijothomas: Got it.. Not the real 🍪 😄

accepted/2021/System.Diagnostics/Metrics-Design.md
victlu: How do you intend to report errors?

tarekgh: I think Exception throwing for non-supported types would be reasonable here. I don't expect someone will use some non-supported type without testing it :-)

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+    /// A helper class used by the Observable instruments Observe method to repro the measurement values
+    /// with the associated tags.
+    /// </summary>
+    public struct Measurement<T> where T : unmanaged
+    {
+        /// <summary>
+        /// Construct the Measurement using the value and the list of tags.
+        /// We'll always copy the input list as this is not perf hot path.
+        /// </summary>
+        public Measurement(T value, IEnumerable<KeyValuePair<string, object>> tags) { throw null; }
```
victlu, line 288: Do `object` here need to be nullable as `object?`

victlu: Should we allow null object?

tarekgh: yes. we already allow that in tags used in tracing.

tarekgh: I found I already correctly annotated it in `public ReadOnlySpan<KeyValuePair<string, object?>> Tags { get { throw null;  } }` but missed the annotations in the other places. Thanks for the catch.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+                            KeyValuePair<string, object> tag1) { throw null; }
+        public void Add(T measurement,
+                            KeyValuePair<string, object> tag1,
+                            KeyValuePair<string, object> tag2) { throw null; }
+        public void Add(T measurement,
+                            KeyValuePair<string, object> tag1,
+                            KeyValuePair<string, object> tag2,
+                            KeyValuePair<string, object> tag3) { throw null; }
+        public void Add(T measurement,
+                            params KeyValuePair<string, object>[] tags) { throw null; }
```
victlu, line 319: Do we want to take a ReadOnlySpan<KeyValuePair<string,Object>> as well? Since we return ReadOnlySpan in the Measurment.Tags

tarekgh: I think this would be a good idea. at least can allow passing a range of items inside the array. I'll add that overload.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,502 @@
...
+        public void Add(T measurement,
+                            KeyValuePair<string, object> tag1) { throw null; }
+        public void Add(T measurement,
+                            KeyValuePair<string, object> tag1,
+                            KeyValuePair<string, object> tag2) { throw null; }
+        public void Add(T measurement,
+                            KeyValuePair<string, object> tag1,
+                            KeyValuePair<string, object> tag2,
+                            KeyValuePair<string, object> tag3) { throw null; }
+        public void Add(T measurement,
```
victlu, line 338: This signature looks EXACTLY like Counter<T>. The main difference is that Add() takes a negative number in UpDownCounter<T>. For clarity, does it make sense in .NET to use Increment() and Decrement()?

PS: I know the OTel spec says Add().

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: Just to tell, I'll add a constructor overload which take ReadOnlySpan to the `Measurement<T>` struct.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: I prefer `Add` here as it will accept positive and negative values and that will be legitimate. But using Increment/Decrement and pass negative values that will be undesired. I know we do the same with Counter as it supposes to get positive values only but at least with UpDownCounter we are reducing this problem when we use `Add` :-)

victlu: So, what do you do if I Add(-10) in a Counter<long>?

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: they can do that if they inherit from `Instrument<T>` and internally calling RecordMeasurement. But if they inherit from `Instrument`, they will not be able to report the measurements.

accepted/2021/System.Diagnostics/Metrics-Design.md
reyang: Give it to @cijothomas and let him handle it 😄

victlu: LOL. I agree. Chuck it as the SDK!

victlu (approved):

tarekgh:
I switched the PR to be not draft.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+```csharp
+    InstrumentListener listener = new InstrumentListener();
+    listener.InstrumentPublished = (instrument) =>
+    {
+        if (instrument.Name == "Requests" && instrument.Meter.Name == "io.opentelemetry.contrib.mongodb")
+        {
+            listener.EnableMeasurementEvents(instrument, null);
+        }
+    };
+    listener.SetMeasurementEventCallback<int>((instrument, measurement, labels, cookie) =>
```
victlu, line 506: labels => tags

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+There are two types of instruments:
+
+- The first type we'll call it just `Instrument` for simplicity. These instruments are called inside a request, meaning they have an associated distributed Context (with Span, Baggage, etc.). OpenTelemetry specs call this type of instrument a synchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here proposes three instrument classes of that type: `Counter`, `UpDownCounter`, and `Histogram`.
+- The second type is called `ObservableInstrument` which reports measurements by a callback, once per collection interval, and lacks Context. OpenTelemetry specs call this type of instrument an asynchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here is proposing three instrument classes of that type: `ObservableCounter`, `ObservableGauge`, and `ObservableUpDownCounter`.
+
+### Meter
+
+Meter is the factory type responsible for creating Instruments. Meter will have a name and optional version strings.
+
+### Listener
```
noahfalk, line 35:
```suggestion
### MeterListener
```

victlu:
Do you want to provide a way to get a list of Instruments?

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+```
+
+### Observable Instrument Base class
+
+```csharp
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// ObservableInstrument<T> is the base class from which all observable instruments will inherit from.
+    /// Mainly It'll support the CLS compliant numerical types
```
noahfalk, line 251:
```suggestion
    /// It will only support the CLS compliant numerical types
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+        /// Protected constructor to create the instrument with the common properties.
+        /// </summary>
+        protected ObservableInstrument(
+                    Meter meter,
+                    string name,
+                    string? description,
+                    string? unit) : base(meter, name, description, unit) { throw null; }
+
+        /// <summary>
+        /// Observe will report the non-additive values of the measurements.
```
noahfalk, line 265:
```suggestion
        /// Observe() fetches the current measurements being tracked by this instrument.
```
(The text about non-additive appears arbitrary, there are additive observable instruments as well)

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+```
+
+### Measurement class
+
+```csharp
+
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// A helper class used by the Observable instruments Observe method to report the measurement values
```
noahfalk, line 281:
```suggestion
    /// A measurement stores one observed value and its associated tags. This type is used by Observable instruments' Observe() method when reporting current measurements. 
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+
+### Listening Example
+
+```csharp
+    InstrumentListener listener = new InstrumentListener();
+    listener.InstrumentPublished = (instrument) =>
+    {
+        if (instrument.Name == "Requests" && instrument.Meter.Name == "io.opentelemetry.contrib.mongodb")
+        {
+            listener.EnableMeasurementEvents(instrument, null);
```
victlu, line 503: nit: Ideally, we don't have to capture listener in this closure.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+
+```csharp
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// ObservableCounter is an observable Instrument that reports monotonically increasing value(s)
+    /// when the instrument is being observed.
+    /// e.g. CPU time (for different processes, threads, user mode or kernel mode).
+    /// </summary>
+    public class ObservableCounter<T> : ObservableInstrument<T> where T : unmanaged
```
noahfalk, line 389: Should we seal all these derived instrument implementations? We haven't planned for any scenarios where users would directly extend them and I assume we could always unseal them in the future if a new scenario emerged.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// A delegate to represent the callbacks signatures used in the listener.
+    /// </summary>
+    public delegate void MeasurementCallback<T>(
+                            Instrument instrument,
+                            T measurement,
+                            ReadOnlySpan<KeyValuePair<string, object?>> tags,
+                            object? cookie);
```
noahfalk, line 428: @tarekgh - do you know if it is common for us to name data like this 'cookie'? I am fine with it as-is, but if there is a better convention such as calling it 'state', 'userData', 'userState', 'userToken', etc. I am glad to change it. For example I see [Timer](https://docs.microsoft.com/en-us/dotnet/api/system.threading.timer.-ctor?view=net-5.0#System_Threading_Timer__ctor_System_Threading_TimerCallback_System_Object_System_Int32_System_Int32_) and [WaitCallback](https://docs.microsoft.com/en-us/dotnet/api/system.threading.waitcallback?view=net-5.0) call their similar parameter 'state'

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+
+        /// <summary>
+        /// Stop listening to a specific instrument measurement recording.
+        /// </summary>
+        public void DisableMeasurementEvents(Instrument instrument) { throw null; }
+
+        /// <summary>
+        /// Set a callback for a specific numeric type to get the measurement recording notification
+        /// from all instruments which enabled listened to and was created with the same specified
+        /// numeric type.
```
noahfalk, line 468:
```suggestion
        /// numeric type. If a measurement of type T is recorded and a callback of type T is registered, that callback is used. If there is no callback for type T but there is a callback for type object, the measured value is boxed and reported via the object typed callback. If there is neither type T callback nor object callback then the measurement will not be reported.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,511 @@
...
+                        base(meter, name, description, unit) { throw null; }
+
+        /// <summary>
+        /// Record measurement overloads allowing passing different numbers of tags.
+        /// </summary>
+
+        protected void RecordMeasurement(T val) { throw null; }
+
+        protected void RecordMeasurement(
+                            T val,
```
victlu, line 223: nit: val => measurement

noahfalk (approved):
LGTM : )

tarekgh:
> Do you want to provide a way to get a list of Instruments?

The listener already supports that. when you call `Start()` on the listener, all instruments will get enumerated and reported to `InstrumentPublished` callback.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: agree. it make sense to seal them.

tarekgh: @noahfalk 
The question now about the listener. can you think in any scenario someone would need to extend it? if not, then we can seal it too.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: I don't think we have a design guideline to name such parameter more than 

✔️ [DO use descriptive parameter names.](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-parameters)

But trying to look at more cases like: [contextcallback](https://docs.microsoft.com/en-us/dotnet/api/system.threading.contextcallback?view=net-5.0), [SendOrPostCallback](https://docs.microsoft.com/en-us/dotnet/api/system.threading.sendorpostcallback?view=net-5.0), and [WaitOrTimerCallback](https://docs.microsoft.com/en-us/dotnet/api/system.threading.waitortimercallback?view=net-5.0) and I am seeing it is using the name `state`. Considering that, I'll change the name from cookie to state.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: Actually, @noahfalk had it before passing the MeterListener object to the `InstrumentPublished` callback and we thought it is not really needed. Thinking more about and expecting users will need to call `EnableMeasurementEvents` inside the `InstrumentPublished`, I have returned it back so users can avoid capture the listener in the closure.

accepted/2021/System.Diagnostics/Metrics-Design.md
noahfalk: I don't see any reason the listener needs to be extendable and I'd be fine to seal it for now. At best I could see scenarios where it is mildly more convenient for the user to extend the listener rather than aggregating it, but I don't see a reason why we need to prioritize it up-front.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,484 @@
...
+
+        /// <summary>
+        /// Start listening to a specific instrument measurement recording.
+        /// </summary>
+        public void EnableMeasurementEvents(Instrument instrument, object? state = null) { throw null; }
+
+        /// <summary>
+        /// Stop listening to a specific instrument measurement recording.
+        /// </summary>
+        public void DisableMeasurementEvents(Instrument instrument) { throw null; }
```
victlu, line 436: Do you want to return the state?

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,484 @@
...
+
+        /// <summary>
+        /// Set a callback for a specific numeric type to get the measurement recording notification
+        /// from all instruments which enabled listened to and was created with the same specified
+        /// numeric type. If a measurement of type T is recorded and a callback of type T is registered, that callback is used. If there is no callback for type T but there is a callback for type object, the measured value is boxed and reported via the object typed callback. If there is neither type T callback nor object callback then the measurement will not be reported.
+        /// </summary>
+        public void SetMeasurementEventCallback<T>(MeasurementCallback<T>? measurementCallback) { throw null; }
```
victlu, line 443: Given MeasurementCallback is a delegate, do you intend to only allow one callback, or multiple?

tarekgh: It will be one callback per numeric type.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: I think this will be a good idea.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,484 @@
...
+        /// <summary>
+        /// Protected constructor to initialize the common instrument properties.
+        /// </summary>
+        protected Instrument(Meter meter, string name, string? description, string? unit) { throw null; }
+
+        /// <summary>
+        /// Publish is to allow activating the instrument to start recording measurements and to allow
+        /// listeners to start listening to such measurements.
+        /// </summary>
+        protected void Publish() { throw null; }
```
cijothomas, line 171: Its not clear what is the use case of a protected Publish() method.. when one does meter.CreateCounter/Gauage, won't that be automatically published? 
Do we expect any instrument to not be published?

cijothomas (approved):
Thanks!
LGTM. 
Will try to share more feedbacks as soon as we learn them by implementing SDKs.

accepted/2021/System.Diagnostics/Metrics-Design.md
cijothomas: `state` sounds better.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: `Publish` is important to tell when start using the instrument after creating and initializing it. We didn't call it automatically inside the abstract base classes constructors because that means the instrument would be published before it is fully initialized in the concrete classes. When using our built-in instruments, you'll not need to call it manually as it will get called from that instrument's constructors. Publish will be needed if someone introduce a custom instrument. They will need to call it from the constructor at that time.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,484 @@
...
+        protected void RecordMeasurement(
+                            T measurement,
+                            KeyValuePair<string, object?> tag1,
+                            KeyValuePair<string, object?> tag2) { throw null; }
+
+        protected void RecordMeasurement(
+                            T measurement,
+                            KeyValuePair<string, object?> tag1,
+                            KeyValuePair<string, object?> tag2,
+                            KeyValuePair<string, object?> tag3) { throw null; }
```
victlu, line 230: Given our Listener doesn't also support these 1, 2, 3 KVP prototypes. Do we really need these 1, 2, 3 KVP prototypes?
Also, here is the protected versions, so I assume derived classes can convert to ReadOnlySpan<> for us.

accepted/2021/System.Diagnostics/Metrics-Design.md
cijothomas: Got it. Somehow I misread it as a something end users need to do.
So for the built-in instruments, simply creating them is enough for end user. 

publish() is mostly for anyone trying to create own instruments.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,484 @@
...
+                            KeyValuePair<string, object?> tag1) { throw null; }
+        public void Add(T measurement,
+                            KeyValuePair<string, object?> tag1,
+                            KeyValuePair<string, object?> tag2) { throw null; }
+        public void Add(T measurement,
+                            KeyValuePair<string, object?> tag1,
+                            KeyValuePair<string, object?> tag2,
+                            KeyValuePair<string, object?> tag3) { throw null; }
+        public void Add(T measurement,
+                            ReadOnlySpan<<string, object?>> tags) { throw null; }
```
victlu, line 317: The KeyValuePair<> got dropped!

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,484 @@
...
+        public void Add(T measurement,
+                            params KeyValuePair<string, object?>[] tags) { throw null; }
```
victlu, line 319: Assuming our end result is ReadOnlySpan<KeyValuePair<>>, I'm wondering if the 1,2,3 KVP prototypes is any faster than a proto that takes param KVP. If it's not faster, then, I don't know we need these 1,2,3 KVP prototypes.

accepted/2021/System.Diagnostics/Metrics-Design.md
victlu: You want to grep these, as it's in multiple places.

accepted/2021/System.Diagnostics/Metrics-Design.md
victlu: Yes. But if our goal is to create ReadOnlySpan<>, we will still need to create an array! So, practically, I'm wondering if there's any difference at this point!

tarekgh: > Yes. But if our goal is to create ReadOnlySpan<>, we will still need to create an array! So, practically, I'm wondering if there's any difference at this point!

No, we don't need to create array for that. we'll just store the values on the stack and create the ReadOnlySpan from that.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: The 1, 2, 3 KVPs overloads are needed to avoid creating array parameter.

accepted/2021/System.Diagnostics/Metrics-Design.md
tarekgh: I sealed the listener too.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,485 @@
...
+}
+```
+
+### Library Measurement Recording Example
+
+```csharp
+    Meter meter = new Meter("io.opentelemetry.contrib.mongodb", "v1.0");
+    Counter<int> counter = meter.CreateCounter<int>("Requests");
+    counter.Add(1);
+    counter.Add(1, KeyValuePair<string, object>("request", "read"));
```
victlu, line 466: I think you need `new` KeyValuePair<>

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
+# Metrics APIs Design
+
+## Overview
+
+This document is discussing the .NET Metrics APIs design which implements the [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/new_api.md).
```
reyang, line 5:
```suggestion
This document is discussing the .NET Metrics APIs design which implements the [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md).
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+```MD040
+Naming in this proposal is picked up from the OpenTelemetry specification to avoid having .NET APIs
+deviating from the standard even if the standard names are not the best option for the .NET ecosystem.
+```
+
+## Metric Terminologies
+
+### Instrument
+
+The Instrument is the type that will be used by the app and library authors to report measurements (e.g. Counter, Gauge...etc.). The instrument will have a name that should be validated as described in the OpenTelemetry specs. The Instrument can optionally have a description and unit of measurements. Instruments will be created with a numerical type parameter (e.g. `Counter<int>`). The Instrument is going to support all CLS-compliant numerical types which are supported on the Full Framework (Byte, Int16, Int32, Int64, Single, Double, and Decimal).
```
reyang, line 24:
```suggestion
The Instrument is the type that will be used by the app and library authors to report measurements (e.g. Counter, ObservableGauge...etc.). The instrument will have a name that should be validated as described in the OpenTelemetry specs. The Instrument can optionally have a description and unit of measurements. Instruments will be created with a numerical type parameter (e.g. `Counter<int>`). The Instrument is going to support all CLS-compliant numerical types which are supported on the Full Framework (Byte, Int16, Int32, Int64, Single, Double, and Decimal).
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+
+There are two types of instruments:
+
+- The first type we'll call it just `Instrument` for simplicity. These instruments are called inside a request, meaning they have an associated distributed Context (with Span, Baggage, etc.). OpenTelemetry specs call this type of instrument a synchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here proposes two instrument classes of that type: `Counter` and `Histogram`.
+- The second type is called `ObservableInstrument` which reports measurements by a callback, once per collection interval, and lacks Context. OpenTelemetry specs call this type of instrument an asynchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here is proposing three instrument classes of that type: `ObservableCounter`, `ObservableGauge`, and `ObservableUpDownCounter`.
```
reyang, line 29:
```suggestion
- The second type is called `ObservableInstrument` which reports measurements by a callback, and lacks Context. OpenTelemetry specs call this type of instrument an asynchronous Instrument but we are trying to avoid confusion with the async feature of the .NET. The proposal here is proposing three instrument classes of that type: `ObservableCounter`, `ObservableGauge`, and `ObservableUpDownCounter`.
```

reyang: There is discussion in the OpenTelemetry metrics SDK if we allow different frequencies (e.g. callback being invoked every 5 seconds, but the report is only happening every 1 minute or based on the scraper pull).

Given this is the SDK details and not related to the API, I suggest that we remove this.

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+        public Meter(string name, string? version) { throw null; }
+
+        /// <summary>
+        /// Getter properties to retrieve the Meter name and version
+        /// </summary>
+        public string Name { get; }
+        public string? Version { get; }
+
+        /// <summary>
+        /// Factory Methods to create Counter and Histogram instruments.
```
reyang, line 68:
```suggestion
        /// Factory methods to create Counter and Histogram instruments.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+                            string? description = null,
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        public Histogram<T> CreateHistogram<T>(
+                            string name,
+                            string? description = null,
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        /// <summary>
+        /// Factory Methods to create an observable Counter instrument.
```
reyang, line 81:
```suggestion
        /// Factory methods to create an ObservableCounter instrument.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        public ObservableCounter<T> CreateObservableCounter<T>(
+                            string name,
+                            Func<IEnumerable<Measurement<T>>> observeValues,
+                            string? description = null,
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        /// <summary>
+        /// Factory Methods to create observable gauge instrument.
```
reyang, line 103:
```suggestion
        /// Factory methods to create ObservableGauge instrument.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        public ObservableGauge<T> CreateObservableGauge<T>(
+                            string name,
+                            Func<IEnumerable<Measurement<T>>> observeValues,
+                            string? description = null,
+                            string? unit = null) where T : unmanaged { throw null; }
+
+        /// <summary>
+        /// Factory Methods to create observable UpDownCounter instrument.
```
reyang, line 124:
```suggestion
        /// Factory methods to create ObservableUpDownCounter instrument.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+```
+
+### Non-observable Instrument Base Class
+
+```csharp
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// Instrument<T> is the base class from which all non-observable instruments will inherit from.
+    /// Mainly It'll support the CLS compliant numerical types
```
reyang, line 201:
```suggestion
    /// Mainly it will support the CLS compliant numerical types.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+```
+
+### Observable Instrument Base class
+
+```csharp
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// ObservableInstrument<T> is the base class from which all observable instruments will inherit from.
+    /// It will only support the CLS compliant numerical types
```
reyang, line 246:
```suggestion
    /// It will only support the CLS compliant numerical types.
```

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+```
+
+### Measurement class
+
+```csharp
+
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// A measurement stores one observed value and its associated tags. This type is used by Observable instruments' Observe() method when reporting current measurements.
```
reyang, line 276:
```suggestion
    /// A measurement stores one observed value and its associated tags. This type is used by Observable instruments' Observe() method when reporting current measurements with associated tags.
```

(and delete the next line)

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+}
+```
+
+### Non-observable Instruments
+
+```csharp
+namespace System.Diagnostics.Metrics
+{
+    /// <summary>
+    /// The counter is a non-observable Instrument that supports non-negative increments.
```
reyang, line 301: nit: "non-observable" sounds very weird

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,483 @@
...
+                            KeyValuePair<string, object?> tag3) { throw null; }
+        public void Add(T measurement,
+                            ReadOnlySpan<KeyValuePair<string, object?>> tags) { throw null; }
+        public void Add(T measurement,
+                            params KeyValuePair<string, object?>[] tags) { throw null; }
+    }
+
+    /// <summary>
+    /// The histogram is a non-observable Instrument that can be used to report arbitrary values
+    /// that are likely to be statistically meaningful. It is intended for statistics such
```
reyang, line 324:
```suggestion
    /// that are likely to be statistically meaningful. It is intended for statistics such as the request duration.
```

(and remove the next line)

accepted/2021/System.Diagnostics/Metrics-Design.md
```diff
@@ -0,0 +1,484 @@
+# Metrics APIs Design
+
+## Overview
+
+This document is discussing the .NET Metrics APIs design which implements the [OpenTelemetry specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md).
```
stephentoub, line 5: How should I think about these APIs vs the DiagnosticCounter APIs we've been using from System.Diagnostics.Tracing?  When does someone choose these vs EventCounter, PollingCounter, etc.?

tarekgh: We already working on publishing the guide for the users currently using EventCounters. That include, having the tools like dotnet-counters work with new Metrics, migrating from EventCounter to Metrics, and allow migrated libraries/apps from EventCounters to Metrics to implement a small shim for any app compat support if necessary. 

CC @noahfalk @shirhatti

stephentoub: Does that mean we're deprecating DiagnosticCounter, changing all of our existing DiagnosticCounters to be based on these new APIs, etc.?  I'm confused.

tarekgh: We'll recommend using Metrics moving forward but we are not deprecating the Event counters in this release at least.

stephentoub: So I hope then we're replacing our own usage in .NET 6?

PR #212, opened 2021-04-30
Include tizen6.0-tizen in precedence rules

WonyoungChoi:
Context: https://github.com/NuGet/NuGet.Client/pull/4011#issuecomment-829301719

There is no mention of `net6.0-tizen` in the `Compatibility rules` section.
These should be the same as `net6.0-ios` except that `tizen` is used instead of `xamarin.ios`.

And the Tizen platform runtime identifiers should also be added to the `RIDs` section.

Tizen is preparing to provide `net6.0-tizen` TargetFramework in https://github.com/Samsung/Tizen.NET  to support .NET6 and MAUI.

### DIFF
--- a/accepted/2021/net6.0-tfms/net6.0-tfms.md
+++ b/accepted/2021/net6.0-tfms/net6.0-tfms.md
@@ -61,6 +61,9 @@ app.
 |                    | (+everything else inherited from net6.0) |
 | net6.0-tvos        | xamarin.tvos                             |
 |                    | (+everything else inherited from net6.0) |
+| net6.0-tizen       | tizen                                    |
+|                    | (+everything else inherited from net6.0) |
+
 
 ### Compatibility rules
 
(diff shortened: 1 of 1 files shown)


terrajobst (approved):
Makes sense to me.

zkat (approved):

Happypig375:
The title says tizen6.0-tizen, not net6.0-tizen?

nkolev92 (approved):

terrajobst:
> The title says tizen6.0-tizen, not net6.0-tizen?

I assume the title is a typo and the desired TFM is `net6.0-tizen`.

terrajobst:
I've filed this issue in NuGet to track the work: https://github.com/NuGet/Home/issues/10819

WonyoungChoi:
> The title says tizen6.0-tizen, not net6.0-tizen?

Yes. It was my mistake.

PR #213, opened 2021-05-01
Add Simple C# Programs

cartermp:
Let's see if we can make C# a little simpler.

Working list of additions/changes to make to the doc:

- [x] Clarify things like build output
- [x] Correct mistakes/assumptions
- [x] Call out primary affected areas
- [x] Propose an order of work
- [x] Call out tooling scenarios
- [x] More attributions (e.g., dotnet-script)

### DIFF
--- a/proposed/simple-csharp-programs-motivation-walkthrough.md
+++ b/proposed/simple-csharp-programs-motivation-walkthrough.md
@@ -110,7 +110,7 @@ dotnet bot
 Or via a shebang,
 
 ```csharp
-#r! /usr/bin/shared/dotnet
+#! /usr/bin/shared/dotnet
 #r "nuget: Newtonsoft.Json"
 
 record Person(string Name);
--- a/proposed/simple-csharp-programs-motivation-walkthrough.md
+++ b/proposed/simple-csharp-programs-motivation-walkthrough.md
@@ -1,0 +1,311 @@
+# Simple C# Programs
+
+This speclet defines "Simple C# Programs" and covers what they can do from the user's standpoint. It does **not** get into detailed design points, a deep dive of a particular mechanisms, etc. The goal is to drive consensus on the user experience and then walk back towards the right design and implementation.
+
+It is also designed to be incremental. Not everything written here must be implemented. For example, we could ship a version without support for packages/frameworks.
+
+## Motivation
+
+C# has lots of great things going for it, but it falls short in a key area: getting started with the least possible friction.
+
+Lots of developers in the world have experience with languages like JavaScript, Python, and Ruby. In these languages, you typically get started in one of these three ways:
+
+1. Launch a REPL and start typing code and getting results
+2. Create a file with the right extension and run the code in that file
+3. Use build tooling to create an application and then run that application
+
+If you look at online guides for the above languages, they tend to start people off with (1) and (2) and rarely with (3).
+
+In the C# world, we start people off with (3) with the base tooling, the .NET CLI.
+
+This is problematic, because these people **do not expect that they need to do (3) to get started**. The tools don't support their instincts and there is very little information online to support them.
+
+Of course, REPLs are another great way to get started. But they are outside the scope of this document.
+
+## What is a Simple C# Program
+
+A Simple C# Program is this:
+
+* A C# file
+* A set of global `using`s inferred by the build environment
+* A tool that can run that file / produce a working program
+* A way to incorporate libraries and frameworks/sdks into that file and run it
+* A way to "graduate" to using a project file
+
+A Simple C# Program is **not** a replacement for project-based C# code or .NET Interactive.
+
+## Simple C# is just C# code
+
+Simple C# Programs are not a dialect of the C# language, like the interactive dialect, CSX. Any code in a Simple C# Program can be built and run in a project-based program.
+
+## User experience walkthrough
+
+Given the following code in `Hello.cs`:
+
+```csharp
+WriteLine("Hello World");
+```
+
+You can just run it:
+
+```bash
+$ dotnet run
+Hello World
+```
+
+Or with a shebang,
+
+```csharp
+#!/usr/bin/shared/dotnet
+WriteLine("Hello World");
+```
+
+and running:
+
+```bash
+$ ./hello.cs
+Hello World
+```
+
+The `using`s that are normally in a C# project template are implicitly available at design-time. This is accomplished with [Global Using Directives](https://github.com/dotnet/csharplang/blob/main/proposals/GlobalUsingDirective.md) that the build tooling will have available.
+
+The shebang is covered in [Shebang (#!) support](https://github.com/dotnet/csharplang/issues/3507).
+
+### Building instead of running
+
+You can also build it with `dotnet build`:
+
+```bash
+$ dotnet build
+$ ls
+Hello.exe
+$ ./Hello.exe
+Hello World
+```
+
+When built, it creates a single executable that it drops in the current working directory.
+
+### Referencing a package
+
+You can also pull in a package (syntax perhaps subject to change):
+
+```csharp
+#r "nuget: Newtonsoft.Json"
+
+record Person(string Name);
+
+var json = "{\"name\": \"dotnet bot\"}";
+var person = JsonConvert.DeserializeObject<Person>(json);
+
+Console.WriteLine(person.Name);
+```
+
+And it'll run after downloading the package:
+
+```bash
+$ dotnet run
+dotnet bot
+```
+
+Or via a shebang,
+
+```csharp
+#r! /usr/bin/shared/dotnet
+#r "nuget: Newtonsoft.Json"
+
+record Person(string Name);
+
+var json = "{\"name\": \"dotnet bot\"}";
+var person = JsonConvert.DeserializeObject<Person>(json);
+
+Console.WriteLine(person.Name);
+```
+
+and running:
+
+```bash
+$ ./hello.cs
+dotnet bot
+```
+
+The `#r "nuget:..."` syntax is already supported by C# today in .NET Interactive. It is a shared implementation between C# and F#. The syntax and mechanism could be re-used here.
+
+### Coding against a framework/sdk like ASP.NET Core
+
+You can also reference a framework/sdk (nomenclature up for debate) via a shebang:
+
+```csharp
+#!/usr/bin/share/dotnet-aspnet
+
+var app = WebApplication.CreateBuilder(args);
+app.MapGet("/", () => "Hello World");
+await app.RunAsync();
+```
+
+The `dotnet-aspnet` gesture indicates two things to the build/run mechanism:
+
+* ASP.NET Core gets loaded and referenced
+* Applicable implicit ASP.NET Core `using`s are brought into scope
+
+And you can run it:
+
+```bash
+$ dotnet run
+Listening on localhost://12345
+```
+
+The syntax for coding against a framework/sdk could also be more similar to the `#r "nuget:"` syntax shown earlier instead of being a shebang.
+
+## `dotnet` command support
+
+Only `dotnet run` and `dotnet build` are supported with Simple C# Programs. All other `dotnet` commands will error out, indicating that you need a project.
+
+## Grow-up story
+
+There is eventually a need for a grow-up story. What's interesting is that this can, in theory, go one of two ways:
+
+1. You "graduate" to using projects, like every other .NET developer
+2. You continue to create loose files, but you instead organize them in meaningful folders
+
+The first option is well-understood today, but could involve a special `dotnet` command that _adds_ a project rather than wipes out your existing directory, and converts special directives in your file:
+
+```bash
+$ dotnet add project
+    Added project 'MyFolder.csproj' to the current workspace
+```
+
+This lets you "grow up" into a project by:
+
+* Creating a project file and placing it in the current directory, with a name that matches
+* Potentially removing shebangs / special directives and converting things to an appropriate SDK attribute or `PackageReference`
+* Provides a nice message saying what it did
+
+After this, you need to run your program with `dotnet run` and friends, but by virtue of being a project, you now have the full `dotnet` toolset at your disposal.
+
+Shebangs wouldn't technically need to be removed. You should still be able to run a single file the same was as before. Having a project file in the current directory shouldn't change that.
+
+The second option has potentially wild implications across our entire build tools and IDE tooling stack. It would require us to fundamentally rethink how .NET applications can be built.
+
+For the purposes of this design, we will require that users create a project-based codebase if they wish to be grown up.
+
+## Design points / pivots
+
+The following is a non-exhuastive list of user experience considerations that go beyond the basic walkthrough above.
+
+### Multiple files
+
+We will need to decide if Simple C# Programs support more than one file. In other languages, like JavaScript and F#, you can have several loose files in your directory (or other directories) and include them into your program. For example:
+
+**utils.js**
+```js
+const chalk = require('chalk')
+const currentTime = (new Date().toLocaleTimeString([], { hour12: false })).split(':').join('-')
+
+// Pretty print console logs with Chalk
+exports.log = {
+  dim: (text) => console.log(chalk.dim(text)),
+  success: (text) => console.log(chalk.green(text))
+}
+
+// Generates a fileName with current time. e.g. fake-names-19:29:17.json
+exports.getFileName = () => `fake-names-${currentTime}.json`
+```
+
+**program.js**
+
+```js
+const { log, getFileName } = require('../lib/utils')
+
+// use stuff from `utils`
+```
+
+Similarly, in F# there is the `#load "path-to-file.fsx"` directive.
+
+This leads to some questions:
+
+* Are multiple file supported? That is, can I define a `server.cs` file that is my web service, and a `data.cs` that handles some data manipulation that is called from code inside of `server.cs`?
+* If multiple files are supported, what are the guardrails? How can people know if they made a mistake and fix it?
+
+### Coding against a framework alternative syntax
+
+The walkthrough proposes the shebang `#!/usr/bin/share/dotnet-aspnet`, but the .NET CLI doesn't have a command to "pull in the ASP.NET Core references", so it wouldn't map to something like that.
+
+An alternative could be to extend `#r` like so:
+
+* `#r "framework:aspnet"`
+* `#r "sdk:aspnet"`
+* etc.
+
+This could cleanly separate the "tell the compiler which references to pull in" world from the "short-cut having to use `dotnet`" world like so:
+
+```csharp
+#!/usr/bin/share/dotnet
+#r "framework:aspnet"
+
+var app = WebApplication.CreateBuilder(args);
+app.MapGet("/", () => "Hello World");
+await app.RunAsync();
+```
+
+### Shebangs and their semantics
+
+The shebangs in the walkthrough all point at `/usr/bin/share/dotnet`. Substitute a path for your appropriate installation of .NET on your operating system of choice.
+
+What this means is that it identifies the .NET environment and implies a default set of references available in the file. However, it does not bring in other things like ASP.NET or other frameworkreferences. It is a contextual, "this is how I run" sort of directive.
+
+It could be extended/changed in a few ways:
+
+* `/usr/bin/share/dotnet run` makes it explicit that the command is to run the file
+* `/usr/bin/share/dotnet-aspnet` is an alternative to `#!aspnet` that brings ASP.NET Core into scope for the file (substitute a different framework/sdk here)
+
+In effect, the shebang's semantics could span a spectrum of "I need the path and command to literally execute this file" towards "I add context that affects the references available to this file".
+
+We'll need to decide if they can affect the set of references available in the file or not.
+
+### Build output
+
+The walkthrough states that the build output is a single executable file in the current working directory. This is based on the hypothesis that this is the least surprising place to put the executable.
+
+* Is this hypothesis based on reality?
+* Is this still true if you pull in a package?
+* Is this still true if you pull in a framework? (e.g., ASP.NET Core is hundreds of assemblies, where do they go?)
+
+### Nuget reference syntax and naming
+
+The current walkthrough proposed `"#r "nuget:..."` for packages.
+
+Syntax could change:
+
+* Use a `#!r` shebang-style syntax
+* Use a custom directive like `#nuget`
+
+Naming could change:
+
+* `package` instead of `nuget`
+
+However, that seems unlikely given that the current syntax is already shipping in .NET Interactive.
+
+### A note on .NET Interactive
+
+.NET Interactive _may_ be able to service these goals, but since it is currently focused on notebooks-based programming it doesn't really address the same scenario like JavaScript, Python, and Ruby. It is highly relevant for people who are used to notebook environments, and since notebooks are breaking into educational spaces, it may also play a big role in getting people started there.
+
+But if a JS/Python/Ruby programmer is looking to try out C# for the first time, the way that they know how to start with a language is not well supported by our tooling.
+
+### A note on F#
+
+With some very real ["Simpsons already did it"](https://knowyourmeme.com/memes/the-simpsons-did-it) energy, we here recognize that F# has already gone down this path with F# scripting and F# Interactive:
+
+**hello.fsx**:
+
+```fsharp
+printfn "Hello, World!"
+```
+
+and run it:
+
+```bash
+$ dotnet fsi hello.fsx
+Hello, World!
+```
+
+However, the model is based on FSX and F# Interactive, which is not a realistic approach for C# to do. The CSX and C# dialects are subtly different, there is legacy behavior to reconcile, and more.
--- a/proposed/simple-csharp-programs.md
+++ b/proposed/simple-csharp-programs.md
@@ -53,7 +53,7 @@ $ dotnet run
 Hello World
 ```
 
-Or with a shebang,
+Or with a `#!` directive,
 
 ```csharp
 #!/usr/bin/shared/dotnet
@@ -67,9 +67,11 @@ $ ./hello.cs
 Hello World
 ```
 
+When running via `dotnet run` or the `#!` directive, artifacts may be created on disk in the temp folder. They could also be generated in memory.
+
 The `using`s that are normally in a C# project template are implicitly available at design-time. This is accomplished with [Global Using Directives](https://github.com/dotnet/csharplang/blob/main/proposals/GlobalUsingDirective.md) that the build tooling will have available.
 
-The shebang is covered in [Shebang (#!) support](https://github.com/dotnet/csharplang/issues/3507).
+The `#!` directive is covered in [Shebang (#!) support](https://github.com/dotnet/csharplang/issues/3507).
 
 ### Building instead of running
 
@@ -107,7 +109,7 @@ $ dotnet run
 dotnet bot
 ```
 
-Or via a shebang,
+Or via a `#!` directive,
 
 ```csharp
 #! /usr/bin/shared/dotnet
@@ -132,7 +134,7 @@ The `#r "nuget:..."` syntax is already supported by C# today in .NET Interactive
 
 ### Coding against a framework/sdk like ASP.NET Core
 
-You can also reference a framework/sdk (nomenclature up for debate) via a shebang:
+You can also reference a framework/sdk (nomenclature up for debate) via a `#!` directive:
 
 ```csharp
 #!/usr/bin/share/dotnet-aspnet
@@ -154,7 +156,7 @@ $ dotnet run
 Listening on localhost://12345
 ```
 
-The syntax for coding against a framework/sdk could also be more similar to the `#r "nuget:"` syntax shown earlier instead of being a shebang.
+The syntax for coding against a framework/sdk could also be more similar to the `#r "nuget:"` syntax shown earlier instead of being a `#!` directive.
 
 ## `dotnet` command support
 
@@ -188,12 +190,12 @@ $ dotnet add project
 This lets you "grow up" into a project by:
 
 * Creating a project file and placing it in the current directory, with a name that matches
-* Potentially removing shebangs / special directives and converting things to an appropriate SDK attribute or `PackageReference`
+* Potentially removing `#!` directives / special directives and converting things to an appropriate SDK attribute or `PackageReference`
 * Provides a nice message saying what it did
 
 After this, you need to run your program with `dotnet run` and friends, but by virtue of being a project, you now have the full `dotnet` toolset at your disposal.
 
-Shebangs wouldn't technically need to be removed. You should still be able to run a single file the same was as before. Having a project file in the current directory shouldn't change that.
+`#!` directives wouldn't technically need to be removed. You should still be able to run a single file the same was as before. Having a project file in the current directory shouldn't change that.
 
 The second option has potentially wild implications across our entire build tools and IDE tooling stack. It would require us to fundamentally rethink how .NET applications can be built.
 
@@ -236,11 +238,24 @@ This would require some design work to arrive at the best syntax and semantics. 
 
 This phase would bring in support for pulling in multiple files for a single program. Imagine a `server.cs` with a `network-utils.cs` file, where `network-utils.cs` is used by `server.cs`.
 
-Some guardrails would have to be built into place. More considerations are listed under 
+Some guardrails would have to be built into place.
+
+### Phase 4 - support in Visual Studio
+
+This is not designed, but eventually support for Simple C# Programs would have to make its way into Visual Studio. This would perhaps be a complicated task though, since .NET tooling in Visual Studio more or less assumes you are in a project/solution-based workspace. It could be that we never support Simple C# Programs in Visual Studio.
 
 ## Design points / pivots
 
 The following is a non-exhuastive list of user experience considerations that go beyond the basic walkthrough above.
+
+### Where do the artifacts go when running?
+
+When using `dotnet run` or a `#!` directive in a Simple C# Program, there's a question of where build artifacts go, because it has to compile the program. There are two primary options:
+
+1. Compile and execute in memory
+2. Place artifacts in the temp directory on your machine
+
+There is precedent for (2) in Golang, which means it's probably a viable option. But there is a concern about cleaning the temp folder as well, lest it get too large. If everything is compile and ran in memory, this isn't a problem.
 
 ### Multiple files
 
@@ -298,7 +313,7 @@ In Directory Two, the `utils.cs` file is used by `reboot-server.cs`. A developer
 
 ### Coding against a framework alternative syntax
 
-The walkthrough proposes the shebang `#!/usr/bin/share/dotnet-aspnet`, but the .NET CLI doesn't have a command to "pull in the ASP.NET Core references", so it wouldn't map to something like that.
+The walkthrough proposes the `#!` directive `#!/usr/bin/share/dotnet-aspnet`, but the .NET CLI doesn't have a command to "pull in the ASP.NET Core references", so it wouldn't map to something like that.
 
 An alternative could be to extend `#r` like so:
 
@@ -328,7 +343,7 @@ It could be extended/changed in a few ways:
 * `/usr/bin/share/dotnet run` makes it explicit that the command is to run the file
 * `/usr/bin/share/dotnet-aspnet` is an alternative to `#!aspnet` that brings ASP.NET Core into scope for the file (substitute a different framework/sdk here)
 
-In effect, the shebang's semantics could span a spectrum of "I need the path and command to literally execute this file" towards "I add context that affects the references available to this file".
+In effect, the `#!` directive's semantics could span a spectrum of "I need the path and command to literally execute this file" towards "I add context that affects the references available to this file".
 
 We'll need to decide if they can affect the set of references available in the file or not.
 
@@ -356,7 +371,7 @@ The current walkthrough proposed `"#r "nuget:..."` for packages.
 
 Syntax could change:
 
-* Use a `#!r` shebang-style syntax
+* Use a `#!r`-style syntax, similar to the `#!` directive
 * Use a custom directive like `#nuget`
 
 Naming could change:
--- a/proposed/simple-csharp-programs.md
+++ b/proposed/simple-csharp-programs.md
@@ -40,7 +40,7 @@ Simple C# Programs are not a dialect of the C# language, like the interactive di
 
 ## User experience walkthrough
 
-Given the following code in `Hello.cs`:
+Given the following code in a singular `Hello.cs` in a directory:
 
 ```csharp
 WriteLine("Hello World");
@@ -56,7 +56,7 @@ Hello World
 Or with a `#!` directive,
 
 ```csharp
-#!/usr/bin/shared/dotnet
+#!/path/to/dotnet
 WriteLine("Hello World");
 ```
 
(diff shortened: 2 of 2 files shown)


proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+
+We'll need to decide if they can affect the set of references available in the file or not.
+
+### Build output
+
+The walkthrough states that the build output is a single executable file in the current working directory. This is based on the hypothesis that this is the least surprising place to put the executable.
+
+* Is this hypothesis based on reality?
+* Is this still true if you pull in a package?
+* Is this still true if you pull in a framework? (e.g., ASP.NET Core is hundreds of assemblies, where do they go?)
```
Tratcher, line 271: Most of those assemblies live in the shared framework, they do not end up in your bin dir by default.

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+and running:
+
+```bash
+$ ./hello.cs
+Hello World
+```
+
+The `using`s that are normally in a C# project template are implicitly available at design-time. This is accomplished with [Global Using Directives](https://github.com/dotnet/csharplang/blob/main/proposals/GlobalUsingDirective.md) that the build tooling will have available.
+
+The shebang is covered in [Shebang (#!) support](https://github.com/dotnet/csharplang/issues/3507).
```
MichalStrehovsky, line 72: The term shebang originates in the US Civil War when it was coined by the Confederate soldiers [with the meaning of](https://en.wiktionary.org/wiki/shebang) "temporary shelter".

[Wikipedia](https://en.wikipedia.org/wiki/Shebang_(Unix)) provides several other names for `#!` that might be more inclusive: sha-bang, hashbang, pound-bang, or hash-pling.

davidfowl: TIL...

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+
+```bash
+$ dotnet run
+dotnet bot
+```
+
+Or via a shebang,
+
+```csharp
+#r! /usr/bin/shared/dotnet
```
333fred, line 113:
```suggestion
#! /usr/bin/shared/dotnet
```

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+
+```bash
+$ dotnet run
+Hello World
+```
+
+Or with a shebang,
+
+```csharp
+#!/usr/bin/shared/dotnet
```
omajid, line 59: This path will most likely be different in different environments. For example, in Fedora it's `/usr/lib64/dotnet/dotnet` right now. Debian [might use `/usr/lib/x86_64-linux-gnu/dotnet`](https://wiki.debian.org/Multiarch/Tuples).

Maybe we should take a page from [a common convention for bash scripts](https://linuxize.com/post/bash-shebang/#using-shebang-in-bash-scripts) and use `/usr/bin/env dotnet`? That will use the `dotnet` in `$PATH`, so it should be more portable.

As a down-side, it means we can't pass any more arguments to the dotnet process via this shebang line.

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+
+For the purposes of this design, we will require that users create a project-based codebase if they wish to be grown up.
+
+## Design points / pivots
+
+The following is a non-exhuastive list of user experience considerations that go beyond the basic walkthrough above.
+
+### Multiple files
+
+We will need to decide if Simple C# Programs support more than one file. In other languages, like JavaScript and F#, you can have several loose files in your directory (or other directories) and include them into your program. For example:
```
omajid, line 197: What about the case of multiple files where each file is a separate/independent program? For example, what if I have a directory with (at least) 2 files:

- `check-server-status.cs`
- `reboot-server.cs`

Both of these would be independent stand-alone programs that can be executed individually via `./check-server-status.cs` and `./reboot-server.cs`. Would dotnet be able to build/run them as such? Or is it assumed that if there are multiple files they are related and need to be built together and only one can be executed?

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+**hello.fsx**:
+
+```fsharp
+printfn "Hello, World!"
+```
+
+and run it:
+
+```bash
+$ dotnet fsi hello.fsx
```
omajid, line 307: If we were to re-bolt the design proposed in this document onto F#, what would the shebang line look like?

proposed/simple-csharp-programs-motivation-walkthrough.md
sylveon: FWIW Wikipedia also says the shebang name might have been coined for Unix scripts independently of that earlier usage, as a contraction of hash bang or sharp bang.

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+$ dotnet build
+$ ls
+Hello.exe
+$ ./Hello.exe
+Hello World
+```
+
+When built, it creates a single executable that it drops in the current working directory.
+
+### Referencing a package
```
loic-sharma, line 88: How was the need for NuGet packages in simple C# evaluated? From my understanding, the goal is to make C# more approachable. The `#r "nuget: X"` syntax is a special one-off feature that newbies will need to "unlearn" when graduating to a project file.

Will this "unlearning" process confuse new users? Would it be better to save NuGet support to project files only?

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+
+```csharp
+#!/usr/bin/shared/dotnet
+WriteLine("Hello World");
+```
+
+and running:
+
+```bash
+$ ./hello.cs
```
jkotas, line 66: Does this create any artifacts on disk or is it all compiled and executed in memory?

proposed/simple-csharp-programs-motivation-walkthrough.md
cartermp: There's upside and downside to doing/not doing it. Not necessarily the same audience, but for F# developers the `#r` nuget support has gone absolutely gangbusters, and ex-python programmers praise the simplicity. It's also supported in .NET Interactive and is the primary way to do things like chart a data set. But your points about learning two different ways is completely valid. Some degree of this may be necessary in the future. But this could be done incrementally as mentioned, so that package support comes later (if at all).

proposed/simple-csharp-programs-motivation-walkthrough.md
davidfowl: I hope it would be like go lang and use the temp folder.

proposed/simple-csharp-programs-motivation-walkthrough.md
cartermp: The name can easily change, although we'll have to balance this out with following/not following a convention used in UNIX land. For now, since the C# proposal uses the name, I'll leave it as-is. If we decide to change the name for other reasons, it will change here.

proposed/simple-csharp-programs-motivation-walkthrough.md
zivkan: If it uses the temp folder will it clean up after itself?  My temp folder is nearly as big as my nuget global packages folder, since nothing cleans the temp folder, not even a reboot. 😞

proposed/simple-csharp-programs-motivation-walkthrough.md
rgwood: I think this depends on the audience. Newcomers might not need NuGet packages, but I can see a lot of experienced C# developers using Simple Programs as a low-ceremony way to knock out ad-hoc programs and scripts. That latter audience will need+want package support.

proposed/simple-csharp-programs-motivation-walkthrough.md
```diff
@@ -0,0 +1,311 @@
...
+
+In effect, the shebang's semantics could span a spectrum of "I need the path and command to literally execute this file" towards "I add context that affects the references available to this file".
+
+We'll need to decide if they can affect the set of references available in the file or not.
+
+### Build output
+
+The walkthrough states that the build output is a single executable file in the current working directory. This is based on the hypothesis that this is the least surprising place to put the executable.
+
+* Is this hypothesis based on reality?
```
danmoseley, line 269: For C I think it is --  `gcc -o test test.c` is the simplest way to play with a test program. If you are running the program as a separate step (ie with `dotnet build`) then there is no more discoverable place than the same directory you're in.

gulshan:
Do the commands `dotnet run` and `dotnet build` without filename param assume only a single C# file in the directory? I think, having multiple C# files in a directory will be more common and running them using their file names will be more useful. And if loading one file from another is supported, this determines which file will be the entry-point, without guessing by dotnet tooling.

Also, is there a chance to evolve CSX to meet these introductory/scripting demands, while keeping C# structured? Because when I see a `.fs` and a `.fsx` file, I know they behave differently, and they are meant to be used differently. There is no cognitive load of getting the context to really know how the file will behave. Where `#load` or `#r "nuget..."` can and cannot be used, no unlearning needed. Just my opinion though. I don't know much about the obstacles here.

proposed/simple-csharp-programs-motivation-walkthrough.md
cartermp: I suspect that we'd need an explicit directive like `#load` in F# or `require` in JS. Otherwise each would be assumed be a separate program, and you'd need to specify which to run (e.g., `dotnet run reboot-server.cs`). I'll likely add this to the walkthrough itself, but I expect this to be a point of debate.

proposed/simple-csharp-programs-motivation-walkthrough.md
cartermp: A "Simple F# Programs" proposal would probably work similarly, assuming that F# supports shebang syntax and global `using`s inferred by the build environment. I don't think there's anything stopping F# from using the same syntax.

proposed/simple-csharp-programs-motivation-walkthrough.md
MichalStrehovsky: > convention used in UNIX land

The Wikipedia article mentions several other names for this that are already established  - we can just pick one of those names. A soldier's "temporary shelter" in a war is often someone else's "home", and they often don't wait for an invitation. I see no good reason to use that language in our repos if there's established alternatives.

It should be easy enough to change in the C# spec as well - @333fred what do you think?

proposed/simple-csharp-programs-motivation-walkthrough.md
cartermp: Thanks, will add this to the design consideration section.

proposed/simple-csharp-programs-motivation-walkthrough.md
cartermp: I clarified the choices here and made it an explicit section under Design points / pivots.

cartermp: I won't make a call in this doc on the right choice though. I am personally for using the temp folder, but I'm cool with whatever folks who are building this feel is the best choice.

proposed/simple-csharp-programs-motivation-walkthrough.md
333fred: It's changed.

cartermp: On this end too

KevinCathcart:
It is also worth pointing out that `#r` and `#load` are part of the language, even if currently only supported in scripting scenarios. Even if this is a new third mode (new `SourceCodeKind` member in the compiler) somewhere between scripting and traditional, it makes sense to still reuse these directives.

Furthermore it makes sense to have the compiler enforce the exact same semantics for these directives as for c# scripting. If new features like framework references are needed here, then scripting could benefit from them too. There are already requests  for scripting (the official csi.exe and/or VS interactive window) to support the same nuget package reference syntax that dotnet-script and .NET Interactive support (see dotnet/roslyn#43918 for one such request).

Other than this needing to be something the host deals with resolving (the compiler could certainly parse this for the host's benefit, but I doubt the compiler wants to be in charge of actually fetching the nuget packages) I see no reason this should not be supported everywhere the `#r` syntax is allowed.

I do have mixed feeling about introducing a third type of csharp mode for this. The simple program language support has already made the scripting and non-scripting a lot more similar than they previously were, yet the scripting language has some "features" that may be undesirable for simple program usage.

On the other hand. the whole point here seems to be making it possible to run these files a lot like an interpreted scripting language (python, ruby, etc), including hash bang support, which kind of points in the direction that this really should just be using the scripting language despite its differences from full c#, or perhaps have this be some form of stricter scripting mode that disables some of the undesirable differences.

cartermp:
@KevinCathcart A key thing, which perhaps isn't expressed well in this document, is that this is **not** another dialect of C# like the scripting dialect is. It's just C#. Everything here (top-level statements, global `using`, hashbang support) is proposed as simply just being C# proper. There's no "simple mode" activated or anything. If you have these in a C# file in a project-based codebase, nothing behaves differently.

There's some more details to write up here, such as also allowing `#r "path-to-assembly"` (no reason for that not to work) and re-using `#load` provided that multiple files are supported.

This is in contrast to the scripting dialect, which very much changes C# semantics such as having shadowing and disallowing namespace declarations. This proposal is not about unifying the scripting dialect with C# proper. It's just about pulling in existing and proposed C# features + updating build tooling to allow using them together without a project.

mhutch:
Making `#r "nuget:[package]"` work correctly in a project context would be difficult as you'd have to reconcile the C# compiler's `#r` package graph with the assets already resolved by NuGet restore, or somehow pull the `#r "nuget:[package]"` references out and handle them in the existing NuGet restore. Similarly you'd have to reconcile `#r "[dl]"` assembly references and RAR.

cartermp:
Yeah, in a project context they'd likely get ignored, at least that's the current thinking. Any `#r` or `#load` would just be ignored. But it's worth thinking about more

zivkan:
Some packages, for example SQLite, need to run msbuild props/targets to work. This is most common for packages that are managed wrappers of native dlls, and the packages support both packages.config as well as PackageReference, hence why they contain msbuild props/targets and not just use NuGet's `runtimes/` folder.

Maybe this question is too detailed for a high level design, but should these packages work with the single file `#r` syntax? Only support packages that do not have msbuild props/targets and warn/error that the customer must use a project file to use the package?

cartermp:
@zivkan that's a good point of consideration. I haven't dived into details like this just yet (maybe that comes later?) but in my head I imagine that an in-memory or temp project file is involved. For example, here's that package working in F# Interactive, which has `#r "nuget:..."`. It's actually the same mechanism that is used in .NET Interactive:

image

If a project file is generated behind the scenes (temp or in memory) then other msbuildy things would also work, like .editorconfig plumbing into the compiler, directory.build.props, etc.

KevinCathcart:
> @KevinCathcart A key thing, which perhaps isn't expressed well in this document, is that this is **not** another dialect of C# like the scripting dialect is. It's just C#. Everything here (top-level statements, global `using`, hashbang support) is proposed as simply just being C# proper. There's no "simple mode" activated or anything. If you have these in a C# file in a project-based codebase, nothing behaves differently.
> 

So you are proposing removing the errors for those directives for normal project based compilation to avoid even being a minor dialect from the language POV? (arguable still a minor dialect none-the-less, since you are applying different meaning to those directives then I would get by typing `csc.exe filename.cs {/reference:...}` but fair enough.)

I don't see any discussion of possibly allowing no-op or fully functional #r or #load directive in normal compilation in the LDM notes, so I'm guessing they have not weighed in on if doing that is sensible? Obviously it is trivially possible to implement, but that does not mean it is a good idea.

333fred:
> I don't see any discussion of possibly allowing no-op or fully functional #r or #load directive in normal compilation in the LDM notes, so I'm guessing they have not weighed in on if doing that is sensible? 

It's tentatively on the schedule for next week. Suffice to say that we have a good deal of interest in the idea, as we've been discussing it in various forms (mostly via the top-level statements feature) for a few years now. https://github.com/dotnet/csharplang/tree/main/meetings/2021#may-12-2021

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,428 @@
...
+```bash
+$ dotnet run
+Listening on localhost://12345
+```
+
+The syntax for coding against a framework/sdk could also be more similar to the `#r "nuget:"` syntax shown earlier instead of being a `#!` directive.
+
+## `dotnet` command support
+
+Only `dotnet run` and `dotnet build` are supported with Simple C# Programs. All other `dotnet` commands will error out, indicating that you need a project.
```
jeffhandley, line 163: `dotnet test` would be interesting too. The file could have top-level methods that are unit tests, and `dotnet test` would run them.

While working in JavaScript, I found standalone unit test files useful when sharing a minimal repro for a library bug--that's something especially hard to do with .NET today.

KevinCathcart:
> > I don't see any discussion of possibly allowing no-op or fully functional #r or #load directive in normal compilation in the LDM notes, so I'm guessing they have not weighed in on if doing that is sensible?
> 
> It's tentatively on the schedule for next week. Suffice to say that we have a good deal of interest in the idea, as we've been discussing it in various forms (mostly via the top-level statements feature) for a few years now. https://github.com/dotnet/csharplang/tree/main/meetings/2021#may-12-2021

Cool. So the question of if this should effectively be a super minor dialect, or if `#r` (and possibly `#load`) should be supported in project based builds is yet to actually be settled. Here are some things that probably ought to be discussed at LDM:
- If those features act as more than no-ops in a project-based build, these would need to be reported back to MSBuild, so it could write them to a cache file. Otherwise MSBuild has no way of doing proper incremental builds, since these would be new inputs it did not know needed to be checked for updates. Off the top of my head, I'm not sure of any other case where the compiler ends up using input files that MSBuild did not explicitly specify, but there could already be such cases I am simply not aware of. There are also project system impacts, etc.
- If they do act as no-ops, then it could be really confusing to have them silently be ignored in project based compilation. They could issue a warning or error message in this case. Issuing an error message would be creating a new minor dialect at the language level, while the warning message would not. Obviously people who have enabled warnings as errors would error in this case. However they would be transitioning from an error message that the directives are only supported by scripting mode into a warning promoted to error that the directives are having no effect, so that is unlikely to be an issue. 
- There are other open questions. Is something like .NET Interactive's `#i` directive to specify additional nuget feeds needed? Or is just requiring a nuget.config in the same folder fine? Unlike nuget based `#r` which is supported by both dotnet-script and .NET Interactive, this one looks to be exclusive to interactive, as are the various pesudo-hashbang directives that interactive supports.

I also note that .NET Interactive does not currently rely on the compiler parsing the NuGet version of `#r` directives, but instead preprocesses them, letting the non-NuGet cases fall through and be processed by the compiler. In the case of C# I'm not sure this behavior is strictly required for the simple case, and a custom MetadataReferenceResolver could probably handle things, but .NET Interactive supports these directives for thinks like PowerShell which don't have any native support for such directives, so it needs to do this as a preprocessing step for them. 

Preprocessing is probably best avoided, since making that work as expected with things like ifdefs is really tricky. On the other hand, for nuget packages that include MSBuild snippets that are required for proper functioning, loading them in the compiler is potentially too late. I'm also not sure if packages including analyzers or source generators would be a problem during compilation, or if those could be successfully added to a compilation already in progress,

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,428 @@
...
+```
+
+The `#r "nuget:..."` syntax is already supported by C# today in .NET Interactive. It is a shared implementation between C# and F#. The syntax and mechanism could be re-used here.
+
+### Coding against a framework/sdk like ASP.NET Core
+
+You can also reference a framework/sdk (nomenclature up for debate) via a `#!` directive:
+
+```csharp
+#!/usr/bin/share/dotnet-aspnet
```
Zaid-Ajaj, line 140: Should this be _shared_ instead of _share_?
```
#!/usr/bin/shared/dotnet-aspnet
```
same potential typo below

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,428 @@
...
+Given the following code in `Hello.cs`:
+
+```csharp
+WriteLine("Hello World");
+```
+
+You can just run it:
+
+```bash
+$ dotnet run
```
st0le, line 52: dotnet run Hello.cs ?

333fred:
> So the question of if this should effectively be a super minor dialect, or if #r (and possibly #load) should be supported in project based builds is yet to actually be settled

Well, it's been a key driving principle that it's _not_ a dialect of C#. Also, while it's certainly up for discussion, I don't believe anyone has discussed or has intentions of putting a NuGet resolver into the compiler itself: it will almost certainly be a separate tool invoked as part of the pipeline. Whether that tool has differences when invoked as part of a project file or when just running a `.cs` file certainly still needs to be designed, but I don't expect roslyn itself to know anything about whether it's just compiling a .cs file or compiling a traditional .csproj project.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,428 @@
...
+Hello World
+```
+
+Or with a `#!` directive,
+
+```csharp
+#!/usr/bin/shared/dotnet
```
tmds, line 59: All of these could/should be `#!/usr/bin/dotnet` if you assume Microsoft packages.
And as noted later in the doc `#!/usr/bin/env dotnet` is the portable way.

proposed/simple-csharp-programs.md
omajid: The path is currently actually `/usr/share/dotnet/dotnet` for Microsoft-built Deb/RPM packages for Linux.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,428 @@
...
+```
+
+However, the downside is that you can't pass any more arguments to the `dotnet` process. It's a matter of portability vs. utility.
+
+### Build output
+
+The walkthrough states that the build output is a single executable file in the current working directory. This is based on the hypothesis that this is the least surprising place to put the executable.
+
+* Is this hypothesis based on reality?
+* Is this still true if you pull in a package?
```
KevinCathcart, line 366: For a normal package based project, the build output consists of several files, only a subset of which are actually required to run the result:
- First we have the resulting DLL. (.NET Core does not create managed EXEs).
- There is (optionally, enabled by default) a pdb file with debugging information.
- There is an executable host that runs the the dll.
- There is the deps.json file that specifies dependency dlls.
- There is the runtimeconfig.json file that specifies which runtime to use to run this, and which shared frameworks are needed.
- There is a runtimeconfig.dev.json file that specifies alternative locations to look for assemblies, such as the nuget cache folders.
- Lastly (optionally, enabled by default) it copies referenced dlls from nuget packages. This is not needed, since they can be accessed from the nuget cache. It can be disabled with `<CopyLocalLockFileAssemblies>false</CopyLocalLockFileAssemblies>`, and everything still works.

This many files is quite simply too many to place along side the code.

As an alternative, publish could be used under the hood.  If you request creating a framework-dependent "single file" executable, on windows you can get as few as 2 files, the .exe and the .pdb (and of course the pdb generation could be supressed). (I've not tested, but I'm guessing Linux also results in 2 files, and I've no clue about macOS). Nuget referenced dlls get embeded in the exe, and the exe relies on the shared runtime installation. This might well be the result that makes the most sense for simple program builds.

proposed/simple-csharp-programs.md
cartermp: Replacing with `/path/to/dotnet` for now

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+The `#!` directive is covered in [Hashbang (#!) support](https://github.com/dotnet/csharplang/issues/3507).
+
+### Building instead of running
+
+You can also build it with `dotnet build`:
+
+```bash
+$ dotnet build
+$ ls
+Hello.exe
```
lambdageek, line 83: nit: `hello.cs` is also in that directory

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+```bash
+$ dotnet run
+Listening on localhost://12345
+```
+
+The syntax for coding against a framework/sdk could also be more similar to the `#r "nuget:"` syntax shown earlier instead of being a `#!` directive.
+
+### Running a specific file
+
+If you have multiple files in the same directory, `dotnet run` won't work by default. You'll need to specify the file you wish to run:
```
lambdageek, line 163: So if I do `dotnet build` followed by `dotnet run` it won't work anymore because there are now 2 files in the directory?

rgwood: I interpreted that as "multiple _C#_ files in the same directory," presumably identified by the .cs extension. But that would be a good point to clarify.

RichiCoder1:
> @zivkan that's a good point of consideration. I haven't dived into details like this just yet (maybe that comes later?) but in my head I imagine that an in-memory or temp project file is involved. For example, here's that package working in F# Interactive, which has `#r "nuget:..."`. It's actually the same mechanism that is used in .NET Interactive:
> 
> image
> 
> If a project file is generated behind the scenes (temp or in memory) then other msbuildy things could (in theory?) also work, like .editorconfig plumbing into the compiler, directory.build.props, etc.

Would a net effect of this also allow something like Source Generators? I could see some interesting applications specifically for using source generators specifically with simple programs.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+However, that seems unlikely given that the current syntax is already shipping in .NET Interactive.
+
+### Should there be support for a user-specified TFM?
+
+So far, the assumption is that a Simple C# Program will simply execute against the TFM in the SDK it's built from. If you run from a .NET 7 SDK, it targets .NET 7. There's no way to configure that unless you use a project file.
+
+Another alternative is to use a hashbang like `dotnet-script` to support a specific TFM, like so:
+
+```csharp
+#!net5
```
smoothdeveloper, line 399: `#runtime "net5"`

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+and run it:
+
+```bash
+$ dotnet fsi hello.fsx
+Hello, World!
+```
+
+However, the model is based on FSX and F# Interactive, which is not a realistic approach for C# to do. The CSX and C# dialects are subtly different, there is legacy behavior to reconcile, and more.
+
+A "Simple F# Programs" proposal could also be implemented. However, it would require the F# language to support `#!` in non-scripts and global `open`s.
```
smoothdeveloper, line 441: `#load "some.cs"` should work in F# and conversely in C# (and VB.NET).

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+$ ./hello.cs
+Hello World
+```
+
+When running via `dotnet run` or the `#!` directive, artifacts may be created on disk in the temp folder. They could also be generated in memory.
+
+The `using`s that are normally in a C# project template are implicitly available at design-time. This is accomplished with [Global Using Directives](https://github.com/dotnet/csharplang/blob/main/proposals/GlobalUsingDirective.md) that the build tooling will have available.
+
+The `#!` directive is covered in [Hashbang (#!) support](https://github.com/dotnet/csharplang/issues/3507).
+
```
smoothdeveloper, line 75: maybe have a reference here to the "Multiple files" section of the document just for heads up.

aelij:
> in a project context they'd likely get ignored, at least that's the current thinking. Any #r or #load would just be ignored. But it's worth thinking about more

Why not support `#r` in project contexts as well? It would make .cs files more encapsulated, easily movable between projects, and avoids manually adding usings. Go does something similar with `import` where you can just specify remote and local packages.

proposed/simple-csharp-programs.md
smoothdeveloper: `dotnet run` should, in case it can't identify "entry file", prompt the user to pick/`dotnet run file.ext` with the actual list of files present.

This is in order to still be more helpful than a bland error message.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+* `#r "nuget:..."` or similar directive that resolves packages. Also works in tooling, so you get IntelliSense for the package
+* `#!/usr/bin/share/dotnet-aspnet` or `#r "framework:..."` support, allowing you to run against ASP.NET Core (`server.cs`)
+
+This would require some design work to arrive at the best syntax and semantics. It is also expected that tooling works when these are incorporated.
+
+### Phase 3 - multiple files
+
+This phase would bring in support for pulling in multiple files for a single program. Imagine a `server.cs` with a `network-utils.cs` file, where `network-utils.cs` is used by `server.cs`.
+
+Some guardrails would have to be built into place.
```
smoothdeveloper, line 250: Considering if `#load "some other.cs"` should work like in fsx scripts, and identify if a single file has top level statements (that would be the program).

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+utils.cs
+```
+
+In Directory Two, the `utils.cs` file is used by `reboot-server.cs`. A developer may expect to be able to just run `dotnet run` in that directory and have the program execute. But if they are in Directory One, they would have to be explicit about which file they are running.
+
+### Coding against a framework alternative syntax
+
+The walkthrough proposes the `#!` directive `#!/path/to/dotnet-aspnet`, but the .NET CLI doesn't have a command to "pull in the ASP.NET Core references", so it wouldn't map to something like that.
+
+An alternative could be to extend `#r` like so:
```
tmds, line 327: This sounds like it's one or the other. Both can be supported at the same time.

smoothdeveloper (commented):
Adding a section about C# compiler error messages, and the expectation for them to be more helpful in context without tooling would also be great.

Some minor comments along the lines.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+It could be extended/changed in a few ways:
+
+* `/path/to/dotnet run` makes it explicit that the command is to run the file
+* `/path/to/dotnet-aspnet` is an alternative to `#!aspnet` that brings ASP.NET Core into scope for the file (substitute a different framework/sdk here)
+
+In effect, the `#!` directive's semantics could span a spectrum of "I need the path and command to literally execute this file" towards "I add context that affects the references available to this file".
+
+We'll need to decide if they can affect the set of references available in the file or not.
+
+### Using `#!` to point to `dotnet` or point to `env`
```
tmds, line 359: The `#!` will make it possible to execute these files on Linux just using the filename.

Is it desired/possible on Windows too? Maybe by having a dedicated extension?

proposed/simple-csharp-programs.md
cartermp: Yeah, but I don't think we want two ways to do the same thing here.

cartermp: I'm mostly in favor of not having a hashbang approach for this, actually. `#r` is the mini-"language" of referencing things in this kind of context today, and to use ASP.NET you do need to reference the right framework reference. It just happens that the way you express that in a project file is with the SDK attribute.

333fred: Additionally, the hashbang approach would require a separate executable for every sdk, which isn't good for 3rd-party sdks.

jmarolf:
>  I haven't dived into details like this just yet (maybe that comes later?) but in my head I imagine that an in-memory or temp project file is involved.

imho the semantics should be similar to how AssemblyInfo.cs works: a default is generated for you, but you can provide your own if you need to customize it. So if the only thing I need to customize in my project file is:
- target framework
- set of nuget packages

Then you don't need a project file and those can be inferred for the C# files themselves.

This would have the consequence that:

```cmd
dotnet build <some folder>
```

would act as if it were running as 

```cmd
dotnet build <some folder>\<some folder>.csproj
```

(assuming only C# files are in that folder, I assume this should error if there are cs/vb/f# files in a directory, same situation as when you run `dotnet build` in a folder with multiple project files.).


This also has several behavioral fallouts:
- `Directory.Build.props` would "Just work" in these projects
- `editorconfig` and `globalconfig` files would as well (though editorconfig will likely work regardless as it is folder-based)

I think the advantage of these semantics is it is easy to explain when to you a project file:
- Do you need some custom build or pack logic? Ok you need a project file to specify those things.
- Is everything you are doing just the defaults or are all your common settings coming from `Directory.Build.props`? No need for a project file just run `dotnet <command>` on that folder

NOTE: I am also assuming that _any_ dotnet cli command that operates on projects today would work following this scheme. That would mean:
- dotnet add
- dotnet build 
- dotnet build-server
- dotnet clean
- dotnet msbuild
- dotnet pack
- dotnet publish
- dotnet remove
- dotnet restore
- dotnet run
- dotnet test
- dotnet vstest

proposed/simple-csharp-programs.md
tmds: I'm also in favor of the `#r` alternative. Maybe the design doc can highlight it is more extensible.

zivkan:
> I am also assuming that any dotnet cli command that operates on projects today would work following this scheme.

`dotnet add reference`, `dotnet remove reference` are used for project (p2p) references. If we're going to support p2p references, then this is no longer just about "simple programs" (yes, in my opinion a program that spans multiple assembly is no longer simple). The spec explicitly calls out that programs can "grow up" (a name which I don't like at all, FWIW), so I personally think it's completely reasonable to not support all dotnet cli commands. Otherwise we're effectively trying to make project files obsolete.

For the same reason, I wouldn't expect `dotnet msbuild` to work with simple programs.

While `dotnet new classlib; dotnet pack` does work, high quality packages should have additional metadata, currently only specified in the project file. Furthermore, I associate "program" with "exe", so "simple programs" to me does not mean "simple libraries" or "simple packages", so I don't make the assumption that `dotnet pack` should work in a directory with only `program.cs`.

It seems to me the ideal/preference of making all dotnet cli commands work consistently for directories without project files does not line up with the spec's proposal that programs will need to "grow up" to have project files.  Otherwise I think this spec should be proposing that project files become optional/obsolete, rather than being only for simple programs.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+```
+
+and running:
+
+```bash
+$ ./hello.cs
+dotnet bot
+```
+
+The `#r "nuget:..."` syntax is already supported by C# today in .NET Interactive. It is a shared implementation between C# and F#. The syntax and mechanism could be re-used here.
```
jcouv, line 133: I think the proposal if for C#-proper to tolerate but ignore such directives. Those directives would be picked up by the SDK to use in synthesized project file.
Let's confirm in LDM on Wednesday

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+
+```bash
+$ dotnet build
+$ ls
+Hello.exe
+$ ./Hello.exe
+Hello World
+```
+
+When built, it creates a single executable that it drops in the current working directory.
```
jcouv, line 88: Would we also drop the synthesized project file?

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,461 @@
...
+Hello World
+```
+
+When running via `dotnet run` or the `#!` directive, artifacts may be created on disk in the temp folder. They could also be generated in memory.
+
+The `using`s that are normally in a C# project template are implicitly available at design-time. This is accomplished with [Global Using Directives](https://github.com/dotnet/csharplang/blob/main/proposals/GlobalUsingDirective.md) that the build tooling will have available.
+
+The `#!` directive is covered in [Hashbang (#!) support](https://github.com/dotnet/csharplang/issues/3507).
+
+### Building instead of running
```
jcouv, line 76: We'll want to clarify what's the TFM for the implicit project file.

jcouv (commented):
Done with review pass (iteration 6)

proposed/simple-csharp-programs.md
cartermp: Maybe, although `#!` really isn't a convention on Windows.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,499 @@
...
+
+## `dotnet` command support
+
+Only `dotnet run` and `dotnet build` are supported with Simple C# Programs. All other `dotnet` commands will error out, indicating that you need a project.
+
+## Synthesized project file
+
+Behind the scenes, a likely implementation will involved a synethesized project file. This would then have all of the information to make the `dotnet` commands work, and open us up to allowing more `dotnet` commands in the future.
+
+A synthezized project file could live either in memory or in the temp folder.
```
jcouv, line 217: I'm sure that'll come up in next round of discussion with SDK team: we'll want to specify what that project looks like.
I assume the `OutputType` would be `Exe`, for instance.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,499 @@
...
+When built, it creates a single executable that it drops in the current working directory.
+
+### Referencing a package
+
+You can also pull in a package (syntax perhaps subject to change):
+
+```csharp
+#r "nuget: Newtonsoft.Json"
+
+record Person(string Name);
```
jcouv, line 100: nit: As you mentioned, the record declaration should come after the top-level statements.

proposed/simple-csharp-programs.md
333fred, line 217:
```suggestion
A synthesized project file could live either in memory or in the temp folder.
```

cheong00:
For me, I think another possibility is to put it like those .NET v2 "website project", where we have a .CONFIG file provide configurations common to files in folder, all .aspx/,asmx files can be treated as single application, and you can freely reference any addition binaries in bin folder.

File location and config file syntax could be rearranged it to make it configuration folder like .git, possibly with simple configuration UI like that of TortoiseGit or IIS management for those less comfortable with editing config files directly.

Now when I work on .CS files inside that folder, I can be sure that the basic settings I used to compile to be the same and I don't need to re-specify it every time.

jmarolf:
> If we're going to support p2p references, then this is no longer just about "simple programs" (yes, in my opinion a program that spans multiple assembly is no longer simple).

There is no technical reason the `#r` or `#load` syntax couldn't be updated to support project references. But supposing we do not allow this, from a tooling perspective, how do you see users adding new packages? Do they get completion in their code file within `#r`?

I think it is good to have the discussion about where the "line" is terms of what is supported. I took a position very much on the "most possible things working" side because it is easy to explain. What works? Everything. My internal model is "If you don't need project files you should not require them". Project to project references very much feel like a case where having project files makes things easier. Perhaps a project-less program could reference other projects, I could see the utility of this. But regardless of the technical hurdles I think trying to reference a folder or file instead of a direct project would be error prone. imho we should allow p2p references so long as it is not ambiguous. So a single file referencing projects would be fine.

> so I personally think it's completely reasonable to not support all dotnet cli commands. Otherwise we're effectively trying to make project files obsolete.

I don't think the goal is to make them obsolete, just unnecessary for "simple" cases. You used to need to specify AssemblyInfo.cs files, now those are generated for you. You used to need to create `Main` methods in C#, now those are implicit. The entire re-design around project files for .NET was about not requiring you to specify the defaults over and over. This just feels akin to that effort. It will be easier to scaffold assemblies and their relationships to each other via project files. Users will migrate to them when they have a need. But they shouldn't be forced to use them if they offer no benefit (which is the same logic used for the features listed above).

cartermp:
I'll close this one since I don't work on the product anymore and, although I'd like to see this through, I don't have the time nor energy to see this proposal through to completion. Anyone else who's interested can give it a whirl, though.

PR #215, opened 2021-05-03
Add proposal design of rate limits

JunTaoLuo:


proposed/resource-limit/resource-limit.md
```diff
@@ -0,0 +1,316 @@
...
+
+We expect that the user can opt between different implementations of the proposed interface to choose between different rate limiting algorithms. For example, between an implementation that uses token bucket vs one that makes rate limiting decisions based on ambient environment such as CPU usage. Also, different implementations may choose to use a queue/stack/priorityqueue when waiting for resources to become available.
+
+### Non-Goals
+
+TDB
+
+## Stakeholders and Reviewers
+
+- .NET team (System.Threading)
```
jkotas, line 27: Networking too? (https://github.com/dotnet/runtime/issues/52079#issuecomment-831375001)

jkotas: Also, add details about how you expect it to be used by the low-level networking stack.

KevinCathcart:
There is a mention of a generic interface to return specific details, but that does not seem present in the proposed API. Is this something that got cut from the design?

JunTaoLuo:
@KevinCathcart The reason that it's not included in the dotnet/runtime proposal is that we haven't identified any use cases for it within dotnet/runtime. So far, the use cases we have identified lie within aspnetcore. If there is a need for it in dotnet/runtime, I'll append it to the API proposal.

KevinCathcart:
I'm confused. I was talking about this line:
> 3. There are scenarios where additional metadata on the result is desired, such as reason phrases, response codes, percentage of rate saturation, etc. In these cases, use the generic interface and specify the `TResult`.

There is no `TResult` anywhere else in the design document. Sure, the aggregating interface is not being proposed for the framework, but that bullet point seems to describe functionality that neither interface mentioned in the design supports. I feel like that is left over from an older iteration of the design.

proposed/resource-limit/resource-limit.md
```diff
@@ -0,0 +1,312 @@
...
+- ASP.NET Core team
+
+## Design
+
+```c#
+    public interface IResourceLimiter
+    {
+        // an inaccurate view of resources
+        long EstimatedCount { get; }
```
drewnoakes, line 35: Count of what? `EstimatedAvailableCount`?

proposed/resource-limit/resource-limit.md
```diff
@@ -0,0 +1,312 @@
...
+        // If unsuccessful, throw
+        ValueTask<Resource> AcquireAsync(long requestedCount, CancellationToken cancellationToken = default);
+    }
+
+    // Represent an aggregated resource (e.g. a resource limiter aggregated by IP), no identified use within dotnet/runtime yet
+    // Abstraction likely to be part of ASP.NET Core
+    public interface IAggregatedResourceLimiter<TKey>
+    {
+        // an inaccurate view of resources
+        long EstimatedCount(TKey resourceID);
```
drewnoakes, line 52:
```suggestion
        long GetEstimatedAvailableCount(TKey resourceID);
```

proposed/resource-limit/resource-limit.md
```diff
@@ -0,0 +1,357 @@
...
+        public void Dispose()
+        {
+            _onDispose?.Invoke(this);
+        }
+
+        public static Resource NoopResource = new Resource(null, null);
+    }
+```
+
+The `Resource` struct is designed to handle the releasing of the resource. This way, the user won't be able to release more resources than was obtained, unlike Semaphores. We've also identified scenarios where additional information need to be returned by the limiter, such as reason phrases, response codes, percentage of rate saturation, retry after etc. For these cases, the object `Resource.State` can be used to store the additional metadata.
```
eerhardt, line 85: > We've also identified scenarios where additional information need to be returned by the limiter, such as reason phrases, response codes, percentage of rate saturation, retry after etc. For these cases, the object `Resource.State` can be used to store the additional metadata.

This scenario doesn't fit with the proposed `bool TryAcquire(long requestedCount, out Resource resource);` API. If a `bool TryXXX` method returns `false`, it isn't a canonical pattern that the caller can use the `out` parameter. You tried to get it, but it failed. In this case, how can information be passed back to the caller in order to tell them the reason or "retry after"?

JunTaoLuo: Ah good point. I was thinking that even when `TryAcquire` returns false, it can choose to return a `Resource`. It's just that it's guaranteed to not be null when `TryAcquire` returns true.

JunTaoLuo:
Hey all, from the feedback I've been getting, it seems like what folks were expecting was closer to a feature proposal so I've updated the doc here with more background information including scenarios and usage samples. I've undrafted the PR to encourage more conversations. Feel free to chime in here or contact me directly for disccussion/feedback!

proposed/resource-limit.md
```diff
@@ -0,0 +1,559 @@
...
+
+#### Partial Acquisition and Release
+
+Current design does not allow for partial acquisition or release of resources. In other words, either all of the request count is acquired or nothing is acquired. Likewise, when the operation is complete, all resources acquired via the limiter is released together. Though there are theoretically potential use cases for partial acquisitions and releases, no concrete examples has been identified in ASP.NET Core or .NET. In case these functionalities are needed, additional APIs can be added indepently from the APIs proposed here.
+
+#### Aggregated limiters
+
+A separate abstraction different from the one proposed here is required to support scenarios where aggregated resources are needed.
+
+For complex resources such as rate limit by IP, we don't want to have a rate limit per bucket (i.e. one rate limiter per IP Address). As such, we need an API where you can pass in a resourceID. This is in contrast to simpler resources where a key is not necessary, such as a n requests/second limit, where requiring a default key to be passed in becomes awkward. Hence the simpler API proposed here is preferred.
```
shirhatti, line 139: s/complex resources/high cardinality. Cardinality is the issue, not complexity

proposed/resource-limit.md
```diff
@@ -0,0 +1,559 @@
...
+
+This limiter keeps track of the number of concurrent resources in use. Conceptually, it will function in a similar way to a Semaphore/SemaphoreSlim but adapted to the proposed API. Like the rate limits above, it will also keep a queue of pending acquisition requests if configured to do so.
+
+#### Resource limit adoption
+
+Two BCL types where adoption of the resource limiters are beneficial and relatively simple are:
+
+- System.Threading.Channels
+- System.IO.Pipelines
+
```
shirhatti, line 124: I would include your proposed work in aspnetcore as part of your goals here

proposed/resource-limit/resource-limit.md
```diff
@@ -84,6 +84,35 @@ TDB
...
+1. Resource limit: this represents how many resource are available to be acquired immediately
+2. Queue/Stack limit: this represents how many requests can be queued
+
+This gives 3 scenarios :
+
+1. Count + RequestedCount <= Resource limit: In this case `TryAcquire` will return `true` and `AcquireAsync` will return immediately
+2. Resource limit < Count + RequestedCount <= Resource limit + Queue/Stack limit: In this case `TryAcquire` will return `false` and `AcquireAsync` will wait in a queue.
+3. Count + RequestedCount > Resource limit + Queue/Stack limit: In this case `TryAcquire` will return `false` and `AcquireAsync` will throw an exception immediately.
+
+In case we want to reduce the allocation of an `Exception` type in scenario 3, we can return a `bool` either as a tuple for `AcquireAsync`, i.e. `ValueTask<(bool, Resource)> AcquireAsync(long requestedCount, CancellationToken cancellationToken = default)`, or add a `bool` to the `Resource` struct.
```
davidfowl: Yes please. I think this could also look like a deconstruct method on Resource:

```C#
public struct Resource : IDisposable
{
    ...
    
    public Resource(object? state, Action<Resource>? onDispose, bool success)
    {
        State = state;
        _onDispose = onDispose;
        _success = success;
    }
    public void Deconstruct(out Resource resource, out bool success)
    {
          resource = this;
          success = _success;
    }

    public static Resource Success = new Resource(null, null, success: true);
    public static Resource Fail = new Resource(null, null, success: false);
}
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,559 @@
...
+### Non-Goals
+
+#### HttpClient, Streams, Sockets
+
+Although we anticipate that there will be applications in many .NET areas and components such as HttpClient, Streams and Sockets, these scenarios will be explored but not committed for adoption in .NET 6 since they may be more complex. The adoption of the resource limiting APIs in these areas can happen independently of this proposal in the future.
```
jkotas, line 129: It would be useful to design the APIs for HttpClient, Streams and Sockets, even if we are not planning the implement them in .NET 6.

proposed/resource-limit.md
```diff
@@ -0,0 +1,559 @@
...
+
+```c#
+namespace System.Threading.Channels
+{
+    public sealed class ResourceLimitedChannelOptions : ChannelOptions
+    {
+        public ResourceLimiter WriteLimiter { get; set; }
+    }
+
+    public sealed class ResourceLimitedChannel<T> : Channel<T> { }
```
jkotas, line 472: Nit: Do we really need `ResourceLimitedChannel<T>` in the public API surface?

davidfowl: No, I don't think it needs to be exposed

proposed/resource-limit.md
jkotas: This is important to decide where it needs to be in the stack. The key question - does it need to be in CoreLib?

davidfowl: @JunTaoLuo for Streams, we should show a hypothetical ResourceLimitedStream decorator (with settings for read and write). For Sockets I'm not sure we have an obvious place to put it at that layer since we don't rate limit at that level in the stack. For HttpClient there are places in the implementation we might want to use this for (HTTP2 flow control for example?).

KevinCathcart:
How composable is this?

Lets say I have some slow and expensive operation that will monopolize a database connection for a while. I may want both a concurrency limiter (to limit the proportion of my connection pool that can be used by this, which might also be shared with certain other slow operations that monopolize a db connection), in addition to a per-operation rate limit, and perhaps also global per IP address rate limit utilizing the aggregate version of this provided by aspnetcore.

There seems to be no way to make acquisition of all three limits an all-or-nothing type event (which might or might not even be desirable in the first place), so I need to carefully choose the order of evaluation of the limits. Should I evaluate my currency limit before or after the operation rate limit? (Doing so after means the operation rate limit is decremented even if it is later blocked by the concurrency late limit, which may be undesirable). Should my per operation rate limit be decremented if the request would already be blocked by my global IP address rate limit? Perhaps not, but that require that the global IP rate limit gets decremented even if the request is blocked at the operation level, since I would need to chose one order or the other.

KevinCathcart:
On the topic of concurrency limiters, can those really be zero allocation for the `count != 1 case`? Currently it sounds like the count is not stored in the resource, so either it is stored in the state object, or the limiter must create a closure over the count for the count for the dispose delegate. Both of those could require allocation, although I suppose it may be possible to have a single allocation per distinct count get cached and reused for future cases with the same count.

Along these same lines, the guidelines recommend that Dispose methods be idempotent, and cause no harm if called multiple times. User code frequently depends on this property. (For example many static analysis tools are unable to understand situations of say stream ownership being passed to the reader, and thus would recommend both the reader, and the stream get disposed.) I'm having trouble seeing how this can be supported allocation free.  

I'm not sure that not supporting idempotent dispose would be a viable option.

<details>
<summary>Why not supporting double dispose could be a problem</summary>
Remember that the user could store the resource as a field of a class, and dispose it when the class is disposed. If this class ("the inner class") is only disposing various fields, it may omit any logic to do nothing on a second dispose, due to relying on dispose of its fields being idempotent. If another class ("the outer class") conceptually owns the instance of the inner class, then it will dispose the inner class when disposed. If the user is using a static analysis tool, it might wrongly suggest that both the inner and outer class be disposed, since it does not know that the outer class is claiming ownership. This will result in inner being disposed twice, and thus the resource being double disposed.

 An alternate scenario that could result in double dispose is if these classes (not designed with DI in mind) got used with DI. A DI container must conservatively assume that it is responsible for disposing of any transient objects its creates. and the inner user class can easily be double disposed (once by the outer class, and once by DI), resulting is double disposing of the resource struct. 
</details>

proposed/resource-limit.md
eerhardt: > does it need to be in CoreLib?

We can start outside of CoreLib, and even outside of the shared framework, and move it down in the future. We've done that with other APIs.

davidfowl: I don't think it needs to be in CoreLib, but it needs to be low enough in the stack for things like Pipelines and channels to be able to use them. I don't know where that is right now. Maybe a new assembly.

davidfowl: System.Threading.ResourceLimits?

eerhardt: > Maybe a new assembly

This was my thinking - a new library.

> but it needs to be low enough in the stack for things like Pipelines and channels to be able to use them

Pipelines isn't in the shared framework today. Channels is. So if we needed Channels to depend on this new library, it would need to be in the shared framework.

Channels is extensible though, so if we really wanted to avoid putting this new library in the shared framework, I think we design a way to allow Channel integration, but without having Channels depend on the new library.

proposed/resource-limit.md
```diff
@@ -0,0 +1,559 @@
...
+// This creates a LimitedChannel<T>. Its operations are limited by the limiter specified,
+// which may be rate/concurrency based. LimitedChannel<T> with a concurrency limiter would
+// be similar to a BoundedChannel if the limiter's queue is disabled, see design section
+// for details on limiter's queue.
+var channel = Channel.CreateLimited<string>(limiter);
+
+channel.Writer.TryWrite("New message");
+```
+
+The benefit of tighter coupling with the BCL types is that it allows for simpler usage. In the former case, the user need to explicitly check the limiter every time an operation to the channel writer is requested. In the latter case, the channel writer can be used in the same manner as today and the channel writer itself will perform the check on the resource limiter on behalf of the user.
```
jkotas, line 63: I do not think that this example demonstrates the need to the tight coupling.

Why this cannot be:
```
var channel = new ResourceLimitedChannel<string>(limitter);

channel.Writer.TryWrite("New message");
```
No coupling required. This is similar pattern used in many other places in BCL. One example from many - `Stream` vs. `BufferedStream`.

proposed/resource-limit.md
jkotas: I agree that it feels like a higher level abstraction (from the runtime point of view), and thus can start outside the core framework or outside CoreLib at least.

From the examples I have seen so far, we will be typically creating wrappers of existing types that are resource limited. These wrappers do not need to live in the next to the type that they are wrapping. For example, the channels one does not really have to live in the System.Threading.Channels assembly.

If you agree, I would suggest to reword "Shared abstractions and implementation for the entire stack" paragraph. This paragraph makes it sound like that the core types have to be tightly coupled with the resource limiters. I have not seen a good reason for that so far.

proposed/resource-limit.md
davidfowl: Agreed, I don't think that's the point though. I think the point is that there are places we would want to be able to both use in the implementation and take as public API as part of the assemblies where these existing BCL types exist. So in your `ResourceLimitedChannel` example, that would live with the channels package and would reference these types. Same with pipelines

jkotas, line 63: The disadvantage of tight-coupling is that it is not pay-for-play. Everybody has to pay for the dependency even if they do not use it.

proposed/resource-limit.md
```diff
@@ -0,0 +1,559 @@
...
+
+4. Concurrency limiter
+
+This limiter keeps track of the number of concurrent resources in use. Conceptually, it will function in a similar way to a Semaphore/SemaphoreSlim but adapted to the proposed API. Like the rate limits above, it will also keep a queue of pending acquisition requests if configured to do so.
+
+#### Resource limit adoption
+
+Two BCL types where adoption of the resource limiters are beneficial and relatively simple are:
+
+- System.Threading.Channels
```
jkotas, line 122: Is there a reason why we are prioritizing the `Channels` here? Is the ASP.NET part of the work going to take advantage of it?

proposed/resource-limit.md
jkotas: I can believe you. The examples in the doc do not demonstrate the point well.

eerhardt: @davidfowl - concretely, is the only concern whether it is `Channel.CreateLimited(...)` vs. `LimitedChannel.Create(...)`?

Either way, you can still pass around the base `Channel` types, whether the limited Channel implementation lives in the Channels assembly or not.

davidfowl: Where it lives matters. When you have 2 types that don't want to take a dependency on each other you force everyone to what I call "bridge packages". Then you end up with a bridge package per concept if those concepts don't live at the same layer in the dependency graph. For example:

- HttpClient + ResourceLimits = one package
- Channels  + ResourceLimits = one package 
- Pipelines + ResourceLimits = one package

You see the pattern.

proposed/resource-limit.md
davidfowl: It's to help flesh out the design, not that we have an immediate need. Channels and Pipelines were the most obvious places to slot this into the BCL to make sure it could be used at this level in the stack (they basically do this in a very specific way today internally)

jkotas: FWIW, if we add a new feature without feedback loop, we typically get it wrong. If we do not have immediate use for Channels, I think it would be better to move it to future with the other BCL types.

I assume that the support in Pipelines will be used by ASP.NET. Is that right?

proposed/resource-limit.md
davidfowl: I would agree generally but the dependency here is too small to matter IMO.

proposed/resource-limit.md
davidfowl: I see this like an interface for SemaphoreSlim, I think it belong at that layer of the stack but I'm willing to not put this in CorLib until we're happy with the design. See my comment here about bridge packages https://github.com/dotnet/designs/pull/215#discussion_r631915858.

The argument being made here breaks down when it wants to be used inside of HttpClient (which is also in the shared framework).

proposed/resource-limit.md
davidfowl: Let me be clear, we already have 3 places where we are *concretely* going to be using it in ASP.NET Core, but we know the abstraction does not belong in ASP.NET Core. To make sure we didn't design something that can only be used in ASP.NET Core, we looked for places that do similar things in the BCL and @JunTaoLuo wrote prototypes to make sure the design could be applied to those types (and it works).

That is the purpose for listing Channels and pipelines.

proposed/resource-limit.md
JunTaoLuo: Good point, I'll update it.

proposed/resource-limit.md
JunTaoLuo: So far I've been prototyping with the assumption it will be a new library in the shared framework, coupling tightly with types such as Channels. But I'm willing to adjust this as necessary based on the discussions here.

proposed/resource-limit.md
jkotas: Is Pipelines support for resource limits going to be used by ASP.NET in .NET 6?

davidfowl: I don't know if we need that in .NET 6 no. What does that change?

JunTaoLuo:
@KevinCathcart In terms of composability, that's up to the user's implementation. Just like locks, the user needs to make sure to query the limiters in a consistent manner as to prevent deadlocks. Also, it's up to the user to ensure on failure to obtain all resources, any obtained resources are released as appropriate. An example of this can be found in our aspnetcore middleware proof of concepts where users can specify multiple resource limiters (both simple and high-ordinality such as per IP rate limit). 

As for dispose, you brought up many good points but they would be part of the implementation discussion. For example, yes we need to be careful about double dispose since the user can explicitly, though incorrectly, call `Resource.Dispose()` twice. However, the abstraction designed here is not opinionated about whether this is idempotent or not.

As for the allocations and dispose delegate capturing, those are great points to consider for specific implementations. For example, I anticipate that concurrency limiters can store the count on the `State` that is used by the dispose delegate.

proposed/resource-limit.md
jkotas: It means that we won't have the feedback loop to ensure that it is done right.

davidfowl: One option could be to replat the current limits on top of resource limiters.

proposed/resource-limit/resource-limit.md
JunTaoLuo: This has been adopted in the proposal

JunTaoLuo:
Hey all, I've updated the proposal with a tweaked API. Instead of `bool TryAcquire(long requestedCount, out Resource resource)`, I'm convinced it's better to put the bool indicating acquisition success on the `Resource` itself. i.e. `Resource Acquire(long requestedCount);`

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+
+Current design does not allow for partial acquisition or release of resources. In other words, either all of the request count is acquired or nothing is acquired. Likewise, when the operation is complete, all resources acquired via the limiter is released together. Though there are theoretically potential use cases for partial acquisitions and releases, no concrete examples has been identified in ASP.NET Core or .NET. In case these functionalities are needed, additional APIs can be added indepently from the APIs proposed here.
+
+For example, this will likely involve a modified `Resource`:
+
+```c#
+public struct Resource : IDisposable
+{
+    // This represents whether resource acquisition was successful
+    public bool IsAcquired{ get; }
```
davidfowl, line 149:
```suggestion
    public bool IsAcquired { get; }
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+    public Resource(long count, bool isAcquired, object? state, Action<Resource>? onDispose);
+
+    // Release a specific amount of resources, if possible
+    public void Release(long releaseCount);
+
+    // Release remaining resources
+    public void Dispose();
+
+    // This static `Resource` is used by rate limiters or in cases where `Acquire` returns false.
+    public static Resource SuccessNoopResource = new Resource(true, null, null);
```
davidfowl, line 168:
```suggestion
    public static Resource Success = new Resource(true, null, null);
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+    public static Resource FailNoopResource = new Resource(false, null, null);
```
davidfowl, line 169:
```suggestion
    public static Resource Failed = new Resource(false, null, null);
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+However, we have not yet found any use cases for aggregated limiters in dotnet/runtime and hence it's not part of this proposal. If we eventually deem that such aggregated limiters are needed in the BCL as well, it can be added independently from the APIs proposed here.
+
+Sample API:
+
+```c#
+// Represent an aggregated resource (e.g. a resource limiter aggregated by IP)
+public abstract class AggregatedResourceLimiter<TKey>
+{
+    // an inaccurate view of resources
+    public abstract long EstimatedCount(TKey resourceID);
```
davidfowl, line 188:
```suggestion
    public abstract long EstimatedCount(TKey resourceId);
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+        // Wait until the requested resources are available
+        // Set requestedCount to 0 to wait until resource is replenished
+        abstract ValueTask<Resource> AcquireAsync(long requestedCount, CancellationToken cancellationToken = default);
+    }
+
+    // Represents a resource obtained from the limiter. The user disposes this type to release the acquired resources.
+    public struct Resource : IDisposable
+    {
+        // This represents whether resource acquisition was successful
+        public bool IsAcquired{ get; }
```
davidfowl, line 235:
```suggestion
        public bool IsAcquired { get; }
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+        // Set requestedCount to 0 to get whether resource limit has been reached
+        abstract Resource Acquire(long requestedCount);
+
+        // Wait until the requested resources are available
+        // Set requestedCount to 0 to wait until resource is replenished
+        abstract ValueTask<Resource> AcquireAsync(long requestedCount, CancellationToken cancellationToken = default);
+    }
+
+    // Represents a resource obtained from the limiter. The user disposes this type to release the acquired resources.
+    public struct Resource : IDisposable
```
davidfowl, line 232: Why is this defined twice?

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+        // Potential uses could include a RetryAfter value or an error code.
+        public object? State { get; }
+
+        // `state` is the additional metadata, onDispose takes `state` as its argument.
+        public Resource(bool isAcquired, object? state, Action<Resource>? onDispose);
+
+        public void Dispose();
+
+        // This static `Resource` is used by rate limiters or in cases where `Acquire` returns false.
+        public static Resource SuccessNoopResource = new Resource(true, null, null);
```
davidfowl, line 247:
```suggestion
        public static Resource Success = new Resource(true, null, null);
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+        public static Resource FailNoopResource = new Resource(false, null, null);
```
davidfowl, line 248:
```suggestion
        public static Resource Failed = new Resource(false, null, null);
```

davidfowl, line 248: Do we need these anymore?

proposed/resource-limit.md
davidfowl, line 169: Do we need these anymore?

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+In an ASP.NET Core endpoint, usage of specific `ResourceLimiter`s will look like the following. Note that in these cases, the resource limiter is not known to be a rate or concurrency limiter and therefore the release mechanics must be followed.
+
+```c#
+ResourceLimiter limiter = new SomeResourceLimiter(options => ...)
+
+// Synchronous checks
+endpoints.MapGet("/acquire", async context =>
+{
+    // Check limiter using `Acquire` that should complete immediately
+    using var resource = limiter.Acquire();
```
davidfowl, line 297: I love this new pattern. It's *very* clean.

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+In cases where it is known that the `ResourceLimiter` is for certain a rate limiter, for example in cases where the user of the limiter is also the instantiator of the limiter, the usage can be simplified to ignore the release mechanic:
+
+```c#
+ResourceLimiter rateLimiter = new SomeRateLimiter(options => ...)
+
+// Synchronous checks
+endpoints.MapGet("/acquire", context =>
+{
+    // Check limiter using `Acquire` that should complete immediately
+    if (rateLimiter.Acquire().IsAcquired)
```
davidfowl, line 341:
```suggestion
    using var resource = rateLimiter.Acquire();
    if (resource .IsAcquired)
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+        // Resource could not be obtained, send 429 response
+        context.Response.StatusCode = StatusCodes.Status429TooManyRequests;
+        return Task.CompletedTask;
+    }
+}
+
+// Async checks
+endpoints.MapGet("/acquireAsync", async context =>
+{
+    if (await rateLimiter.AcquireAsync().IsAcquired)
```
davidfowl, line 357:
```suggestion
    using var resource = await rateLimiter.AcquireAsync();
    if (resource.IsAcquired)
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+        public ResourceDepletedMode ResourceDepletedMode { get; set; }
+        // Queue size when queuing is enabled
+        public long MaxQueueSize { get; set; }
+    }
+
+    // Rate limiter options
+    internal abstract class RateLimiterOptions : ResourceLimiterOptions
+    {
+        // This can be used to externally trigger rate limit window/replenishment updates
+        public EventHandler<DateTimeOffset> TickEventHandler { get; set; }
```
davidfowl, line 412: I'm not a fan of event handlers. Why do we need this? It's not async friendly in it's current state either...

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+    public ValueTask<Resource> AcquireAsync(long requestedCount, CancellationToken cancellationToken = default)
+    {
+        while(true)
+        {
+            cancellationToken.ThrowIfCancellationRequested();
+            if (cpuCounter.NextValue() <= _threshold)
+            {
+                return Resource.SuccessNoopResource;
+            }
+            Thread.Sleep(1000); // Wait 1s before rechecking.
```
davidfowl, line 569:
```suggestion
            await Task.Delay(1000); // Wait 1s before rechecking.
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+    public Resource Acquire(long requestedCount)
+    {
+        if (cpuCounter.NextValue() > _threshold)
+        {
+            return Resource.FailNoopResource;
+        }
+        return Resource.SuccessNoopResource;
+    }
+
+    public ValueTask<Resource> AcquireAsync(long requestedCount, CancellationToken cancellationToken = default)
```
davidfowl, line 560:
```suggestion
    public async ValueTask<Resource> AcquireAsync(long requestedCount, CancellationToken cancellationToken = default)
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+## Proof-of-concepts
+
+### Customized resource limiter based on ambient state that won't ship in box
+
+e.g. The limiter fails when CPU usage is above 90%
+
+```c#
+class CPUUsageLimiter : ResourceLimiter
+{
+    private PerformanceCounter _cpuUsageCounter = new PerformanceCounter("Processor", "% Processor Time", "_Total");
```
davidfowl, line 541: This is windows specific

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+}
+
+// Usage scenario
+
+// API Controller:
+[ApiController]
+[RequestLimit(new ResourceLimiterWrapperReasonPhrase("Too many booking requests", new RateLimiter(requestPerSecond: 100)))]
+public class BookingController : ControllerBase
+{
+    [RequestLimit(new ResourceLimiterWrapperReasonPhrase("Too many hotel booking requests", new RateLimiter(requestPerSecond: 5)))]
```
davidfowl, line 614: You can't new up things in attributes.

proposed/resource-limit.md
```diff
@@ -0,0 +1,843 @@
...
+            if (!resource.IsAcquired)
+            {
+                context.Response.StatusCode = StatusCodes.Status429TooManyRequests;
+                if (resource.State is string reasonPhrase)
+                {
+                    await context.Response.WriteAsync(reasonPhrase);
+                }
+                return;
+            }
+            resources.push(resource);
```
davidfowl, line 648:
```suggestion
            resources.Push(resource);
```

proposed/resource-limit.md
JunTaoLuo: Yes I think so, these are convenient for quick default limiter responses. For example, they would be used by all rate limiters and when concurrency limiters fail. FYI I've also updated the names to `SuccessfulAcquisition` and `FailedAcquisition`. I'm still not 100% happy with the names though.

proposed/resource-limit.md
JunTaoLuo: Are you referring to the section above ~line 150ish? That was an example of additional APIs required for partial release, but it's a non-goal for now. I've hidden it under <detail> tags for now.

proposed/resource-limit.md
JunTaoLuo: Here I'm explicitly illustrating when you know for sure it is a rate limiter, for example when you were the one to create the instance, you can skip the using pattern for even simpler usage.

proposed/resource-limit.md
JunTaoLuo: Same as above.

proposed/resource-limit.md
JunTaoLuo: What's a better way to trigger a resource replenishment externally for a rate limiter? I think it's worse to have each limiter create a Timer to trigger self-replenishment but I'm open to ideas.

proposed/resource-limit.md
JunTaoLuo: Yea I'm illustrating a possibility here, it's certainly not how we would actually go about implementing this.

proposed/resource-limit.md
JunTaoLuo: Doh, should have verified that this compiles instead coding in a md file.

proposed/resource-limit.md
```diff
@@ -0,0 +1,883 @@
...
+<details>
+
+```c#
+public struct ResourceLease : IDisposable
+{
+    // This represents whether resource lease acquisition was successful
+    public bool IsAcquired { get; }
+
+    // Number of acquired resources in the lease. Alternatively this can be stored on `State`
+    public long ResourceCount { get; }
```
davidfowl, line 156:
```suggestion
    public int ResourceCount { get; }
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,883 @@
...
+    public ResourceLease(long count, bool isAcquired, object? state, Action<ResourceLease>? onDispose);
+
+    // Release a specific amount of resources, if possible
+    public void Release(long releaseCount);
+
+    // Release remaining resources
+    public void Dispose();
+
+    // This static `ResourceLease` is used by rate limiters or in cases where `Acquire` returns false.
+    public static ResourceLease SuccessfulAcquisition = new ResourceLease(true, null, null);
```
davidfowl, line 172:
```suggestion
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,883 @@
...
+    public static ResourceLease FailedAcquisition = new ResourceLease(false, null, null);
```
davidfowl, line 173:
```suggestion
```

JunTaoLuo: I'll finish updating these in the morning.

proposed/resource-limit.md
JunTaoLuo: Removed

proposed/resource-limit.md
```diff
@@ -0,0 +1,868 @@
...
+        public object? State { get; }
+
+        // Private fields to be used by `onDispose()`, this is not a public API shown here for completeness
+        private ResourceLimiter? _resourceLimiter;
+        private Action<ResourceLimiter?, int, object?>? _onDispose;
+
+        // Constructor which sets all the readonly values
+        public ResourceLease(
+            bool isAcquired, 
+            int Count,
```
davidfowl, line 266:
```suggestion
            int count, 
```

proposed/resource-limit.md
```diff
@@ -0,0 +1,868 @@
...
+
+```c#
+// Represent an aggregated resource (e.g. a resource limiter aggregated by IP)
+public abstract class AggregatedResourceLimiter<TKey>
+{
+    // an inaccurate view of resources
+    public abstract int EstimatedCount(TKey resourceID);
+
+    // Fast synchronous attempt to acquire resources
+    public abstract ResourceLease Acquire(TKey resourceID, int requestedCount);
```
davidfowl, line 205: I think the ResourceLease needs to be generic so it can store the `AggregatedResourceLimiter<TKey>`

proposed/resource-limit/resource-limit.md
JunTaoLuo: Naming was updated.

proposed/resource-limit.md
JunTaoLuo: Likely it will involve the implementation having an explicit `Replenish(DateTimeOffset timestamp)` that can be called.

JunTaoLuo:
I've updated the proposal based on the latest feedback from API reviews. I'll merge this proposal as is now, but I will respond to further comments here or in API review with new PRs to this proposal separately..

PR #217, opened 2021-05-06
Add plan for x64 emulation

richlander:
We're working on a plan to support .NET for x64 emulation on Arm64, on macOS and Windows. There are multiple decision points that we are needing to navigate to make a decision. This document is intended to explore those.

### DIFF
--- a/accepted/2021/x64-emulation-on-arm64/x64-emulation.md
+++ b/accepted/2021/x64-emulation-on-arm64/x64-emulation.md
@@ -2,7 +2,7 @@
 
 We're working on a plan to support .NET for x64 emulation on Arm64, on macOS and Windows. There are multiple decision points that we are needing to navigate to make a decision. This document is intended to explore those.
 
-The following issues are the fundamental decision points. The choices we make for each will have significant downstream and user observable impact.
+The following issues are the fundamental decision points. The choices we make for each will have significant downstream and user observable impact. They are also co-dependent.
 
 - Install location
 - Model for targeting architecture
@@ -37,7 +37,7 @@ We could install .NET x64 to another location, but the implications would be the
 
 Goal: Enable developers to produce architecture-specific assets correctly and with confidence.
 
-The .NET application model is oriented around rid-specific apps. In particular, the apphost is rid-specific, and the apphost ia a core part of the experience. In some scenarios, you don't have to pay much attention to the rid of the apphost. For example, if you exclusively develop on *and* target Windows x64, then rid-targeting isn't really important, even though it is present. If your development and target environment differ, then you need to directly participate in rid-targeting.
+The .NET application model is oriented around rid-specific apps. In particular, the apphost is rid-specific, and the apphost is a core part of the experience. In some scenarios, you don't have to pay much attention to the rid of the apphost. For example, if you exclusively develop on *and* target Windows x64, then rid-targeting isn't really important, even though it is present. If your development and target environment differ, then you need to directly participate in rid-targeting.
 
 In the typical scenario, you can develop on your machine, for example Windows x64, and then not need to consider rid-targeting until you are ready to test on or deploy to Linux x64, for example. x64 emulation places new requirements on developers to participate in RID-targeting during inner loop development. That's new.
 
@@ -45,11 +45,13 @@ The core issue is that some .NET versions are available for one rid and not the 
 
 ### Rely on the implicit rid of the SDK
 
-In theory, the easiest approach is to simply use the matching SDK for the rid you want to target. That's the model used in the "developing and targeting on Windows x64" example discussed earlier.
+In theory, the easiest approach is to simply use the matching SDK for the rid you want to target.
+
+Note: Arm64 and x64 .NET builds will not be installed together. If you need Arm64 .NET, install it. If you need x64 .NET, install it. If you install both, they need to be updated separately.
 
 There are problems with this model:
 
-- Using the x64 SDK will be a substandard experience, primarily due to it not being in the `PATH`. On macOS, you can either type `/usr/local/share/dotnet/x64/dotnet`, prepend that directory to the `PATH`, create a shell alias, or create a symbolic link (in `/usr/local/bin`). Some of those same options exist on Windows.
+- Using the x64 SDK will be a substandard experience, primarily due to it not being in the `PATH`. On macOS, you can either type `/usr/local/share/dotnet/x64/dotnet`, prepend that directory to the `PATH`, create a shell alias, or create a symbolic link (for example, in `/usr/local/bin`). Some of those same options exist on Windows.
 - Users will find it confusing and unpleasant to need to pivot between the x64 and Arm64 SDKs as a means of targeting a .NET version (short-term problem) or RID (long-term problem).
 - This form of targeting will be untenable for many open source projects. Our GitHub repos don't have this problem because we always download the correct SDK via our build and test scripts. Assumption: Many other projects don't.
 
@@ -84,7 +86,7 @@ The `UseAppHost` property isn't strictly needed, but is a topic we need to resol
 The upside of this plan is that this model is supported today and doesn't require any work (beyond resolving the apphost challenges on macOS). The downside is that is regression in experience. It has the following problems:
 
 - Projects files get longer and more complicated, for users that need (or are required) to use x64 emulation.
-- Project files are no longer portable across machines. This would be a deal killer for teams (or open source projects) that has users with multiple machine types. For example, dotnet org repos would never be able to implement this technique.
+- Project files are no longer portable across machines. This would be a deal killer for teams (or open source projects) that have users (or CI test legs) with multiple machine types. For example, dotnet org repos would never be able to implement this technique.
 
 ### Rely on implicit RID defaults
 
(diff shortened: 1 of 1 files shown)


accepted/2021/x64-emulation-on-arm64/x64-emulation.md
```diff
@@ -0,0 +1,243 @@
...
+
+In the typical scenario, you can develop on your machine, for example Windows x64, and then not need to consider rid-targeting until you are ready to test on or deploy to Linux x64, for example. x64 emulation places new requirements on developers to participate in RID-targeting during inner loop development. That's new.
+
+The core issue is that some .NET versions are available for one rid and not the other. That means that you need to explicitly target the matching rid in some way. There are three choices for that, two of which are already supported today and one not.
+
+### Rely on the implicit rid of the SDK
+
+In theory, the easiest approach is to simply use the matching SDK for the rid you want to target. That's the model used in the "developing and targeting on Windows x64" example discussed earlier.
```
jkotas, line 48: What is the "developing and targeting on Windows x64" example that this talks about?

richlander: It just means that you develop on and Windows x64 and deploy to Windows x64 for prod, rid targeting isn't something you need to think about. Same thing with Linux x64 (for development and prod). You need to think more about rid targeting when the development environment and prod environment don't match in operating system and/or architecture dimensions. Make sense?

jkotas: Yes, makes sense. I was commenting more on the fact that this example is not actually discussed earlier in this doc, and so people reading this doc for the first time will be lost.

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
```diff
@@ -0,0 +1,243 @@
...
+- If x64 builds of Visual Studio for Windows or Visual Studio for Mac are installed on Arm64, then they will need to install the Arm64 .NET SDK and/or adapt those products to .NET x64 being installed in the `x64` directory.
+
+We could install .NET x64 to another location, but the implications would be the same.
+
+## Model for targeting architecture
+
+Goal: Enable developers to produce architecture-specific assets correctly and with confidence.
+
+The .NET application model is oriented around rid-specific apps. In particular, the apphost is rid-specific, and the apphost ia a core part of the experience. In some scenarios, you don't have to pay much attention to the rid of the apphost. For example, if you exclusively develop on *and* target Windows x64, then rid-targeting isn't really important, even though it is present. If your development and target environment differ, then you need to directly participate in rid-targeting.
```
k15tfu, line 40: typo: _apphost ia a core part_

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
```diff
@@ -0,0 +1,243 @@
...
+The core issue is that some .NET versions are available for one rid and not the other. That means that you need to explicitly target the matching rid in some way. There are three choices for that, two of which are already supported today and one not.
+
+### Rely on the implicit rid of the SDK
+
+In theory, the easiest approach is to simply use the matching SDK for the rid you want to target. That's the model used in the "developing and targeting on Windows x64" example discussed earlier.
+
+There are problems with this model:
+
+- Using the x64 SDK will be a substandard experience, primarily due to it not being in the `PATH`. On macOS, you can either type `/usr/local/share/dotnet/x64/dotnet`, prepend that directory to the `PATH`, create a shell alias, or create a symbolic link (in `/usr/local/bin`). Some of those same options exist on Windows.
+- Users will find it confusing and unpleasant to need to pivot between the x64 and Arm64 SDKs as a means of targeting a .NET version (short-term problem) or RID (long-term problem).
```
k15tfu, line 53: Does it mean osx-arm64 installer will also include osx-x64 .NET SDK?

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
richlander: Fixed.

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
richlander: fixed.

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
richlander: Great q. No. I updated the text.

yaakov-h:
On the macOS front, do universal binaries ("fat binaries") factor in at all, or is .NET going to ignore its existence entirely?

richlander:
Great q @yaakov-h. Search for "universal binaries" @ https://github.com/dotnet/sdk/issues/16896.

The short version is that they are not a silver bullet and add significant cost. We'd also have to invent a way for .NET 6 Arm64 and x64 to co-exist that would likely end up looking a lot like this plan. I think that universal binaries end up being mostly an implementation detail as opposed to a fundamentally different direction.

I should say that we discussed universal binaries only related to the dotnet host and its direct native dependencies (not the runtime). We firmly rejected compiling the entire product as universal. That has massive cost for us and significant UX implications. We could imagine compiling the runtime that way, but our ready-to-run native code would have to participate, too. The product would become huge. There is no chance we go down that path.

yaakov-h:
Another question related to that, which I haven't seen mentioned anywhere yet:

On macOS and similar platforms (iOS, potentially tvOS?) with multi-architecture binaries, a .NET app may have a dependency on a native binary which has multiple architecture slices.

Can/will .NET deal with these binaries as-is, or does/will .NET require the binary to be either duplicated across or split into architecture-specific directories in NuGet packages etc.?

richlander:
That's a good question. As long as OS library loader is used to load the native binary, then the correct slice should be selected. I believe that's what will happen by default.

Make sense?

yaakov-h:
I believe so too, but is there a common folder for each OS where .NET runtimes/SDKs will search regardless of architecture?

richlander:
Are you asking if there is a universal binary concept (for a given OS) for NuGet packages so that you can single instance a binary?

yaakov-h:
For NuGet, but also for published applications.

Can I have / will I be able to have, for example, a universal binary at `runtimes/osx/native/libfoo.dylib`, or do I need both `runtimes/osx-x64/native/libfoo.dylib` and `runtimes/osx-arm64/native/libfoo.dylib`?

richlander:
Right. That is what I was getting at. The answer is that you need both copies, even though they'd be identical. If we start to get significant requests for this, we can add support for this scenario. I don't think it is very hard since it's just another (less specific) entry in the RID graph. Feel free to file an issue for a universal RID for macOS in dotnet/runtime. Same thing would apply for Windows. This is also a change that we could take in servicing, post RTM.

rhuijben:
What about the Windows ARM64EC platform support?

This is x64 compatible ARM64 on latest Windows platforms, but with an ABI that looks more like x64 than ARM64.

I can see scenarios where you want native .Net support on this architecture for performance reasons.

richlander:
We didn't adopt ARM64EC because it's non obvious which architecture `dotnet.exe` should choose.

ddavidebor:
@richlander to clarify, is there any way to use X64 dlls from a .net app compiled to Arm64 ?

Because that's the use case for Arm64EC, but is seems that currently if an app uses .net, it's not possible to take advantage of Arm64EC and the app has to be 100% recompiled, including any SDKs that might be supplied as DLLs by 3rd parties.

jkotas:
@ddavidebor You can run your x64 app with full emulation. The app is still going to take advantage of Arm64EC for the Windows OS .dlls that are all compiled with Arm64EC option. Our experience is that the x64 emulator has good enough performance.

Arm64EC is just a performance optimization of the x64 emulation. Arm64EC is always going to be slower than native Arm64 when everything else is equal. Our focus has been on the native Arm64 performance.

ddavidebor:
So the only way to transition a large app written in C# with both C# and C++ dependencies without having to port everything all at once would be:

1. Run it in fully x64 emulated mode.
1. Recompile some of those SDKs/DLLs C/C++ to ARM64EC
1. Run it in x64 emulated mode, but it will take advantage of the newly recompiled DLLs
1. Finish porting everything to Arm64
1. Run it in Arm64 mode

is that correct?

jkotas:
Correct. This whole process is a performance optimization, and it is not specific to .NET. If the performance is good enough in step 1, you do not have to do the rest.

jozefizso:
> Our experience is that the x64 emulator has good enough performance.

We see extremely bad performance in MS Office add-ins built using .NET and targeting x64 platform.

richlander:
I assume that's running x64 Office. Yes?

rwg0:
I would potentially have a use case for an ARM64EC version of the runtime. I have a relatively large .NET WPF app that includes a considerable amount of my own C++ code for performance and also needs to make use of DLLs provided by 3rd parties. 

I can (and have) rebuilt my C++ code as ARM64, but getting ARM64 versions of the 3rd party DLLs is hard. Often these DLLs control hardware and the drivers are actually easier to find than the control DLLs (winusb, hid devices need no special drivers, etc). Being able to run the .NET runtime and my C++ code in ARM64 gives much better performance than full x64 emulation. Being able to emulate x64 for DLLs that are either unavailable or unported to ARM64 lets the ARM version progress incrementally.

From my point of view, an additional platform target of 'ARM64EC' target for build and publish would be sufficient - I publish self contained, so wouldn't worry if the installed runtime didn't support ARM64EC.

driver1998:
> I assume that's running x64 Office. Yes?

@richlander The problem is there is no ARM64 Office, only ARM64EC Office. So if you want to build Office addins with .NET, it can only be x64.

toptensoftware:
@richlander Hi Rich, I have a situation very similar to that described by @rwg0 above.  It's a music application built on .NET that loads VST2 and VST3 plugins.  Most plugin developers currently only ship x64 binaries but Microsoft is encouraging them to start shipping ARM builds too.  In practice however there are literally thousands of old plugins that will probably never be ported - seems a shame to have to run the entire application emulated just to support the occasional plugin that hasn't been ported.

Pete Brown's article explains how DAW developers should support ARM64EC ([see here](https://devblogs.microsoft.com/windows-music-dev/load-x64-plug-ins-like-vsts-from-your-arm-code-using-arm64ec/)) but given that my app is built on .NET that's currently not possible as far as I can tell.

Any chance of a solution similar to that described by @rwg above? ie: additional platform target of "ARM64EC".

Failing that what are my other options?

PR #220, opened 2021-05-24
Remove compiler enforcement and let the analyzer handle it

terrajobst:
In an earlier design I proposed that the analyzer handles use of preview APIs while the compiler handles use of language features that depend on preview runtime features. This proposal shifts the enforcement entirely to the analyzer.

### DIFF
--- a/accepted/2021/preview-features/preview-features.md
+++ b/accepted/2021/preview-features/preview-features.md
@@ -139,29 +139,6 @@ namespace System
 }
 ```
 
-> What will the compiler do when the attribute is not applied at the assembly level?
-
-I don't think the compiler/analyzer should do anything in this case. The SDK
-will generate the assembly level attribute based on the project setting, but
-this can be turned, off which is what we will do to build the BCL. For obvious
-reasons we don't want `System.Runtime` as whole to have an assembly level
-attribute -- we only want this on specific types/members.
-
-> If assembly A has a method M marked with [RequiresPreviewFeatures], but no
-> assembly level attribute, and it references assembly B that has the assembly
-> level attribute, will the compiler warn?
-
-Yes. The assembly level attribute is meant to protect the user that just uses
-the project level setting. Advanced users (like us) can turn off the generation
-of the assembly level attribute and selectively apply this attribute to
-particular parts of the library. In this case it will only warn users that uses
-those APIs.
-
-The advanced user is responsible for making sure that they very well understand
-which parts are using preview features and ensure they get the attribution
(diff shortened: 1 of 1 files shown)


accepted/2021/preview-features/preview-features.md
```diff
@@ -264,48 +241,59 @@ that the runtime can use this information if ever needed:
 }
 ```
 
-### Compilation context
+### Compiler tracking
+
+*We propose that the compiler does no tracking of preview features at this
```
tannergooding, line 246: Isn't this problematic for the scenario of:
* .NET 6 and C# 10 ship `StaticVirtualsInInterfaces` as "preview"
* .NET X and C# Y ship `StaticVirtualsInInterfaces` as "stable"
* C# Y downtargets .NET 6, it things `StaticVirtualsInInterfaces` is "stable" but .NET 6 considers it "preview"

tannergooding: That is, I read the below part about the complications, but I don't understand how having the analyzer recognize the differences is safe/sound.

Only the language understands when a given feature is used and therefore when the corresponding feature flag needs to be checked. In certain cases, the use of these feature flags may be internal to the compiler (possibly as part of another feature/syntax) and not something you can deduce from the exposed syntax.

And so simply having the compiler warn if the language considers a feature preview/non-preview and the runtime annotation doesn't match seems straightforward and still allows users to workaround the issue by disabling the warning in "power" scenarios.

terrajobst: The analyzer will warn whenever you define statics on interfaces and the targeted reference assemblies marks has the `StaticVirtualsInInterfaces` field marked with `[RequiresPreviewFeatures]`. So in your scenario the analyzer will warn.

> Only the language understands when a given feature is used and therefore when the corresponding feature flag needs to be checked.

In principle yes. However, for .NET 6 we only really care about statics in interfaces which are easy to detect. In general, we don't care about arbitrary language features. We only care about language features that require runtime support. Those will likely leave an easy to identify mark in the resulting metadata, such as statics interfaces. DIMs would also be easy to detect.

But yes, the trade of this direction is that analyzer will have to understand more than just APIs. But given where we are, it seems easier to move logic to the analyzer rather than trying to bring analyzer features into the compiler. We can, however, revisit this later if (1) we think preview features is something we do more off and (2) we believe there is value in doing so.

AlekseyTs (approved):

pgovind (approved):
Approving for now. I will have the analyzer test-able today/tomorrow, so we can revisit this later this week if things need to change

PR #221, opened 2021-05-26
Add mechanism for renaming workloads

mhutch:


wli3:
The design looks good. But we should caveat that the support may come after net7.0

dsplaisted (approved):

PR #222, opened 2021-05-27
Add special case for Xamarin assets to nuget precedence rules

akoeplinger:
While working on .NET 6 support for Xamarin we hit a build break in a test project when referencing the System.ComponentModel.Composition v4.7.0 nuget package.

```
xamarin-macios/tests/linker/ios/link all/MEFTests.cs(3,29): error CS0234: The type or namespace name 'Composition' does not exist in the namespace 'System.ComponentModel' (are you missing an assembly reference?)
```
 
We've tracked it down to the nuget restore storing this in obj/project.assets.json:
 
```
      "System.ComponentModel.Composition/4.7.0": {
        "type": "package",
        "compile": {
          "ref/xamarinios10/_._": {}
        },
        "runtime": {
          "lib/xamarinios10/_._": {}
        }
      },
```

I.e. it wants to pick the assembly from the GAC/framework, which is what we want for the legacy Xamarin (since System.ComponentModel.Composition shipped in the framework there) but _not_ for the new .NET6-based one since it is an out-of-band package there and doesn't ship in the framework.

We're proposing an addition to the nuget precedence rules to skip the Xamarin TFM asset in this case, so we fallback to the netcoreapp2.0 assembly.

The other alternative would be to add an explicit `net6.0` asset to the package, however this would only apply to new releases since we can't fix the existing nugets.

**Note:** we discovered additional issues while working on this, see https://github.com/dotnet/designs/pull/222#issuecomment-852178771 so we're also proposing to change the precedence rules to prefer `netstandard` assets before `net5.0`

/cc @marek-safar @ericstj @rolfbjarne

### DIFF
--- a/accepted/2021/net6.0-tfms/net6.0-tfms.md
+++ b/accepted/2021/net6.0-tfms/net6.0-tfms.md
@@ -99,6 +99,9 @@ follows:
     - However, it should still prefer `net6.0` assets over `xamarin.ios`.
     - It also shouldn't be compatible with any other existing Xamarin TFM (such
       as `xamarin.mac`)
+* It should prefer `netstandard2.1` and earlier assets before `net5.0`
+    - Packages with `net5.0`, `netcoreapp3.1` and earlier assets assumed desktop-like environments,
+      since this was the only place where .NET Core existed. This causes issues on mobile.
 * Generate NuGet warning [NU1701] when a `xamarin.ios` asset is being used
     - Package 'packageId' was restored using 'xamarin.ios' instead the project
       target framework 'net6.0-maccatalyst'. This package may not be fully
@@ -128,9 +131,9 @@ This reasoning results in these precedence rules:
 1. `net6.0-ios`
 1. `net6.0`
 1. `xamarin.ios` (no warning, skip if it points to `_._`)
+1. `netstandard2.1` – `1.0`
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
-1. `netstandard2.1` – `1.0`
 1. `net4.x` – `1.0` (warning)
 
 **net6.0-maccatalyst**
@@ -138,9 +141,9 @@ This reasoning results in these precedence rules:
 1. `net6.0-maccatalyst`
 1. `net6.0`
 1. `xamarin.ios` ([NU1701] warning, skip if it points to `_._`)
+1. `netstandard2.1` – `1.0`
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
-1. `netstandard2.1` – `1.0`
 1. `net4.x` – `1.0 ([NU1701] warning)
 
 **net6.0-macos**
@@ -148,9 +151,9 @@ This reasoning results in these precedence rules:
 1. `net6.0-macos`
 1. `net6.0`
 1. `xamarin.macos` (no warning, skip if it points to `_._`)
+1. `netstandard2.1` – `1.0`
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
-1. `netstandard2.1` – `1.0`
 1. `net4.x` – `1.0` ([NU1701] warning)
 
 **net6.0-tvos**
@@ -158,9 +161,9 @@ This reasoning results in these precedence rules:
 1. `net6.0-tvos`
 1. `net6.0`
 1. `xamarin.tvos` (no warning, skip if it points to `_._`)
+1. `netstandard2.1` – `1.0`
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
-1. `netstandard2.1` – `1.0`
 1. `net4.x` – `1.0` ([NU1701] warning)
 
 **net6.0-android**
@@ -168,9 +171,9 @@ This reasoning results in these precedence rules:
 1. `net6.0-android`
 1. `net6.0`
 1. `monoandroid12.0` - `1.0` (no warning, skip if it points to `_._`)
+1. `netstandard2.1` – `1.0`
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
-1. `netstandard2.1` – `1.0`
 1. `net4.x` – `1.0` ([NU1701] warning)
 
 The currently supported `monoandroid` TFMs are:
--- a/accepted/2021/net6.0-tfms/net6.0-tfms.md
+++ b/accepted/2021/net6.0-tfms/net6.0-tfms.md
@@ -114,13 +114,20 @@ not generate a warning. `net6.0-android` will behave in the same way as `net6.0-
 except that the Xamarin.Android TFM is `monoandroid` (1.0 - 12.0). `net6.0-tizen` is
 also the same as `net6.0-ios` except the TFM is `tizen`.
 
+A special case are packages which have a Xamarin TFM asset that points to `_._` which normally
+means that the asset should be picked from the framework/GAC instead of the package.
+However we have cases like `System.ComponentModel.Composition` where in "legacy" Xamarin this
+assembly shipped as part of the framework (hence e.g. `xamarin.ios` in the package points to `_._`)
+but in .NET Core the assembly is _not_ in the shared framework. We need to skip picking the
+Xamarin TFM asset in this case to avoid a build break due to missing references.
+
 This reasoning results in these precedence rules:
 
 **net6.0-ios**
 
 1. `net6.0-ios`
 1. `net6.0`
-1. `xamarin.ios` (no warning)
+1. `xamarin.ios` (no warning, skip if it points to `_._`)
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
 1. `netstandard2.1` – `1.0`
@@ -130,7 +137,7 @@ This reasoning results in these precedence rules:
 
 1. `net6.0-maccatalyst`
 1. `net6.0`
-1. `xamarin.ios` ([NU1701] warning)
+1. `xamarin.ios` ([NU1701] warning, skip if it points to `_._`)
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
 1. `netstandard2.1` – `1.0`
@@ -140,7 +147,7 @@ This reasoning results in these precedence rules:
 
 1. `net6.0-macos`
 1. `net6.0`
-1. `xamarin.macos` (no warning)
+1. `xamarin.macos` (no warning, skip if it points to `_._`)
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
 1. `netstandard2.1` – `1.0`
@@ -150,7 +157,7 @@ This reasoning results in these precedence rules:
 
 1. `net6.0-tvos`
 1. `net6.0`
-1. `xamarin.tvos` (no warning)
+1. `xamarin.tvos` (no warning, skip if it points to `_._`)
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
 1. `netstandard2.1` – `1.0`
@@ -160,7 +167,7 @@ This reasoning results in these precedence rules:
 
 1. `net6.0-android`
 1. `net6.0`
-1. `monoandroid12.0` - `1.0` (no warning)
+1. `monoandroid12.0` - `1.0` (no warning, skip if it points to `_._`)
 1. `net5.0`
 1. `netcoreapp3.1` – `1.0`
 1. `netstandard2.1` – `1.0`


mhutch:
Does this affect just this assembly or are there other cases where Xamarin frameworks included out-of-band assemblies in-box?

I don't think it would be good to set a precedent of special-casing individual assemblies in the precedence rules. Perhaps we could generalize this by having the `net6.0-*` precedence rules skip `_._` assets for Xamarin frameworks?

If we choose to go the route of updating the affected packages, we should consider adding a warning when affected projects directly or indirectly reference the old versions of the packages.

akoeplinger:
> Does this affect just this assembly or are there other cases where Xamarin frameworks included out-of-band assemblies in-box?

It should affect these assemblies/packages:

```
System.ComponentModel.Composition
System.Data.SqlClient
System.Diagnostics.PerformanceCounter
System.Drawing.Common
System.Json
System.Runtime.Caching
System.Runtime.InteropServices.WindowsRuntime
System.Security.Cryptography.ProtectedData
System.ServiceModel.Duplex
System.ServiceModel.Http
System.ServiceModel.NetTcp
System.ServiceModel.Primitives
System.ServiceModel.Security
System.Xml.XPath.XmlDocument
```

> I don't think it would be good to set a precedent of special-casing individual assemblies in the precedence rules. Perhaps we could generalize this by having the net6.0-* precedence rules skip _._ assets for Xamarin frameworks?

Sorry if this wasn't clear but my proposal was about making this a generalized rule like you described.

akoeplinger:
Another thing that we noticed which would not be fixed by this proposal is that with the current precedence rules you also get transitive dependencies for the `Xamarin.iOS` group instead of e.g. `netcoreapp3.1`.

This manifests itself when referencing e.g. System.Text.Json 5.0.2, now you get this in project.assets.json:
```
"System.Text.Json/5.0.2": {
"type": "package",
"dependencies": {
    "Microsoft.Bcl.AsyncInterfaces": "5.0.0",
    "System.Buffers": "4.5.1",
    "System.Runtime.CompilerServices.Unsafe": "5.0.0",
    "System.Text.Encodings.Web": "5.0.1"
},
"compile": {
    "lib/netcoreapp3.0/System.Text.Json.dll": {}
},
"runtime": {
    "lib/netcoreapp3.0/System.Text.Json.dll": {}
}
}
```

i.e. unnecessary `Microsoft.Bcl.AsyncInterfaces` while a normal net5.0 project gets no dependency nugets.

danmoseley:
cc @joperezr since he was on the email.

mhutch:
So NuGet selects the TFM for dependencies independently of selecting the TFM for assets? I don't know how hard it would be to change that, either in the general case or for this specific pattern. @nkolev92 ?

Maybe it's okay to special case the package names, given it's a small and limited set and this is an exceptional scenario.

akoeplinger:
Here's another case where we run into issues now due to nuget picking netcoreapp assets instead of netstandard: https://github.com/xunit/xunit/issues/2314

I'm starting to wonder whether we should tweak the rules to prefer netstandard assets before netcoreapp <= 5, since these earlier netcoreapp assets could've justifiably made the assumption that they're running on "desktop" .net core.

nkolev92:
> So NuGet selects the TFM for dependencies independently of selecting the TFM for assets? I don't know how hard it would be to change that, either in the general case or for this specific pattern. @nkolev92 ?

This is a known behavior. There's a lot of packages that author their assets correctly, but don't use dependency groups for their dependencies.
We have not messed with it, because we haven't gotten a lot of feedback around this topic and it would be a breaking change. 

I'd really prefer us to update the packages. 

cc @zkat @zivkan @aortiz-msft @JonDouglas

akoeplinger:
So to summarize I think there are three issues right now:

1) Using the `_._` GAC assembly which doesn't exist for .NET6-based Xamarin (tentative fix in this proposal)
2) Preferring netcoreapp assets over netstandard assets even though they might be assuming "desktop" .NET behavior (not fixed by this proposal)
3) Transitive dependencies via Xamarin.iOS dependency group are not ideal (not fixed by this proposal, though unclear how bad this is)

ViktorHofer:
> The other alternative would be to add an explicit net6.0 asset to the package, however this would only apply to new releases since we can't fix the existing nugets.

FWIW I'm doing this in https://github.com/dotnet/runtime/pull/53439 but that will only cover packages that we build live.

cc @ericstj

ericstj:
> FWIW I'm doing this in dotnet/runtime#53439 but that will only cover packages that we build live.

And only those that happen to have a `NetCoreAppCurrent` TargetFramework already.  I'm more inclined to remove the placeholders in the latest packages instead; so long as all existing customers who want to consume the new packages will be on net6.0-based SDKs.

akoeplinger:
Existing Xamarin will still be supported for quite some time (https://dotnet.microsoft.com/platform/support/policy/xamarin) so removing the placeholders would break the packages for them.

ericstj:
> so removing the placeholders would break the packages for them.

Unless conflict resolution worked for them like it does for .NETCore and .NETFramework (where we have packages that overlap with shared framework and "best" wins).  Not sure what the state of this is: do you know?

JonDouglas:
Just so this PR contains all the suggested changes talked about, should we also introduce the ideal precedence order into the PR as it sounded like @Redth and @mattleibow saw some inconsistencies with this already?

Is there a proposal on how to resolve the precedence issues here? If `netcoreapp3.1-1.0` assets exist in a package and are preferred with current rules, is that going to break too many things? Do we need to adjust the current proposal to skip certain generations to bring back the desired precedence?

You can use a tool like https://nugettoolsdev.azurewebsites.net/5.10.0-preview.2.7185/framework-precedence to check out existing precedence rules.

/cc @davidortinau

mhutch:
> Unless conflict resolution worked for them like it does for .NETCore and .NETFramework (where we have packages that overlap with shared framework and "best" wins). Not sure what the state of this is: do you know?

@ericstj how exactly does this work? I can't think why it should be any different for ios/android,

ericstj:
> @ericstj how exactly does this work? I can't think why it should be any different for ios/android

When there are two sources of the same binary it tries to [decide which is better](https://github.com/dotnet/sdk/blob/17da35a869826254123d9ed518dc1c72f5298ba7/src/Tasks/Common/ConflictResolution/ConflictResolver.cs#L173-L360) based on higher version.  Reasons it could be different: if the build for these frameworks doesn't use these targets, if the files are not *visible* to the build (like the runtime packs in core that have to be represented in manifests passed to the resolver), or if for some reason the framework files should be better (eg more API, specialized implementation) but have a lower version than the package.  You can try this out pretty easily by modifying the packages locally (deleting the placeholders in both the nupkg) and seeing how it works.

mhutch:
Ah, it's a related problem but kinda the inverse?

The problem here is that the fallbacks are causing the package to resolve to a TFM where the assets _are_ in the framework, but in the current TFM they _aren't_ in the framework and instead need to come from the package. So no assets are resolved at all and there is no conflict to be resolved.

ericstj:
>  Ah, it's a related problem but kinda the inverse? ... So no assets are resolved at all and there is no conflict to be resolved.

I'm discussing what can be done to fix those no assets cases: we can just remove placeholders from the packages and let things apply.  Conflict resolution can make those additional assets a non-issue on the frameworks where they exists (Xamarin* frameworks).  IOW: conflict resolution works better and replaces the uses of placeholder files.

mhutch:
Oh, I wasn't aware the conflict resolution targets were available for non-SDK style projects. It looks they get conditionally imported for `.NETFramework` projects in [`Microsoft.NET.Build.Extensions.targets`](https://github.com/dotnet/sdk/blob/c53c8ac81e1bcb8118c31bc4f5622396c72e9959/src/Tasks/Microsoft.NET.Build.Extensions.Tasks/msbuildExtensions/Microsoft/Microsoft.NET.Build.Extensions/Microsoft.NET.Build.Extensions.targets), which is present in the MSBuild  shipped inside Visual Studio on Windows.

Unfortunately `Microsoft.NET.Build.Extensions.targets` is not present on Mac, so we'll have to fix that.

Redth:
@JonDouglas here's what I believe the precedence rules should really look like if we want to support the cases where packages expected net5.0/netcoreapp3.1-1.0 to run on desktop/coreclr and that netstandard2.1-1.0 would be selected in xamarin/mobile scenarios:

```
net6.0-ios|net6.0-android|net6.0-maccatalyst|net6.0-tvos|net6.0-macos
net6.0
xamarin.ios
netstandard2.1-1.0
net5.0 (warning)
netcoreapp3.1-1.0 (warning)
net4.x-1.0 (warning)
```

akoeplinger:
@ericstj @mhutch if I understand your suggestion correctly it'd still only apply to newer versions of the packages right? At that point I'd rather add `net6.0` assets to the packages so we don't need to touch legacy mono/Xamarin to fix conflict resolution.

ViktorHofer:
Do we have the full list of affected packages? Presumably most of those don't exist in dotnet/runtime's main branch anymore or don't generate a package anymore as configurations were removed.

akoeplinger:
@ViktorHofer I posted the list of packages that run into the `_._` problem here: https://github.com/dotnet/designs/pull/222#issuecomment-849857649

I'll update the proposal to include the precedence of netcoreapp vs. netstandard changes as those likely affect a lot more packages.

ViktorHofer:
> @ViktorHofer I posted the list of packages that run into the _._ problem here: #222 (comment)

You mentioned that this list might not be complete:

> It should affect these assemblies/packages (based on dotnet/runtime main, so might be missing older packages we don't build there anymore):

akoeplinger:
@ViktorHofer I cross-referenced the list with a Kusto query of affected packages which @JonDouglas shared over email, and all of the additional packages in that query are unlisted so it should be complete :)

nkolev92:
> @JonDouglas here's what I believe the precedence rules should really look like if we want to support the cases where packages expected net5.0/netcoreapp3.1-1.0 to run on desktop/coreclr and that netstandard2.1-1.0 would be selected in xamarin/mobile scenarios:
> 
> ```
> net6.0-ios|net6.0-android|net6.0-maccatalyst|net6.0-tvos|net6.0-macos
> net6.0
> xamarin.ios
> netstandard2.1-1.0
> net5.0 (warning)
> netcoreapp3.1-1.0 (warning)
> net4.x-1.0 (warning)
> ```

If I'm interpreting this correctly, this is exactly what the precedence of the old xamarin frameworks was. 

The fact that net5.0 isn't consider compatible makes me question the fact that the platform specific frameworks like ios/android report a .NETCoreApp target framework identifier.

akoeplinger:
@nkolev92 no, the old Xamarin frameworks weren't compatible with `netcoreapp`/`net5.0+` TFMs at all.

nkolev92:
I was mostly referring to the following part: 

> net6.0-ios|net6.0-android|net6.0-maccatalyst|net6.0-tvos|net6.0-macos
> net6.0
> xamarin.ios
> netstandard2.1-1.0

mhutch:
@Redth that's not what this issue is about though?

mhutch:
@nkolev92 the precedence we have is correct. the problem here is that we've essentially moved framework assemblies to out of box packages but we're resolving to the framework where they were in box.

akoeplinger:
@mhutch yup, that's true for solving the `_._` issue. There's another issue which is about the precedence rules that is unrelated to that one.

mhutch:
I really think the precedence discussion belongs in another issue. However, here's my read of the situation:

AFAICT the only advantage of preferring `netstandard` assets over `net5.0`/`netcoreapp*` assets is that it *may* fix packages for which *all* of the following are true:
* do not have xamarin assets
* have `netstandard` assets that are portable
* have `net5.0`/`netcoreapp*` assets that assume they're running on desktop

However, it *will* result in inferior assets being selected for packages for which all the following are true:
* do not have xamarin assets
* have `netstandard` assets that are portable
* have `net5.0`/`netcoreapp*` assets that don't assume they're running on desktop

And it will *break* valid packages for which all the following are true:
* do not have xamarin assets
* `netstandard` asset is reference assembly
* `net5.0`/`netcoreapp*` assets have concrete impl

It also introduces a massive inconsistency into the precedence.

filipnavara:
I've to strongly agree with @mhutch here that it would be very weird if changing TFM from `net6.0` to `net6.0-macos` (ie. to consume some Cocoa API) would suddenly start pulling in `netstandard2.1` packages instead of `net5.0`/`netcoreapp3.1` ones.

Redth:
I actually tend to agree that the precedence rules are correct as currently designed, and not the ones I suggested in my last comment to collect more evidence of the correct decision either way.

The unfortunate reality with the current rules is that many packages will need updating. Some of these are more impactful due to their prevalence in the ecosystem (xUnit for instance). 

Is there anything else we can do for the user experience here when this is encountered? Is there a way for consumers of packages in this scenario to work around it?

I want to be sure this scenario is well understood and messaged appropriately /cc @terrajobst @richlander @davidortinau

mhutch:
@Redth re. third party libraries, maybe NuGet should emit a warning when resolving `netcoreapp*`/`net5.0` packages from `net6.0-(ios|android|etc)`, like you suggested, to make the developer aware that there may be issues. i don't love this idea though, as it will result in many false positives and will be unactionable.

there isn't really much else we can do. we could statically determine _some_ broken cases but that would be too expensive to do on restore or build and wouldn't catch everything. perhaps we could have the linker validate the [platform checks](https://github.com/dotnet/designs/blob/main/accepted/2020/platform-checks/platform-checks.md) and warn if unavailable APIs are found to be reachable? it would be a bit late in the dev cycle (especially for android, where linking may not happen till publish) but better than nothing. it would also be a generalized feature that would be useful long after any of this ceases to be an issue.

AFAIK there's no way to override the framework NuGet resolves for a package. it would definitely be a nice feature to have and would make problems like this actionable. i image it would mesh nicely with a generalized graph override capability.

re. the framework assemblies that moved out of box, I think all we can do is detect the unfixed versions of these specific packages and emit a build error instructing the developer to update those packages. maybe we could have a nuget feature where the targets could instruct NuGet to force package references to a minimum version if found anywhere in the graph e.g. something like  `<ForceUpgradePackage Id="SomeId" MinVersion="SomeVersion" />`?

mhutch:
@filipnavara ah, my bad, looks like we do in fact ship the conflict resolution targets on Mac and enable them unconditionally for all Xamarin frameworks since late 2019.

So maybe it's safe to just remove all those `_._` files from the NuGet packages.

akoeplinger:
I think for fixing the `_._` issue I'd prefer adding `net6.0` assets to the 14 packages mentioned in https://github.com/dotnet/designs/pull/222#issuecomment-849857649 rather than relying on conflict resolution, seems much easier to understand and requires no testing on old Xamarin frameworks.

Warning when restoring `netcoreapp*`/`net5.0` packages indeed sounds like it will cause a _lot_ of warnings so probably not something we want to do.

> Is there anything else we can do for the user experience here when this is encountered? Is there a way for consumers of packages in this scenario to work around it?

Yeah I think giving users an escape hatch to override might be a good compromise, basically something similar to `AssetTargetFallback`: https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#assettargetfallback

akoeplinger:
Having the "override asset" feature would actually allow us to solve the `_._` issue for existing packages too since then we could just ship a hardcoded list of the 14 bad packages and do something like this in the iOS/Android SDKs:

```
<PackageReference Update="System.ComponentModel.Composition" Condition="Version == 5.0.0">
  <AssetTargetOverride>netcoreapp3.1</AssetTargetOverride>
</PackageReference>
```

JonDouglas:
Just to circle back and re-iterate on a potential consensus of the [initial 3 issues](https://github.com/dotnet/designs/pull/222#issuecomment-852178771) & what's been discussed thus far:

1. We believe fixing the libraries themselves is the best path forward. Whether that's removing the `_._` or adding `net6.0` assets.
2. We believe the precedence rules as implemented today are ideal and keep the precedence integrity for the future.
    - There are concerns around mobile/MAUI user expectations & UX of previous experiences & new experiences until the ecosystem broadly supports `net6.0-(ios|android|etc)` assets.
3. We believe this is a known behavior & do not have plans to resolve it at this time.

Does this sound right? If so, I propose we find some owners for the various items above & work to close out this PR if there's nothing in addendum to precedence rules.

akoeplinger:
It sounds good to me, however we'd need the asset override nuget feature to make it a good experience.

@JonDouglas @nkolev92 how hard do you think that would be to implement?

nkolev92:
It's non trivial as NuGet does asset selection per asset group + the fact that we'd need to consider what this means upstream for consumers of the project in question, be it as a projref or packref.

mhutch:
> Having the "override asset" feature would actually allow us to solve the `_._` issue for existing packages too since then we could just ship a hardcoded list of the 14 bad packages and do something like this in the iOS/Android SDKs:
> 
> ```
> <PackageReference Update="System.ComponentModel.Composition" Condition="Version == 5.0.0">
>   <AssetTargetOverride>netcoreapp3.1</AssetTargetOverride>
> </PackageReference>
> ```

Conditionally applying metadata to `PackageReference` items wouldn't work for transitive refs.

I think my proposed `<ForceUpgradePackage Id="SomeId" MinVersion="SomeVersion" />` mechanism would be pretty straightforward? The idea is to clamp the version of any reference in the graph matching the ID.

zivkan:
> I think my <ForceUpgradePackage Id="SomeId" MinVersion="SomeVersion" /> suggestion would be pretty straightforward?

The NuGet Client team had been working on a Central Package Version Management feature, when we had fewer urgent tasks, but currently it doesn't make the cut when prioritized given our resources. Anyway, it contains a "transitive package pinning" feature, which we should use, once implemented, rather than providing another syntax.  However, it took months to prototype, and once released, library authors immediately hated it given it's impact on pack & package dependencies, plus it broke features such as `PrivateAssets`. So, I would not say it's straightforward to implement.  As a result, I don't imagine this could be delivered in time for net6.0.

akoeplinger:
Ok so it seems the consensus is that we shouldn't do this 😃 

We'll be doing this instead:
- Change the libraries mentioned in https://github.com/dotnet/designs/pull/222#issuecomment-849857649 to add `net6.0` assets which is tracked by https://github.com/dotnet/runtime/issues/54012
- Work with affected nuget package owners to fix their packages, e.g. https://github.com/xunit/xunit/issues/2314
- Find out whether we can have some kind of escape hatch for unfixed packages to override the asset: https://github.com/NuGet/Home/issues/10937

ViktorHofer:
> Change the libraries mentioned in #222 (comment) to add net6.0 assets which is tracked by dotnet/runtime#54012

Correct me if I'm wrong but dotnet/runtime#54012 won't help with this in any way. None of the libraries mentioned above are part of the work tracked in that issue.

akoeplinger:
Talked with Viktor offline, here's the status for the packages:

- Already fixed in dotnet/runtime or fixed shortly:

```
System.Drawing.Common
System.Diagnostics.PerformanceCounter
System.ComponentModel.Composition
System.Runtime.Caching
System.Security.Cryptography.ProtectedData
```

- Owned by dotnet/wcf, adding the net6.0 assets is tracked by https://github.com/dotnet/wcf/issues/4237

```
System.ServiceModel.Duplex
System.ServiceModel.Http
System.ServiceModel.NetTcp
System.ServiceModel.Primitives
System.ServiceModel.Security
```
---

The following packages however are not building in dotnet/runtime main anymore and are basically deprecated:

```
System.Data.SqlClient (replaced by Microsoft.Data.SqlClient)
System.Runtime.InteropServices.WindowsRuntime (never ran on Xamarin)
System.Json
System.Xml.XPath.XmlDocument
```

From talking to @ViktorHofer we don't service these packages anymore so we need to work with package owners to remove their dependencies on these packages.

PR #181, opened 2021-02-23
Proposal to add Framework specific inbox source generators

ericstj:
This document describes desired characteristics for source generators we
wish to add to .NET 6.0.

PR #223, opened 2021-05-28
Streamline Windows Forms application configuration and bootstrap

RussKie:
This document describes the application configuration and bootstrap experience utilising Roslyn source generators that facilitates sharing of configurations with the VS Designer (so that it can continue to provide the true WYSIWYG experience) we wish to add to .NET 6.0.


accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,304 @@
...
+The bolded roles makes it easier for people to understand who the driver of the
+proposal is (PM) and who can be asked for technical questions (Dev). At
+Microsoft, these roles happen to match to job titles too, but that's irrelevant.
+-->
+
+**Owner** [Igor Velikorossov](https://github.com/RussKie)
+
+# Overview
+
+Historically Windows Forms designer was rendering its design surface based on hardcoded assumptions, such as control default font and dpi settings. However over the past 10-15 years not only multi-monitor configurations became a norm, monitor tech has made significant progress in making 4K, 5K, 8K, etc. monitors widely used by our consumers. The Windows itself has been innovating and tinkering with UI configurations, such default font metrics, which over the years deviated from "Microsoft Sans Serif, 8pt" and now is set to "Segoe UI, 9pt". Incidentally the office team is currently [in search of the next deafult font](https://www.microsoft.com/microsoft-365/blog/2021/04/28/beyond-calibri-finding-microsofts-next-default-font/).
```
jkotas, line 19:
```suggestion
Historically Windows Forms designer was rendering its design surface based on hardcoded assumptions, such as control default font and dpi settings. However over the past 10-15 years not only multi-monitor configurations became a norm, monitor tech has made significant progress in making 4K, 5K, 8K, etc. monitors widely used by our consumers. The Windows itself has been innovating and tinkering with UI configurations, such default font metrics, which over the years deviated from "Microsoft Sans Serif, 8pt" and now is set to "Segoe UI, 9pt". Incidentally the office team is currently [in search of the next default font](https://www.microsoft.com/microsoft-365/blog/2021/04/28/beyond-calibri-finding-microsofts-next-default-font/).
```

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,303 @@
...
+    Application
+        .Configure(() =>
+        {
+            // application configuration logic
+        })
+        .Run(new Form1());
+}
+```
+
+Indeed, this sounds like a grand idea. However in reality Windows Forms app often have elaborate application bootsrap logic (e.g. [this](https://github.com/gitextensions/gitextensions/blob/master/GitExtensions/Program.cs), [this](https://github.com/luisolivos/Halconet/blob/4ed328a15ba69aab00e92076774977143790b5d3/HalcoNET%205.8.0/Ventas/Program.cs), or [this](https://github.com/aaraujo2019/PEQMIMERIA/blob/e11b6d57e3bbbe505834cbfeb717fe87066d6850/DBMETAL_SHARP/DBMETAL_SHARP/Program.cs)) that won't be possible in this scenario, and will require deeper investigations and feasibility studies.
```
drewnoakes, line 302:
```suggestion
Indeed, this sounds like a grand idea. However in reality Windows Forms app often have elaborate application bootstrap logic (e.g. [this](https://github.com/gitextensions/gitextensions/blob/master/GitExtensions/Program.cs), [this](https://github.com/luisolivos/Halconet/blob/4ed328a15ba69aab00e92076774977143790b5d3/HalcoNET%205.8.0/Ventas/Program.cs), or [this](https://github.com/aaraujo2019/PEQMIMERIA/blob/e11b6d57e3bbbe505834cbfeb717fe87066d6850/DBMETAL_SHARP/DBMETAL_SHARP/Program.cs)) that won't be possible in this scenario, and will require deeper investigations and feasibility studies.
```

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,302 @@
...
+
+The .NET Core/.NET-powered Windows Forms runtime has been (more or less) keeping up with the changes, e.g. by providing new API to [set high dpi mode](https://github.com/dotnet/winforms/pull/396), or by updating the [default font](https://github.com/dotnet/winforms/pull/656). The default font change has unearthed numerous issues in the runtime (e.g. different font-based scaling), which we addressed by providing (yet another) API to set an [application-wide default font](https://github.com/dotnet/winforms/pull/4911). 
+
+However during numerous discussions the team has identified probably the biggest flaw in the current separation of the runtime and the designer, and the overall move from .NET Framework assumptions - the lack of a mechanism to share Windows Forms project-level configuration settings between the runtime and the designer. That is, if the app is configured to run with disabled visual styles, in PerMonitorV2, and have an application-wide default font set to "Arial, 14pt" there is no way to show the designer surface with the same settings to truly provide the WYSIWYG experience.
```
stephentoub, line 23: What is PerMonitorV2?

stephentoub: Nevermind, found the docs (didn't realize it was an existing setting).

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,302 @@
...
+think should be involved in the design process and ensure they are involved
+(which might mean being tagged on GitHub issues, invited to meetings, or sent
+early drafts).
+-->
+
+* Windows Forms runtime/designer: @dotnet/dotnet-winforms @OliaG @DustinCampbell 
+* Visual Basic: @KathleenDollard  @KlausLoeffelmann
+* Roslyn: @jaredpar @sharwell
+* Project System: @drewnoakes 
+* SDK: @ericstj
```
ericstj, line 147: I work on libraries not SDK. It would be good to involve @dsplaisted for SDK

dsplaisted (approved):
Looks good from the .NET SDK side.

It looks like there isn't any work required in the .NET SDK targets, it will all be handled by the analyzers.

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,302 @@
...
+# Proposal
+
+The purpose of this proposal is to:
+
+1. **streamline a Windows Forms application configuration and bootstrap**, 
+2. with the view that this will facilitate the **sharing of the information between the runtime and the designer during the development phase**.<br/>
+That is, whenever the designers surface process is started configuration information is read from a known location, and necessary configurations are applied (e.g. run the design surface in PerMonitorV2 mode, or set a form/usercontrol default font to "Arial, 14pt").<br/>
+:warning:  **out of scope**, tracked under https://github.com/dotnet/winforms-designer/issues/3192.
+
+**NOTE:** The new functionality is opt-in, i.e. unless a developer makes a conscious decision to use the new configuration and the bootstrap mechanism existing applications will continue to work as-is, and the current developer experience will remain the same.
```
Pilchie, line 36: Do you actually mean opt-in, or do you mean "non-breaking"?  I presume we'd want to update templates to start with this for example?

KlausLoeffelmann:
We need to make sure, not to exclude the typical VB customer in that and make sure VB‘s application framework fits into this. Actually, I’d propose to have two templates for VB Apps and phase out the current principle of storing defaults in the application framework in favor of this concept. But still we should keep the Program.vb-less and application-events concept alive and also I am against the requirement for a VB dev to patch vbproj files directly.

merriemcgaw:
>We need to make sure, not to exclude the typical VB customer in that and make sure VB‘s application framework fits into this. 
Do you have concerns that as proposed we are excluding them? I agree with the principle that we should keep VB customers in mind with the design and update the application framework as necessary to support it.

> I am against the requirement for a VB dev to patch vbproj files directly.
I would have agreed with you with the old style project files. But with SDK style projects editing is so much more streamlined, it could/should very well be an option.

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,302 @@
...
+
+You should avoid splitting this into various product stages (like MVP, crawl,
+walk, run) because that usually indicates that your proposal tries to cover too
+much detail. Keep it high-level, but try to paint a picture of what done looks
+like. The design section can establish an execution order.
+-->
+
+1. The new bootstrap experience (this includes source generators and analyzers) must come inbox with the Windows Desktop SDK, and be available without any additional references or NuGet packages from get go (i.e. `dotnet new winforms && dotnet build`).<br/>Related: [dotnet/designs#181](https://github.com/dotnet/designs/pull/181)
+1. The new bootstrap API must only work for Windows applications projects (i.e. `OutputType = WinExe`). These projects have an entry point, where an app is initialised; and they also specify an application manifest, if there is one.
+2. The new bootstrap must be correctly generated for .NET Framework and .NET projects taking in consideration available
```
Pilchie, line 87: Does this bootstrap actually need to support .NET Framework at all? I'm wondering why we would change things there?

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: Yes, the plan is to update templates. For existing apps the behaviour is "opt-in" in the sense that developers will need to write the new method, and remove existing `Application.*` methods. If they don't do this - nothing happens, apps will continue to work as before.

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,302 @@
...
+    Application
+        .Configure(() =>
+        {
+            // application configuration logic
+        })
+        .Run(new Form1());
+}
+```
+
+Indeed, this sounds like a grand idea. However in reality Windows Forms app often have elaborate application bootstrap logic (e.g. [this](https://github.com/gitextensions/gitextensions/blob/master/GitExtensions/Program.cs), [this](https://github.com/luisolivos/Halconet/blob/4ed328a15ba69aab00e92076774977143790b5d3/HalcoNET%205.8.0/Ventas/Program.cs), or [this](https://github.com/aaraujo2019/PEQMIMERIA/blob/e11b6d57e3bbbe505834cbfeb717fe87066d6850/DBMETAL_SHARP/DBMETAL_SHARP/Program.cs)) that won't be possible in this scenario, and will require deeper investigations and feasibility studies.
```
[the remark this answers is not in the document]
RussKie: This example isn't about top-level programs like those currently discussed for ASP.NET Core apps. 
The point I was trying to make here is that in these examples there is a lot going on before the main form is started, as well as `try..catch` block around `Application.Run(new Form1())`. Such bootstraps are difficult to express with the example snippet.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: Here I've been thinking about the multi-targeting scenarios. I'm still trying to work out how the analyzers weave into the SDK though, may be this point isn't possible after all.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: Apologies. Corrected.
I tagged you primarily because of your proposal https://github.com/dotnet/designs/pull/181

RussKie:
> It looks like there isn't any work required in the .NET SDK targets, it will all be handled by the analyzers.

I may need some help/guidance/reviews with analyzer inboxing, so we can provide a seamless exprience.

RussKie:
> I am against the requirement for a VB dev to patch vbproj files directly.

Could please explain why not? 
Even though some VB developers may be coming from non CS backgrounds and may not necessarily identify themselves as software engineers, I am sure, they are intelligent individuals, and would understand what a project file is and modify properties in it. May be it is time we should start breaking the stereotype about VB developers...

That said, I do agree with you that we should offer a GUI configuration in VS, and that's where the Project System comes into play. I've had a chat with @drewnoakes about it, and for C# it doesn't appear as a big job. Sadly for VB  it maybe not the case - I was told (and I'm sure you know it all too well) that the AF makes it very hard to modernise in this area.

> Actually, I’d propose to have two templates for VB Apps and phase out the current principle of storing defaults in the application framework in favor of this concept.

I'm happy to explore this. Will chat with you when you're back.

accepted/2021/winforms/streamline-application-bootstrap.md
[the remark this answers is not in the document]
RussKie: I added a little more to this section, hopefully it makes it clearer.

KlausLoeffelmann:
> Sadly for VB it maybe not the case - I was told (and I'm sure you know it all too well) that the AF makes it very hard to modernise in this area.

And that’s exactly why I think we should try to get rid of the current UI experience for the AF in VB altogether and maybe have a somewhat identical experience (if not shareable) for C# and VB. 

Don’t get me wrong, though! I don’t want to get rid of the AF. Just trying to make the UI part easier!

eriawan:
@RussKie 

Apologize to barge in the middle of this PR review 🙏  
Is this .NET design PR is supposed to be fully "open" as fully open, or it is meant for MS employee only?

Because in the `streamline-application-bootstrap.md` file, the link on this line 34 of  `https://github.com/dotnet/winforms-designer/issues/3192` is pointing to a private repo, 
Since I'm not MS employee, I could not open this link at all,

KlausLoeffelmann:
@eriawan It's basically only a reference to the internal Winforms-Designer repo. Yes, the Designer repo is not open to the public, but we need to track the work, since there is Designer work which needs to be done.

For this discussion, it doesn't contain any relevant information except, that the planning/discussion for the designer work for this is out of scope (😄) and that the additional work, which would become necessary, if we did this, is tracked there.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: I think I've been overthinking this. If analyzers are placed under dotnet\packs\Microsoft.WindowsDesktop.App.Ref\\X.Y.<version\>\ref\netX.Y then there is no backwards compatibility or multitargeting scenarios to consider - the analyzers will be imported by the Windows Desktop SDK for the given TFM. And if there was no analyzers available for .NET Framework or .NET 5.0 SDKs, then those can't/won't be imported.

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,326 @@
...
+You should avoid splitting this into various product stages (like MVP, crawl,
+walk, run) because that usually indicates that your proposal tries to cover too
+much detail. Keep it high-level, but try to paint a picture of what done looks
+like. The design section can establish an execution order.
+-->
+
+1. The new bootstrap experience (this includes source generators and analyzers) must come inbox with the Windows Desktop SDK, and be available without any additional references or NuGet packages from get go (i.e. `dotnet new winforms && dotnet build`).<br/>Related: [dotnet/designs#181](https://github.com/dotnet/designs/pull/181)
+1. The new bootstrap API must only work for Windows applications projects (i.e. `OutputType = WinExe`). These projects have an entry point, where an app is initialised; and they also specify an application manifest, if there is one.
+2. The new bootstrap must be correctly generated for .NET Framework and .NET projects taking in consideration available 
+API in each framework version:
```
RussKie, line 88:
```suggestion
```

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,324 @@
...
+        {
+            // application configuration logic
+        })
+        .Run(new Form1());
+}
+```
+
+Indeed, this sounds like a grand idea. However in reality Windows Forms app often have elaborate application bootstrap logic (e.g. [this](https://github.com/gitextensions/gitextensions/blob/master/GitExtensions/Program.cs), [this](https://github.com/luisolivos/Halconet/blob/4ed328a15ba69aab00e92076774977143790b5d3/HalcoNET%205.8.0/Ventas/Program.cs), or [this](https://github.com/aaraujo2019/PEQMIMERIA/blob/e11b6d57e3bbbe505834cbfeb717fe87066d6850/DBMETAL_SHARP/DBMETAL_SHARP/Program.cs)) that won't be possible in this scenario, and will require deeper investigations and feasibility studies.
+
+In the long term we could consider supporting a host/builder model, similar to the suggestion in [dotnet/winforms#2861](https://github.com/dotnet/winforms/issues/2861) and implementation in [alex-oswald/WindowsFormsLifetime](https://github.com/alex-oswald/WindowsFormsLifetime). E.g.:
```
mhutch, line 305: I'm very much in favor of standardizing the host/builder model across app types where appropriate. Is there any particular reason this should be a long term option rather than an immediate option?

RussKie: This requires a thorough design and planning, and with the current workload the team doesn't have sufficient capacity for this to be delivered in .NET 6.0 timeframe.

Tragetaschen (changes requested):

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,324 @@
...
+The bolded roles makes it easier for people to understand who the driver of the
+proposal is (PM) and who can be asked for technical questions (Dev). At
+Microsoft, these roles happen to match to job titles too, but that's irrelevant.
+-->
+
+**Owner** [Igor Velikorossov](https://github.com/RussKie)
+
+# Overview
+
+Historically Windows Forms designer was rendering its design surface based on hardcoded assumptions, such as control default font and dpi settings. However over the past 10-15 years not only multi-monitor configurations became a norm, monitor tech has made significant progress in making 4K, 5K, 8K, etc. monitors widely used by our consumers. The Windows itself has been innovating and tinkering with UI configurations, such default font metrics, which over the years deviated from "Microsoft Sans Serif, 8pt" and now is set to "Segoe UI, 9pt". Incidentally the office team is currently [in search of the next default font](https://www.microsoft.com/microsoft-365/blog/2021/04/28/beyond-calibri-finding-microsofts-next-default-font/).
```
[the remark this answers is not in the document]
RussKie: Thank you!

RussKie, line 19:
```suggestion
Historically Windows Forms designer was rendering its design surface based on hardcoded assumptions, such as control default font and dpi settings. However over the past 10-15 years not only multi-monitor configurations became a norm, monitor tech has made significant progress in making 4K, 5K, 8K, etc. monitors widely used by our consumers. The Windows itself has been innovating and tinkering with UI configurations, such default font metrics, which over the years deviated from "Microsoft Sans Serif, 8pt" and now is set to "Segoe UI, 9pt". Incidentally the office team is currently [in search of the next default font](https://www.microsoft.com/en-us/microsoft-365/blog/2021/04/28/beyond-calibri-finding-microsofts-next-default-font/).
```

RussKie:
Here's a proposed implementation: https://github.com/dotnet/winforms/pull/5035

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,362 @@
...
+
+### 2. C#
+
+```diff
+    static class Program
+    {
+        [STAThread]
+        static void Main()
+        {
++           ApplicationConfiguration.Initialize();
```
terrajobst, line 189: It's unfortunate that this can't be `Application.Initialize()`.

Alternatively, we could follow the pattern that we have used in the JSON generator where the user controls the type. This could look as follows:


```C#
// Program.cs, written by the user

static partial class Program
{
    [STAThread]
    static void Main()
    {
        InitializeApplication();
        Application.Run(new MainForm());
    }

    [GenerateInitializeApplication]
    private partial void InitializeApplication();
}

// Program.g.cs, generated

partial class Program
{
    private void InitializeApplication()
    {
        Application.EnableVisualStyles();
        Application.SetDefaultFont(new Font(....));
        Application.SetHighDpiMode(HighDpiMode.PerMonitorV2);
        Application.SetCompatibleTextRenderingDefault(false);
    }
}
```

We could also decide that we don't need the user to opt-in at the source level and decide that setting the MSBuild properties is sufficient, which would reduce it to just this:

```C#
// Program.cs, written by the user

static partial class Program
{
    [STAThread]
    static void Main()
    {
        InitializeApplication();
        Application.Run(new MainForm());
    }
}

// Program.g.cs, generated

partial class Program
{
    private void InitializeApplication()
    {
        Application.EnableVisualStyles();
        Application.SetDefaultFont(new Font(....));
        Application.SetHighDpiMode(HighDpiMode.PerMonitorV2);
        Application.SetCompatibleTextRenderingDefault(false);
    }
}
```

accepted/2021/winforms/streamline-application-bootstrap.md
terrajobst: Yeah, it's too late for .NET 6

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,362 @@
...
+```
+
+Indeed, this sounds like a grand idea. However in reality Windows Forms app often have elaborate application bootstrap logic (e.g. [this](https://github.com/gitextensions/gitextensions/blob/master/GitExtensions/Program.cs), [this](https://github.com/luisolivos/Halconet/blob/4ed328a15ba69aab00e92076774977143790b5d3/HalcoNET%205.8.0/Ventas/Program.cs), or [this](https://github.com/aaraujo2019/PEQMIMERIA/blob/e11b6d57e3bbbe505834cbfeb717fe87066d6850/DBMETAL_SHARP/DBMETAL_SHARP/Program.cs)) that won't be possible in this scenario, and will require deeper investigations and feasibility studies.
+
+In the long term we could consider supporting a host/builder model, similar to the suggestion in [dotnet/winforms#2861](https://github.com/dotnet/winforms/issues/2861) and implementation in [alex-oswald/WindowsFormsLifetime](https://github.com/alex-oswald/WindowsFormsLifetime). E.g.:
+```cs
+[main: STAThread]
+CreateHostBuilder().Build().Run();
+
+public static IHostBuilder CreateHostBuilder() =>
```
richlander, line 343: Why is this method definition here? It seems unnecessary.

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,362 @@
...
+
+
+The benefit the new approach provides is by facilitating the sharing of the information between the runtime and the designer during the development phase, as well as unifying how dpi settings are configured. This benefit is believed to outweigh the need to remove several lines of code from `Main()` method and adding several MSBuild properties.
+
+### Go further?
+
+At this point an astute reader may ask why no go further and offer an API similar to a ASP.<!-- -->NET Core bootstrap? Wouldn't it be nice to write something like:
+```cs
+[STAThread]
+static void Main()
```
richlander, line 325: Why does the "Go further" still use the old syntax? It seems like there is no value to using "Main" here.

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,362 @@
...
+
+```diff
+    static class Program
+    {
+        [STAThread]
+        static void Main()
+        {
++           ApplicationConfiguration.Initialize();
+
+-           Application.EnableVisualStyles();
```
mhutch, line 191: Can the designer determine the effective configuration from these existing method calls?

accepted/2021/winforms/streamline-application-bootstrap.md
mhutch: Agreed. It would be great if this was as streamlined as the latest [ASP.NET Core templates](https://github.com/dotnet/aspnetcore/blob/543eb19/src/ProjectTemplates/Web.ProjectTemplates/content/EmptyWeb-CSharp/Program.cs).

accepted/2021/winforms/streamline-application-bootstrap.md
Pilchie: Today there is no way to apply the `[STAThread]` to the Main method of a C# program using top level statements. We have a thread with the language design team to discuss options there.

mhutch: You could do something like this 😈

```
await Application.SwitchToSTAThread();
```

Pilchie: What would the implementation of that look like?  AFAIK, that requires there to already be an STA thread in the application, which there isn't in this case.  Instead you'd have to create a thread, set it's apartment state, and run the rest of the method as a lambda/local function, which isn't really any better than a Main method.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: This would require the designer compile and parse user code, which will severely impact performance.

ericstj: Would it?  What if it were done with an analyzer/source-generator as part of the compilation that already happens?

accepted/2021/winforms/streamline-application-bootstrap.md
jnm2: The single statement `await Application.SwitchToSTAThread();` could start a new STA thread and run the rest of the method on it following the await.

jnm2: Evidence:
image

<details>
<summary>Click to expand</summary>

```cs
using System;
using System.Runtime.CompilerServices;
using System.Threading;

Console.WriteLine(Thread.CurrentThread.GetApartmentState());
await Application.SwitchToStaThread();
Console.WriteLine(Thread.CurrentThread.GetApartmentState());

public static class Application
{
    public static StaThreadAwaitable SwitchToStaThread() => default;

    public readonly struct StaThreadAwaitable
    {
        public StaThreadAwaiter GetAwaiter() => default;
    }

    public readonly struct StaThreadAwaiter : ICriticalNotifyCompletion
    {
        public bool IsCompleted => Thread.CurrentThread.GetApartmentState() == ApartmentState.STA;

        public void OnCompleted(Action continuation) => UnsafeOnCompleted(continuation);

        public void UnsafeOnCompleted(Action continuation)
        {
            // Real impl would also use existing thread if a message pump has already started

            var thread = new Thread(parameter => ((Action)parameter).Invoke());
            thread.SetApartmentState(ApartmentState.STA);
            thread.Start(parameter: continuation);
        }

        public void GetResult() { }
    }
}
```

</details>

accepted/2021/winforms/streamline-application-bootstrap.md
davidwengier: In theory this generator could be flipped on its head, and have it analyze the source (existing method calls) and output a manifest file for the designer to read. However there is currently no support in Roslyn for source generators that output non-source files, though it has been talked about.

accepted/2021/winforms/streamline-application-bootstrap.md
jkotas: > The single statement await `Application.SwitchToSTAThread();` could start a new STA thread

Creating new thread is unnecessary overhead. You can switch the main thread apartment state by doing this as first thing in Main:

```
// It is unfortunate that you have to set it to Unknown first. If there was a good reason, 
// we may be able to change the runtime to make it unnecessary.
Thread.CurrentThread.SetApartmentState(ApartmentState.Unknown);

Thread.CurrentThread.SetApartmentState(ApartmentState.STA);
```

> Today there is no way to apply the [STAThread] to the Main method of a C# program using top level statements.

The attributes on the Main method are legacy. I would not try to complicate the language to support them.

jkotas: We did not respect the attributes on Main in .NET Core 1.0 at all and instead tell people to set the apartment explicitly via a method call. We have added back the support to respect the attributes on Main to make existing (mostly WinForms/WPF) apps work.

accepted/2021/winforms/streamline-application-bootstrap.md
ericstj: It could emit source just the same, attributes if that’s easier for designer to read…. Just an idea if that’s a better user experience. Doesn’t feel too far fetched.

RussKie: There's another reason why we can't rely on code parsing. It won't work in a multi-project setup, where user UI controls reside in one project, and the the `Main` method lives in another project. We can't possibly go and parse all `Main` methods that may exist in a solution, and try to guess which one the designer should base its configuration on.
This proposal makes it possible for the designer to load UI controls in the designer with required settings.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: Great discussion! Just worth noting, the top level application is orthogonal to this design proposal, which is about bridging the gap between the runtime and the designer. 
It doesn't look there's anything that prevents developers from writing Windows Forms apps as top level apps now (modulo some additional boiler plate). This topic is being actively discussed, and it is something we will explore separately.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: Also the designer perf is on a critical path right now, and we need to be extremely mindful of anything we do in a designer that may require additional processing.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: > It's unfortunate that this can't be `Application.Initialize()`.

Yep.

> Alternatively, we could follow the pattern that we have used in the JSON generator where the user controls the type. This could look as follows: ...
> We could also decide that we don't need the user to opt-in at the source level and decide that setting the MSBuild properties is sufficient, which would reduce it to just this: ...

I have no strong preference either way.

accepted/2021/winforms/streamline-application-bootstrap.md
mhutch: That's not guaranteed to give correct results. It relies on the user knowing to use `Directory.Build.props` for those settings and not (intentionally or accidentally) overriding them in any individual project. The controls project might even be referenced by multiple exe projects with different settings.

I also don't really think it's an appropriate use of `Directory.Build.props`. These settings have nothing to do with the build at all, and will be adding overhead to evaluation, binlogs, etc.

In the case of controls projects this seems almost like editorconfig in intent 🤷‍♀️

RussKie: > That's not guaranteed to give correct results. It relies on the user knowing to use `Directory.Build.props` for those settings and not (intentionally or accidentally) overriding them in any individual project. 

I don't see this is as a significant problem. Nor being anything significantly different from the current developer experience with SDK-style project. 
AFAIK it is quite common to move shared properties from individual projects to Directory.Build.props.

By default these properties can be configured via the project property page (see: https://github.com/dotnet/project-system/issues/7279), and it will be up to a developer to move them, if necessary.

> The controls project might even be referenced by multiple exe projects with different settings.

The current designer experience is "_we show something which likely doesn't reflect the reality_", and with this proposal it will be "_we do our best to make it WYSIWYG_".
There will be further work required to make it "truly WYSIWYG" and most of this work will be on the VS side, e.g. VS will need to be updated to allow selection of the context, just like it did for the project selection.



> In the case of controls projects this seems almost like editorconfig in intent 🤷‍♀️

If anything with this we'll be sending a wrong message. From  https://editorconfig.org/: "_EditorConfig helps maintain consistent coding styles for multiple developers working on the same project across various editors and IDEs_".

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: image

RussKie: @jkotas FWIW the app seems to run fine without setting thread apartment states ¯\\_(ツ)\_/¯, though I guess the lack of it may lead to some random failures somewhere else...

jkotas: Windows GDI does not depend on COM at all. It explains why classic Win32 controls work just fine without COM initialized. (In fact, the classic old Win32 apps tried to avoid initializating COM back in the day since COM initialization was a heavy operation that took considerable amount of time and memory.)

The STA appartment has to be only set to use COM and only in some cases. I would expect that you may see bugs with web browser control that is heavily COM-based.

jnm2: I think the file dialogs may also rely on it. There are also single-threaded unit tests involving DevExpress controls that fail if that thread is not an STA thread, and I just assumed this was a general Windows Forms thing. Clipboard operations fail if not on an STA thread.

terrajobst: > The STA appartment has to be only set to use COM and only in some cases. I would expect that you may see bugs with web browser control that is heavily COM-based.

The first one I usually hit is clipboard :-)

RussKie:
I have restructured and in parts rewritten the proposal reflecting and incorporating the feedback sourced in a number of offline discussions.

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: Opened https://github.com/dotnet/winforms/issues/5071 for further discussions.

OliaG (approved):

accepted/2021/winforms/streamline-application-bootstrap.md
AraHaan: Ah Actually I have a better idea:
- ``Application.RegisterMainForm(Form form)`` (could be added to winforms)
- ``Application.Run()`` (no args) could be made to check if a main form is registered then do as if ``Application.Run(Form form)`` was called.
- The generated ``Initialize()`` method calls ``Application.Run()`` and leaves the developer to call ``Application.RegisterMainForm(Form form)`` (which then stores it for use in ``Application.Run()`` as a variable that is then checked for null in the implementation of ``Application.Run()``)

In theory it should work, but in general it could be a breaking change for .NET 6 (not to mention possibly too late to propose ``Application.RegisterMainForm(Form form)``) and having the generator call ``Application.Run()`` for the developer inside of the generated code.

99.9999999% of cases there are only a single main form anyway with the main form spinning up the child forms directly at startup of that main form.

RussKie: @terrajobst if we're going for top level statements I don't think the partial `Program` will work:
image

accepted/2021/winforms/streamline-application-bootstrap.md
RussKie: FYI https://devblogs.microsoft.com/oldnewthing/20140811-00/?p=283

RussKie:
The designer using the `ApplicationFont` property:
designer-fontstyle

RussKie:
If there's no more feedback I'm planning to merge this in the next couple of days.

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,516 @@
...
+E.g.:
+
+```cs
+Application.SetDefaultFont(new Font(new FontFamily("Calibri"), 11f, FontStyle.Italic | FontStyle.Bold, GraphicsUnit.Point));
+```
+
+It would rather be impractical to provide an MSBuild properties for each argument the `Font` constructor takes, and instead the proposal is to configure fonts via a single property `ApplicationFont`. This property will have a format equivalent to the output of `FontConverter.ConvertToInvariantString()` API e.g.:
+
+```
+Name=fontName[, Size=size[, Units=units[, GDiCharSet=gdiCharSet[, GdiVerticalFont=boolean]]]]
```
RussKie, line 433:
```suggestion
<FontName>, <Size><Size units>[, Style=[style1 [, style2[, ...]]]]
```

accepted/2021/winforms/streamline-application-bootstrap.md
```diff
@@ -0,0 +1,516 @@
...
+#### Default scenario
+
+Initial thinking was to allow configuration of only [`Font.FamilyName`](https://docs.microsoft.com/dotnet/api/system.drawing.font.fontfamily) and [`Font.Size`](https://docs.microsoft.com/dotnet/api/system.drawing.font.size) properties. However these two properties may be insufficient in some use cases, which otherwise would be achievable if `Application.SetDefaultFont(Font)` API was invoked manually.
+E.g.:
+
+```cs
+Application.SetDefaultFont(new Font(new FontFamily("Calibri"), 11f, FontStyle.Italic | FontStyle.Bold, GraphicsUnit.Point));
+```
+
+It would rather be impractical to provide an MSBuild properties for each argument the `Font` constructor takes, and instead the proposal is to configure fonts via a single property `ApplicationFont`. This property will have a format equivalent to the output of `FontConverter.ConvertToInvariantString()` API e.g.:
```
RussKie, line 430:
```suggestion
It would rather be impractical to provide an MSBuild properties for each argument the `Font` constructor takes, and instead the proposal is to configure fonts via a single property `ApplicationFont`. This property will have a format equivalent to the output of [`FontConverter.ConvertToInvariantString()` API](https://github.com/dotnet/runtime/blob/00ee1c18715723e62484c9bc8a14f517455fc3b3/src/libraries/System.Drawing.Common/src/System/Drawing/FontConverter.cs#L29-L86) e.g.:
```

RussKie:
In the top level scenario absent the ability to set `[STAThread]` right now, instead of requiring a developer to write the the following code by hand...
```diff
-   // Set STAThread
-   Thread.CurrentThread.SetApartmentState(ApartmentState.Unknown);
-   Thread.CurrentThread.SetApartmentState(ApartmentState.STA);

    ApplicationConfiguration.Initialize();
    Application.Run(new MainForm());
```
...we can emit it as part of the source generated `Initialize()` method. E.g.:
```diff
[CompilerGenerated]
internal static partial class ApplicationConfiguration
{
    public static void Initialize()
    {
+       // Set STAThread
+       Thread.CurrentThread.SetApartmentState(ApartmentState.Unknown);
+       Thread.CurrentThread.SetApartmentState(ApartmentState.STA);
+
        Application.EnableVisualStyles();
        Application.SetCompatibleTextRenderingDefault(false);
        Application.SetHighDpiMode(HighDpiMode.PerMonitorV2);
    }
}
```

Does anyone see any issues with this?

eriawan:
@RussKie 

Nice proposal!

If the 2 calls of `SetApartmentState()` is included in a method in the top level, does the generated Initialize() method cover this too? Should there be additional analyzer for this?

For example:

```
stativ void SetSTA()
{
    Thread.CurrentThread.SetApartmentState(ApartmentState.Unknown);;
    Thread.CurrentThread.SetApartmentState(ApartmentState.STA);
}

SetSTA();
ApplicationConfiguration.Initialize();
Application.Run(new MainForm());
```

cnmade:
As a user and CSharp programmer. I would like to suggest, the Designer UI should learning from **Sketch app**, The professional design software, what you see is what you get.

image


I hope you guys really don't limited your mind locked down with Visual Studio traditional UI Style. Which not for human behaviour.

```
Sketch is a vector graphics editor for macOS developed by the Dutch company Sketch B.V
```

jkotas:
> Does anyone see any issues with this?

It will work, but it is not pretty and it is inefficient. The COM will be first initialized as MTA, then uninitialized, and then initialized again as STA. If it is a temporary solution that we will cleanup later, I think it would ok.

AraHaan:
I feel like best course of action is to have COM initialized by the runtime as STA by default (and an backwards compat switch to have it use MTA for COM). This way both lines of code for Winforms applications would be needless (as COM would be initialized properly to STA from the beginning) without needing 1. ``STAThreadAttribute``, and 2. setting the current thread appartment state to ``ApartmentState.Unknown`` and then to ``ApartmentState.STA`` which I would agree is not a very efficient hack than just having the runtime just default to STA thread mode by default for every application. I seen no issues before with adding the attribute in 1. to any type of applications (including console applications) so it should be a reasonable option to default it to that within the runtime itself.

terrajobst:
@jkotas 

> > Does anyone see any issues with this?
> 
> It will work, but it is not pretty and it is inefficient. The COM will be first initialized as MTA, then uninitialized, and then initialized again as STA. If it is a temporary solution that we will cleanup later, I think it would ok.

Could we support the configuration as an assembly level attribute? This might eliminate the need for language syntax for putting attributes on `Main` and allows the generator to generate the attribute as well, removing it from the user code entirely.

jkotas:
> I feel like best course of action is to have COM initialized by the runtime as STA by default 

STA is wrong default for console and server applications. STA is the right default for Windows UI frameworks only.

> Could we support the configuration as an assembly level attribute?

There are number of solutions possible that can be used to solve the STAThreadAttribute problem. I agree that allowing the STAThreadAttribute at assembly level would work fine and look nice.

The argument for solving the problem using a language feature has been that there are other attributes that one may want to apply to Main method (e.g. debugger or MethodImpl attributes) and the language feature addresses them all.

RussKie:
@jkotas @terrajobst would this something we'd expect done in .NET6 timeframe, or should I amalgamate  `SetApartmentState` into `Initialize` (https://github.com/dotnet/designs/pull/223#issuecomment-882361786)?

> It will work, but it is not pretty and it is inefficient.

It is both, and a developer will have to have it one way or another until we have a better solution.

terrajobst:
@jaredpar would have to comment on the feasibility of language features in .NET 6.

AraHaan:
> > I feel like best course of action is to have COM initialized by the runtime as STA by default
> 
> STA is wrong default for console and server applications. It would be major breaking change for those. STA is the right default for Windows UI frameworks only.
> 
> > Could we support the configuration as an assembly level attribute?
> 
> There are number of solutions possible that can be used to solve the STAThreadAttribute problem. I agree that allowing the STAThreadAttribute at assembly level would work fine and look nice.
> 
> The argument for solving the problem using a language feature has been that there are other attributes that one may want to apply to Main method (e.g. debugger or MethodImpl attributes) and the language feature addresses them all.

I mean unless you are a console application that then also needs to pull in the UI frameworks for things like ``MessageBox.Show`` or ``TaskDialog``'s.

I also know of other console applications (QuickBMS for example) which utilizes the ``OpenFileDialog`` / ``SaveFileDialog``'s from the Windows APIs (which basically makes it an console-ui application hybrid.

Shayan-To:
> The argument for solving the problem using a language feature has been that there are other attributes that one may want to apply to Main method (e.g. debugger or MethodImpl attributes) and the language feature addresses them all.

@jkotas Is there a proposal for the language feature? Or a relevant issue? Haven't been able to find any.

jkotas:
I have created one: https://github.com/dotnet/csharplang/issues/5045

PR #224, opened 2021-06-04
Workload Target Imports Design

sfoslund:


accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
+# .NET Workloads Target Import Ordering
+
+**Owners** [Sarah Oslund](https://github.com/sfoslund) | [Daniel Plaisted](https://github.com/dsplaisted)
+
+## Background
+
+To support .NET SDK workloads, we [changed the order of targets imports](https://github.com/dotnet/sdk/pull/14393) to allow SDK workloads to change property defaults. When we did this, we also changed the import order of some Windows and WPF targets, as we want to make it a workload in the future, which caused a breaking change. It appears that it is not possible to allow workloads to change property defaults and make windows support a .NET SDK workload without introducing a breaking change. As a result, this document explores possible solutions to minimize the user impact.
```
terrajobst, line 7:
```suggestion
To support .NET SDK workloads, we [changed the order of targets imports](https://github.com/dotnet/sdk/pull/14393) to allow SDK workloads to change property defaults. When we did this, we also changed the import order of some Windows and WPF targets, as we want to make it a workload in the future, which caused a breaking change. It appears that it is not possible to allow workloads to change property defaults and make support for Windows a .NET SDK workload without introducing a breaking change. As a result, this document explores possible solutions to minimize the user impact.
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Original Ordering Change
+
+Originally, the Windows, WindowsDesktop, and workload targets were imported at the end of Microsoft.NET.Sdk.targets, which was almost the last files imported.  However, this was not an appropriate place for the imports if those targets were to override default property values that were set in .NET SDK or MSBuild common targets.  The workload needs a chance to set the property if its not already set before the default logic would do so.
```
terrajobst, line 11:
```suggestion
Originally, the Windows, WindowsDesktop, and workload targets were imported at the end of `Microsoft.NET.Sdk.targets`, which was almost the last files imported.  However, this was not an appropriate place for the imports if those targets were to override default property values that were set in .NET SDK or MSBuild common targets.  The workload needs a chance to set the property if its not already set before the default logic would do so.
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+Some properties depend on other properties, so the order in which they are declared matters. For example, the default value for the `GenerateDependencyFile` property depends on the `TargetFrameworkIdentifier` property, which itself is derived from the `TargetFramework` property. So these declarations need to be evaluated in the correct order for the values to be set correctly.
+
+Likewise, project imports (typically used to import `.targets` or `.props` files) can be conditioned on property values. So if WindowsDesktop targets are imported if `UseWPF` or `UseWindowsForms` is true, then those properties need to be set before the WindowsDesktop project import is evaluated.
+
+The evaluation ordering of MSBuild Targets is usually not important. The order that the Targets are executed is determined by the dependencies between them, not by the order they come in evaluation. The exception to this is that you can override a Target by defining another target with the same name later in evaluation.
```
terrajobst, line 29:
```suggestion
The evaluation ordering of MSBuild targets is usually not important. The order that the targets are executed is determined by the dependencies between them, not by the order they come in evaluation. The exception to this is that you can override a target by defining another target with the same name later in evaluation.
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Exploring Import Order
+
+There are many .props and .targets filed that get imported when building a .NET project. One way to see what is imported and in what order is by preprocessing a project, either with the "-pp:" [command line argument](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference?) or with [MSBuild Structured Log Viewer](https://msbuildlog.com/). This will create a single aggregated project file with all project imports expanded inline.
```
terrajobst, line 33:
```suggestion
There are many `.props` and `.targets` files that get imported when building a .NET project. One way to see what is imported and in what order is by preprocessing a project, either with the `-pp:` [command line argument](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-command-line-reference) or with [MSBuild Structured Log Viewer](https://msbuildlog.com/). This will create a single aggregated project file with all project imports expanded inline.
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+It is also possible to explore the project imports in the Visual Studio Solution Explorer. Clicking on the "Show All Files" button will add an "Imports" node to the solution tree under the project. You can expand this node and explore the tree of imports active in the project. You can also double click on an imported file to open it up in the editor and view its contents.
```
terrajobst, line 35:
```suggestion
It is also possible to explore the project imports in the Visual Studio Solution Explorer. Clicking on the **Show All Files** button will add an **Imports** node to the solution tree under the project. You can expand this node and explore the tree of imports active in the project. You can also double click on an imported file to open it up in the editor and view its contents.
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+![Project Imports in Solution Explorer](./solution-explorer-project-imports.png)
+
+#### Order of Imports in the .NET SDK
+
+The following is a simplified list of the files that are imported in an SDK-style .NET project:
+
+- `Directory.Build.props`
+- Main project file
+- Version logic (Microsoft.NET.DefaultAssemblyInfo.targets)
```
terrajobst, line 45:
```suggestion
- Version logic (`Microsoft.NET.DefaultAssemblyInfo.targets`)
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+- Output path logic (Microsoft.NET.DefaultOutputPaths.targets)
```
terrajobst, line 46:
```suggestion
- Output path logic (`Microsoft.NET.DefaultOutputPaths.targets`)
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+- Publish profile
+- Target Framework parsing (Microsoft.NET.TargetFrameworkInference.targets)
```
terrajobst, line 48:
```suggestion
- Target Framework parsing (`Microsoft.NET.TargetFrameworkInference.targets`)
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+  - Also appends target framework to output and intermediate paths
+- Runtime identifier inference (Microsoft.NET.RuntimeIdentifierInference.targets)
```
terrajobst, line 50:
```suggestion
- Runtime identifier inference (`Microsoft.NET.RuntimeIdentifierInference.targets`)
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+  - Also appends Runtime Identifier to output and intermediate paths
+- Workload targets imports
+- Language targets (e.g. Microsoft.CSharp.targets) and MSBuild common targets
```
terrajobst, line 53:
```suggestion
- Language targets (e.g. `Microsoft.CSharp.targets`) and MSBuild common targets
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+A possible mitigation for some of the cons could be to say that we don’t look for the new file to import independently. Rather, we could say that you need to have a `Directory.Build.props` file and it needs to be in the same folder as `Directory.Build.props`. So a `Directory.AfterTargetFrameworkInference.targets` file outside the repo root wouldn’t automatically be imported unless a `Directory.Build.props` file was going to be imported from outside the repo anyway.
+
+### Change import location of `Directory.Build.targets`
+
+Change`Directory.Build.targets` to be imported after TargetFramework parsing but before workloads and most of the .NET SDK and common targets. This would be a big breaking change but might not affect most people who use `Directory.Build.targets`. It would break people who use `Directory.Build.targets` to override properties or targets in the MSBuild common targets.
+
+#### Pros:
+
+- Logic to enable workloads in `Directory.Build.targets` would “just work”, which is likely what developers expect
```
terrajobst, line 109: I strongly prefer this option. It took a while to evangelize the use of `Directory.Build.props` and `Directory.Build.targets`. I believe we should strive to optimize for letting developers do 99% of their tweaks by just these files. Having to put more stuff in new files whose import order is slightly different is a complexity nightmare. Yes, this is a breaking change but I think that's a perfect example where the net harm of the breaking change is *much* lower than the additional cognitive load over more files over time.

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Cons
+
+- This would be a breaking change, possibly a very big one
+- Implementation would be somewhat complicated:
+  - Logic would be split betweenthe .NET SDK and MSBuild, as the .NET SDK would now be responsible for importing `Directory.Build.targets` for projects that use the .NET SDK, but other projects would still need MSBuild to import `Directory.Build.targets`
+  - The new behavior would need to be behind a change wave and ideally a separate opt-out flag
+- This goes against the original intent behind `Directory.Build.targets`, which was that it would be imported after all built-in targets.  However, the .NET SDK (or any MSBuild SDK) already break this.
+  - As part of this work we could possibly add another extension point that allows you to specify `.targets` files that are imported after all other projects or imports.  This would likely be an MSBuild engine feature.
```
terrajobst, line 118: It would be good if you could state a preference from your team as you're the experts.

terrajobst (commented):
This is very well written document!

jaredpar:
Have we evaluated how this would impact some of our repos? Basically will it cause us to change? I think that would be instructive for understanding the potential scope of the change for customers.

accepted/2021/workloads/workload-target-imports.md
aaronla-ms: Would it be at all possible to give users a bit more control over shimming each of these stages? E.g. in addition to Directory.Build.targets, maybe also Directory.Build.AfterSdks.targets and Directory.Build.BeforeSdk.targets. Sure, you'll still probably break things when you change things around, but then at least users that relied on the order would have a destination to move their definitions to.

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+- Language targets (e.g. `Microsoft.CSharp.targets`) and MSBuild common targets
+- `Directory.Build.targets`
+- (Rest of) .NET SDK targets
+- Old location for Windows, WindowsDesktop, and workload targets imports
+
+#### `Directory.Build.targets` import location
+
+Conceptually, `Directory.Build.targets` is imported after the body of the main project file. The exact location it is imported is not something most developers likely think about, but it is a good place to put common build logic that depends on properties set in the project file, such as the `TargetFramework`.
+
+There's not a perfect place to import `Directory.Build.targets`. However, given what we've learned, it may be that the best place to import it is after the TargetFramework parsing, and before the workloads are imported.  That way the logic in `Directory.Build.targets` would still be able to depend on the parsed TargetFramework information, but would not be able to override all of the targets and properties set by the .NET SDK and MSBuild common targets that it can today.
```
aaronla-ms, line 62: Would it be at all possible to give users a bit more control over shimming each of these stages? E.g. in addition to Directory.Build.targets, maybe also Directory.Build.AfterSdks.targets and Directory.Build.BeforeSdk.targets. Sure, you'll still probably break things when you change things around, but then at least users that relied on the order would have a destination to move their definitions to.

aaronla-ms: Also, very related to ordering, is there any chance that msbuild might eventually support lazily-evaluated properties? A large minority of the build ordering issues I help peers with revolves around reading properties too early (e.g. trying to read OutputPath in Directory.Build.props). Being able to set these using lazy evaluation (e.g. like gnu make `=` vs `:=`) in these cases would resolve a few issues. Admittedly, would create its own issues too, but worth a thought here.

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,118 @@
...
+
+- Possible performance impact
+- Adds another way for a build to “leak out” of a repo root (which could be a security issue)
+- More extension points may be needed in the future, and we probably don’t want to add a new auto-imported file for each one
+
+A possible mitigation for some of the cons could be to say that we don’t look for the new file to import independently. Rather, we could say that you need to have a `Directory.Build.props` file and it needs to be in the same folder as `Directory.Build.props`. So a `Directory.AfterTargetFrameworkInference.targets` file outside the repo root wouldn’t automatically be imported unless a `Directory.Build.props` file was going to be imported from outside the repo anyway.
+
+### Change import location of `Directory.Build.targets`
+
+Change`Directory.Build.targets` to be imported after TargetFramework parsing but before workloads and most of the .NET SDK and common targets. This would be a big breaking change but might not affect most people who use `Directory.Build.targets`. It would break people who use `Directory.Build.targets` to override properties or targets in the MSBuild common targets.
```
poke, line 105: Do we have a list of which targets would be affected here? (Alternatively, a link to the files, so we could check if this affected any of our use cases?)

accepted/2021/workloads/workload-target-imports.md
jaredpar: It's hard to grok how significant this is or is not though. Can we get some examples of which properties we can no longer override in `Directory.Build.targets` if we take this approach. At the moment it's pretty nebulous hence it's hard to decide if this is or is not the right approach.

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,120 @@
+# .NET Workloads Target Import Ordering
+
+**Owners** [Sarah Oslund](https://github.com/sfoslund) | [Daniel Plaisted](https://github.com/dsplaisted)
+
+## Background
+
+To support .NET SDK workloads, we [changed the order of targets imports](https://github.com/dotnet/sdk/pull/14393) to allow SDK workloads to change property defaults. When we did this, we also changed the import order of some Windows and WPF targets, as we want to make it a workload in the future, which caused a breaking change. It appears that it is not possible to allow workloads to change property defaults and make support for Windows a .NET SDK workload without introducing a breaking change. As a result, this document explores possible solutions to minimize the user impact.
```
RussKie, line 7: Did you mean "Windows Forms and WPF"?
```suggestion
To support .NET SDK workloads, we [changed the order of targets imports](https://github.com/dotnet/sdk/pull/14393) to allow SDK workloads to change property defaults. When we did this, we also changed the import order of some Windows Forms and WPF targets, as we want to make it a workload in the future, which caused a breaking change. It appears that it is not possible to allow workloads to change property defaults and make support for Windows a .NET SDK workload without introducing a breaking change. As a result, this document explores possible solutions to minimize the user impact.
```

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,120 @@
...
+#### Proposed Breaking Change Mitigations
+
+- An opt out property would be implemented to allow for preserving existing behavior.
+- We would like to make this change only apply to projects targeting .NET 6 and higher.
+
+## Alternative solutions
+
+### Extension Point via Property
+
+Support an `AfterTargetFrameworkInferenceTargets` property. This property could be used by creating a `Directory.AfterTargetFrameworkInference.targets` file and putting the following in the `Directory.Build.props` file located in the same folder:
```
RussKie, line 90: If it is to accommodate workloads, why not introduce something like `Workload.Build.props` and `Workload.Build.targets`?

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,120 @@
...
+
+Change`Directory.Build.targets` to be imported after TargetFramework parsing but before workloads and most of the .NET SDK and common targets. This would be a big breaking change but might not affect most people who use `Directory.Build.targets`. It would break people who use `Directory.Build.targets` to override properties or targets in the MSBuild common targets.
+
+#### Pros:
+
+- Logic to enable workloads in `Directory.Build.targets` would “just work”, which is likely what developers expect
+
+#### Cons
+
+- This would be a breaking change, possibly a very big one
```
RussKie, line 74: A massive one...

accepted/2021/workloads/workload-target-imports.md
```diff
@@ -0,0 +1,120 @@
...
+
+The Windows and WindowsDesktop targets were moved together with the workload targets import, as we expect them to eventually become part of a workload.  However, this moved those imports before the import of `Directory.Build.targets`, which meant that properties set in that (such as `UseWPF` and `UseWindowsForms`) would no longer take effect.
+
+This is an issue not just because it's a breaking change, but because it means that `Directory.Build.targets` can never be used to set a property that determines whether a workload is used.
+
+### MSBuild Importing Context
+
+#### MSBuild Evaluation
+
+We are primarily concerned with "[Pass 1](https://github.com/dotnet/msbuild/blob/6f9e0d620718578aab8dafc439d4501339fa4810/src/Build/Evaluation/Evaluator.cs#L613)" of [MSBuild Evaluation](https://docs.microsoft.com/en-us/visualstudio/msbuild/build-process-overview#evaluation-phase), where properties are evaluated and project imports are loaded.
```
RussKie, line 23: https://review.docs.microsoft.com/help/contribute/links-how-to?branch=master#locale-codes-in-external-links
```suggestion
We are primarily concerned with "[Pass 1](https://github.com/dotnet/msbuild/blob/6f9e0d620718578aab8dafc439d4501339fa4810/src/Build/Evaluation/Evaluator.cs#L613)" of [MSBuild Evaluation](https://docs.microsoft.com/visualstudio/msbuild/build-process-overview#evaluation-phase), where properties are evaluated and project imports are loaded.
```

ditto below

PR #226, opened 2021-06-08
Add a design document for System.Text.Json polymorphic serialization

eiriktsarpalis:
[Draft Implementation PR](https://github.com/dotnet/runtime/pull/53882)

### DIFF
--- a/accepted/2021/json-polymorphism.md
+++ b/accepted/2021/json-polymorphism.md
@@ -57,7 +57,7 @@ public class Foo
 
 public class Bar : Foo
 {
-  public int B { get; set; }
+    public int B { get; set; }
 }
 
 public class Baz : Bar
--- a/accepted/2021/json-polymorphism.md
+++ b/accepted/2021/json-polymorphism.md
@@ -88,8 +88,8 @@ public class Foo
 which will result in the above values now being serialized as follows:
 ```csharp
 JsonSerializer.Serialize<Foo>(foo1); // { "A" : 1 }
-JsonSerializer.Serialize<Foo>(foo1); // { "A" : 1, "B" : 2 }
-JsonSerializer.Serialize<Foo>(foo2); // { "A" : 1, "B" : 2, "C" : 3 }
+JsonSerializer.Serialize<Foo>(foo2); // { "A" : 1, "B" : 2 }
+JsonSerializer.Serialize<Foo>(foo3); // { "A" : 1, "B" : 2, "C" : 3 }
 ```
 Note that the `JsonPolymorphic` attribute is not inherited by derived types.
 In the above example `Bar` inherits from `Foo` yet is not polymorphic in its own right:
@@ -194,7 +194,7 @@ var options = new JsonSerializerOptions
 
 ### Open Questions
 
-The type discriminator feature has two possible design alternatives to be considered,
+The type discriminator semantics could be implemented following two possible alternatives,
 which for the purposes of this document I will be calling "strict mode" and "lax mode".
 Each approach comes with its own sets of trade-offs.
 
--- a/accepted/2021/json-polymorphism.md
+++ b/accepted/2021/json-polymorphism.md
@@ -42,7 +42,7 @@ largely orthogonal features:
     arbitrary classes that can be specified by the user. It trivially dispatches to the converter corresponding 
     to the runtime type without emitting any metadata on the wire and does not provide any provision for 
     polymorphic deserialization.
-2. 	Polymorphism with type discriminators ("tagged polymorphism"): classes can be serialized and deserialized
+2.  Polymorphism with type discriminators ("tagged polymorphism"): classes can be serialized and deserialized
     polymorphically by emitting a type discriminator ("tag") on the wire. Users must explicitly associate each
     supported subtype of a given declared type with a string identifier.
 
--- a/accepted/2021/json-polymorphism.md
+++ b/accepted/2021/json-polymorphism.md
@@ -8,43 +8,25 @@ This documents describes the proposed design for extending [polymorphism support
 
 ## Background
 
-By default, System.Text.Json will serialize a value using a converter derived from its declared type,
-regardless of what the runtime type of the value might be. This behavior is in line with the
-Liskov substitution principle, in that the serialization contract is unique (or "monomorphic") for a given type `T`,
-regardless of what subtype of `T` we end up serializing at runtime.
+By default, System.Text.Json will serialize a value using a converter based on its declared type, regardless of what the runtime type of the value might be. This behavior is in line with the Liskov substitution principle, in that the serialization contract is unique (or "monomorphic") for a given type `T`, regardless of what subtype of `T` we end up serializing at runtime.
 
-A notable exception to this rule is members of type `object`, in which case the runtime type of the value
-is looked up and serialization is dispatched to the converter corresponding to that runtime type.
-This is an instance of _polymorphic serialization_, in the sense that the schema might vary depending on
-the runtime type a given `object` instance might have. 
+A notable exception to this rule is members of type `object`, in which case the runtime type of the value is looked up and serialization is dispatched to the converter corresponding to that runtime type. This is an instance of _polymorphic serialization_, in the sense that the schema might vary depending on the runtime type a given `object` instance might have. 
 
-Conversely, in _polymorphic deserialization_ the runtime type of a deserialized value might vary depending on the
-shape of the input encoding. Currently, System.Text.Json does not offer any form of support for polymorphic 
-deserialization.
+Conversely, in _polymorphic deserialization_ the runtime type of a deserialized value might vary depending on the shape of the input encoding. Currently, System.Text.Json does not offer any form of support for polymorphic deserialization. It should be noted that deserialization `object` _is_ currently supported, however that is implemented via the `JsonElement` or `JsonNode` types rather than using a true form of polymorphic deserialization.
 
-We have received a number of user requests to add polymorphic serialization and deserialization support 
-to System.Text.Json. This can be a useful feature in domains where exporting type hierarchies is desirable,
-for example when serializing tree-like data structures or discriminated unions.
+We have received a number of user requests to add polymorphic serialization and deserialization support to System.Text.Json. This can be a useful feature in domains where exporting type hierarchies is desirable, for example when serializing tree-like data structures or discriminated unions.
 
 It should be noted however that polymorphic serialization comes with a few security risks:
 
-* Polymorphic serialization applied indiscriminately can result in unintended data leaks, 
-  since properties of unexpected derived types may end up written on the wire.
-* Polymorphic deserialization can be vulnerable when deserializing untrusted data,
-  in certain cases leading to remote code execution attacks.
+* Polymorphic serialization applied indiscriminately can result in unintended data leaks, since properties of unexpected derived types may end up written on the wire.
+* Polymorphic deserialization can be vulnerable when deserializing untrusted data, in certain cases leading to remote code execution attacks.
 
 ## Introduction
 
-The proposed design for polymorphic serialization in System.Text.Json can be split into two
-largely orthogonal features:
+The proposed design for polymorphic serialization in System.Text.Json can be split into two largely orthogonal features:
 
-1.  Simple Polymorphic serialization: extends the existing serialization infrastructure for `object` types to 
-    arbitrary classes that can be specified by the user. It trivially dispatches to the converter corresponding 
-    to the runtime type without emitting any metadata on the wire and does not provide any provision for 
-    polymorphic deserialization.
-2.  Polymorphism with type discriminators ("tagged polymorphism"): classes can be serialized and deserialized
-    polymorphically by emitting a type discriminator ("tag") on the wire. Users must explicitly associate each
-    supported subtype of a given declared type with a string identifier.
+1.  Simple Polymorphic serialization: extends the existing serialization infrastructure for `object` types to arbitrary classes that can be specified by the user. It trivially dispatches to the converter corresponding to the runtime type without emitting any metadata on the wire and does not provide any provision for polymorphic deserialization.
+2.  Polymorphism with type discriminators ("tagged polymorphism"): classes can be serialized and deserialized polymorphically by emitting a type discriminator ("tag") on the wire. Users must explicitly associate each supported subtype of a given declared type with a string identifier.
 
 ## Simple Polymorphic Serialization
 
@@ -65,8 +47,7 @@ public class Baz : Bar
     public int C { get; set; }
 }
 ```
-Currently, when serializing a `Bar` instance as type `Foo`
-the serializer will apply the JSON schema derived from the type `Foo`:
+Currently, when serializing a `Bar` instance as type `Foo` the serializer will apply the JSON schema derived from the type `Foo`:
 ```csharp
 Foo foo1 = new Foo { A = 1 };
 Foo foo2 = new Bar { A = 1, B = 2 };
@@ -76,8 +57,7 @@ JsonSerializer.Serialize<Foo>(foo1); // { "A" : 1 }
 JsonSerializer.Serialize<Foo>(foo2); // { "A" : 1 }
 JsonSerializer.Serialize<Foo>(foo3); // { "A" : 1 }
 ```
-Under the new proposal we can change this behaviour by annotating 
-the base class (or interface) with the `JsonPolymorphicType` attribute:
+Under the new proposal we can change this behaviour by annotating the base class (or interface) with the `JsonPolymorphicType` attribute:
 ```csharp
 [JsonPolymorphicType]
 public class Foo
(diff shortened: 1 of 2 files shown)


accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+## Introduction
+
+The proposed design for polymorphic serialization in System.Text.Json can be split into two
+largely orthogonal features:
+
+1.  Simple Polymorphic serialization: extends the existing serialization infrastructure for `object` types to 
+    arbitrary classes that can be specified by the user. It trivially dispatches to the converter corresponding 
+    to the runtime type without emitting any metadata on the wire and does not provide any provision for 
+    polymorphic deserialization.
+2. 	Polymorphism with type discriminators ("tagged polymorphism"): classes can be serialized and deserialized
```
watfordsuzy, line 45:
```suggestion
2.  Polymorphism with type discriminators ("tagged polymorphism"): classes can be serialized and deserialized
```

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+Consider the following type hierarchy:
+```csharp
+public class Foo
+{
+    public int A { get; set; }
+}
+
+public class Bar : Foo
+{
+  public int B { get; set; }
```
watfordsuzy, line 60:
```suggestion
    public int B { get; set; }
```

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+}
+
+public class Baz : Bar
+{
+  public int C { get; set; }
```
watfordsuzy, line 65:
```suggestion
    public int C { get; set; }
```

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+
+This feature allows users to opt in to polymorphic serialization for a given type
+by associating string identifiers with particular subtypes in the hierarchy.
+These identifiers are written to the wire so this brand of polymorphism is roundtrippable.
+
+At the core of the design is the introduction of `JsonKnownType` attribute that can
+be applied to type hierarchies like so:
+```csharp
+[JsonKnownType(typeof(Derived1), "derived1")]
+[JsonKnownType(typeof(Derived2), "derived2")]
```
layomia, line 138: Ditto, could a user want to ignore these attributes for non-owned types? Is that a valid scenario?

In this case it looks like they could provide a `TypeDiscriminatorConfiguration` for the type with no known types... assuming we like that behavior and runtime `JsonSerializerOptions` win over design-time attributes.

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+This documents describes the proposed design for extending [polymorphism support](https://github.com/dotnet/runtime/issues/45189) in System.Text.Json.
+
+[Draft Implementation PR](https://github.com/dotnet/runtime/pull/53882).
+
+## Background
+
+By default, System.Text.Json will serialize a value using a converter derived from its declared type,
+regardless of what the runtime type of the value might be. This behavior is in line with the
+Liskov substitution principle, in that the serialization contract is unique (or "monomorphic") for a given type `T`,
+regardless of what subtype of `T` we end up serializing at runtime.
```
layomia, line 14: nit: avoid line-breaks within a paragraph, here and below.
```suggestion
By default, System.Text.Json will serialize a value using a converter derived from its declared type, regardless of what the runtime type of the value might be. This behavior is in line with the Liskov substitution principle, in that the serialization contract is unique (or "monomorphic") for a given type `T`, regardless of what subtype of `T` we end up serializing at runtime.
```

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+        new TypeDiscriminatorConfiguration(typeof(Base))
+          .WithKnownType(typeof(Derived1), "derived1")
+          .WithKnownType(typeof(Derived2), "derived2")
+    }
+};
+```
+
+### Open Questions
+
+The type discriminator feature has two possible design alternatives to be considered,
```
layomia, line 197: Are these alternatives to the design of the type discriminator feature, or choices regarding how it works?

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+Bar bar = new Baz { A = 1, B = 2, C = 3 };
+JsonSerializer.Serialize<Bar>(bar); // { "A" : 1, "B" : 2 }
+```
+If annotating the base class with an attribute is not possible,
+polymorphism can alternatively be opted in for a type using the
+new `JsonSerializerOptions.SupportedPolymorphicTypes` predicate:
+```csharp
+public class JsonSerializerOptions
+{
+  public Func<Type, bool> SupportedPolymorphicTypes { get; set; }
```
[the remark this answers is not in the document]
eiriktsarpalis: I was having similar thoughts. What would be some suggested alternatives? `SupportedPolymorphicTypesPredicate` perhaps?

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+}
+```
+Applied to the example above:
+```csharp
+var options = new JsonSerializerOptions { SupportedPolymorphicTypes = type => type == typeof(Foo) };
```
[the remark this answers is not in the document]
eiriktsarpalis: Correct, `typeof(object)` behaviour is always hardcoded.

accepted/2021/json-polymorphism.md
eiriktsarpalis: Similarly, one could specify a custom converter to opt out from these attributes. Using an empty `TypeDiscriminatorConfiguration` instance would be a valid alternative, provided we opt for "lax mode" semantics as described in the "open questions" section.

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+{
+    TypeDiscriminatorConfigurations =
+    {
+        new TypeDiscriminatorConfiguration<Base>()
+          .WithKnownType<Derived1>("derived1")
+          .WithKnownType<Derived2>("derived2")
+    }
+};
+```
+or alternatively
```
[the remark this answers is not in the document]
eiriktsarpalis: Having just the non-generic one should be sufficient, however the generic one is type-safe so users might appreciate having that option.

accepted/2021/json-polymorphism.md
eiriktsarpalis: Good point, I've changed the wording to make this more clear.

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+## Polymorphism with type discriminators
+
+This feature allows users to opt in to polymorphic serialization for a given type
+by associating string identifiers with particular subtypes in the hierarchy.
+These identifiers are written to the wire so this brand of polymorphism is roundtrippable.
+
+At the core of the design is the introduction of `JsonKnownType` attribute that can
+be applied to type hierarchies like so:
+```csharp
+[JsonKnownType(typeof(Derived1), "derived1")]
```
FiniteReality, line 137: Would users be able to specify the name of the discriminator field in the serialized model here? The examples use `$type`, but not all web APIs which use discriminated union-like types use that naming.

eiriktsarpalis: > Would users be able to specify the name of the discriminator field in the serialized model here? The examples use $type, but not all web APIs which use discriminated union-like types use that naming.

Yes, that is something we are considering although it has not been included in the prototype for now. There are a few infrastructural problems to be sorted out, for example our metadata parsing implementation currently requires that all metadata properties start with `$` and are placed before any other properties in the object.

> Furthermore: are non-string values permitted? It's also often the case that web APIs use integers for these discriminator values, which would often be mapped to an enum in C# code.

Not in the prototype, but you're making a good point that this should be supported. We would probably need to make the `JsonKnownTypeAttribute` accept `object` parameters but require that each type uses a unique identifier type, so the following configuration would be invalid:
```csharp
[JsonKnownType(typeof(Bar), 1)]
[JsonKnownType(typeof(Baz), "baz")]
public class Foo {}
```
As far as supported identifier types go, am I right to assume that only `string` and `int` values should be accepted?

FiniteReality: > for example our metadata parsing implementation currently requires that all metadata properties start with `$` and are placed before any other properties in the object.

Yeah - in my own converters I've resorted to making a copy of the reader, finding and deserializing the discriminator, and then using the reader copy to deserialize as the actual type, updating the original reader with the copy after the fact. While this fixes the "must be the first field in the JSON object" issue, it's not that great because it parses the value multiple times.

> As far as supported identifier types go, am I right to assume that only string and int values should be accepted?

As far as I'm aware, yes, only `string` and `int` are used for these. However, I think it may be convenient to support enum types too, as they're easily converted to `int`, and are (IMO) the natural type to use for a discriminator.

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+Foo foo3 = new Baz { A = 1, B = 2, C = 3 };
+
+JsonSerializer.Serialize<Foo>(foo1); // { "A" : 1 }
+JsonSerializer.Serialize<Foo>(foo2); // { "A" : 1 }
+JsonSerializer.Serialize<Foo>(foo3); // { "A" : 1 }
+```
+Under the new proposal we can change this behaviour by annotating 
+the base class (or interface) with the `JsonPolymorphicType` attribute:
+```csharp
+[JsonPolymorhicType]
```
jeffhandley, line 82:
```suggestion
[JsonPolymorphicType]
```

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+  ...
+}
+```
+which will result in the above values now being serialized as follows:
+```csharp
+JsonSerializer.Serialize<Foo>(foo1); // { "A" : 1 }
+JsonSerializer.Serialize<Foo>(foo2); // { "A" : 1, "B" : 2 }
+JsonSerializer.Serialize<Foo>(foo3); // { "A" : 1, "B" : 2, "C" : 3 }
+```
+Note that the `JsonPolymorphic` attribute is not inherited by derived types.
```
jeffhandley, line 94:
```suggestion
Note that the `JsonPolymorphicType` attribute is not inherited by derived types.
```

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+```csharp
+Foo foo1 = new Foo { A = 1 };
+Foo foo2 = new Bar { A = 1, B = 2 };
+Foo foo3 = new Baz { A = 1, B = 2, C = 3 };
+
+JsonSerializer.Serialize<Foo>(foo1); // { "A" : 1 }
+JsonSerializer.Serialize<Foo>(foo2); // { "A" : 1 }
+JsonSerializer.Serialize<Foo>(foo3); // { "A" : 1 }
+```
+Under the new proposal we can change this behaviour by annotating
```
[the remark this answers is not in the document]
layomia: Right, that's true. The custom converter escape hatch should be sufficient.

accepted/2021/json-polymorphism.md
layomia: 👍 https://github.com/dotnet/designs/pull/226#discussion_r650055156

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+
+**Owner** [Eirik Tsarpalis](https://github.com/eiriktsarpalis)
+
+This documents describes the proposed design for extending [polymorphism support](https://github.com/dotnet/runtime/issues/45189) in System.Text.Json.
+
+[Draft Implementation PR](https://github.com/dotnet/runtime/pull/53882).
+
+## Background
+
+By default, System.Text.Json will serialize a value using a converter derived from its declared type,
```
steveharter, line 11: "using a converter derived from its declared type"
->
"using a converter based on its declared type"

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+regardless of what subtype of `T` we end up serializing at runtime.
+
+A notable exception to this rule is members of type `object`, in which case the runtime type of the value
+is looked up and serialization is dispatched to the converter corresponding to that runtime type.
+This is an instance of _polymorphic serialization_, in the sense that the schema might vary depending on
+the runtime type a given `object` instance might have. 
+
+Conversely, in _polymorphic deserialization_ the runtime type of a deserialized value might vary depending on the
+shape of the input encoding. Currently, System.Text.Json does not offer any form of support for polymorphic 
+deserialization.
```
steveharter, line 23: "System.Text.Json does not offer any form of support for polymorphic 
deserialization."

It may be useful to explain that a System.Object property is materialized with either a `JsonElement` or a `JsonNode`. Although not polymorphic in the true sense, it does fill the feature gap (and provide a workaround) to be able to deserialize arbitrary JSON that may be aligned to a polymorphic model.

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+Bar bar = new Baz { A = 1, B = 2, C = 3 };
+JsonSerializer.Serialize<Bar>(bar); // { "A" : 1, "B" : 2 }
+```
+If annotating the base class with an attribute is not possible,
+polymorphism can alternatively be opted in for a type using the
+new `JsonSerializerOptions.SupportedPolymorphicTypes` predicate:
+```csharp
+public class JsonSerializerOptions
+{
+  public Func<Type, bool> SupportedPolymorphicTypes { get; set; }
```
steveharter, line 106: I assume the result is cached on the corresponding JsonTypeInfo, in order to prevent potential perf issues with a complex func.

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+JsonSerializer.Serialize<Base>(new Derived3()); // throws NotSupportedException
+JsonSerializer.Serialize<Base>(new OtherDerived1()); // throws NotSupportedException
+JsonSerializer.Serialize<Base>(new Base()); // throws NotSupportedException
+```
+Any runtime type that is not associated with a type discriminator will be rejected, 
+including instances of the base type itself. This approach has a few drawbacks:
+
+* Does not work well with open hierarchies: any new derived types will have to be explicitly opted in.
+* Each runtime type must use a separate type identifier.
+* Interfaces or abstract classes cannot specify type discriminators.
```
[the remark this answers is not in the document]
FiniteReality: I think the first point should work just fine, if reflection is done at runtime - it probably wouldn't work so well with a source generator though.

As for interfaces/abstract classes, how would you instantiate one when deserializing a JSON payload?

accepted/2021/json-polymorphism.md
```diff
@@ -0,0 +1,275 @@
...
+
+JsonSerializer.Deserialize<Base>(json1); // uses Derived1 as runtime type
+JsonSerializer.Deserialize<Base>(json2); // uses Derived2 as runtime type
+```
+Alternatively, users can specify known type configuration using the
+`JsonSerializerOptions.TypeDiscriminatorConfigurations` property:
+```csharp
+public class JsonSerializerOptions
+{
+    public IList<TypeDiscriminatorConfiguration> TypeDiscriminatorConfigurations { get; }
```
steveharter, line 167: I was thinking this should be a dictionary, but then I thought if we expose a way to get `JsonTypeInfo`, someone could set the list directly for the type. e.g.:
```cs
JsonTypeInfo typeinfo = options.GetTypeInfo<Base>();
IList<KnownType> knownTypes = typeinfo.KnownTypes;
knownTypes.Add<Derived1>("derived1");
knownTypes.Add<Derived2>("derived2");
```

eiriktsarpalis: FWIW the configuration is used to construct [a class that is backed by a bidirectional map](https://github.com/dotnet/runtime/pull/53882/files#diff-ea8cdec34c5ef04cf0ce79e23e722bccb113c41251c53d36e4d7697dc39abf36).

accepted/2021/json-polymorphism.md
eiriktsarpalis: Correct, cf. https://github.com/dotnet/runtime/pull/53882/files#diff-8d841bf86f2a9da83758276bd17f55d002ee5b55f58b72da658df351a7c22bc0R553

accepted/2021/json-polymorphism.md
eiriktsarpalis: Makes sense, added clarification.

accepted/2021/json-polymorphism.md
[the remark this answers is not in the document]
eiriktsarpalis: The wording might be slightly ambiguous here, updated to better clarify.

eiriktsarpalis:
I've split this design document into to separate API proposals in the relevant issues: https://github.com/dotnet/runtime/issues/29937 and https://github.com/dotnet/runtime/issues/30083.

PR #228, opened 2021-06-16
Add ability to sign extractable binaries included in a singlefile app.

VSadov:
This is a proposal to add a build-time hook that would provide an opportunity to sign files included in a single file application.

In general the solution could be used for any kind of post-build instrumentation. For example obfuscation.

Related: https://github.com/dotnet/sdk/issues/16939


VSadov:
CC: @agocke, @jkotas, @vitek-karas, @hoyosjs

agocke (approved):
LGTM

proposed/before_bundle_build_hook.md
```diff
@@ -0,0 +1,49 @@
...
+
+In addition to signing a similar difficulties may affect other scenarios that rely on post-build modification of build artefacts. For example obfuscation and instrumentation.
+
+## Current state (as of .NET 6)
+
+The problem of signing applies primarily to files that are extracted to the file system as a part of execution. For the assemblies that are not extracted the problem is generally not a concern. 
+With transitioning to superhost technology on all platforms in 6.0 a need to extract files is greatly reduced as native binaries that belong to the dotnet platform are no longer extracted and in fact cannot be extracted, since they are monolithically linked with the app. 
+
+The files that still can be extracted:
+-	Native dependencies included via a nuget references. (this is, technically, the only truly supported scenario to include a native dependency)
```
hoyosjs, line 17: What about P2P references to VCX projects? It's not obvious. The other thing that should be done is tell the user to hook their signing targets to AfterBuild, but that has it's own set of limitations like machine marshalling (which was described in the original issue).

proposed/before_bundle_build_hook.md
```diff
@@ -0,0 +1,49 @@
...
+-	A Property `AppHostFile` that will specify the apphost.  
+The reason for this is that the host will be in the list of files to bundle, but generally need not be signed.
+
+It is possible that the signing tool will need to copy files in the process of signing. That could happen if the file is a shared item not owned by the build, for example the file comes from a nuget cache. In such case it is expected that the tool will modify the path of the corresponding `FilesToBundle` item to point to the signed copy.
+
+## Alternative approaches considered:
+
+-	Unbundle/Rebundle support
+
+Provide a way to unbundle/rebundle a singlefile app.
```
hoyosjs, line 44: There's one scenario where this is valuable

- Apps that need to get signed on different platforms: dylib requiring signatures on macOS, dll's needing signing but that can't be done on the machine (like we do for security and compliance reasons).
- The general tool-needed machine marshaling issue - a lot of things need an apple device such as `codesign` and notarization/hardening, or a windows device because the tool operates on Windows only, etc...

proposed/before_bundle_build_hook.md
```diff
@@ -0,0 +1,49 @@
+# Add ability to sign extractable binaries included in a singlefile app.
+
+There is a known issue around singlefile apps when the user needs to sign included artefacts or otherwise modify them. A typical solution for regular build artefacts is to sign binaries after the build is completed.
+A problem arises for files that are included in a singlefile bundle. The fact that included files are not signed is observable for extractable files, but the user cannot easily address that.
```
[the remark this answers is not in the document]
vitek-karas: All versions of single-file bundle support extraction, it's just that for 5 and 6 it's typically not enabled by default, but it's still possible to get that behavior. So this applies to all versions of single-file.

I don't know what "zipped components" refer to...

proposed/before_bundle_build_hook.md
vitek-karas: Do P2P references to VCX actually work? The typical problem with native dependencies in .NET apps is that the .NET app when consuming the dependency needs to know with RID the dependency supports and potentially consume multiple versions of the native dependency (for multiple RIDs - for portable apps).

I agree that if there's a supported scenario in this area it should be mentioned in the doc though.

proposed/before_bundle_build_hook.md
vitek-karas: +1 that this should be mentioned in the doc. We decided to implement a simple solution for .NET 6 and collect more feedback before investing into a more complex solution which could support "async" processing of the files in the bundle.

vitek-karas (approved):

proposed/before_bundle_build_hook.md
hoyosjs: There's the issue of I am not sure if it's supported, but I know it works and it is used by and large by customers (at least internal ones). They do P2P and then add the dll's needed to Content* ItemGroups.

VSadov: I assume that whatever supported or unsupported mechanism used, ultimately it would need to arrange the inclusion of a native file into the list of bundled files. 

I'd like to not get deep into how an app may gain a native dependency as it is a separate concern and should be orthogonal to this proposal. I just wanted to mention that native dependencies do exist, but most likely signed as a part of their build and prior to building of the app that consumes them. 

In either case the proposed mechanism should be sufficient to sign native dependencies if such need arises.

proposed/before_bundle_build_hook.md
VSadov: Right. The proposal assumes that signing can be logically done as a part of a build, even if files need to be sent to a service or copied around.  
If build and sign machines are different, the copying would need to happen regardless of the solution, but not having to do it while the build is waiting may have some advantages. I can imagine situations when the service is very slow, unreliable or need to happen on a disconnected machine and holding up the build could become a problem. How big such a problem is in practice - it would be useful to listen to feedback here.

I will mention this concern.

proposed/before_bundle_build_hook.md
VSadov: > There's the issue of I am not sure if it's supported, but I know it works and it is used by and large by customers (at least internal ones). They do P2P and then add the dll's needed to Content* ItemGroups.

Assuming this works with singlefile apps, the files must make it to the current `_FilesToBundle` ItemGroup. In such case the proposed solution will work if these files need to be signed in `BeforeBundle`

VSadov:
Should this be merged? Is "proposed" the right location?

proposed/before_bundle_build_hook.md
```diff
@@ -0,0 +1,53 @@
...
+I believe this is the case that causes the most inconveniences.
+
+The abscense of documented solutions in this area forces users to implement solutions that rely on internal implementation details.  
+For example: [`CollectBundleFilesToSign` target](https://github.com/dotnet/diagnostics/blob/18a8e986ddf28ee3c950b5acf00abc652d87be74/src/Tools/Directory.Build.targets).  
+
+## Proposal for .NET 6
+
+Provide a build-time hook to implement signing logic. The use pattern is expected to be similar to `CollectBundleFilesToSign`, but without relying on undocumented behavior.  
+We will provide:
+-	A target `BeforeBundle` that will be called between `_ComputeFilesToBundle` and `GenerateBundle`
```
vitek-karas, line 33: Looking at the current PR for implementing this - I was under the assumption that the `BeforeBundle` would be target defined by the application - it would be similar to what `BeforeBuild` does - meaning that the right properties would be already computed before it is triggered.
I don't think this design is wrong - it's just that the PR doesn't match it (or I don't understand how MSBuild works - which is very possible).

vitek-karas:
If it's an accepted design we should move the file to the `accepted` folder (and merge).

proposed/before_bundle_build_hook.md
VSadov: MSBuild scripting is not something that I do every day either :-). I am not sure how a user can make their target be called directly, there is probably a way. It is however pretty simple to insert a target into the dependence order between existing targets via `BeforeTargets`, `DependsOnTargets`, so I always assumed that defining documented points between which user inserts their stuff _is_ the way to provide a stable extensibility point in MSBuild.

This is also the way the existing signing support works, although it uses undocumented `_ComputeFilesToBundle`/`_FilesToBundle`

vitek-karas: Great - we should just include a sample of how this will be done in this doc then.

hoyosjs: CallTarget is how you do it. But that's something we actively discourage except for some really specific cases. Indeed, the Before + Depends is probably the best way.

VSadov:
I think I have addressed all suggestions and this is ready to merge.

VSadov:
Thanks everybody!!!

PR #229, opened 2021-06-16
Spec for resolving '-p' in 'dotnet run'

KathleenDollard:
`dotnet run` has an abbreviation inconsistent with its sister commands `dotnet build` and `dotnet publish`. This proposal is to change that with a deprecation path for the current usage.


accepted/2021/dotnet-run-resolve-p.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Design
+
+The basic design is described in the opening: 
+
+* `-p` with syntax that is valid for MSBuild properties will be treated as `--property`.
+* `-p:` will be passed to MSBuild so the an MSBuild error is displayed when the argument is illegal.
+* Any other usage will be treated as `--project` and will receive a warning that this use is deprecated.
+* Usage in the form `-p:` will be allowed but not encouraged or documented. 
+* At some future point, we will remove `-p` for `--project`.
```
dsplaisted, line 35: I'm not sure we need to commit to eventually removing `-p` as `--project`.

```suggestion
* At some future point, we may remove `-p` for `--project`.
```

accepted/2021/dotnet-run-resolve-p.md
```diff
@@ -0,0 +1,49 @@
...
+
+### Warning text
+
+When -p is used (with no colon) in .NET 6, the following warning will be displayed: 
+
+*“The abbreviation of -p for --project is deprecated. Please use --project.”*
```
dsplaisted, line 41:
```suggestion
> Warning NETSDKNNNN: The abbreviation of -p for --project is deprecated. Please use --project.

Where `NNNN` will be replaced with an SDK error code.  In contrast to many other errors and warnings generated by the .NET SDK, this warning will be generated by the .NET CLI parsing code instead of flowing through MSBuild.  So it will not be affected by "warn as error" or "ignore warning" MSBuild settings.
```

I'm not sure about the exact formatting we should use for the warning and error code.

dsplaisted (commented):
Looks good.  I think we should also describe how this will be a breaking change:

- In cases where the output of `dotnet run` is being parsed
- In cases where the path to a project file passed in via `-p` includes an `=`

accepted/2021/dotnet-run-resolve-p.md
```diff
@@ -0,0 +1,64 @@
...
+
+During a deprecation phase, which will be at least the length of .NET 6, older usage of `-p` as `--project` will be recognized as when there is no trailing colon (`-p:`) and the argument is not legal for MSBuild. Users will receive a warning that the abbreviation is deprecated. 
+
+We will interpret `-p` as `--property` when the argument value is legal MSBuild syntax. Legal MSBuild syntax will be defined as: 
+
+* A comma or semi-colon delimited strings of values in the format <name>=<value> or <name>:<value>. We may simplify this to only checking for the equals - the simplest sufficient check will be used.
+
+If you have any scripts that are using “dotnet run” and process the output you could encounter a break. dotnet run typically doesn’t output anything of its own if there are no errors, so you only get the output of the program that is being run. If you have a script or other program wrapping “dotnet run” and parsing the output, the warning would be unexpected text and may cause a failure.
```
dsplaisted, line 34:
```suggestion
If you have any scripts that are using `dotnet run` and process the output you could encounter a break. `dotnet run` typically doesn’t output anything of its own if there are no errors, so you only get the output of the program that is being run. If you have a script or other program wrapping `dotnet run` and parsing the output, the warning would be unexpected text and may cause a failure.
```

dsplaisted (approved):

dsplaisted:
@KathleenDollard Are you going to look at the feedback and eventually merge this?

PR #230, opened 2021-06-25
Clarify workload extension & platform restrictions

mhutch:
Expand explanation of extending workloads and clarify how platform restrictions interact with it.

### DIFF
--- a/accepted/2020/workloads/workload-manifest.md
+++ b/accepted/2020/workloads/workload-manifest.md
@@ -176,8 +176,8 @@ Workload definitions take the following form:
 | `kind` | string | Either `build` or `dev`. Default is `dev`. | No |
 | `description` | string | User-visible description for the workload. | Yes if dev and non-abstract |
 | `packs` | string array | IDs of the packs that are included in the workload. | No |
-| `platforms` | string array | Limits the workload and workloads that extend it to only be shown and installed on these host platforms. The strings are RIDs. | No |
 | `extends` | string array | IDs of "base" workloads whose packs should be included in this workload. | No |
+| `platforms` | string array | Restricts the workload and workloads that extend it to only be shown and installed on these host platforms. Inherits any restrictions imposed by base workloads. The strings are RIDs. | No |
 |`redirect-to` | string | The ID of another workload with which to replace this workload. Cannot coexist with any other keys. | No |
 
 ### Workload Composition
@@ -194,6 +194,7 @@ Although `extends` is a composition system, it may be useful to conceptualize it
 
 Workload composition may also be compared to package managers such as apt-get and NuGet. In this analogy, workloads are metapackages with unversioned dependencies and packs are packages that are only installable transitively.
 
+Platform restrictions are transitive across workload composition: a workload will inherit `platforms` restrictions from the workloads that it extends. The effective set of supported platforms for any workload is the intersection of its supported platforms and the platforms supported by base workloads.
 
 ### Workload Kinds
 
--- a/accepted/2020/workloads/workload-manifest.md
+++ b/accepted/2020/workloads/workload-manifest.md
@@ -176,15 +176,24 @@ Workload definitions take the following form:
 | `kind` | string | Either `build` or `dev`. Default is `dev`. | No |
 | `description` | string | User-visible description for the workload. | Yes if dev and non-abstract |
 | `packs` | string array | IDs of the packs that are included in the workload. | No |
-| `extends` | string array | IDs of workloads whose packs should be included in this workload. | No |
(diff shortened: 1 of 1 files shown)


accepted/2020/workloads/workload-manifest.md
```diff
@@ -82,7 +82,9 @@ Library packs are normal NuGet packages. When a library pack is installed, the n
 
 This location is used as a local feed by NuGet, not a fallback folder. If ones of these packages gets used, it will be extracted into the global packages folder. If the SDK is updated and the library pack is removed or replaced with a newer version, projects that have already been created and restored will continue to be able to use the extracted copy from the global packages folder.
 
-This is intended to be used to pre-download NuGet packages referenced by templates so that a workload can work offline after installation. It should be used sparingly i.e. only for NuGets referenced by the core templates with default or popular options.
+Library packs are intended to be used to pre-download NuGet packages referenced by templates so that a workload can work offline after installation. Core workloads should include library packs very sparingly, if at all. It is strongly recommended that library packs are contained in workloads with the ID suffix `-offline-templates`. An offline template workload should only contain library packs referenced by templates, and should extend the workload that contains the templates.
```
akoeplinger, line 85:
```suggestion
Library packs are intended to be used to pre-download NuGet packages referenced by templates so that a workload can work offline after installation. Core workloads should include library packs very sparingly, if at all. Instead, it is strongly recommended that library packs are contained in separate workloads with the ID suffix `-offline-templates`. An offline template workload should only contain library packs referenced by templates, and should extend the workload that contains the templates.
```

accepted/2020/workloads/workload-manifest.md
```diff
@@ -176,15 +178,25 @@ Workload definitions take the following form:
 | `kind` | string | Either `build` or `dev`. Default is `dev`. | No |
 | `description` | string | User-visible description for the workload. | Yes if dev and non-abstract |
 | `packs` | string array | IDs of the packs that are included in the workload. | No |
-| `extends` | string array | IDs of workloads whose packs should be included in this workload. | No |
-| `platforms` | string array | Limits the workload and workloads that extend it to only be shown and installed on these host platforms. The strings are RIDs. | No |
+| `extends` | string array | IDs of "base" workloads whose packs should be included in this workload. | No |
+| `platforms` | string array | Restricts the workload and workloads that extend it to only be shown and installed on these host platforms. Inherits any restrictions imposed by base workloads. The strings are RIDs. | No |
```
akoeplinger, line 182: does this only support portable RIDs or the full RID graph? e.g. `osx-arm64` vs. `osx.11.0-arm64`

mhutch: It uses the chain in `sdk/[VERSION]/NETCoreSdkRuntimeIdentifierChain.txt`

akoeplinger (approved):

accepted/2020/workloads/workload-manifest.md
```diff
@@ -82,7 +82,9 @@ Library packs are normal NuGet packages. When a library pack is installed, the n
...
 
-This is intended to be used to pre-download NuGet packages referenced by templates so that a workload can work offline after installation. It should be used sparingly i.e. only for NuGets referenced by the core templates with default or popular options.
+Library packs are intended to be used to pre-download NuGet packages referenced by templates so that a workload can work offline after installation. Core workloads should include library packs very sparingly, if at all. Instead, it is strongly recommended that library packs are contained in separate workloads with the ID suffix `-offline-templates`. An offline template workload should only contain library packs referenced by templates, and should extend the workload that contains the templates.
+
+> NOTE: SDK targets should **not** implicitly add `PackageReference` items. They should instead add `FrameworkReference` items that resolve to runtime packs and targeting packs.
```
dsplaisted, line 87: I wouldn't phrase this as strongly, as I think .NET MAUI is going to have implicit PackageReference items.

@jonathanpeppers

dsplaisted (approved):

accepted/2020/workloads/workload-manifest.md
jonathanpeppers: Right now the plan with dotnet/maui is to have some implicit `PackageReference` for things like:
* https://www.nuget.org/packages/Microsoft.ProjectReunion/
* https://www.nuget.org/packages/Xamarin.Google.Android.Material/
* https://www.nuget.org/packages/Xamarin.AndroidX.AppCompat/
* https://www.nuget.org/packages/Microsoft.Maui.Graphics/ - supposed to be a standalone NuGet

I agree with `workloads should include library packs very sparingly`, but it doesn't seem like dotnet/maui is doing that at the moment. dotnet/maui probably wouldn't be a good fit with the `-offline-templates` naming for these either, because they are dependencies for any app.

/cc @Redth

PR #233, opened 2021-07-08
Add architecture targeting proposal

richlander:
This proposal supports [x64 emulation on Arm64 operating systems](https://github.com/dotnet/designs/pull/217). In particular, it enables targeting x64 runtimes with the Arm64 SDK. It also resolves general architecture targeting challenges with the SDK.

Notably, it includes:

- Additive changes in .NET 6.
- Breaking changes in .NET 7.


accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,466 @@
...
+- `dotnet build`
+- `dotnet build -a x64`
+- `dotnet build --os Linux`
+- `dotnet build -a arm64 --os osx`
+
+Assuming a Windows Arm64 SDK, the following RIDs would be used in these examples:
+
+- `win-arm64`
+- `win-x64`
+- `linux-x64`
```
jkotas, line 197:
```suggestion
- `linux-arm64`
```
I would expect this to default to linux-arm64.

richlander: You are right. Oops.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,492 @@
...
+- `dotnet test -r win-x64`
+- `dotnet tool install -r win-x64`
+
+Note: `dotnet watch` is effectively a wrapper over `dotnet run` and `dotnet test` so would follow the same model as they do.
+
+Where appropriate, the SDK should encode a different implicit RID, such as `osx-x64` when targeting .NET Core 3.1 on Apple Silicon macOS machines with the Arm64 .NET 6 SDK. The SDK should understand that .NET Core 3.1 only supports x64 and not Arm64 for macOS. The x64 SDK should provide the same behavior in reverse.
+
+Tool installation with `dotnet tool install` is a particularly problematic scenario. At the time .NET 6 is released, one can expect that close to 100% of .NET tools (packaged via NuGet) will target a verion prior to .NET 6 and therefore be incapable on running on Arm64 (at least on macOS). However, when installed, the SDK will generate an apphost that matches the SDK RID. As a result, the tool will consistently fail to launch (since the required runtime will be missing). The SDK needs to apply the same implicit RID rules for .NET tools that it will apply for `dotnet build`. In addition, `dotnet tool install` should enable explicit RID targeting.
```
jkotas, line 174:
```suggestion
Tool installation with `dotnet tool install` is a particularly problematic scenario. At the time .NET 6 is released, one can expect that close to 100% of .NET tools (packaged via NuGet) will target a version prior to .NET 6 and therefore be incapable on running on Arm64 (at least on macOS). However, when installed, the SDK will generate an apphost that matches the SDK RID. As a result, the tool will consistently fail to launch (since the required runtime will be missing). The SDK needs to apply the same implicit RID rules for .NET tools that it will apply for `dotnet build`. In addition, `dotnet tool install` should enable explicit RID targeting.
```
Typo: verion

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,492 @@
...
+
+Note: Portable apps are not a meaningful option for client apps since almost all client apps require an executable launcher, and there is only ever one generated. Portable apps are only meaningful for commandline apps.
+
+Note: In theory, we could consider an option to create multiple apphosts for different RIDs, either of different names or in different directories. Again, this capability would need to be opt-in.
+
+### Enable RID folders for RID-specific and portable apps
+
+There have been some issues filed (see appendix) that suggest that managed libraries are not resilient to RID-specific libraries changing location as a result of changing build type (portable to RID-specific). We should consider adding an option that creates `runtimes` folder (for compatible RIDs) for RID-specific builds. Perhaps this challenge just needs to be investigated more.
```
jkotas, line 126: This also means that the library is not compatible with single-file. Single-file does not preserve the deep directory structure either.

We have seen both native libraries not compatible with the deep directory structure, and native libraries not compatible with the flat directory structure. We consider these issues to be a design problem with the native library.

For example: https://github.com/dotnet/runtime/issues/25670

richlander: That resonates with me.

This is probably better docs and tools.

richlander: Updated the text based on your feedback. https://github.com/dotnet/designs/pull/233/commits/92af8d18e7f2759bb4a1a94681de30405ae97c1e

terrajobst:
FYI, I've changed the author tag so it works with the indexing tools and designsof.net.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,496 @@
...
+   - The macOS Arm64 SDK should treat pre .NET 6 versions as X64 by default.
+   - The Windows Arm64 SDK should treat pre .NET 5 versions as X64 by default.
+- Parity syntax between `dotnet build` and `dotnet publish` for pivoting between self-contained and framework-dependent modalities, for example:
+   - `dotnet build -r win-x64 --self-contained`
+   - `dotnet build -r win-x64 --self-contained true`
+   - `dotnet build -r win-x64 --self-contained false`
+   - `dotnet build -r win-x64 --no-self-contained`
+- Add warning when using `-r` without a `--self-contained` or `--no-self-contained`. The warning is for .NET 6+ apps only.
+
+The addition of the shorthand RID syntax and the parity syntax between `build` and `publish` will provide a satisfactory set of gestures to enable migration with .NET 6 to a non-breaking syntax with respect to .NET 7. In particular, users need to migrate any uses or `-r` to always be accompanied by one of the `--self-contained` or `--no-self-contained` switches.
```
marcpopMSFT, line 272:
```suggestion
The addition of the shorthand RID syntax and the parity syntax between `build` and `publish` will provide a satisfactory set of gestures to enable migration with .NET 6 to a non-breaking syntax with respect to .NET 7. In particular, users need to migrate any uses of `-r` to always be accompanied by one of the `--self-contained` or `--no-self-contained` switches.
```

marcpopMSFT, line 272: To confirm, if we add the shorthand to just build and publish, that would be sufficient for .net 6?  Sarah will start on this next week and I wanted to ensure we had the .net 6 part down.

richlander: We need the shorthand syntax for all the verbs listed in the .NET 6 section. If not, we don't have a story for `test`, `run` and other experiences.

richlander: Put another way, the spec is clear on ordering, including on that point. I'm happy to review any of those sections in more detail.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,496 @@
...
+Portable framework-dependent apps are the default scenario today. As already stated, it's not a good default choice because it produces larger apps with niche benefits.
+
+### Enable gesture to build portable apps
+
+If rid-specific apps become the default, then we need a new gesture for portable apps. There are two (non-mutually-exclusive) models for that:
+
+- Use the existing `any` RID to mean portable.
+- Expose a new CLI arg and MSBuild property for portable apps, like  `--portable` and `<Portable>true</Portable>`.
+
+Recommendation: Start with using just the RID. Wait for feedback to enable more polished experiences to validate that portable is an important scenario.
```
tannergooding, line 112: I believe we need the MSBuild option from day 1 or some alternative.

`dotnet build --runtime ... MyRepo.sln` fails today with `error NETSDK1134: Building a solution with a specific RuntimeIdentifier is not supported. If you would like to publish for a single RID, specifiy the RID at the individual project level instead.` and so you cannot actually decide to build an entire solution for a given RID. Doing this per project is slow and unmaintainable for larger projects.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,496 @@
...
+
+There have been some issues filed (see appendix) that suggest that managed libraries are not resilient to RID-specific libraries changing location as a result of changing build type (portable to RID-specific). This would also be true for single file deployment builds.
+
+We could enable (as an option) creating `runtimes` folders for RID-specific apps to make the folder layout the same across deployment types. Doing that would put single file at a significant disadvantage. We'd rather encourage the ecosystem to support both folder layouts. If we had to choose one of these deployments types to disadvantage, it would be portable (with the `runtimes` folders), not RID-specific, which is the likely the opposite of what developers are tuning for today.
+
+We need to invest more in docs and tools to the end of clear and actionable guidance.
+
+### Use SDK RID as default RID
+
+RID-specific apps should default to the SDK RID as an implicit RID, and no longer require a RID to be specified. The SDK already does this for the apphost for portable apps. This proposal is a straightforward extension of that behavior.
```
tannergooding, line 134: Just as an FYI, NuGet has some issues today with RID specific assets: https://github.com/NuGet/Home/issues/10571

Ideally these would get fixed as part of this to ensure that pulling in native dependencies "just works".

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,496 @@
...
+- `dotnet publish --self-contained`
+- `dotnet publish /p:PublishSingleFile=true`
+
+Note: As already suggested, this change would make Dockerfiles MUCH easier to write. Needing to specify RIDs is probably the worst part of using .NET with Docker.
+
+Specifying a RID should have no effect on the build type/deployment model. It should simply refine the existing model. This change largely follows from using the SDK RID as the implicit default for all types of RID-specific apps. When a RID is specified, it is substituted for the implicit RID.
+
+This means that the following commands would produce a RID-specific framework-dependent app:
+
+- `dotnet build -r linux-x64`
```
tannergooding, line 147: How much thought has been put forward on `linux-x64`? Unlike Windows, there can be a lot of differences between the various distros and writing a portable Linux app can be difficult.

In some cases, you might have different native dependencies for Ubuntu vs RHEL vs SLES vs etc (and sometimes even different dependencies between each of these due to differences in what ships in the system by default).

accepted/2021/architecture-targeting.md
richlander: Interesting. I didn't know that. I didn't include `.sln` files in my experiments/research. Fail on that.

I take it that you'd be fine with `-r any` as long as it is respected on a `.sln` file. Yes?

accepted/2021/architecture-targeting.md
richlander: Yup. We need a followon spec that improves building and consuming RID-specific packages. So many problems. @terrajobst has tried to resolve this in the past. I'd like us to take another run at it for .NET 7.

accepted/2021/architecture-targeting.md
richlander: Very little thought. I'd heard very few requests for build-time differentiation. We are forced to do it for Alpine (musl libc), but that's logically separate from what you are getting at. Certainly, we could extend the RID graph for distros. We need compelling scenarios.

accepted/2021/architecture-targeting.md
richlander: https://github.com/dotnet/designs/pull/233/commits/9a9528228faabd954e353b0efbd8269501194fe7

tannergooding:
One of the concerns I have is that not everyone has the resources that Microsoft (or "large enterprise") has. Successfully publishing and deploying for every architecture can be a pain and may require additional hardware or resources that are not easily available or commonplace. Not everyone, in fact I'd say most non enterprises, can afford both a Windows/Linux [dual boot] and an Apple machine, let alone an Android + iPhone nor to mention x86 vs x64 vs Arm32 vs Arm64 vs <other architectures> (and not getting into chipset specific optimizations that frameworks like the BCL have, which can greatly impact overall perf). Likewise, when you start looking at native dependencies (particularly on Linux) you can quickly explode into supporting 10+ operating system/architecture combinations.

and so, IMO, it would be nice if we had the option of managing what is portable and what is not at a more granular level and if certain "post processing" (e.g. crossgen and linking/triming) tools were made available for use as "separate" steps/tools. That is, I should be able to create a "portable" windows app that has all my IL. I should then be able to separate that "portable" code from any non-portable bits (crossgen'd or trimmed/linked binaries, native executable hosts, native dependencies, etc). When deploying, I should be able to easily take those bits and produce something that is "xcopyable" (that is, you can put it on a USB drive and move it between two Windows machines and have it still "work"), that is mostly xcopyable minus selected dependencies (e.g. I can make it framework dependent or requiring some native dependency to be pre-installed by the user), or I can package it up in an installer and leave steps like crossgen or trimming/linking to install time. This allows me to optimize the app more and can get even more benefits when you know the exact machine you are targeting. This is fine because installers are there own apps that often setup things like config files or registry keys and are typically not expected to be copyable elsewhere

This would give users the option to target and deploy in the model that works best for them and their scale. If perf isn't a huge concern for my small CLI tool, I can produce something that is framework dependent and where the only non-portable bits are the native hosts that make it easy to launch the tool (that is, xcopyable zips are identical minus the single PE/Elf executable). While if I'm doing a game or an app where optimizations can really make or break things, I can either do the full end to end on my CI machines and produce pre-crossgen'd, linked, and trimmed binaries for each platform or I can ship an installer that gives users the option to spend some extra install time to get faster startup times and that last 15fps.

accepted/2021/architecture-targeting.md
tannergooding: > I take it that you'd be fine with -r any as long as it is respected on a .sln file. Yes?

Yes, provided you can build the entire solution as desired with a single command, I think that "fits the bill".

However, it's generally nice to be able to do just `dotnet build` or just `dotnet run` and have things work as expected and not have to pass command line args for it to be "correct" (assuming the project wants portable by default).

jkotas:
> Likewise, when you start looking at native dependencies (particularly on Linux) you can quickly explode into supporting 10+ operating system/architecture combinations.

Agree. It has been a problem for arm64 and wasm, both of which we want to be growing. I believe that adding [source distribution](https://github.com/NuGet/Home/issues/9631) as fallback option for native dependencies is the only scallable way to address this.

> I can package it up in an installer and leave steps like crossgen or trimming/linking to install time. 

It would an option for CLI tools or desktop app installers only, and it is getting increasingly problematic even for those due to security issues with installers producing unsigned native binaries.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+
+Today, we talk about self-contained and framework-dependent as the two [deployment types](https://docs.microsoft.com/dotnet/core/deploying/). This model is misleading for two different reasons: the set isn't complete, and it's oriented exclusively on deployment.
+
+The following are the full set of "build types" (I'm changing the term) that can be used to greater or lesser degrees across development, testing, and production scenarios.
+
+* Portable framework-dependent app
+* RID-specific framework-dependent app
+* RID-specific self-contained app
+
+Note: The first build type is the default today.
```
marek-safar, line 55: I don't think this is true (e.g. Client Blazor) and in .NET7 e.g. MAUI

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+* `win10-x64`
+* `win10-x86`
+* `win7-x64`
+* `win7-x86`
+
+See the appendix for an analysis of portable app size metrics.
+
+Portable apps are easier to deploy to production for PaaS scenarios, like Azure Websites. As soon as developers adopt containers, it makes more sense to adopt our [documented RID-specific container patterns](https://github.com/dotnet/dotnet-docker/tree/main/samples/aspnetapp), which results in optimized apps.
+
+The reason to keep portable apps as the default is that it is easier for non-container web developers that target a different operating system than the one they use locally for development. Portable apps are significantly worse for all other scenarios, particularly client.
```
marek-safar, line 79: > The reason to keep portable apps as the default is that it is easier for non-container web developers that target a different operating system than the one they use locally for development.

Do we have any evidence that this workflow is used by developers?

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+
+Some people may argue that portable apps benefit new developers since they are simpler. That's potentially true, if they are deploying to a PaaS service that runs another operating system or uses a different CPU type. On the other hand, portable apps have the worst metrics (they can be significantly larger) such that their utility for the new developer experience is questionable. They are not our "best foot forward".
+
+## Proposal for behavior changes
+
+There are multiple proposed changes.
+
+### Enable RID-specific as default build type
+
+The default build type should be changed to the RID-specific framework-dependent model. This approach has the following benefits:
```
marek-safar, line 91: Is the intention to change every existing template? In other words will unit test project not produce RID specific framework model?

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+
+There have been some issues filed (see appendix) that suggest that managed libraries are not resilient to RID-specific libraries changing location as a result of changing build type (portable to RID-specific). This would also be true for single file deployment builds.
+
+We could enable (as an option) creating `runtimes` folders for RID-specific apps to make the folder layout the same across deployment types. Doing that would put single file at a significant disadvantage. We'd rather encourage the ecosystem to support both folder layouts. If we had to choose one of these deployments types to disadvantage, it would be portable (with the `runtimes` folders), not RID-specific, which is the likely the opposite of what developers are tuning for today.
+
+We need to invest more in docs and tools to the end of clear and actionable guidance.
+
+### Use SDK RID as default RID
+
+RID-specific apps should default to the SDK RID as an implicit RID, and no longer require a RID to be specified. The SDK already does this for the apphost for portable apps. This proposal is a straightforward extension of that behavior.
```
marek-safar, line 134: How is this going to work on configurations where SDK is built for the more generic RID than I'm on (e.g. some Linux flavours) ?

tannergooding:
> It would an option for CLI tools or desktop app installers only, and it is getting increasingly problematic even for those due to security issues with installers producing unsigned native binaries.

@jkotas, I don't disagree here. However, I do feel there is a large difference in expectations and potential acquisition avenues between things like Android/iOS or xbox/playstation which provide official "app stores" and sometimes their own recommendations/tools/procedures for optimizing per device and which offset or completely cover the costs of CDN, global distribution, caching, etc.

With standard desktop installs users are often fronting the costs themselves or have some "free/low tier" limitation on bandwidth/CDN to help keep costs low. Even NuGet has a limit (of like 250MB) per package and even compressed, there are apps that can hit that relatively quickly (particularly with native dependencies).

Some of this is improving with things like the Microsoft Store or winget, but its still not perfect and many users still prefer to use things like MSIs or self extracting installers instead and I feel that ensuring the options are available (and maybe even providing guidance and tooling to help make "correct" and "safe" installers that do this viable) might be something worth looking into.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+   - `true` (for all TFMs)
+   - Note: Enables changes #2, #3 and #4 for all versions.
+
+Note: These config names are not great. They are placeholders for better names.
+
+### .NET 8
+
+The following changes should be included in an early .NET 8 preview.
+
+- Remove support for the three config options added in .NET 7, just just above.
```
rolfbjarne, line 313:
```suggestion
- Remove support for the three config options added in .NET 7, just above.
```

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+- Do NOT enable the `-r` RID syntax on these same verbs. Doing so would cause a breaking change in .NET 7, which seems like a unforgivable mistake.
+- Enable implicit RID where there is one known choice:
+   - The macOS Arm64 SDK should treat pre .NET 6 versions as X64 by default.
+   - The Windows Arm64 SDK should treat pre .NET 5 versions as X64 by default.
+- Parity syntax between `dotnet build` and `dotnet publish` for pivoting between self-contained and framework-dependent modalities, for example:
+   - `dotnet build -r win-x64 --self-contained`
+   - `dotnet build -r win-x64 --self-contained true`
+   - `dotnet build -r win-x64 --self-contained false`
+   - `dotnet build -r win-x64 --no-self-contained`
+- Add warning when using `-r` without a `--self-contained` or `--no-self-contained`. The warning is for .NET 6+ apps only.
```
rolfbjarne, line 270: For platforms that only support self-contained apps (or vice versa if that exists), it seems redundant to have to specify it.

accepted/2021/architecture-targeting.md
akoeplinger: One scenario that comes to my mind is dealing with openssl since there can be different incompatible versions/variants between distros and versions of the same distro.

accepted/2021/architecture-targeting.md
zivkan: Please also consider how NuGet lock files should work. Today, the following commands result in an error: https://github.com/NuGet/Home/issues/9195

```text
dotnet new console
dotnet restore --use-lock-file
dotnet restore --locked-mode -r linux-x64
```

While that looks like a contrived example (why would someone restore locked mode for a specific RID?), consider that customers want to use locked mode on CI build pipelines, and that `dotnet publish -r linux-x64` does an implicit restore. Bang, restore failed, so publishing failed, and CI failed.

Customers want to use lock files and locked mode so they can use floating versions, but still get consistent restores across all machines, and also some package substitution prevention, as lock files ensure the package's content hash is the same when packages are extracted.

There needs to be a way for the SDK to ensure that the set of packages NuGet is asked to restore (plus the set of RIDs that NuGet is informed about) is consistent, regardless of what RID is being built/published by the comment line arguments.

This spec seems to be a higher level than these details, but lock files need to be designed to be workable from the top. NuGet can't have a workable lock file experience when the SDK reports a different set of packages on different actions.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+- `dotnet publish /p:PublishSingleFile=true`
+
+Note: As already suggested, this change would make Dockerfiles MUCH easier to write. Needing to specify RIDs is probably the worst part of using .NET with Docker.
+
+Specifying a RID should have no effect on the build type/deployment model. It should simply refine the existing model. This change largely follows from using the SDK RID as the implicit default for all types of RID-specific apps. When a RID is specified, it is substituted for the implicit RID.
+
+This means that the following commands would produce a RID-specific framework-dependent app:
+
+- `dotnet build -r linux-x64`
+- `dotnet publish -r linux-x64 /p:PublishReadyToRun=true`
```
zivkan, line 148: Please also consider how NuGet lock files should work. Today, the following commands result in an error: https://github.com/NuGet/Home/issues/9195

```text
dotnet new console
dotnet restore --use-lock-file
dotnet restore --locked-mode -r linux-x64
```

While that looks like a contrived example (why would someone restore locked mode for a specific RID?), consider that customers want to use locked mode on CI build pipelines (via Directory.Build.props or environment variable), and that `dotnet publish -r linux-x64` does an implicit restore. Bang, restore failed, so publishing failed, and CI failed.

Customers want to use lock files and locked mode so they can use floating versions, but still get consistent restores across all machines, and also some package substitution prevention, as lock files ensure the package's content hash is the same when packages are extracted.

There needs to be a way for the SDK to ensure that the set of packages NuGet is asked to restore (plus the set of RIDs that NuGet is informed about) is consistent, regardless of what RID is being built/published by the comment line arguments.

This spec seems to be a higher level than these details, but lock files need to be designed to be workable from the top. NuGet can't have a workable lock file experience when the SDK reports a different set of packages on different actions.

accepted/2021/architecture-targeting.md
richlander: We already have that problem solved. Put another way, I don't think enable `linux` as any any-style RID would change our approach to OpenSSL. Same with ICU. Perhaps I'm missing something.

accepted/2021/architecture-targeting.md
richlander: Totally true. My thinking is that these app types already have bespoke build experiences so many of these default experiences don't apply. Fair?

accepted/2021/architecture-targeting.md
richlander: I don't, but I expect it happens a lot. For example, anyone who publishes directly from Visual Studio to a cloud may benefit from this. That said, I'm not arguing this point just conceding that this could be a legit reason.

accepted/2021/architecture-targeting.md
richlander: This isn't a template change. I don't think unit tests care one way or the other. Unless I'm missing something, I'd expect them to be the least impacted.

accepted/2021/architecture-targeting.md
richlander: For the most part, we don't offer RIDs for different Linux distros (AFAIK). The key outlier is Alpine. Outside of Alpine, I don't see that mattering. Today, the product only cares about glibc and musl c.

accepted/2021/architecture-targeting.md
richlander: It does. Someone has to win ;)

More seriously, `-r` is an equally legit option for framework dependent apps and self-contained ones. IMO, it is incoherent to change the deployment mode as a result of specifying a RID.

accepted/2021/architecture-targeting.md
richlander: This is a good issue. I hadn't thought about that. Fortunately, this change doesn't come until .NET 7 so we'll have time properly design this experience. I agree with you.

richlander:
@tannergooding -- the CLI/msbuild offers significant options for application portability today. None of those are being taken away. I'm only moving them to non-default. In fact, making portable non-default may enable us to reason about it better. That said, I think there is a small set of apps where it makes sense. I talked to multiple people on the team that don't understand how portable works, in part, because it is the default.

I see a spectrum of choices:

1. x64 only and fully optimized (relying on x64 emulation on Arm64).
1. portable with architecture-specific launchers (which would be a new feature) and not optimized (no R2R).
1. Same as above but with architecture-specific composite R2R.

The middle option is the sweet spot, IMO. The first one will likely have noticeable perf challenges due to emulation. The last may suffer from size challenges. R2R is only meaningful for large apps and that's where size is likely to show up.

Note: I'm thinking desktop client apps. If it's tools, then we already have a good enough story, IMO. You'll need to convince me otherwise on that front.

This scenario really only applies to Windows. I'm waiting for Arm64 adoption to demonstrate itself before spending any time thinking about this. We're doing our part to support Windows Arm64 and are not holding anyone back (IMO) with the lack of this experience. This scenario is very much a "++" experience that requires more justification before investing.

accepted/2021/architecture-targeting.md
jkotas: We have solved it for the .NET Runtime dependencies on icu and openssl by using non-standard mechanisms to bind to these libraries. The solution was non-trivial to setup and requires constant maintenance. (In theory, we could have used similar non-standard mechanism to bridge the glibc/musl differences too, but that was too complex even for our taste, so we choose to let glibc/musl bleed out in the UX.)

We have not solved it for 3rd party packages with native libraries that depend on openssl, icu and other native libraries. For example, the dependency on libtinfo was giving me a hard time in one of my projects. I think it is rare for authors of 3rd party packages with native dependencies to go through the pains of targeting generic linux RID. The authors of 3rd party packages with native dependencies typically end up picking up a set of Linux distros, configurations and architectures to build and package the native dependencies for. If your project target platform is lucky to be on this list, great! If it is not, you are on a unhappy path.

The only scalable way to solve this is to add fallback to build-from-source for the native dependencies (https://github.com/dotnet/designs/pull/233#issuecomment-877676493). Fallback to build-from-source would work well for current-RID-specific publish, so this proposal is good for it. It would not work well (without additional manual configuration) for RID-neutral publish.

AraHaan:
Also with this being a thing could it also be valid to allow AnyCPU self-contained apps where if they do not set an RID then bundle in the x86, x64, and arm64 runtimes into it and have it's loader a select which runtime to use based on:

- is the cpu x86? (or i386)
- is the cpu x64?
- is the cpu arm64?

I would love this so much so that way I do not have to worry then, plus if my application later pulls in some C++ bits then I could *consider* that part of the runtime and as such separate them like that as well without issues.

richlander:
@AraHaan ... I like the idea and it's possible to achieve but it is unclear how. The executables all need to be RID specific. That's a requirement of the operating system. If you've got ideas on how to get around this, can you share the file layout that you'd expect?

AraHaan:
I would figure the layout of the runtime parts would be like so:

- root of the self contained application
- runtimes
   - x64
      - runtime dll's here (including user made C++ dll's, this also includes dll's made using C++/CLI).
   - x86
      - runtime dll's here (including user made C++ dll's, this also includes dll's made using C++/CLI).
   - arm64
      - runtime dll's here (including user made C++ dll's, this also includes dll's made using C++/CLI).
- Application's files (All Managed without any C++ targeting AnyCPU)

However I think the Application Host exe part would he a snag however.

But it *could* be possible to somehow combine multiple cpu targets into that single binary then have it somehow select the right one by checking the CPU's features using some asm code (if it's arm64, x64, or x86)?

As for loading C++/CLI dll's the dll loader of the runtime could *check* first in the root of the self contained application, then if not found load from the directory of the chosen runtime used to run it.

richlander:
Yes, the exes are the key thing. This article might help better explain the problem. https://news.ycombinator.com/item?id=24256883

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+* Portable apps will be bigger if the app depends on packages with architecture-specific dependencies, since the app will (likely) contain the native assets for multiple architectures (that's one of three key aspects of what makes it portable). The only reason to build portable apps is to accommodate multi-architecture packages.
+* Portable apps and RID-specific builds [do not place native assets in the same location](https://github.com/dotnet/sdk/issues/11754). That should be configurable and is a [confusing behavior](https://github.com/dotnet/sdk/issues/3903).
+* It is possible to make an app portable for one operating system, such as for Windows x86, x64, and Arm64, but that isn't enabled.
+* Portable apps -- for the apphost -- don't support the universal binaries feature of Windows or macOS, which reduces their utility, particularly for client apps.
+* Some users want to target a specific architecture and/or operating system. The SDK produces a self-contained app by default when you do this. Many users (likely most) do not want that.
```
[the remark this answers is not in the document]
richlander: Totally agree. I think most people don't (A) realize that this model even exists, (B) don't realize why they would use it, and (C) don't understand the syntax. My take is that RID specific fx-dependent apps are the best default for the product. That's much of the thesis of this proposal.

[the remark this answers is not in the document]
eerhardt: I wonder if we can get some data around which deployments are used more often.

I would hope that the more used scenarios would be easier for the user - less settings to pass to the command line / set in the .csproj.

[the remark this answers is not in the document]
tannergooding: I think getting deployments for real world desktop apps is also important.

What you want for a simple command line tool and what you want for a game or a more advanced app might be very different. I'm not sure how well this model fits for Paint.NET for example where, from what I've seen from Rick brew in the C# Community Discord, it seems to be a bit of a different setup and deployment model.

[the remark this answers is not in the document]
richlander: It would be very hard to reason about usage patterns/data and I don't think it was lead us to the right decision, in part due to what @tannergooding raises.

The current system doesn't align well with user needs. I don't think most users want portable by default and the focus on `-r` is not good UX. Instead, users can focus on `--self-contained` and specialize it with a architecture/OS info as an option.  I don't see data pointing us in that direction given how challenging the current UX is.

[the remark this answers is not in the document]
AraHaan: Also sometimes you want self contained without only AnyCPU parts of the BCL and not the complete runtime that is cpu specific too (and so you would require them to need to have the cpu specific runtime bits either installed or in separate folders within an ``runtime`` subfolder within the self contained application where it would *somehow* select the right bit of runtime to load the thing).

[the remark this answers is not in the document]
eerhardt: > I don't see data pointing us in that direction given how challenging the current UX is.

Data isn't going to point us to the right UX. But data is going to point us to "what do users want to do more often?", which will help us create the right UX.

Making users specify 3 arguments instead of 1 in a hardly used scenario isn't that big of a deal. But making users specify 3 arguments instead of 1 in an often used scenario is a big deal IMO.

`dotnet publish -r win-x64`

vs.

`dotnet publish --self-contained --os win -a x64`

Which would you rather type?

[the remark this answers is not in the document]
richlander: Sure, but that's cherry-picking the proposal.

The new proposal does the following, across .NET 6 and 7:

- Improves the default (portable fx-dependent to rid-specific fx-dependent).
- Improves self-contained build UX when the target RID and the SDK RID match.
- Regresses self-contained build UX when the target RID and SDK RID do not match.

Do you agree with that assessment?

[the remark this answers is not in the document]
eerhardt: I agree that the proposal does that. And I definitely agree I want the first two.

But the proposal also does:

* When the target RID and SDK RID do not match, prioritize fx-dependent apps over self-contained 

That is the part of the proposal that I don't want for myself as a user. And so that's where my ask for data/evidence is coming from.

In my experience, when I publish for a RID that isn't the SDK RID, every time I want self-contained. I honestly can't think of a time when I've wanted a fx-dependent app. So I want to verify that most users do want that behavior.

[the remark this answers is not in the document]
tannergooding: In practice, I'm not typing either. I'm coding this up in a CI script so I don't need to remember what to type everytime and relying on F5 debugging for local testing.
If I need to test the "shipping product" I'll submit a PR to my repo and have CI spit out some artifacts for me to try out (or I can simulate that experience locally with `.\scripts\cibuild.cmd` for my own repos and that takes care of all the options)

I'm not a big enterprise and as an OSS project likely don't have the ability to cover all the CDN and deployment costs for all the self-contained binaries. So for me, I'd ideally be able to produce small portable deployments that the installer can then finish doing things like AOT or R2R on at install time so that startup is fast and its taking advantage of all the cool features .NET does have available. This helps with any costs I need to pay, fits well within the limitations of most free/low tier services (e.g. NuGet package size or GitHub artifact size limitations), and avoids needing to maintain or explicitly support deployments for every OS (windows, linux, macos, android, ios, etc), every distro (win7, win10, win11, ubuntu18.04, ubuntu20.04, rhel, suse, freebsd, etc), and every architecture (x86, x64, Arm32, Arm64, etc) or needing to consider what the "average" hardware a target user has for cases where myself or a dependency use hardware acceleration (SSE4.1, AVX, AVX2 on x86/x64 or AdvSIMD and friends on A64, or WebSIMD for WASM in the future, etc).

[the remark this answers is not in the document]
tannergooding: -- For reference, the above is also not be speaking abstractly. I maintain several OSS libraries and tools independently of work and these are the kinds of problems I actively run into already.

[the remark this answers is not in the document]
richlander: > I'd ideally be able to produce small portable deployments that the I can then finish doing things like AOT or R2R on at install time or first launch so that startup is fast

This is separate from the overall conversation, but we don't intend to build those experiences. We're focused on dev-time crossgen and awesome JIT features w/nothing in-between. Today, crossgen has a couple characteristics that we really like:

- The build produces final optimized artifacts, per OS/arch.
- Those artifacts can be signed.

The NGEN experience had neither of those characteristics and had a ton of UX problems. We're not going to recreate those challenges.

> I honestly can't think of a time when I've wanted a fx-dependent app. So I want to verify that most users do want that behavior.

The thinking behind this change is super simple (as I articulate in the doc):

- fx dependent is the default build type, since .NET Core 1.0.
- rid-specific is relevant to fx-dependent apps, using an implicit RID, and this will be the default in .NET 7.
- specifying an explicit rid shouldn't change the deployment type.

I respect and sympathize with the point you are raising. It's legit. My view is coming from a very different direction, as I've articulated. I am a fan of self-contained. Judging by our container image pulls (10s of millions/month), self-contained is not the primary developer scenario.

I think the new model is generally better. It's not better for all cases but that's typical of change.

Also, for the x64 emulation scenario, rid-specific targeting absolutely benefits from self-contained not being the default.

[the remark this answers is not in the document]
tannergooding: > This is separate from the overall conversation, but we don't intend to build those experiences.

I think that's unfortunate and is overlooking the independent developer scenario. In particular certain things like signing simply aren't a consideration for many independent developers. Authenticode certs tend to start at $500-600/yr and that's out of reach for a lot of projects (assuming they aren't in the .NET Foundation or similar). Having the ability to easily perform steps at any stage, including post-install, is general goodness IMO and allows developers to setup their CI flow however works best for them, even if they are signing, etc.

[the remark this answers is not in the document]
richlander: Sure. I'd rather us focus on dynamic PGO than a post-install NGEN replacement. BTW: I don't think that NGEN was very popular either, particularly among the audience you are describing.

accepted/2021/architecture-targeting.md
```diff
@@ -0,0 +1,498 @@
...
+Note: As already suggested, this change would make Dockerfiles MUCH easier to write. Needing to specify RIDs is probably the worst part of using .NET with Docker.
+
+Specifying a RID should have no effect on the build type/deployment model. It should simply refine the existing model. This change largely follows from using the SDK RID as the implicit default for all types of RID-specific apps. When a RID is specified, it is substituted for the implicit RID.
+
+This means that the following commands would produce a RID-specific framework-dependent app:
+
+- `dotnet build -r linux-x64`
+- `dotnet publish -r linux-x64 /p:PublishReadyToRun=true`
+
+Developer would add `--self-contained` or `--self-contained true` to create a self-contained app, like with the following command.
```
eerhardt, line 150: Since `--self-contained` is kind of a lot to type. What do you think about adding a short option `-s` for this switch?

`dotnet publish -r linux-x64 -s`

accepted/2021/architecture-targeting.md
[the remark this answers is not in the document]
tannergooding: I think that might have been for other reasons and that dynamic PGO isn't really a viable alternative for handling fast startup, its more about tuning the running application to achieve higher end throughput.

I think its worth considering what would be required to make our core build tools, including `crossgen`, the `linker`/`trimmer`, and others usable either standalone or as part of the SDK. I imagine its not much and that its mostly just packaging them as `dotnet tools`. This would cover the normal "publish" scenario, running them as post-build or independent CI steps, or even integration into custom build pipelines (which frameworks like Unity have).

accepted/2021/architecture-targeting.md
richlander: Yup. Fan of that (or something like it).

If/when we do that, I'd suggest that we enable an implicit RID by default. That would enable this scenario:

```console
PS C:\Users\rich\app> dotnet publish --self-contained
Microsoft (R) Build Engine version 16.8.3+39993bd9d for .NET
Copyright (C) Microsoft Corporation. All rights reserved.

  Determining projects to restore...
  Restored C:\Users\rich\app\app.csproj (in 61 ms).
C:\Program Files\dotnet\sdk\5.0.103\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.RuntimeIdentifierInference.targets(136,5): error NETSDK1031: It is not supported to build or publish a self-contained application without specifying a RuntimeIdentifier. You must either specify a RuntimeIdentifier or set SelfContained to false. [C:\Users\rich\app\app.csproj]
```

AraHaan: I would have the RID default to the host OS's platform and architecture.

richlander: Filed an issue on this. https://github.com/dotnet/sdk/issues/19576

eerhardt: Perfect! Thanks.

PR #234, opened 2021-07-15
Make `System.Drawing.Common` only supported on Windows

terrajobst:


safern:
FYI: @ericstj @tarekgh @danmoseley @marek-safar @akoeplinger

chucker:
I find “Common” too confusing, then.

Lots of stuff currently depends on it and expects cross-platform. IMHO, you should instead deprecate the package and consider a clearer name.

terrajobst:
Renaming a package/assembly is a breaking change. The marginal gain of a clearer name is far outperformed with the impact on the ecosystem IMHO.

Also, the platform compatibilty analyzer will inform you at build time that this component is Windows-only, so it's not like this would be a landmine for cross-platform code.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,183 @@
...
+already had a cross-platform implementation, we decided to bring that over. The
+cross-platform functionality is provided via the native library `libgdiplus`.
+Sadly, `libgdiplus` is a community supported library and at this point
+effectively unmaintained.
+
+While `libgdiplus` has helped a lot of customers expand to cross-platform it has
+also caused some headaches. `System.Drawing.Common` is very Windows driven as it
+is mostly a wrapper around GDI+ and is heavily coupled to Windows Forms, which
+keeps evolving and diverging from cross-platform graphics. This has caused some
+confusion over why some things don't work well on Unix or don't work not at all.
```
jkotas, line 18:
```suggestion
confusion over why some things don't work well on Unix or don't work at all.
```

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,183 @@
...
+* `libgdiplus` doesn't ship with all operating systems by default, just like
+  other native libraries we depend on (icu, openssl).
+
+* `libgdiplus` is community driven, so a lot of issues/PRs just go stale and are
+  never looked at.
+
+* There is no servicing story for bugs and security issues.
+
+However, making `libgdiplus` a first-class citizen is costly. It's around 30k
+lines of pretty old C code, virtually untestable, and lacks a lot of
```
jkotas, line 36:
```suggestion
lines of pretty old C code, virtually untested, and lacks a lot of
```
all code is testable, with enough effort...

atrauzzi:
The rationale behind the deprecation is sound, but the action taken -- well, I'll just say that the optics are suspect.

(If for some reason this same issue was affecting Windows, I'd put some money on Microsoft being more willing to invest effort to shore things up.)

Is this going to end up being something that people expect to have available, but are blindsided because it touches some parts of the MS ecosystem which are viewed as conduits to Windows adoption?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,183 @@
...
+on non-Windows platforms and instead propose we only keep evolving it in the
+context of Windows Forms and GDI+.
+
+## Requirements
+
+### Goals
+
+* Consumers of `System.Drawing.Common` will be informed that starting with .NET
+  6, it's only supported on Windows, via the platform compatibility analyzer
+* Cross-platform consumers have the ability to turn on a compatibility switch to
```
stephentoub, line 68: If we're recommending SkiaSharp and ImageSharp instead, can we create some kind of guidance that provides migration samples for the common use cases we see?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,183 @@
...
+>   `libgdiplus` separately. For more information, see Install .NET on Linux or
+>   Install .NET on macOS.
+>
+>   If you can't use System.Drawing with your application, recommended
+>   alternatives include ImageSharp, SkiaSharp, and Windows Imaging Components.
+
+However, this disclaimer doesn't feel clear enough to steer people towards using
+other graphics libraries.
+
+The proposal is to do the following for .NET 6:
```
jkotas, line 102: What is going to be the plan for .NET 7+? Are we going to keep the libgdiplus in increasingly more broken state forever?

Can we be more aggressive in .NET 6 to be done with this? Make it throw unconditionally, and tell people who need the Unix support to use the older version of the package.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,183 @@
...
+The challenge is that `System.Drawing.Common` is used by 1st party consumers
+such as ML.NET, which uses it for image manipulation and even exposes an API
+that returns a Bitmap. At the same time, `System.Devices.Iot` is in the process
+of moving to use ImageSharp and so far they had a great migration experience. We
+need to engage with 1st party customers to see what impact this proposal will
+have on them.
+
+***Note:** This only affect `System.Drawing.Common` and would leave
+`System.Drawing.Primitives` as-is. That assembly on contains primitive types
+that don't depend on GDI+, such `Rectangle`, `Point`, and `Size`.*
```
stephentoub, line 128: Is there anything else in System.Drawing.Common that doesn't depend on libgdiplus and we should push down?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,183 @@
...
+3. Add an AppContext switch customers can set in their apps or at runtime to
+   re-enable `System.Drawing.Common` via `libgdiplus` on non-Windows platforms.
+
+4. Add targets to the package that when it is restored using a non-windows RID,
+   it warns that we no longer support it on non-windows and point to the aka.ms
+   link with more info.
+
+The challenge is that `System.Drawing.Common` is used by 1st party consumers
+such as ML.NET, which uses it for image manipulation and even exposes an API
+that returns a Bitmap. At the same time, `System.Devices.Iot` is in the process
```
jkotas, line 121:
```suggestion
that returns a Bitmap. At the same time, https://github.com/dotnet/iot is in the process
```
`System.Devices.Iot` is not the right name for dotnet/iot repo

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
jkotas, line 128:
```suggestion
that don't depend on GDI+, such `Rectangle`, `Point`, and `Size`.
```

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,183 @@
...
+is mostly a wrapper around GDI+ and is heavily coupled to Windows Forms, which
+keeps evolving and diverging from cross-platform graphics. This has caused some
+confusion over why some things don't work well on Unix or don't work not at all.
+
+Since the inclusion of the Mono cross-platform implementation we spent lot of
+time redirecting issues to `libgdiplus` that never got fixed, or helping people
+install `libgdiplus` correctly (because it's not part of .NET Core proper). We
+have thought about our options, and to support `System.Drawing.Common` properly
+on Unix we'd need to make `libgdiplus` a first-class citizen and ship it as part
+of the product:
```
richlander, line 25: I think the first two issues are red-herrings. For example, the same things could be said about OpenSSL.

The difference with OpenSSL is that we're reacting to it, whereas we would need libgdiplus to react to us. Same comparison with ICU. That's the unworkable part. Also, the level of coupling is high.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
richlander: Agree. This is another red herring.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
eerhardt: > Can we be more aggressive in .NET 6 to be done with this? Make it throw unconditionally

I don't think we should do that until we at least migrate our own things. ML.NET suddenly throwing an exception when you upgrade to .NET 6 doesn't sound like a great experience to me.

safern: I support this, the library hasn't evolved for Unix in any release honestly, so using a previous version of the package for Unix support makes sense and there aren't any features to be lost.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
safern: Yes. We should put out a doc with guidance for the migration.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
safern: I haven't seen any, but still doing an analysis to see if we need to push types down.

Tornhoof:
Aren't you a little late in the .NET 6 release cycle for such a breaking change? Considering that you want to release in ~4 months and you're already in Preview 6 and afaik Feature Development is mostly done. I imagine quite a few people are already either testing their applications with the preview bits or planning to do so and that change might throw a wrench into their plans considering that .NET 5 will be out of maintenance ~3 months later (and Xmas is right in the middle of these 3 months).

Looking at the products, I had my sticky fingers in, in the last 12 months, I immediately recognize the QR libraries and atleast one PDF library (Aspose.PDF, I expect pretty much all the pdf libraries on the market to use System.Drawing.Common) and these will make migrating to .NET 6 a lot more complicated.

(FWIW: I personally wholeheartedly agree with this step, from an enterprise dev pov I think it's too late.)

safern:
> Looking at the products, I had my sticky fingers in, in the last 12 months, I immediately recognize the QR libraries and atleast one PDF library (Aspose.PDF, I expect pretty much all the pdf libraries on the market to use System.Drawing.Common) and these will make migrating to .NET 6 a lot more complicated.

While that is a valid concern,`System.Drawing.Common` is a standalone NuGet package which doesn't ship as part of the shared framework, so these libraries can definitely upgrade to .NET 6 and use the 5.0.0 version of the `System.Drawing.Common` package. 

`System.Drawing.Common` doesn't have anything .NET 6 special and the only thing that couples it with .NET 6 is the release cycle.

Tornhoof:
> While that is a valid concern,System.Drawing.Common is a standalone NuGet package which doesn't ship as part of the shared framework
May I recommend that you point that out in the Q/A section of the document? I was not immediately aware of that fact while reading the document.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: Sounds good.

@safern do you want to update the doc? You can just push to this branch.

safern: Yup.

hughbe:
As one of the most active contributors to libgdiplus, I hate to feel like I've wasted my time, but I do support this change. Libgdiplus doesn't have any maintainers, is riddled with incompatibilities, bugs, security flaws and missing features (e.g. metafiles, lots of brush features, regions).

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+
+The challenge is that `System.Drawing.Common` is used by 1st party consumers
+such as ML.NET, which uses it for image manipulation and even exposes an API
+that returns a Bitmap. At the same time, https://github.com/dotnet/iot is in the process
+of moving to use ImageSharp and so far they had a great migration experience. We
+need to engage with 1st party customers to see what impact this proposal will
+have on them.
+
+***Note:** This only affect `System.Drawing.Common` and would leave
+`System.Drawing.Primitives` as-is. That assembly on contains primitive types
```
jkotas, line 124:
```suggestion
`System.Drawing.Primitives` as-is. That assembly contains primitive types
```

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+that don't depend on GDI+, such `Rectangle`, `Point`, and `Size`. Any type in `System.Drawing.Common`
```
jkotas, line 125:
```suggestion
that don't depend on GDI+, such `Rectangle`, `Point`, and `Size`. Any types in `System.Drawing.Common`
```

MaximRouiller:
> > Looking at the products, I had my sticky fingers in, in the last 12 months, I immediately recognize the QR libraries and atleast one PDF library (Aspose.PDF, I expect pretty much all the pdf libraries on the market to use System.Drawing.Common) and these will make migrating to .NET 6 a lot more complicated.
> 
> While that is a valid concern,`System.Drawing.Common` is a standalone NuGet package which doesn't ship as part of the shared framework, so these libraries can definitely upgrade to .NET 6 and use the 5.0.0 version of the `System.Drawing.Common` package.
> 
> `System.Drawing.Common` doesn't have anything .NET 6 special and the only thing that couples it with .NET 6 is the release cycle.

Are we expecting the 5.0.0 to have ongoing support past .NET support timeline? Just saying because otherwise it means you're stopping support for non-Windows platforms.

That kind of breaking change is exactly the type of scenario in my opinion that deserves a namespace/assembly change. Especially if we want Linux/Mac support.

No?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+   link with more info.
+
+The challenge is that `System.Drawing.Common` is used by 1st party consumers
+such as ML.NET, which uses it for image manipulation and even exposes an API
+that returns a Bitmap. At the same time, https://github.com/dotnet/iot is in the process
+of moving to use ImageSharp and so far they had a great migration experience. We
+need to engage with 1st party customers to see what impact this proposal will
+have on them.
+
+***Note:** This only affect `System.Drawing.Common` and would leave
```
jkotas, line 123:
```suggestion
***Note:*** This only affect `System.Drawing.Common` and would leave
```
The rest of the doc is rendered in bold for some reason. Is this going to fix it?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+`System.Drawing.Primitives` as-is. That assembly contains primitive types
+that don't depend on GDI+, such `Rectangle`, `Point`, and `Size`. Any types in `System.Drawing.Common`
+that don't depend on GDI+, will be moved to `System.Drawing.Primitives`.
```
terrajobst, line 126:
```suggestion
that don't depend on GDI+, will be moved to `System.Drawing.Primitives`.*
```

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: There was a missing `*` at the end of the paragraph (in Markdown, `*` and `_` can span multiple lines). The intention is that this entire paragraph is italicized while the first word is bolded. This is how it renders here: 

image

jkotas: It still does not render correctly in the markdown with syntax highlighting view:

image

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: > I think the first two issues are red-herrings

It's a bit unclear what the two issues are you're referring to. Could you quote those?

The way I read the doc is that @safern meant to explain how the experience of using `libgdiplus` is very much subpar because it requires manual installation. If we wanted to make `System.Drawing.Common` work well, we'd need to invest to make this either much simpler or eliminate it entirely by bundling it. Neither are cheap.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: Yeah, the syntax highlighter uses some weird non-compliant Markdown interpretation 🤷

safern:
> Are we expecting the 5.0.0 to have ongoing support past .NET support timeline? Just saying because otherwise it means you're stopping support for non-Windows platforms.

Yes, that is true statement. However, IMHO, `System.Drawing.Common` was already in a weird state with regards to support for non-Windows platforms as it is really a thin wrapper around `libgdiplus` and most of the bugs are coming from the native library itself; so even with our support sometimes we went beyond and investigated/reported these issues and still, there was no resolution on those bugs or PRs were stale and the fix was never shipped because `libgdiplus` has no maintainers.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
safern: > Neither are cheap.

Nor worth the effort given `System.Drawing.Common` is pretty legacy and Winforms driven; and specially with the great managed libraries we have in the open source community already for xplat imaging/graphics.

terrajobst: I guess that was Rich's argument: the real argument is that `System.Drawing.Common` for cross-platform use is a square peg in a round hole, no the complexities/cost around us making the experience great.

safern: Any suggestions to change the text here to express Rich's point?

lukemcdo:
This is an awful change because there are a substantial number of packages that were brought to .NET Core which have non-primary functionality that depends on System.Drawing.Common. By making this breaking change, those packages will no longer work on Linux, despite that their functionality was intact. 

It would be much better to emit warnings when cross-platform code is reached that calls System.Drawing.Common. This informs the developer without forcing their entire project into a substantial rewrite, or, more likely, causing them to drop support for their users.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
eerhardt: Do we really have to put a poison pill into the package? Instead can't we just say it targets `windows`, and users will get warnings from the analyzer automatically when they try using it cross-plat. If they don't care, they can suppress the warnings. And we can make the statement that we don't support it on non-Windows, your mileage may vary. If it continues to work for you, great. But if you run into issues, we won't be fixing any bugs or new features going forward.

Basically, it is deprecated. Use at your own risk.

safern: While I understand the concern, how would that be different from just recommending to pin the package version to a specific version? The fact that you communicate that we recommend a package that is potentially out of support for both Windows and non-Windows?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+image processing and text rendering, such as `cairo`, `pango`, and other native
+libraries, which makes shipping even more challenging.
+
+We've also hit some memory leaks or double free'd memory that weren't fixed in
+`libgdiplus` and that we had to work around because they were potential security
+issues. This makes it harder as we are basically on our “own” for those security
+issues. `libgdiplus` is a library we don't own, but that we must spend time
+investigating issues and proposing fixes.
+
+We have noticed that `System.Drawing.Common` is used on cross-platform mostly for
```
danmoseley, line 47:
```suggestion
We have noticed (such as from analysis of NuGet packages) that `System.Drawing.Common` is used on cross-platform mostly for
```

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+confusion over why some things don't work well on Unix or don't work at all.
+
+Since the inclusion of the Mono cross-platform implementation we spent lot of
+time redirecting issues to `libgdiplus` that never got fixed, or helping people
+install `libgdiplus` correctly (because it's not part of .NET Core proper). We
+have thought about our options, and to support `System.Drawing.Common` properly
+on Unix we'd need to make `libgdiplus` a first-class citizen and ship it as part
+of the product:
+
+* `libgdiplus` doesn't ship with all operating systems by default, just like
```
danmoseley, line 27:
```suggestion
* `libgdiplus` doesn't ship with all operating systems by default, unlike
```

I think this is what you meant? icu and openssl are typically included in distros or at least on feeds, whereas libgdiplus is on a custom feed and presumably not by default in any distro.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+
+* `libgdiplus` is community driven, so a lot of issues/PRs just go stale and are
+  never looked at.
+
+* There is no servicing story for bugs and security issues.
+
+However, making `libgdiplus` a first-class citizen is costly. It's around 30k
+lines of pretty old C code, virtually untested, and lacks a lot of
+functionality. `libgdiplus` also has a lot of external dependencies for all the
+image processing and text rendering, such as `cairo`, `pango`, and other native
```
danmoseley, line 38: Do we know how actively supported (and widely distributed) those dependencies are too? All the same requirements apply to them.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+Thus, we believe it is not worth investing in making `System.Drawing.Common` work
+on non-Windows platforms and instead propose we only keep evolving it in the
+context of Windows Forms and GDI+.
+
+## Requirements
+
+### Goals
+
+* Consumers of `System.Drawing.Common` will be informed that starting with .NET
+  6, it's only supported on Windows, via the platform compatibility analyzer
```
danmoseley, line 67:
```suggestion
  6, it's only supported on Windows, via the [platform compatibility analyzer](https://docs.microsoft.com/en-us/dotnet/standard/analyzers/platform-compat-analyzer)
```

BTW, are we supposed to use hard breaks in markdown? I haven't been..

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+
+>   The `System.Drawing.Common` NuGet package works on Windows, Linux, and macOS.
+>   However, there are some platform differences. On Linux and macOS, the GDI+
+>   functionality is implemented by the `libgdiplus` library. This library is not
+>   installed by default in most Linux distributions and doesn't support all the
+>   functionality of GDI+ on Windows on macOS. There are also platforms where
+>   `libgdiplus` is not available at all. To use types from the
+>   `System.Drawing.Common` package on Linux and macOS, you must install
+>   `libgdiplus` separately. For more information, see Install .NET on Linux or
+>   Install .NET on macOS.
```
danmoseley, line 94: This guidance mentions that it's not necessarily available or feature complete. Just as important is that it's not supported, which is a reliability and security concern.
That's important because it would be easy to read this and conclude that if it's available and working today for a customer, they can stick with it. We want to keep customers safe so we should explicitly recommend that they migrate off it.

danmoseley: I guess you'll be updating this text anyway to explain the deprecation actions described below.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
danmoseley: > your mileage may vary. If it continues to work for you, great. But if you run into issues, we won't be fixing any bugs or new features going forward

Feels to me we need to be more prescriptive. We led customers to use this library on Unix. They had every reason to believe it would be supported on the same lifecycle, but it will not be, potentially opening stability or security concerns. In my mind we should say essentially

* when you target your app to .NET 6 or later, any use of System.Drawing on Unix platforms will fail.
* to keep existing apps working while adopting .NET 6, you can pin to the previous version of System.Drawing as a temporary measure <instructions>. you should pin only for Unix deployed applications, so that Windows deployed applications recieve the latest updates and bug fixes contained in the latest System.Drawing.
* however long term, we strongly encourage migrating such code to an alternative library on Unix. Alternatives include SkiaSharp and ImageSharp. These work well on both Windows and Unix.
* as a reminder, usage of System.Drawing on Windows remains fully supported as it has always been.
?

danmoseley: We can afford to be more aggressive in our deprecation than BinaryFormatter, which became engrained throughout our platform and ecosystem over 20 years.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,193 @@
...
+In order to assess how `System.Drawing.Common` we looked at the API usage from
+libraries published to nuget.org.
+
+We looked at member refs to types that begin with `System.Drawing.` and
+aggregated the hits by type and member name. We also pulled in the TFM of the
+assemblies in question to help categorize the hits. There's a lot of data (there
+are a lot of members * TFMs), so instead of just huge raw tables of nonsense,
+we'll provide a summary of what we observed. 
+
+If we look at .NET Core family of TFMs, the top APIs are:
```
danmoseley, line 153: Was there any feasible way to guess at Unix vs Windows usage levels - RID specific libraries that depend on them? @marklio

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
eerhardt: At least users get bug fixes/improvements that are common across platforms. For example, nullable annotations and ILLink annotations.

A concrete advantage: using `System.Drawing.Common` v5 in a trimmed app produces ILLink warnings from inside System.Drawing.Common that developers can't fix. Using v6 does not.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
richlander: These were the two bullet points I was referencing. I agree that I wasn't clear.

> * `libgdiplus` doesn't ship with all operating systems by default, just like other native libraries we depend on (icu, openssl).
> * `libgdiplus` is community driven, so a lot of issues/PRs just go stale and are
  never looked at.

danmoseley:
@lukemcdo 

> This is not a good change because there are a substantial number of packages that were brought to .NET Core which have non-primary functionality that depends on System.Drawing.Common. By making this breaking change, those packages will no longer work on Linux, despite that their functionality was intact.

Those packages would still work on Linux in .NET 6 apps but the app would need to pin the version of S.D.C to 5.0 as mentioned above. Does that alleviate some of the concern? The runtime exception messages could have a link to instructions.

danmoseley:
@atrauzzi 

> (If for some reason this same issue was affecting Windows, I'd put some money on Microsoft being more willing to invest effort to shore things up.)

I hear you. However in this case it's really an artefact of Windows history: System.Drawing is fundamentally a wrapper around Windows GDI+ . That meant the Unix port was always a stretch, even if the projects it required were as actively maintained as the Windows APIs themselves. I think if the platforms were reversed, the dynamics for us would be the same. 

Where concepts and API shapes are not platform specific (eg HTTP), we generallyu make great efforts to have good experiences on both Linux and Windows.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
safern: Unfortunately there is not as NuGet is mostly libraries and we don't know who consumes them and how, but from the APIs we see are mostly used on NuGet we can see that this is definitely replaceable by the other libraries mentioned in this doc.

richlander:
One of the most important deliverables from our team is clarity. "You will get this if you do this." I see this change as falling under that umbrella and support it because of that. We're saying "the you will get" part is too severely bad that we're no longer willing to support/deliver it.

I think we're also saying that we'd never support a Microsoft team using this component. That should perhaps be part of our message, that we're only willing to maintain components that we'd feel comfortable with Microsoft using, for example in an Azure service or an app/tool. That's a pretty good test, and I think one that we implicitly always use.

We should be targeting high-quality and industry-accepted libraries as a rule, and only adopting libraries (or tools) that don't satisfy that as intentionally chosen, highly scrutinized, and rare exceptions. From what I'm reading, this choice was probably a mistake. ICU, OpenSSL, and LTTNG are examples of components that hit that high bar. Same with curl although we're only using it as a tool (as opposed to libcurl) at this point.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
marklio: Right, it is exceedingly rare for libraries in NuGet to target RIDs, and even more rare for that to be leveraged as a means to constrain the applicability of a library to an OS. In my experience, targeting RIDs typically means that you're providing a platform abstraction layer of sorts as an implementation detail of a cross-plat API rather than constraining a library to a particular platform. And again, it's so rare that it has no predictive power for what we're trying to determine.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
safern: After giving it a second thought. Given that 5.0.0 package will be out of support 3 months after the 6.0 release, I think it makes sense to do the original proposal which is:

1. Mark the assembly as windows only so that the platform analyzer warns
2. Throw PNSE by default when loading libgdiplus with a link to the aka.ms speaking about unix not being officially supported by us.
3. Providing a compat switch to opt-in to the unix implementation and don't throw when that is enabled.
4. In 7.0 nuke the unix implementation entirely.

The reason I thin we should follow that plan is because there is no way to cross target by RID so you can not say on a project, use this package version for windows but this package version for unix, so all these libraries that depend on our package will be out of support on windows in 3 months after the release and I believe we should give them more time to adjust, with the constraint that _Unix is no longer supported by us and we will not fix any unix bugs in our product nor libgdiplus_.

Thoughs, @eerhardt @danmoseley @terrajobst @jkotas

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: Yes, because Markdown is supposed to be readable in clear text and unwrapped lines aren't friendly in this regard.

danmoseley: Ugh -- which text editor doesn't have a wrapping toggle. But, OK.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
danmoseley: Got it.

terrajobst:
@richlander 

> I think we're also saying that we'd never support a Microsoft team using this component. That should perhaps be part of our message, that we're only willing to maintain components that we'd feel comfortable with Microsoft using, for example in an Azure service or an app/tool. That's a pretty good test, and I think one that we implicitly always use.

That's great framing. I'll try to incorporate that into the doc.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: Right. Also, we're looking at libraries, not applications. Even if there were RIDs, this still wouldn't tell us what consumers are using. All it would tell us is that the library is offering an asset for a given RID, not that it's actually used.

In a sense, that's an actual blind spot in our data gathering and every time we look at NuGet this is a bias we need to consider.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
jkotas: Sounds good to me.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: OK, I've reworded the intro to remove the red herrings and focus on why we believe `libgdiplus` isn't a viable option.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: @safern ?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
danmoseley: That sounds good to me given 5.0 will be going out of support next year. Opt-in to Unix.

terrajobst: I like the plan.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
lukemcdo: Cairo is a core library to GTK still, I believe. I think that's also the case with Pango but I'm not sure.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,213 @@
...
+`System.Drawing.Primitives`.*
+
+## Q & A
+
+### Will this prevent applications to move to .NET 6?
+
+No, `System.Drawing.Common` is a standalone package that does not ship as part
+of the shared framework. For that reason if an application wants to keep using
+`System.Drawing.Common` on `Unix` and move to .NET 6, they can reference the
+`5.0.x` version of the package.
```
safern, line 137:
```suggestion
No, `System.Drawing.Common` is a standalone package that does not ship as part
of the shared framework. For that reason if an application wants to keep using
`System.Drawing.Common` on `Unix` and move to .NET 6, they can either pin the version of the package to `5.0.x` or use the `6.0.x` package and set a runtime config switch `System.Drawing.UnixSupport` to `true`. 

However application should be aware that we will not provide any support for Unix issues starting from .NET 6.
```

safern: @terrajobst what do you think?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,213 @@
...
+   platform compatibility analyzer to warn when consumed from a project that
+   isn't targeting net6.0-windows (or marked as Windows-specific).
+
+2. Throw `PlatformNotSupportedException` when trying to load `libgdiplus`
+   pointing to an aka.ms link with the reasons behind it and recommending
+   stable, well-maintained open-source library alternatives
+
+3. Add targets to the package that when it is restored using a non-windows RID,
+   it warns that we no longer support it on non-windows and point to the aka.ms
+   link with more info.
```
safern, line 115:
```suggestion
1. **Mark the assembly as only being supported on Windows**. That will cause the
   platform compatibility analyzer to warn when consumed from a project that
   isn't targeting net6.0-windows (or marked as Windows-specific).

2. Throw `PlatformNotSupportedException` when trying to load `libgdiplus`
   pointing to an aka.ms link with the reasons behind it and recommending
   stable, well-maintained open-source library alternatives

3. Provide a runtime config switch named `System.Drawing.UnixSupport`, that when is set to `true` we will no longer throw a `PlatformNotSupportedException` when loading `libgdiplus`.

4. In .NET 7+ all of `Sytem.Drawing.Common` APIs dependent of `libgdiplus` will throw `PlatformNotSupportedException` and the runtime config switch will no longer work.
```

lukemcdo:
> 
> 
> @atrauzzi
> 
> > (If for some reason this same issue was affecting Windows, I'd put some money on Microsoft being more willing to invest effort to shore things up.)
> 
> I hear you. However in this case it's really an artefact of Windows history: System.Drawing is fundamentally a wrapper around Windows GDI+ . That meant the Unix port was always a stretch, even if the projects it required were as actively maintained as the Windows APIs themselves. I think if the platforms were reversed, the dynamics for us would be the same.
> 
> Where concepts and API shapes are not platform specific (eg HTTP), we generally make great efforts to have good experiences on both Linux and Windows.

The current implementation, libgdiplus, is a shim over Cairo. The thing I'm seeing is that no one wants to maintain the community C code. That to me doesn't sound like the same thing as "we can't support Drawing.Common on Unix." Since MS helps maintain SkiaSharp and SkiaSharp is expected to be maintained in some capacity for years, would a managed shim for the most common uses for Drawing.Common on Unix be viable? Or is providing only part of the API seen as an anti-feature?

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
safern: Yes cairo is actively supported and widely distributed on Debian and Fedora distros. For macOS you still need to install it manually.

And yes, pango is `GTK` still.

lukemcdo:
> 
> 
> One of the most important deliverables from our team is clarity. "You will get this if you do this." I see this change as falling under that umbrella and support it because of that. We're saying "the you will get" part is too severely bad that we're no longer willing to support/deliver it.
> 
> I think we're also saying that we'd never support a Microsoft team using this component. That should perhaps be part of our message, that we're only willing to maintain components that we'd feel comfortable with Microsoft using, for example in an Azure service or an app/tool. That's a pretty good test, and I think one that we implicitly always use.
> 
> We should be targeting high-quality and industry-accepted libraries as a rule, and only adopting libraries (or tools) that don't satisfy that as intentionally chosen, highly scrutinized, and rare exceptions. From what I'm reading, this choice was probably a mistake. ICU, OpenSSL, and LTTNG are examples of components that hit that high bar. Same with curl although we're only using it as a tool (as opposed to libcurl) at this point.

While this makes sense from Microsoft's perspective, from a Mono user's perspective, their SDK just got absorbed by a larger project, and a core component just got deprecated. I'd like to see the possibility of a replacement shim considered (Skia or other), or a first-party experience using a replacement library with some conversion guidelines for non-WinForms projects.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
eerhardt: +1

AraHaan:
I find this breaking, System.Drawing.Common if I remember right uses gdiplus, and Mac and linux can get gdiplus thanks to mono making System.Drawing.Common usable out of box.

So if anything I think System.Drawing.Common should stay as is. I know of a ton of cross platform software that depends on it being a cross platform dependency.

I could argue however that I think the package should include an reference only "native" version however for C++/CLI users can "reference" the library but have it use the one from the set TargetFramework inside of that C++/CLI project that was made in C# (since it is possible to do that).

danmoseley:
> Or is providing only part of the API seen as an anti-feature?

I think it would be an anti-feature, due to confusion.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
terrajobst: I'd remove the pinning as that is a non-standard workflow and some of our packages (i.e. most of ASP.NET Core won't let you do that). I'd just leave the runtime config switch as the primary mitigation.

huoyaoyuan:
And the state of managed code of System.Drawing.Common is also not good. I tried to do some modernization for it. There are full of old-fashioned marshaling, `HandleRef` instead of `SafeHandle` which was from .NET 2.0, and weird api from Win98. Maintaining work is also horrible for it.

danmoseley:
> Maintaining work is also horrible for it.

This is another reason tht we do not plan to make any innovations in this library. It is for compatibility with existing code, like eg System.Speech.

AraHaan:
Besides I also know of actual commercial software that expects System.Drawing.Common to be xplat as well too.

Perhaps there could be middle ground to modernize it but to also keep support for Unix.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
safern: Yes we will be updating this text and also provide extra documentation with a breaking change notice and guidance on basic samples on how to migrate to SkiaSharp or ImageSharp.

danmoseley:
> Besides I also know of actual commercial software that expects System.Drawing.Common to be xplat as well too.

@AraHaan could you share some of these popular applications/libraries? You mentioned SVG.NET.

> Perhaps there could be middle ground to modernize it but to also keep support for Unix so that way those programs can benefit from said enhancements indirectly.

The problem is not modernization, it is that it's disproportionately expensive to service stability and potentially security fixes into libgdiplus and its dependencies. We are not comfortable supporting a recommending a library in such a situation.

haefele:
@danmoseley I'm not @AraHaan, but our business project uses [DevExpress](https://www.devexpress.com/) and also [FastReport](https://www.fast-report.com/de/) which both seem to depend upon `System.Drawing.Common` for their cross-platform support.

Just checked it today because I was curious how it might affect me at work.

filipnavara:
As one of the last major libgdiplus contributors along with @hughbe I hesitantly have to agree that it's a reasonable move to deprecate it. However, as mentioned above, the 6.0.0 package made some fixes for ILlink-ability and few other small things so I don't think it's the right move to cut it for .NET 6 without releasing this fixed up package. It's a shame [ObsoletedInOSPlatformAttribute](https://github.com/dotnet/runtime/issues/47601) didn't make it since that would have been a perfect solution here by marking all the types with it.

safern:
> However, as mentioned above, the 6.0.0 package made some fixes for ILlink-ability and few other small things so I don't think it's the right move to cut it for .NET 6 without releasing this fixed up package (opt-in compat switch is ok).

Yes, completely agreed. That's why we decided to make it opt-in with the runtime config switch `System.Drawing.EnableUnixSupport` . That will give people a package that is LTS and give bug fixing for Windows and enough time to migrate to another solution for xplat. We just want to make sure that it is known we will not be supporting bug fixing for Unix at 6.0 as it will be an opt-in switch  knowing that.

0xced:
The document mentions the *runtime config switch* but never shows how to enable it. If I understand correctly that would be `AppContext.SetSwitch("System.Drawing.EnableUnixSupport", true);` ?
I think it would be a good thing to _explicitly_ show how to enable this runtime config switch, also maybe with a link to the AppContext documentation.

filipnavara:
@0xced I suppose that most people would enable the switch through runtimeconfig.json. There's reasonably good documentation for it at https://docs.microsoft.com/en-us/dotnet/core/run-time-config/. I agree it may be worth linking but not sure if that's necessary in the design doc itself.

safern:
> I think it would be a good thing to explicitly show how to enable this runtime config switch, also maybe with a link to the AppContext documentation.

We will make sure to mention how to enable the switch in the documentation which will be linked from the exception thrown when the switch is not enabled.

0xced:
> We will make sure to mention how to enable the switch in the documentation which will be linked from the exception thrown when the switch is not enabled.

This is a very good point!

I think it would be a plus to also include the link to https://docs.microsoft.com/en-us/dotnet/core/run-time-config/ in the design document. I can see how someone unfamiliar with runtime config switches reading the design document will have to google "runtime config switch System.Drawing.Common" only to find this exact GitHub issue, then skim through dozen of comments before finding out about the magic documentation link.

It could be as simple as [runtime config switch](https://docs.microsoft.com/en-us/dotnet/core/run-time-config/) the first time the term appears in the document.

Providing links to more documentation is never a bad idea, wherever it's linked from.

My 2 cents.

RussKie (commented):
My main concern is that we're proposing/making changes like this so late in the release cycle.

AraHaan:
I think the best bet is to actually add the switch in .NET 6 (since it is too late for .NET 6), and then remove it and all support in .NET 8 for ``System.Drawing.Common`` on non-Windows OS's.

accepted/2021/system-drawing-win-only/system-drawing-win-only.md
```diff
@@ -0,0 +1,214 @@
...
+
+1. **Mark the assembly as only being supported on Windows**. That will cause the
+   platform compatibility analyzer to warn when consumed from a project that
+   isn't targeting net6.0-windows (or marked as Windows-specific).
+
+2. Throw `PlatformNotSupportedException` when trying to load `libgdiplus`
+   pointing to an aka.ms link with the reasons behind it and recommending
+   stable, well-maintained open-source library alternatives
+
+3. Provide a runtime config switch named `System.Drawing.UnixSupport`, that when is set to `true` we will no longer throw a `PlatformNotSupportedException` when loading `libgdiplus`.
```
safern, line 113:
```suggestion
3. Provide a [runtime config switch](https://docs.microsoft.com/dotnet/core/run-time-config/) named `System.Drawing.EnableUnixSupport`, that when is set to `true` we will no longer throw a `PlatformNotSupportedException` when loading `libgdiplus`.
```

safern:
> @RussKie: My main concern is that we're proposing/making changes like this so late in the release cycle.

> @AraHaan: I think the best bet is to actually add the switch in .NET 7 (since it is too late for .NET 6), and then remove it and all support in .NET 8 for System.Drawing.Common on non-Windows OS's.

Thanks for your concerns. So there are some reasons why we think the timing is OK and we will proceed with this plan:
1. We have stated in our release cycles that we reserve the right from making breaking changes on any Preview and we promise we will not make breaking changes when we are in an RC, so we are still on time to make such changes.
2. We are not removing functionality, the feedback that it was too late to remove the functionality and that .NET 5 was non LTS we do agree with, so that's why we decided to not remove the functionality but just put it behind a runtime switch. The reason and motivation to put it behind the switch is to drive people's attention to what's the key important things here: a) State that we no longer support Unix bugs/issues on `System.Drawing.Common` and give time for libraries/apps to move to the recommended libraries. b) Next package major version will not have this functionality.

Given that we are not removing functionality; it is not blocking, we are just trying to communicate what we had in our documentation since a couple of releases ago more aggressively as that was somewhat ignored.

safern:
@terrajobst good to merge?

danmoseley:
@terrajobst I think we can merge this now?

ANahr:
> > Besides I also know of actual commercial software that expects System.Drawing.Common to be xplat as well too.
> 
> @AraHaan could you share some of these popular applications/libraries? You mentioned SVG.NET.

Not directed to me, But e.g. ClosedXML uses it which alone has more than 12 Mio Downloads.

danmoseley:
@safern we should probably reach out proactively to some of these top users like ClosedXML

safern:
? @safern we should probably reach out proactively to some of these top users like ClosedXML

Yup, I'll do that.

danmoseley:
Thanks. that will help us validate the decision to remove the code next release. Also perhaps we can aggregate best practice suggestions for moving off it by looking at what they choose to do.

ericstj:
@Pangamma maybe you could open an issue in dotnet/runtime which describes the difficulty porting to those other frameworks?  Perhaps we could share some resources to make that easier.  I know @safern was working on some documentation around this and we're open to helping folks make the transition by contributing to some of those cross platform libraries.

Note that the only thing done in this release is to introduce an exception which can be disabled with `System.Drawing.EnableUnixSupport`, so you can use the latest package if you set this.

JimBobSquarePants:
>The main things I needed to swap out was palette quantization, image resizing, dithering, and image save/load. Also combining tiled images onto larger images.

@Pangamma  ImageSharp has all that functionality baked in. If you do decide to port please ping me. It'll be useful to observe and help out.

> we're open to helping folks make the transition by contributing to some of those cross platform libraries. 

@ericstj I think that has to absolutely happen, and sooner rather than later. That's a lot of pressure to put on OSS projects who receive little to no assistance/funding.

PR #237, opened 2021-08-04
Proposal for Blazor/wasm crypto support in .NET 6 and beyond

GrabYourPitchforks:
This document discusses how we intend to support cryptographic scenarios in Blazor/wasm apps in .NET 6 and beyond. It intends to address the common use case of developers wanting access to HMACSHA\* and AES in Blazor/wasm apps.


GrabYourPitchforks:
For feedback on the plan for .NET 7 and beyond, please leave a comment on https://github.com/dotnet/runtime/issues/40074. That issue has been the most active with respect to gathering user scenarios and providing requests. We can also split off feedback items into new individual issues as needed.

PR #238, opened 2021-08-05
Update the default dpi value

RussKie:
We had further internal discussions and it was concluded that we're not ready for HightDpi.PerMonitorV2 to be set as a default as this stage. Updating the default value to "system aware".
Relates to https://github.com/dotnet/winforms/pull/5392


dreddy-work (approved):

merriemcgaw (approved):
Thanks @RussKie!

PR #242, opened 2021-08-23
Package shading draft

mhutch:
For early review of behavior.

Intro and scenarios sections are WIP and will be added

### DIFF
--- a/accepted/2021/experimental-package-shading.md
+++ b/accepted/2021/experimental-package-shading.md
@@ -1,0 +1,59 @@
+# Experiment: NuGet Package Shading
+
+(intro and scenarios to be added)
+
+## Usage
+
+### Package authors
+
+A *package author* may enable shading on any package reference as long as its assets are configured such that it is effectively private. Any other package references that have a shaded package reference as a transitive dependency must also be shaded. When the project is restored, shaded copies of the packages are created and swapped in. When the project is packed, the shaded assets are embedded in the resulting package.
+
+### Package consumers
+
+Shading is superficially invisible to consumers of a package that has shaded dependencies. They may notice the shaded assets in the build logs or output directory, and their app size may increase, but they do not need to be aware that shading exists. From the consumer's perspective, a shaded dependency is no different than any other private asset.
+
+## Behavior
+
+### Private assets
+
+This shading mechanism is built upon the existing NuGet concept of *private assets*. A shaded asset is inherently private and must be treated the same way as other private assets. Handling shaded assets using the private assets mechanisms reuses existing behavior and concepts, and should result in a consistent experience.
+
+Shading may be enabled on any package reference that is effectively a *private package reference* i.e. all assets consumed from the package are consumed as private assets, and the reference is not exposed transitively. A future proposal will explore making private package references a first class concept.
+
+### Shaded packages
+
+To ensure consistency in package assets, shading applies to an entire package rather than individual assets or asset types. The core of this feature is a mechanism that can take a package and synthesize a *shaded package* by copying the original package and renaming the copy it and its assets to have a new identity: the *shaded name*.
+
+The shaded name is a mangled name specific to the context in which the package is shaded and is designed such that the shaded package's assets do not collide at runtime with assets from other shaded copies and from the original unshaded package.
+
+The name mangling is an implementation detail, and although it is deterministic to allow for deterministic builds, it may change in future versions of the .NET SDK. Developers should not depend on the mangling format or specific mangled IDs outside of references created by the .NET SDK itself.
+
+When a shaded package has a dependency on a package that is also shaded, its assets will be processed so that any references to the dependency's unshaded assets are updated to reference the shaded assets.
+
+### Shade-on-restore
+
+During NuGet restore, any package reference in the graph that is shaded will be replaced with a reference to a shaded package created locally in the intermediate output directory.
+
+A package may only be shaded when all other packages in the graph that depend on it are also shaded. Their assets must be shaded in order to be rewritten to reference its shaded assets.
+
+### Rename safety
+
+Renaming a package's assets so that multiple renamed versions can be used at runtime is not something that can be performed safely. Some assets may inherently have singleton behavior, and assets may embed the original name in ways that cannot automatically be detected and updated, for example when using reflection to load an assembly by name.
+
+The shading tools will detect known unsafe patterns and warn when assets cannot be renamed safely, for example calls to [`AssemblyLoadContext.Load`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext.load?view=net-5.0) with values that cannot be determined statically. However, the shading tools will not be able to detect all problematic cases. Authors of packages with shaded dependencies are expected to test their package thoroughly and verify that it works correctly with shaded dependencies.
+
+A particularly problematic case is when assets are not inherently unsafe, but are used by the consumer in an unsafe way. For example, types from a library with shaded dependencies may get serialized in a way that embeds a shaded shaded assembly's shaded ID into the serialization output. This is unlikely to occur in practice as  reflection based serialization of private fields is generally considered problematic, but it represents an example of the kinds of problems that cannot easily be detected automatically.
+
+### Controlling shading
+
+Shading will be controlled by a boolean parameter named `ShadePrivateAssets` will alter the behavior of NuGet restore in .NET SDK based projects. Any package reference for which `ShadePrivateAssets` is `true` will be shaded: replaced with a reference to a synthetic package created by copying the package and renaming it and its assets. The name of this parameter is based on the existing `PrivateAssets` package reference metadata that causes assets to be consumed by the project and not surfaced transitively to referencing packages. Certain types of private assets such as runtime libraries are embedded in any package created by running the Pack target on the project.
+
+The `ShadePrivateAssets` parameter exists both as an MSBuild property and as MSBuild item metadata. The property defaults to an empty value, and the metadata defaults to the property value. Empty values are interpreted as `False`. The property can be used to set the behavior for all package references in a project, and the metadata can set or override the behavior for individual package reference items.
+
+Shading will only be performed on private package references. Package references may be made private via setting the the existing `PrivateAssets`, `IncludeAssets` and `ExcludeAssets` package reference metadata such that compile-time assets and the package reference itself do not transitively escape the project that contains the private reference. Attempting to enable shading for non-private package references via MSBuild metadata will cause a restore error, and non-private package references ignore the MSBuild property.
+
+> *NOTE:* Creating a private reference currently requires a deep understanding of `PrivateAssets`, `IncludeAssets` and `ExcludeAssets`. A future proposal will make it easier to mark a package reference as private and verify that types from private references are not surfaced in public API.
+
+To allow deterministic builds, the assets in a synthetic shaded package are renamed using a deterministic ID created by mangling the name of the project and the original ID of the package: `__Shaded_{ProjectName}_{OriginalPackageId}`. The ID of the synthetic package is an internal implementation detail and may be subject to change, but this does not matter as it should never end up in any artifacts created from the project.
+
+Only direct package references may be shaded. Shaded direct package references may be unified with transitive package references, but only when those transitive package references are transitive via other shaded package references.


accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,59 @@
+# Experiment: NuGet Package Shading
+
+(intro and scenarios to be added)
+
+## Usage
+
+### Package authors
+
+A *package author* may enable shading on any package reference as long as its assets are configured such that it is effectively private. Any other package references that have a shaded package reference as a transitive dependency must also be shaded. When the project is restored, shaded copies of the packages are created and swapped in. When the project is packed, the shaded assets are embedded in the resulting package.
+
```
anangaur, line 10: Can we introduce the term `shading` in the intro section?

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,59 @@
...
+### Package consumers
+
+Shading is superficially invisible to consumers of a package that has shaded dependencies. They may notice the shaded assets in the build logs or output directory, and their app size may increase, but they do not need to be aware that shading exists. From the consumer's perspective, a shaded dependency is no different than any other private asset.
+
+## Behavior
+
+### Private assets
+
+This shading mechanism is built upon the existing NuGet concept of *private assets*. A shaded asset is inherently private and must be treated the same way as other private assets. Handling shaded assets using the private assets mechanisms reuses existing behavior and concepts, and should result in a consistent experience.
```
anangaur, line 19: Can we link it to docs on Private assets?

accepted/2021/experimental-package-shading.md
anangaur: > Any other package references that have a shaded package reference as a transitive dependency must also be shaded.

I was lost here. Can we have examples?

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,59 @@
...
+### Shaded packages
+
+To ensure consistency in package assets, shading applies to an entire package rather than individual assets or asset types. The core of this feature is a mechanism that can take a package and synthesize a *shaded package* by copying the original package and renaming the copy it and its assets to have a new identity: the *shaded name*.
+
+The shaded name is a mangled name specific to the context in which the package is shaded and is designed such that the shaded package's assets do not collide at runtime with assets from other shaded copies and from the original unshaded package.
+
+The name mangling is an implementation detail, and although it is deterministic to allow for deterministic builds, it may change in future versions of the .NET SDK. Developers should not depend on the mangling format or specific mangled IDs outside of references created by the .NET SDK itself.
+
+When a shaded package has a dependency on a package that is also shaded, its assets will be processed so that any references to the dependency's unshaded assets are updated to reference the shaded assets.
+
```
anangaur, line 32: Looks like we need a couple of examples to explain this better.

accepted/2021/experimental-package-shading.md
mhutch: Added a link :)

accepted/2021/experimental-package-shading.md
mhutch: I rewrote this section; hopefully it's clearer? I don't want to complicate this section too much. Maybe I could just omit that detail and say "(subject to constraints detailed later in this spec)."

I will try to add a more detailed example with a diagram elsewhere.

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
+# Experiment: NuGet Package Shading
+
+(intro and scenarios to be added)
+
+## Usage
+
+### Package authors
+
+A *package author* may enable shading on any package reference as long as its assets are configured such that it is effectively private, and as long as any transitive references to that package are via shaded package references. When they restore their project, its shaded package references will be substituted for renamed copies of those packages and their assets. When the project is packed, these renamed assets are bundled in the resulting package.
+
```
anangaur, line 10: > A *package author* may enable shading on any package reference 

"How?" A walkthrough of experience is needed?

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
+# Experiment: NuGet Package Shading
+
+(intro and scenarios to be added)
+
```
anangaur, line 4: We need probably a "Problem" statement section.  I am assuming that will be added as part of the Intro?

accepted/2021/experimental-package-shading.md
anangaur: May be link to the "Controlling Shading" section from here is that's where the actual experience is explained.

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
...
+### Controlling shading
+
+Shading will be controlled by a boolean parameter named `ShadePrivateAssets` will alter the behavior of NuGet restore in .NET SDK based projects. Any package reference for which `ShadePrivateAssets` is `true` will be shaded: replaced at restore time with a reference to a synthetic package created by copying the package and renaming it and its assets. The name of this parameter is based on the existing `PrivateAssets` package reference metadata that causes assets to be consumed by the project and not surfaced transitively to referencing packages. Certain types of private assets such as runtime libraries are embedded in any package created by running the Pack target on the project.
+
```
anangaur, line 64: Can we have the full `<PackageReference>` node with these property set listed here?

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
+# Experiment: NuGet Package Shading
+
```
anangaur, line 2: Reading through the doc makes me feel this is actually a `PrivatePackage` feature similar to `PrivateAssets` feature we got today. So can we call it so i.e. "NuGet Private Packages" or "NuGet Local Packages". I like the Private better though.

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
...
+### Controlling shading
+
+Shading will be controlled by a boolean parameter named `ShadePrivateAssets` will alter the behavior of NuGet restore in .NET SDK based projects. Any package reference for which `ShadePrivateAssets` is `true` will be shaded: replaced at restore time with a reference to a synthetic package created by copying the package and renaming it and its assets. The name of this parameter is based on the existing `PrivateAssets` package reference metadata that causes assets to be consumed by the project and not surfaced transitively to referencing packages. Certain types of private assets such as runtime libraries are embedded in any package created by running the Pack target on the project.
```
anangaur, line 63: Can `ShadedPrivateAssets` be just `Private`? Or `Shaded` if we want to use this term.

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
...
+
+The `ShadePrivateAssets` parameter exists both as an MSBuild property and as MSBuild item metadata. The property defaults to an empty value, and the metadata defaults to the property value. Empty values are interpreted as `False`. The property can be used to set the behavior for all package references in a project, and the metadata can set or override the behavior for individual package reference items.
+
+Shading will only be performed on private package references. Package references may be made private via setting the the existing `PrivateAssets`, `IncludeAssets` and `ExcludeAssets` package reference metadata such that compile-time assets and the package reference itself do not transitively escape the project that contains the private reference. Attempting to enable shading for non-private package references via MSBuild metadata will cause a restore error, and non-private package references ignore the MSBuild property.
+
```
anangaur, line 68: Not clear how `PrivateAssets`, `IncludeAssets` and `ExcludeAssets` works in conjunction with `ShadePrivateAssets` on a `PackageReference` node. Or are these exclusive?

accepted/2021/experimental-package-shading.md
anangaur: I guess an example will make it clear.

accepted/2021/experimental-package-shading.md
mhutch: `PrivateAssets` already enables making package references private. This is about renaming private packages and their assets so they don't conflict with other copies of those assets.  I wanted to build on the existing mechanism for private references rather than defining a new one.

I don't think `PrivateAssets` is particularly user friendly, and I want to fix that in a separate, dependent spec.

accepted/2021/experimental-package-shading.md
mhutch: Yes, it will.

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
...
+### Package authors
+
+A *package author* may enable shading on any package reference as long as its assets are configured such that it is effectively private, and as long as any transitive references to that package are via shaded package references. When they restore their project, its shaded package references will be substituted for renamed copies of those packages and their assets. When the project is packed, these renamed assets are bundled in the resulting package.
+
+The package author can be confident that when their library is used in an app, it will use the bundled copy of the shaded dependencies, and will not be affected by any other versions of those dependencies used elsewhere in the app.
+
+### Package consumers
+
+Shading is superficially invisible to consumers of a package that has shaded dependencies. They may notice the shaded assets in the build logs or output directory, and their app size may increase, but they do not need to be aware that shading exists. From the consumer's perspective, a shaded dependency is no different than any other private asset.
```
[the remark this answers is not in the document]
mhutch: Good catch - I don't think shading needs to affect those commands, but I hadn't considered it.

AFAICT the use cases for that are finding package updates and vulnerable versions, and showing transitive package references. I don't think whether a package reference is shaded is relevant to either of those, and I can't think of a scenario where listing shading information at the CLI would be useful.

accepted/2021/experimental-package-shading.md
mhutch: `Private` will be used by my upcoming spec for private package references. I picked `ShadedPrivateAssets` to make it absolutely clear that shading only affected private assets, but we could assume the user knows that and use  `Shaded`.

accepted/2021/experimental-package-shading.md
mhutch: Shading is mutually exclusive with `IncludeAssets`. The developer sets `PrivateAssets` and `ExcludeAssets` however they want, and if they enable shading, then the private assets are shaded.

The existing default values of these metadata are incompatible with shading, but we _could_ use different defaults for them all when shading is enabled. I'll add that to the spec.

accepted/2021/experimental-package-shading.md
mhutch: Now I think about it, `Shade` might be better than `Shaded`. It's a command, not a description.

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Controlling shading
+
+Shading will be controlled by a boolean parameter named `ShadePrivateAssets` will alter the behavior of NuGet restore in .NET SDK based projects. Any package reference for which `ShadePrivateAssets` is `true` will be shaded: replaced at restore time with a reference to a synthetic package created by copying the package and renaming it and its assets. The name of this parameter is based on the existing `PrivateAssets` package reference metadata that causes assets to be consumed by the project and not surfaced transitively to referencing packages. Certain types of private assets such as runtime libraries are embedded in any package created by running the Pack target on the project.
+
+The `ShadePrivateAssets` parameter exists both as an MSBuild property and as MSBuild item metadata. The property defaults to an empty value, and the metadata defaults to the property value. Empty values are interpreted as `False`. The property can be used to set the behavior for all package references in a project, and the metadata can set or override the behavior for individual package reference items.
```
anangaur, line 65: Will `ShadePrivateAssets` be just `Shade` now based on discussions above?

accepted/2021/experimental-package-shading.md
```diff
@@ -0,0 +1,73 @@
...
+
+> *NOTE:* Creating a private reference currently requires a deep understanding of `PrivateAssets`, `IncludeAssets` and `ExcludeAssets`. A future proposal will make it easier to mark a package reference as private and verify that types from private references are not surfaced in public API.
+
+To allow deterministic builds, the assets in a synthetic shaded package are renamed using a deterministic ID created by mangling the name of the project and the original ID of the package: `__Shaded_{ProjectName}_{OriginalPackageId}`. The ID of the synthetic package is an internal implementation detail and may be subject to change, but this does not matter as it should never end up in any artifacts created from the project.
+
+Only direct package references may be shaded. Shaded direct package references may be unified with transitive package references, but only when those transitive package references are transitive via other shaded package references.
```
anangaur, line 73: Should there be a FAQ section at the end? For example, who is this feature for? Or the cons of using this feature? What are the considerations or precautions to use this feature? etc.

accepted/2021/experimental-package-shading.md
mhutch: Rewrote this bit

accepted/2021/experimental-package-shading.md
mhutch: Stubbed out an FAQ, will add to it

accepted/2021/experimental-package-shading.md
mhutch: Added MSBuild snippet here

accepted/2021/experimental-package-shading.md
mhutch: Added examples with diagrams and an intro.

accepted/2021/experimental-package-shading.md
mhutch: This now has an MSBuild example.

accepted/2021/experimental-package-shading.md
mhutch: Simplified the metadata to `Shade="true"` and added an example of the full element

accepted/2021/experimental-package-shading.md
mhutch: Changed it :)

accepted/2021/experimental-package-shading.md
mhutch: Yes, changed it

accepted/2021/experimental-package-shading.md
mhutch: Added examples of transitive shading

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+Once this feature exists, setting `Shade="true"` on a package reference should implicitly mark it as a private dependency.
+
+## Other considerations
+
+### Rename safety
+
+Renaming a package's assets so that multiple renamed versions can be used at runtime is not something that can be performed safely. Some assets may inherently have singleton behavior, and assets may embed the original name in ways that cannot automatically be detected and updated, for example when using reflection to load an assembly by name.
+
+The shading tools will detect known unsafe patterns and warn when assets cannot be renamed safely, for example calls to [`AssemblyLoadContext.Load`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.loader.assemblyloadcontext.load?view=net-5.0) with values that cannot be determined statically. However, the shading tools will not be able to detect all problematic cases. Authors of packages with shaded dependencies are expected to test their package thoroughly and verify that it works correctly with shaded dependencies.
```
dominoFire, line 184: Does this mean that NuGet should do the static analysis ?

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+NuGet performs unification based on the dependency versions defined in the referencing packages or projects. Dependency versions may be exact, an explicit range, or a simple version that implicitly means "equal to or greater than". In the above example, the `v1.0` dependency means `>= v1.0`, so it is compatible with `v2.0`, and NuGet can unify them:
+
+![block2](experimental-package-shading.md.2.png)
+
+ ### Impact on package consumers
+
+ When different versions of a dependency cannot be unified, NuGet restore fails with errors that are often difficult to understand (`NU1605`, `NU1107`). These errors currently make up a majority of NuGet restore errors. In some cases unification is possible but not automatic, in which case a developer may opt in by adding a direct reference to the dependency, but this is not straightforward.
```
dominoFire, line 39: Are there any approximate percentage of NU1605, NU1107 restore errors from customers?

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Why only allow shading private references?
+
+We have made an explicit scoping decision to disallow shading of public references. Shading of public references leads to scenarios where a project has references to multiple types with the same fully qualified names, differing only by assembly name. While this could be handled using [`extern alias`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/extern-alias), it massively complicates the experience for consumers of the library, and would need better support from tooling before being made mainstream.
+
+Changing the *namespace* of shaded assemblies would also make public shaded references possible, but this would substantially complicate the assembly rewriting and bake the rewritten namespaces into the APIs, samples, and docs. This is not scalable and doesn't align with the long term goal of consumer-side shading.
+
+### As a package author, when should I shade a dependency?
+
+> TODO
```
dominoFire, line 234: What about multitargeting scenarios ? Would you shade for earch TFM ?

accepted/2021/experimental-package-shading/experimental-package-shading.md
mhutch: NU1605 and NU1107 together are about 2/3 of install errors in VS.

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+![block5](experimental-package-shading.md.5.png)
+
+## Concerns
+
+Package shading is fundamentally inefficient. Its entire purpose is to cause an app to include and load multiple copies of the same library, and the inefficiency is the cost you pay to solve incompatibilities in your dependencies.
+
+Loading multiple copies of the same library increases an app's download size and memory use, which is particularly problematic in mobile and client-side web applications. Mobile devices have limited memory and will terminate backgrounded applications if they use too much memory, and users may be reluctant to download an app if it's too large, particularly on client namespaces.
+
+A major downside to shading is that an app will end up with multiple copies of the same library, even when those copies are compatible or identical. This is particularly problematic in mobile and WebAssembly where app size is important and significant effort has been invested in .NET to reduce app size. This solution should be considered an intermediate step towards a more complete, whole-app solution that is able to unify compatible shaded dependencies.
```
akoeplinger, line 81: Size/memory isn't the only downside, it also means a consumer is stuck with the version of the shaded package which might have bugs or security issues.

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+Producer-side shading is superficially invisible to consumers of a package that has shaded dependencies. They may notice the shaded assets in the build logs or output directory, and their app size may increase, but they do not need to be aware that shading exists. From the consumer's perspective, a shaded dependency is no different than any other run-time asset.
+
+### Package authors
+
+A package author may enable producer-side shading on any package reference in their project by setting the `Shade` metadata to `True`:
+
+```xml
+<ItemGroup>
+   <PackageReference Include="SomeDependency" Version="v1.0" Shade="True" />
```
akoeplinger, line 95:
```suggestion
   <PackageReference Include="SomeDependency" Version="1.0.0" Shade="True" />
```

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Creating the shaded package
+
+To maintain coherence within a package's assets and across package dependencies, shading is always performed on an entire package and its contents rather than only on individual assets or asset types.
+
+A shaded package and the shading process that creates it are specific to the *shading context* of the project with the shaded package reference. A project may have multiple shaded package references, and those packages may depend on each other. If a shaded package depends on a package that is shaded in the same shading context, then the shading process must retarget the package's dependency and any references in its assets so that they target the shaded package and its assets.
+
+Retargeting a dependency involves finding references in the any references in the package and its assets so that they refer to the shaded version of that dependency. For example, an assembly reference in an assembly asset's metadata table would be updated to reference the shaded version.
```
akoeplinger, line 129:
```suggestion
Retargeting a dependency involves finding any references in the package and its assets so that they refer to the shaded version of that dependency. For example, an assembly reference in an assembly asset's metadata table would be updated to reference the shaded version.
```

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+The shaded name is a mangled name specific to the shading context and is designed such that the shaded package's assets do not collide at runtime with assets from other shaded copies and from the original unshaded package. The mangling is an implementation detail, and may change in future versions of the .NET SDK. Developers should not depend on the mangling format or specific mangled IDs. However, for a given version of the SDK, the mangled ID is deterministic to allow for deterministic builds.
```
akoeplinger, line 131: Hm this might be problematic for cases where the assembly name is used in an external format, e.g. the linker's XML descriptor: https://github.com/mono/linker/blob/7793a2367790bfbd8418e725e8b9751a11278cfb/docs/data-formats.md#xml-examples

accepted/2021/experimental-package-shading/experimental-package-shading.md
akoeplinger: Is there a limit to what it will try to rewrite, e.g. `Assembly.LoadFrom("foo.dll)"` or `Type.GetType()` usages?

The IL Linker already deals with cases like this, there might be some overlap between the tools.

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+As explained earlier, producer-side shading will be implemented as packing of consumer-side shaded assets. We will build on NuGet's existing functionality for controlling and packing assets, expanding where necessary.
+
+### Characteristics
+
+We start by considering the required and desired characteristics of a producer-side shaded package reference.
+
+1. **It is not transitive**. It does not flow to projects that reference the project that contains the shaded package reference, nor is it a dependency of the package created by packing that project.
+2. **Its runtime assets are bundled**. They are copied into the output directory of the project that contains the shaded package reference, and are packed into the package created by packing that project.
+3. **Its runtime assets are renamed** such that they are specific to the referencing project. When the a package created from that project is consumed as a package reference, the renamed runtime assets must not collide with copies of those assets from any other shaded or unshaded package reference to the original package.
```
akoeplinger, line 145:
```suggestion
3. **Its runtime assets are renamed** such that they are specific to the referencing project. When the package created from that project is consumed as a package reference, the renamed runtime assets must not collide with copies of those assets from any other shaded or unshaded package reference to the original package.
```

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,234 @@
...
+
+### Asset renaming
+
+Asset renaming will be handled by restore-time shading. Setting `Shade="true"` on a package reference will cause restore-time shading to shade it. In the scope of this experiment, this is the only flag that will affect restore-time shading.
+
+### Non-transitivity
+
+NuGet's existing concept of [private assets](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets) already provides a way to make a package reference non-transitive. If a package reference has all its assets set to be private assets, then the package reference will not be transitive.
+
+As producer-side shading requires the shaded package be private, setting `Shade="true"` on a package reference should implicitly set `PrivateAssets="all"` on that reference. Explicitly setting any other value should be cause an error.
```
akoeplinger, line 158:
```suggestion
As producer-side shading requires the shaded package be private, setting `Shade="true"` on a package reference should implicitly set `PrivateAssets="all"` on that reference. Explicitly setting any other value should cause an error.
```

accepted/2021/experimental-package-shading/experimental-package-shading.md
akoeplinger: Ah I see it is explained later on, might be useful to link to that section.

accepted/2021/experimental-package-shading/experimental-package-shading.md
akoeplinger: This overlaps with the analysis the IL linker does to figure out `Type.GetType(string)` usages so in theory those could be rewritten to use the shaded identity.

accepted/2021/experimental-package-shading/experimental-package-shading.md
mhutch: Yes, the shading will affect all assemblies in the package.

accepted/2021/experimental-package-shading/experimental-package-shading.md
mhutch: Yes, for sure. We can build in support for some things (XAML files, linker descriptors, etc.) but supporting third party formats would necessitate an extensibility system like the Maven shade plugin's resource transformers.

accepted/2021/experimental-package-shading/experimental-package-shading.md
mhutch: Excellent point, I'd forgotten to include this

SimonCropp:
i did a little spike to try and solve this problem for my scenarios. any feedback would be appreciated https://github.com/SimonCropp/Alias

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,256 @@
...
+
+Copying runtime assets from a project's package reference into the package is currently nontrivial. It requires setting the [`CopyLocalLockFileAssemblies` property](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#copylocallockfileassemblies) to `True` so that private runtime assets are copied to the output directory, and then either manually adding [or writing an MSBuild target](https://github.com/mono/t4/blob/e2b0252ee958a4c1ec4394773f020b3e5e0b095f/Mono.TextTemplating.Build/Mono.TextTemplating.Build.csproj#L27) to do so.
+
+There is no existing metadata for a package reference that causes `Pack` to include its runtime assets in the package. We need a flag that does this, for example `BundleRuntimeAssets="true"` or `RepackAssets="runtime"`. We could expose this flag as an option for general purpose use, or keep it internal, but either way setting `Shade="true"` should enable it.
+
+### Private dependencies
+
+While a package _should_ not expose references to shaded compile assets, as consumers will be unable to resolve them, this does not have to be enforced by tooling. Within the scope of this experiment, we will leave it up to package authors.
```
rainersigwald, line 171: Is it possible for a package author to do this? Right now I think it would work like this:

1. The package doesn't expose the references
2. NuGet restores the primary assembly and the shaded assets into a folder
3. At build time, the SDK tasks read `project.assets.json` and nominates the primary assembly as a `@(Reference)`
4. RAR opens the primary assembly, sees that it references assemblies with shaded names, and tries to resolve them (looking next to the primary assembly)
5. If RAR finds them it exposes them to the compiler

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,256 @@
...
+
+![block9](experimental-package-shading.md.9.png)
+
+This is **explicitly disallowed** at this time for the same reason as the previous example: the package project's direct and transitive references to `Bar` must unify. The only way to do this is by shading `Foo` so that its references to `Bar` can be updated to target the shaded version, and we require the developer do this explicitly so that all shading operations are clear and intentional.
+
+## FAQ
+
+### Why not runtime side-by-side assembly loading?
+
+By extending the runtime to support side-by-side loading of different versions of the same assembly, we would remove the need to manipulate assemblies and their bytecode. However, the other parts of this proposal would still be needed - specifying which dependencies are to be shaded and including the the shaded dependencies in the package. It would limit shading to a future version of the runtime that supported such a feature, which means it would not be usable for many package authors who need to support older runtimes such as including .NET Framework.side-by-side assembly loading.
```
rainersigwald, line 222:
```suggestion
By extending the runtime to support side-by-side loading of different versions of the same assembly, we would remove the need to manipulate assemblies and their bytecode. However, the other parts of this proposal would still be needed - specifying which dependencies are to be shaded and including the the shaded dependencies in the package. It would limit shading to a future version of the runtime that supported such a feature, which means it would not be usable for many package authors who need to support older runtimes such as including .NET Framework.
```

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,256 @@
...
+
+This proposal is also largely about how shading is specified in the project file, how it fits into the restore/build/pack pipeline, and how it affects the dependency graph. ILMerge/il-repack do not by themselves answer any of those.
+
+### How does this compare to Maven shading?
+
+The Maven shade plugin shades the dependencies specified by [explicit includes](https://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html), or shades all dependencies except those explicitly excluded. The include and exclude declarations permit glob patterns but do not take transitivity into account. There's a an option (`promoteTransitiveDependencies`) to promote transitive dependencies of shaded dependencies to direct dependencies, without which they are omitted. This NuGet proposal supports only explicit shading of direct dependencies, to avoid the size increase caused by unnecessary/accidental shading. It always promotes transitive dependencies as this is required for correct behavior.
```
rainersigwald, line 236:
```suggestion
The Maven shade plugin shades the dependencies specified by [explicit includes](https://maven.apache.org/plugins/maven-shade-plugin/examples/includes-excludes.html), or shades all dependencies except those explicitly excluded. The include and exclude declarations permit glob patterns but do not take transitivity into account. There's an option (`promoteTransitiveDependencies`) to promote transitive dependencies of shaded dependencies to direct dependencies, without which they are omitted. This NuGet proposal supports only explicit shading of direct dependencies, to avoid the size increase caused by unnecessary/accidental shading. It always promotes transitive dependencies as this is required for correct behavior.
```

rainersigwald, line 236: > This NuGet proposal supports only explicit shading of direct dependencies

While this is true, since it's exposed as an item in MSBuild one could easily get the inverted Maven behavior with an ItemDefinitionGroup

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Why only allow shading private references?
+
+We have made an explicit scoping decision to disallow shading of public references in the initial version of this feature. Shading of public references leads to scenarios where a project has references to multiple types with the same fully qualified names, differing only by assembly name. While this could be handled using [`extern alias`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/extern-alias), it massively complicates the experience for consumers of the library, and would need better support from tooling before being made mainstream.
```
rainersigwald, line 250: I am 100% on board with this (if it's contentious and you need any backup)

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Why not ILMerge or il-repack?
+
+Both [ILMerge](https://github.com/dotnet/ILMerge) and [il-repack](https://github.com/gluck/il-repack) perform the same kind of identity-changing IL/asset manipulation required by this proposal, and we may be able to leverage them for those mechanics. However, their goal is to _merge_ assemblies, and while pack-time merging could be used to implement a producer-side shading solution, it would not offer a transition path to the consumer-side shading model into which this is intended to evolve.
+
+This proposal is also largely about how shading is specified in the project file, how it fits into the restore/build/pack pipeline, and how it affects the dependency graph. ILMerge/il-repack do not by themselves answer any of those.
+
+### How does this compare to Maven shading?
```
rainersigwald, line 234: The other other-ecosystem system I was thinking of while reading this was Rust's Cargo. It looks like it (effectively) _always_ shades each distinct version and provides tooling to help audit this happening: https://doc.rust-lang.org/cargo/reference/resolver.html#version-incompatibility-hazards

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,256 @@
...
+
+Changing the *namespace* of shaded assemblies would also make public shaded references possible, but this would substantially complicate the assembly rewriting and bake the rewritten namespaces into the APIs, samples, and docs. This is not scalable and doesn't align with the long term goal of consumer-side shading.
+
+### As a package author, when should I shade a dependency?
+
+You should only shade a dependency if you are reasonably certain that it is likely to conflict with other incompatible versions of that dependency in apps that consume your package. Producer-side shading comes at a cost to the consuming application: the download/install/startup/memory cost of your shaded dependencies will no longer be shared with other copies of those dependencies, even if they are identical. By shading a dependency, you are also taking on responsibility for its bugs and security issues, as your consumers can no longer update it directly.
```
rainersigwald, line 256: I would like some motivating examples with considerations walked through.

For example, your library has configuration expressed in some format (JSON, YAML), and internally uses a library to parse that format

1. How do you assess the likelihood of breakage vs. the benefits of unification?
2. For instance, Newtonsoft.Json has a high version number, indicating "many breaking changes have been made"--but I bet you'd almost always be safe for a simple config json parser taking any version, and applications would be understandably upset if they depended on 8 copies of it from various references
3. When would you choose to add shading rather than update to a newer version of a package, if a new version of that package breaks your library?

accepted/2021/experimental-package-shading/experimental-package-shading.md
mhutch: Yes! I ran across that while researching this and see it as a validation of the consumer side mangling approach. It's definitely a very similar model to where I would see this going long term.

I'm not certain that such a strict interpretation of semver would work for reasoning about compatibility of NuGet packages in general, and I would like to see ways to force things to unify (e.g. to reduce size of a wasm app, when you know you're using a compatible subset of two incompatible versions) and to force things to separate (e,g. when you know there was an accidental breaking change in a minor version). But that's a problem for the future :)

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+
+**Owner** [Mikayla Hutchinson](https://github.com/mhutch)
+
+## Introduction
+
+Producer-side package shading is an experimental feature that allows a NuGet package author to "shade" a dependency: embed a renamed copy of it in their package. This ensures that consumers of the package get the exact same version that the package author intended, regardless of any other direct or indirect references to that dependency.
+
+**The major downside to producer-side shading** is that the shaded assemblies are included in the app even when they are identical or completely compatible with other shaded or non-shaded copies of the same dependency. This forced redundancy is a problem for scenarios where app size and memory use is highly important such as mobile apps and client-side web apps, **so producer side shading is not intended to ever become non-experimental or recommended for general use**.
+
+The preferred long-term solution is consumer-side shading, where shading is performed holistically for an app, where it shade packages only when necessary, and deduplicate compatible and identical shaded assemblies.
```
jeffkl, line 11:
```suggestion
The preferred long-term solution is consumer-side shading, where shading is performed holistically for an app, where it shades packages only when necessary, and deduplicates compatible and identical shaded assemblies.
```

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+
+Producer-side shading uses same building blocks as consumer-side shading, but is substantially smaller in scope. It also has the advantage that consumers of a producer-side shading package do not need tooling that understands shading.
```
jeffkl, line 13:
```suggestion
Producer-side shading uses the same building blocks as consumer-side shading, but is substantially smaller in scope. It also has the advantage that people who reference a package with producer-side shading do not need tooling that understands shading.
```

jeffkl: I had to read this sentence several times to understand, I think because it contains the word "consumer" when talking about produce-side shading.

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+P --> |2.0| C2[Foo v2.0]
+B --> |1.0| C1[Foo v1.0]
+```
+
+The project *transitively* depends on `Foo v1.0` and directly depends on `Foo v2.0`. A project cannot depend on multiple versions of the same package, because this would cause conflicts and ambiguity. NuGet must *unify* the two differently versioned Foo dependencies to a single version in the context of that project.
+
+> **NOTE** Unification is independent for each project. All the direct and transitive dependencies for each project must be unified in the context of that project, but the result of unification is specific to that set of dependencies. Another project in the same solution with different dependencies and/or dependency versions may unify to different versions of the same dependencies.
+
+NuGet performs unification based on the dependency versions defined in the referencing packages or projects. Dependency versions may be exact, an explicit range, or a simple version that implicitly means "equal to or greater than". In the above example, the `v1.0` dependency means `>= v1.0`, so it is compatible with `v2.0`, and NuGet can unify them:
+
```
jeffkl, line 40: I feel like we need a whole section talking about the difference between package versions and assembly versions.  Some packages don't rev their assembly versions at all and its worth noting that the file system only supports a single file in each directory with a unique name which reflects how NuGet unifies the dependency graph.  I don't want people to think that its bad design, its just how it is.

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+flowchart LR
+P[Project]
+P --> B[Bar]
+P --> |2.0| C[Foo v2.0]
+B --> |1.0| C
+```
+
+ ### Impact on package consumers
+
+ When different versions of a dependency cannot be unified, NuGet restore fails with errors that are often difficult to understand (`NU1605`, `NU1107`). These errors currently make up a majority of NuGet restore errors. In some cases unification is possible but not automatic, in which case a developer may opt in by adding a direct reference to the dependency, but this is not straightforward.
```
jeffkl, line 51:
```suggestion
 When different versions of a dependency cannot be unified, NuGet restore fails with errors that can be difficult to resolve (`NU1605`, `NU1107`). These errors currently make up a majority of NuGet restore errors. In some cases unification is possible but not automatic, in which case a developer may opt in by adding a direct reference to the dependency, but this is not straightforward.
```

Its very subjective to say the errors are difficult to understand.  In most case the error explicitly states that A depends on B1 and C depends on B2 so you're getting downgraded or whatever.  The difficulty in my opinion is usually in resolving the conflicts.  That's also subjective I suppose

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+
+Even when dependencies are unified, the result may not be correct. The versions of the dependencies may make them appear to be compatible when they are not. This can result in compiler errors, runtime errors, and behavioral differences.
```
jeffkl, line 53:
```suggestion
Even when dependencies are unified, the result could still result in compiler errors, runtime errors, or behavioral differences. Since restore and build succeeded, the versions of the dependencies may make them appear to be compatible when they might not. This is because only the current source is checked not the code paths that dependent assemblies execute because these dependencies were compiled and tested against different versions potentially.
```

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
+# Experiment: NuGet Package Shading
+
+**Owner** [Mikayla Hutchinson](https://github.com/mhutch)
+
+## Introduction
+
+Producer-side package shading is an experimental feature that allows a NuGet package author to "shade" a dependency: embed a renamed copy of it in their package. This ensures that consumers of the package get the exact same version that the package author intended, regardless of any other direct or indirect references to that dependency.
+
+**The major downside to producer-side shading** is that the shaded assemblies are included in the app even when they are identical or completely compatible with other shaded or non-shaded copies of the same dependency. This forced redundancy is a problem for scenarios where app size and memory use is highly important such as mobile apps and client-side web apps, **so producer side shading is not intended to ever become non-experimental or recommended for general use**.
```
jeffkl, line 9: I would put a note in here that in some cases, it's impossible for a package author to know ahead of time that their package could cause conflicts.  Sometimes it's the combination of random packages that cause conflicts, so it doesn't really make sense for the producer to have to handle it.

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+
+### Impact on package authors
+
+Unification problems create issues for package authors, not just consumers:
+
+* **Package consumers may be averse to installing or updating** a reference to a package if it depends on another package used in their app, as unification may change the version of that package and cause problems elsewhere in their app.
+* **Package consumers may not be able to install or update** a reference to a package if it depends on a package that their project already depends on but the versions are different and cannot be unified.
+* **Unification issues may appear to be bugs in the package itself** if unification causes one of the package's dependencies to have a different version than it specified and it is not compatible with that version.
+
+Some package authors go out of their way to avoid dependencies so that they will not cause unification problems for their consumers.
```
jeffkl, line 65: I've certainly had to face this before, sometimes taking a dependency on something just isn't worth it

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+
+## Shading
```
jeffkl, line 67: Should we put a note in about doing _just_ consumer-side assembly shading only instead of shading an entire package?

accepted/2021/experimental-package-shading/experimental-package-shading.md
```diff
@@ -0,0 +1,339 @@
...
+Foo["Foo"]
+Bar["Bar\n(shaded)"]
+P --> Foo
+P --> |Shaded| Bar
+Foo --> Bar
+```
+
+This is **explicitly disallowed** at this time for the same reason as the previous example: the package project's direct and transitive references to `Bar` must unify. The only way to do this is by shading `Foo` so that its references to `Bar` can be updated to target the shaded version, and we require the developer do this explicitly so that all shading operations are clear and intentional.
+
+## FAQ
```
jeffkl, line 301: Not sure the right place for this but we should address the fact that re-writing an assembly is going to strip its authenticode signature and change its public key token.  This could be a problem for high security infrastructures.

SimonCropp:
@foxyseta here is my effort at an implementation https://github.com/getsentry/dotnet-assembly-alias

foxyseta:
Thank you! This definitely sounds like a good tool to use while waiting for shading (or similar functionalities) to be natively integrated into NuGet. Hopefully, I'll have time to take a deeper look into it and drop by.

PR #248, opened 2021-09-21
Change the configuration to use file per arch

vitek-karas:
Change the way multi-arch configuration files are written. Use one file per architecture (instead of one file with multiple lines in it). This makes it much simpler for installers to read and write these configuration files. It also somewhat simplified the product as it doesn't need as complex parsing logic.

Minor cleanup in other areas of the doc.

### DIFF
--- a/accepted/2021/install-location-per-architecture.md
+++ b/accepted/2021/install-location-per-architecture.md
@@ -151,6 +151,8 @@ on Linux and macOS.
 Pre-.NET 6 Windows installer writes registry keys for all architectures
 (even though it only installs one of them). Ideally the installer would
 only write the key for the architecture it's actually installing.
+The plan is to eventually  modify the installers such that only the "runtime" installers
+write the registry key (and only for the architecture they install).
 
 #### Linux/macOS installer behavior
 
--- a/accepted/2021/install-location-per-architecture.md
+++ b/accepted/2021/install-location-per-architecture.md
@@ -161,11 +161,10 @@ with the product location, in the format `install_location_<arch>`. So for examp
 `install_location_x64` (all lower case). The content of the file should be
 a single line with the path to the product.
 
-*Note: Installers should probably not overwrite any of these files if
-they're already present. The best option would be to read the file and install
-into the location specified in the file instead of the default (this is how Windows
-installers should work already). The next best option is to not overwrite
-as the file may contain intentional user changes.*
+*Note: Installers will overwrite existing files. This is to provide
+predictable experience: Installing a given SDK followed by trying to use it
+should just work. These files are maintained by the installer. Custom modifications
+will be overwritten by the next installation.*
 
 #### Support for downlevel apphost
 
@@ -175,6 +174,9 @@ the architecture which .NET supports in versions lower than .NET 6
 should also write the product location to the arch-less location file
 `install_location`. This is needed for downlevel applications to correctly
 find the product.
+
+*Note: Installers will overwrite the `install_location` file to make the installed
+product work as expected. Custom modification will be discarded in that case.*
 
 For example this should be present on macOS arm64 systems if both x64 and arm64
 products are installed.
@@ -263,3 +265,16 @@ and so on.
 
 We don't like this pattern as it feels too Windows-centric and would look
 weird on non-Windows platforms.
+
+### Preserve configuration
+
+The installers would read the config files to determine the location
+of already installed product. And they would either install to that location
+or provide feedback to the user.
+
+This approach is problematic because currently the macOS installers are not capable
+of customizing their install location when they run, so the installer would have
+to basically fail if the config file contained a non-default location.
+
+Also, it's expected that custom modifications to the config files will be very rare.
+In general the config files are maintained by the installer to make the product work.
--- a/accepted/2021/install-location-per-architecture.md
+++ b/accepted/2021/install-location-per-architecture.md
@@ -178,7 +178,7 @@ products are installed.
 
 | File name | Content |
 | --- | --- |
-| `install_location` | `usr/local/share/dotnet/x64` |
+| `install_location` | `/usr/local/share/dotnet/x64` |
 | `install_location_arm64` | `usr/local/share/dotnet` |
 | `install_location_x64` | `usr/local/share/dotnet/x64` |
 
@@ -188,7 +188,7 @@ that is the right location.
 * arm64 6 apphost will react to the `install_location_arm64`
 * x64 6 apphost will react to the `install_location_x64`
 
-*Note: All supported downlevel installers (3.1 and 5) will need to be updated
+*Note: All supported downlevel installers (3.1) will need to be updated
 to install the product into a different location and to always write
 the arch-less configuration file `install_location`.*
 
--- a/accepted/2021/install-location-per-architecture.md
+++ b/accepted/2021/install-location-per-architecture.md
@@ -161,8 +161,11 @@ with the product location, in the format `install_location_<arch>`. So for examp
 `install_location_x64` (all lower case). The content of the file should be
 a single line with the path to the product.
 
-*Note: If installers can write the file, they should also read it and validate
-that they don't overwrite it with different information.*
+*Note: Installers should probably not overwrite any of these files if
+they're already present. The best option would be to read the file and install
+into the location specified in the file instead of the default (this is how Windows
+installers should work already). The next best option is to not overwrite
+as the file may contain intentional user changes.*
 
 #### Support for downlevel apphost
 
@@ -179,8 +182,8 @@ products are installed.
 | File name | Content |
 | --- | --- |
 | `install_location` | `/usr/local/share/dotnet/x64` |
-| `install_location_arm64` | `usr/local/share/dotnet` |
-| `install_location_x64` | `usr/local/share/dotnet/x64` |
+| `install_location_arm64` | `/usr/local/share/dotnet` |
+| `install_location_x64` | `/usr/local/share/dotnet/x64` |
 
 * 3.1 and 5 apphosts will only read the arch-less file `install_location`,
 which points to x64 install. Since 3.1 and 5 are only supported on x64 on macOS,
@@ -250,8 +253,7 @@ Apphosts would have to contain a parser for this format and implement additional
 logic to handle duplicate entries and such.
 
 But more importantly installers would have to be able to read and edit this file
-which is rather challenging given the currently used installer technologies
-(which basically rely on shell scripts).
+which adds complexity to the installers.
 
 ### Use existing format for architecture specific environment variables
 
--- a/accepted/2021/install-location-per-architecture.md
+++ b/accepted/2021/install-location-per-architecture.md
@@ -65,22 +65,26 @@ line, which is the registered location path. There's nothing else in this file.
 
 Already shipped apphost (.NET Core 3.1 and .NET 5) only knows about one install location
 and only about one registration mechanism. The current proposal is to use
-the existing default install location for the native architecture (that is Arm64)
+the existing default install location for the native architecture (that is arm64)
 and have a new location for the non-native architecture (x64). Existing apphost
 is always x64 (.NET hasn't yet shipped Arm64 release with this) and as such
 will not work when using just the default install location knowledge.
 
 On Linux and macOS the apphost only reads the `DOTNET_ROOT`, in all cases.
 
-## Proposal for .NET 6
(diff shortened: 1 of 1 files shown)


vitek-karas:
@mateoatr @agocke

accepted/2021/install-location-per-architecture.md
```diff
@@ -138,34 +141,56 @@ any specific knowledge of where to find the install location, it would just
...
+| `install_location_arm64` | `usr/local/share/dotnet` |
+| `install_location_x64` | `usr/local/share/dotnet/x64` |
+
+* 3.1 and 5 apphosts will only read the arch-less file `install_location`,
+which points to x64 install. Since 3.1 and 5 are only supported on x64 on macOS,
+that is the right location.
+* arm64 6 apphost will react to the `install_location_arm64`
+* x64 6 apphost will react to the `install_location_x64`
+
+*Note: All supported downlevel installers (3.1 and 5) will need to be updated
```
ericstj, line 191: .NET 5 will not get an update per https://github.com/dotnet/designs/blob/main/accepted/2021/x64-emulation-on-arm64/x64-emulation.md.  Not sure if we're making the same concession to the host or not. @richlander 
```suggestion
*Note: All supported downlevel installers (3.1) will need to be updated
```

accepted/2021/install-location-per-architecture.md
```diff
@@ -138,34 +141,56 @@ any specific knowledge of where to find the install location, it would just
...
 
-*Note: If installers write the file, they should also read it and validate
-that they don't overwrite it with different information. It's unclear yet
-if this is necessary and in which situations.*
+#### Windows installer behavior
 
-#### Support for downlevel apphost
+Pre-.NET 6 Windows installer writes registry keys for all architectures
+(even though it only installs one of them). Ideally the installer would
+only write the key for the architecture it's actually installing.
```
ericstj, line 153: This key is also only written by the SDK and not the runtime installers.  So if someone installs just runtimes the host won't have these keys.  Is that a problem?

Did you consider having the host install a key that you might use?  For instance adding InstallLocation here: https://github.com/dotnet/runtime/blob/2f68bf6a57890ac957943579c55a377265035b6d/src/installer/pkg/sfx/installers/host.wxs#L30-L32

accepted/2021/install-location-per-architecture.md
```diff
@@ -138,34 +141,56 @@ any specific knowledge of where to find the install location, it would just
...
+should also write the product location to the arch-less location file
+`install_location`. This is needed for downlevel applications to correctly
+find the product.
+
+For example this should be present on macOS arm64 systems if both x64 and arm64
+products are installed.
+
+| File name | Content |
+| --- | --- |
+| `install_location` | `usr/local/share/dotnet/x64` |
```
mateoatr, line 181: nit:
```suggestion
| `install_location` | `/usr/local/share/dotnet/x64` |
```

accepted/2021/install-location-per-architecture.md
```diff
@@ -138,34 +141,56 @@ any specific knowledge of where to find the install location, it would just
...
+with the product location, in the format `install_location_<arch>`. So for example
+`install_location_x64` (all lower case). The content of the file should be
+a single line with the path to the product.
 
-* 3.1 and 5 apphost would only read the first line, which points to x64 install.
-Since 3.1 and 5 are only supported on x64 on macOS, that is the right location.
-* arm64 6 apphost would react to the `arm64` location
-* x64 6 apphost would react to the `x64` location
+*Note: If installers can write the file, they should also read it and validate
+that they don't overwrite it with different information.*
```
ericstj, line 165: Can you clarify this?  Are you saying that whenever any of the 3 files exist, we should leave them alone and not write them at all?  "Verify they don't overwrite with different information" sounds unusual to me: if it isn't different writing is a noop.   I'd like to understand if you're expecting anything more than just skip when it exists. (like testing the path for presence of dotnet.exe) 

We should really decide if we want all/some of these files to be _user_ state or _product_ state.  I understand that the legacy path was _user_ state before, but we're turning it into both _user_ and _product_ state since it suits our purposes to allow old apps to work.  We can avoid stomping on a user value by skipping to write the file at all if it exists.

We should decide if we want the same behavior for the new files.

The downside of having files in ambiguous _user/product_ state is that the installer cannot repair them easily (without encoding more rules about what is valid and thus complicating the install) since it can't distinguish a value written by the user from a default written by the product.

accepted/2021/install-location-per-architecture.md
```diff
@@ -204,17 +230,28 @@ for .NET 3.1 (which will read the non-architecture-specific variable only).
...
-one doesn't exist.
+Apphosts would have to contain a parser for this format and implement additional
+logic to handle duplicate entries and such.
 
-The effect of this is almost identical to the proposed solution.
-The only downside is that this doesn't follow to "native first" approach
-where the native architecture gets the existing "nice" names. It also adds more
-files into `/etc/dotnet` which doesn't feel necessary.
+But more importantly installers would have to be able to read and edit this file
+which is rather challenging given the currently used installer technologies
```
ericstj, line 253: I wouldn't call this a limitation of the installer technology.  All installer technologies have primitives they support (files, registry keys, environment, etc).  When you design a product with installation state that doesn't easily map to one of those primitives it requires a custom action.  Any custom action is difficult, the complexity of the custom action makes it more difficult.  Adding ambiguous state (eg: mix of user/product) makes it more difficult.  Making it a shared component between multiple products makes it more difficult, especially when those products have a desired state that is different than just exists.

The general goal for installers it to get them as simple as possible.  Make it so that they are unlikely to fail and deal in very simple primitives.

accepted/2021/install-location-per-architecture.md
```diff
@@ -138,34 +141,56 @@ any specific knowledge of where to find the install location, it would just
...
+that they don't overwrite it with different information.*
+
+#### Support for downlevel apphost
+
+In order to support running downlevel apphost apps on multi-arch non-Windows
+architectures (currently only applies to macOS), the installer for
+the architecture which .NET supports in versions lower than .NET 6
+should also write the product location to the arch-less location file
+`install_location`. This is needed for downlevel applications to correctly
+find the product.
```
ericstj, line 174: I'm assuming we should only do this if the user hasn't already set this path.  See below discussion of product/user state.

accepted/2021/install-location-per-architecture.md
vitek-karas: So if I install just the runtime, we don't write `HKEY_LOCAL_MACHINE\SOFTWARE\WOW6432Node\dotnet\Setup\InstalledVersions\x64\InstallLocation`?

That would definitely be a problem - that is the one value which the host looks for - nothing else.

accepted/2021/install-location-per-architecture.md
vitek-karas: This comment was driven by the Windows behavior. On Windows the rule should be (I actually don't know the installer well enough to tell this) that if I already installed for example 3.1 x64 into a custom path (`D:\dotnet`), then if I try to install 5.0 x64, it should install into the same location `D:\dotnet` and not "override" the location in the registry.

Currently we don't do this on Linux/macOS - and maybe we don't want to - not sure.

I honestly always considered both the registry on Windows and this file on Linux/macOS to be a "product state". There could be some users setting this, but it should be VERY rare (unless you're on one of the special distros which wants to install into different location, but those distros probably don't use our packages by default anyway).

accepted/2021/install-location-per-architecture.md
vitek-karas: Probably - but as mentioned that should be very rare. But I agree that probably should not overwrite it.

accepted/2021/install-location-per-architecture.md
vitek-karas: We're not updating any hosts with this (it would be sort of moot, we can't really ask people to rebuild all of their osx-x64 apps to make them work on macOS arm64, so it has to work on its own without changes).

accepted/2021/install-location-per-architecture.md
ericstj: Correct, it's only written by the SDK.  If you need to move we need to do that, or create a new key.  Moving can get tricky but it's might be possible.  Was discussing here https://github.com/dotnet/installer/issues/11999

accepted/2021/install-location-per-architecture.md
vitek-karas: I tried to make it clearer.

accepted/2021/install-location-per-architecture.md
vitek-karas: I thought that issue was about the fact that the installer writes keys for architectures it doesn't install.
Should I create new issue to track this (whether we fix it in 6 or not)?

accepted/2021/install-location-per-architecture.md
ericstj: > This comment was driven by the Windows behavior. 

Yes, Windows does this, however it's not well advertised.  We definitely don't tell folks to write those keys, they change the path through arguments passed to the bundle.  It coincidentally uses the same key as input in the SDK bundle (though not the runtime).  In these cases it's preserving the path of the single machine-wide installer-managed runtime.

cc @sfoslund On mac we don't support generic retargeting and the mechanism we're using today to retarget to `x64` doesn't lend itself to that since it must have a [static absolute path](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW47).  We tried to use [relocatable bundles](https://developer.apple.com/library/archive/documentation/DeveloperTools/Reference/DistributionDefinitionRef/Chapters/Distribution_XML_Ref.html#//apple_ref/doc/uid/TP40005370-CH100-SW87) and searches but couldn't get macos to honor them: our best guess is that this is a deprecated feature because it doesn't work as documented and we couldn't find anyone successfully using it.  If we could get those to work, then the installer could implement a javascript search that read the file and returned the value, or a default (with x64 appended as appropriate).  Since we cannot, it doesn't seem plausible to "preserve" custom directory installations on mac.

> but it should be VERY rare 

I think we just need to make the call what we should do in that rare case.  We can't honor it as the single installer managed location as we do on windows.  Do we allow it to remain untouched and let folks have some custom location that's side-by-side?  Or do we overwrite it to point to the installer managed location.

accepted/2021/install-location-per-architecture.md
ericstj: I mentioned this as a solution here: 
> If I could go back in time, I'd put this whole thing inside the MSIs. Host would register the "InstallLocation" and persist it with registry search on repair/upgrade. All the other MSI's would read it and honor it. They could block if the key wasn't present. Bundle would read it only for the purposes of blocking a move on upgrade, or a architecture clash. The only property the bundle would ever pass down would be "DOTNETHOME" in the case the user's choice made it past those blocks.

> It might be possible to do all this retroactively, but I'm not sure doing it in 6.0 RC2 is the best time.

accepted/2021/install-location-per-architecture.md
vitek-karas: Thanks for the explanation. Thinking about it some more - I would probably prefer we overwrite actually. Reasoning: I just explicitly installed for example .NET 6 x64. Immediately after I try to run 6.0 x64 app, but it doesn't work because I had my own config and the installer effectively silently didn't update it. That feels like a bad experience. We just position the config files as "maintained by the installer" and so if somebody modifies them, a new installation overwriting makes sense.

@richlander for another opinion.

accepted/2021/install-location-per-architecture.md
ericstj: Current thinking is to overwrite

accepted/2021/install-location-per-architecture.md
vitek-karas: I changed the text to explicitly state that we overwrite and why.

accepted/2021/install-location-per-architecture.md
vitek-karas: Added a note that installers will overwrite.

accepted/2021/install-location-per-architecture.md
ericstj: Per offline discussion we need to try to fix the MSI installers here.  I'll let you decide if you want to change the doc here at all.

richlander (approved):

PR #249, opened 2021-09-23
Update x64 emulation plan

richlander:


accepted/2021/x64-emulation-on-arm64/x64-emulation.md
```diff
@@ -245,3 +245,34 @@ Actual Plan:
...
+- .NET 6+ will default Arm64
+
+Windows:
+
+- .NET Core 3.1 will default to x64
+- .NET 5+ will default to Arm64
+
+What is meant by "default to x64?"
+
+Assume a project that target `netcoreapp3.1`. `dotnet run` should generate an x64 apphost that attempts to discover a compatible runtime (like .NET Core 3.1) in the x64 installation location. This same model applies to other commands like `dotnet test` and `dotnet tool install`.
```
yaakov-h, line 278:
```suggestion
Assume a project that targets `netcoreapp3.1`. `dotnet run` should generate an x64 apphost that attempts to discover a compatible runtime (like .NET Core 3.1) in the x64 installation location. This same model applies to other commands like `dotnet test` and `dotnet tool install`.
```

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
```diff
@@ -245,3 +245,34 @@ Actual Plan:
...
 
 Note: These plans assume that Visual Studio (Dev17) will not be supported on Windows Arm64 (with x64 emulation) before .NET 7. They also assume that Visual Studio for Mac will adapt to this plan, aligned with .NET 6. If those are not true, then we need to re-assess this plan.
+
+## Update
+
+Update: September 23, 2021
+
+The following is what we've decided to support, in terms of x64 emulation and native Arm64 architecture builds and their necessary co-existence.
+
+Supported products:
```
ericstj, line 255:
```suggestion
Supported products for installation on ARM64 Windows and macOS:
```

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
```diff
@@ -245,3 +245,34 @@ Actual Plan:
...
+
+- .NET Core 3.1 x64 runtimes
+- .NET 5 x64 runtimes
+- .NET 5 Arm64 runtimes (Windows only)
+- .NET 6+ x64 and Arm64 runtimes
+- .NET 6+ x64 and Arm64 SDKs
+
```
ericstj, line 262: It's important to note that the 3.1 x64, 5.0 x64, and 5.0 ARM64 SDKs will not block installation on ARM64 but they are not present in this list.  This means that they are all "unsupported" in their latest serviced version and folks should avoid installing them.

It's also important to note that versions of all these products prior to fixes will not be supported and should be uninstalled manually.

richlander: Excellent points.

richlander: https://github.com/dotnet/designs/pull/249/commits/28cf293b4a6d6c628ab5932382cb05eab4b2a91c

accepted/2021/x64-emulation-on-arm64/x64-emulation.md
```diff
@@ -245,3 +245,40 @@ Actual Plan:
...
+Supported products for installation on ARM64 Windows and macOS:
+
+- .NET Core 3.1 x64 runtimes
+- .NET 5 x64 runtimes
+- .NET 5 Arm64 runtimes (Windows only)
+- .NET 6+ x64 and Arm64 runtimes
+- .NET 6+ x64 and Arm64 SDKs
+
+Note: Pre-.NET 6 x64 SDKS will be installable on macOS and Windows x64 but are not supported. They have not been updated in terms of this plan and will break Arm64 installs. If they are installed on a machine, all .NET versions will need to be uninstalled/removed and you'll need to start again with installing supported .NET SDK versions.
```
ericstj, line 263: You mentioned x64 SDKs.  What about the .NET 5.0 ARM64 SDK?  Is that OK for folks to use?  Sorry to nit-pick here, but it makes a real difference in what work we do in servicing.  Also if we do decide not to support .NET 5 ARM64 SDK at all we should probably stop publishing it since it won't be supported anywhere.

richlander: Not a nit-pick. You are right.

Let's cut this one in the middle, for .NET 5. Proposal:

- Support .NET 5 x64 runtimes.
- Do not support .NET 5 SDKs (either x64 or Arm64).

That enables all scenarios and limits our cost. Good?

vitek-karas: I think it makes sense to not support 5.0 ARM64 SDK Side-by-Side with 6.0 ARM64 SDK. But I think we should still publish at least the archive version (zip), so that it can be installed via `dotnet-install.ps1` or similar.

vitek-karas (approved):

PR #251, opened 2021-10-14
Clarify inbox source generators' effect on source-build

dagood:
* Split existing paragraph to emphasize example's subject.
* Add paragraph to highlight that users aren't affected and contrast against .NET component.
* Fix "Deterministic failure" link.

I think that with the right context in mind, the existing text is clear. But, I read this doc a few times with the wrong frame of reference and I got it very wrong! I'm trying some changes that I *think* would have made it more obvious to me. (But it's hard to be sure, of course.)


ericstj (approved):
Thanks for this clarification

PR #261, opened 2022-04-12
Initial draft for a capability-based analyzer

terrajobst:


accepted/2022/capability-analyzer/capability-analyzer.md
```diff
@@ -0,0 +1,205 @@
...
+    {
+        public RequiresDynamicCodeAttribute();
+    }
+
+    [RequiresDynamicCode]
+    public partial class DynamicMethod : MethodInfo
+    {
+    }
+
+    [RequiresDynamicCode]
```
jkotas, line 193: We typically only mark the constructor or factory to limit number of places that need to be annotated.

accepted/2022/capability-analyzer/capability-analyzer.md
```diff
@@ -0,0 +1,205 @@
...
+        syncContext.Post(_ => ValidationComplete(result));
+    });
+    
+    thread.Start();
+}
+```
+
+She immediately gets a diagnostic:
+
+    CAXXXX: The API 'Thread.Start()' requires the capability 'ThreadCreation'. Either mark the consumer with `[RequiresThreadCreation]` or check by calling 'Thread.IsCreationSupported'.
```
jkotas, line 52: Are there other APIs except `Thread.Start` that would be marked as requiring the thread create capability?

accepted/2022/capability-analyzer/capability-analyzer.md
```diff
@@ -0,0 +1,205 @@
...
+    {
+    }
+
+    // Would also mark
+    // - Other *Builder types
+    // - AppDomain.DefineDynamicAssembly
+    // - ILGenerator
+}
+```
+
```
jkotas, line 204: We also have `RequiresAssemblyFilesAttribute` and `RequiresUnreferencedCodeAttribute` and associates analyzers. Are these attributes in scope or out of scope for this proposal?

accepted/2022/capability-analyzer/capability-analyzer.md
```diff
@@ -0,0 +1,205 @@
...
+    public sealed class CapabilityCheckAttribute : Attribute
+    {
+        public CapabilityCheckAttribute(Type type, string propertyName);
+        public Type Type { get; }
+        public string PropertyName { get; }
+    }
+
+    [AttributeUsage(AttributeTargets.Method |
+                    AttributeTargets.Property |
+                    AttributeTargets.Field,
```
marek-safar, line 107: Do you have an example of which capability would be controlled by field?

accepted/2022/capability-analyzer/capability-analyzer.md
jkotas: I see that `Task.WaitAll` is marked as supported on browser, but `Task.Wait` is not. Why is that?

I would expect `Task.WaitAll` to be just a more efficient version of `foreach (Task t in tasks) t.Wait()`.

Are both `Task.Wait` and `Task.WaitAll` going to be marked with this capability?

marek-safar: I believe the idea behind is that if you are in a single-threaded setup Task.WaitAll will never return as you wait on more than 1 concurrent thing (yeah there are some exceptions). Therefore if you are calling it you are doing something wrong.

/cc @lewing

stephentoub: > I believe the idea behind is that if you are in a single-threaded setup Task.WaitAll will never return as you wait on more than 1 concurrent thing (yeah there are some exceptions).

That should not be the case.  As Jan suggested, behaviorally there shouldn't be a difference between using Task.WaitAll on multiple things and using Task.Wait on each of those multiple things.

Further, both WaitAll and Wait can "inline" the thing they're waiting on if it's backed by a delegate to be executed (e.g. Task.Run(action)), so if you are in a single-threaded environment and you're waiting on such tasks, they should both "just work".  If they don't in such cases, that suggests we're missing handling of this case in browser.

lewing: I'm surprised WaitAll isn't marked there are definitely certain circumstances where it can block.  I suspect it got overlooked.

> 
> Further, both WaitAll and Wait can "inline" the thing they're waiting on if it's backed by a delegate to be executed (e.g. Task.Run(action)), so if you are in a single-threaded environment and you're waiting on such tasks, they should both "just work". If they don't in such cases, that suggests we're missing handling of this case in browser.

In the browser we have tasks (e.g. any JS interop involving a Promise) that cannot complete without yielding execution back to the browser main loop, so they won't "just work".   Wait and WaitAll _can_ work under many circumstances but  in practice they can and do break in confusing ways.

stephentoub: > so they won't "just work"

Right, that's the category I mentioned where they're not backed by delegates.

Both WaitAll and Wait would have the same issue there; WaitAll isn't special in that regard.

accepted/2022/capability-analyzer/capability-analyzer.md
terrajobst: We have done this for `[UnsupportedOSPlatformGuard]`, for [example here](https://source.dot.net/#System.Runtime.Caching/System/Runtime/Caching/MemoryCache.cs,44). The intent is to cache the result of the call.

accepted/2022/capability-analyzer/capability-analyzer.md
terrajobst: Depends on whether what they track can be mapped to the capability analyzer. The goal wasn't to replace the existing linker analyzers as much as it was to replace the `[UnsupportedOSPlatform("browser")]` annotations with something more scoped/future proof. I merely added the dynamic code capability because it was a recent example of a capability-based API that this design *should* be able to accommodate as well.

agocke:
FYI, we're doing a lot of this with `RequiresUnreferencedCode`, `RequiresAssemblyFiles`, and `RequiresDynamicCode` for the runtime. I love this direction and I think it would be great if we had a system like this for generalizing the concept of capability semantics and defining the APIs necessary for recognizing the patterns.

I haven't read through the proposal yet, but I hope the semantics would be compatible with what we've come up with for the above attributes.

vitek-karas:
Is there any connection to build properties?
If I'm building an app for net7.0-windows threading is always available and so I should not need to add any attributes or runtime checks to my app code. As described though, the analyzer would still warn.
There should be some connection to some MSBuild property which should enable/disable specific capability checks.

accepted/2022/capability-analyzer/capability-analyzer.md
```diff
@@ -0,0 +1,205 @@
...
+        public RequiresThreadCreationAttribute();
+    }
+
+    public partial class Thread
+    {
+        public static bool IsCreationSupported { get; }
+
+        [RequiresThreadCreation]
+        public void Start();
+    }
```
buyaa-n, line 163: This example makes sense to me, @terrajobst could you add example of `CapabilityGuardAttribute` usage?

PR #265, opened 2022-05-18
Proposal on release labels

richlander:
The "Current" label is confusing. This proposal aims to change that.

[Rendered view](https://github.com/dotnet/designs/blob/release-labels/accepted/2022/dotnet-release-labels.md)


danmoseley (approved):
Thanks for driving to consensus. This seems like a good plan to me. (I’m not a decision maker here of course)

monoman:
Very nice proposal!

My only remark is:
_Does the "Latest" label really need to exist? As that normally is implied as being the highest-numbered version that is not in Preview (although the go-live phase can mess a bit with this notion)._

richlander:
> Does the "Latest" label really need to exist? 

It doesn't need to. It's a question of whether folks want it.

SteveDesmond-ca:
I like "latest", it matches the general naming pattern for Docker image tags. And if everything else is labeled except "latest", what would an unlabeled release mean? Not immediately clear, so probably better to be explicit, IMO.

monoman:
Tweaking again the start of your examples with that latest thought of two properties:
```
- .NET Core 3.1 [LTS](Active)
- .NET 5 [STS](EOL)
- .NET 6 [LTS](Active)
- .NET 7 [STS](Preview)

Just before .NET 7 GA (October, 2022):

- .NET Core 3.1 [LTS](Maintenance)
- .NET 5 [STS](EOL)
- .NET 6 [LTS](Active)
- .NET 7 [STS](Go-Live; Preview)

At .NET 7 GA (November 2022):

- .NET Core 3.1 [LTS](Maintenance)
- .NET 5 [STS](EOL)
- .NET 6 [LTS](Active)
- .NET 7 [STS](Active)
```

monoman:
Reading the end of my last comment, I see that Latest != Active, as there are possibly more than one Active version at a given point in time, because of the overlaps on cycles. So OK to have Latest, but I still would like to see more explicitness for the kind of cycle. Anyway that is just me thinking out loud...

pebezo:
What if instead of STS/LTS you'd use "Supported until Nov 2022" for STS or "Supported until Oct 2024" for LTS, and instead of EOL it would turn into "Out of support" (or "No longer supported")?

richlander:
> What if instead of STS/LTS you'd use "Supported until Nov 2022" 

All other projects use labels. That's what most users expect. We should have consistent labels, too.

poke:
Since you have included that screenshot from Visual Studio, I would say that actually including the support time frame there would be very helpful. Just from knowing “Current” or “Long-term support”, I don't actually know how long something is supported. 

E.g. last year, shortly before 6.0 was released, I would have suggested to pick the non-LTS 5.0 release over the 3.1 since both would be running out of support soon-ish and migrating to 6.0 would be more straightforward from 5.0. But just judging by LTS vs non-LTS, this decision would be different.

An LTS that is soon running out of support isn't really worth more than a non-LTS version. So maybe, instead of giving each version a fixed label, it should get a label depending on how long the support is still going on.

cheenamalhotra:
Great proposal!

Maybe it would be worth considering rolling out the new labels and releasing cadence changes to other dotnet repos/products too for a consistent community experience! 🙂

dotMorten:
Overall great, love the confusion of current being gone. However regarding this :

> In some views, we can only show one label. In those cases, we will show the most relevant label, which is the first one shown in each example.

Some affected examples are:
 -Maintenance, LTS
- LTS, Latest
- Go-Live; Preview

Is the first label here the most important? I'd rather know something is latest than LTS, or Preview rather than go-live, but I'm finding myself really wanting to know both. Why not consider all the labels "a single label" so this limitation isn't an issue? Now and at ".NET 8 GA" you'll actually have two LTS labels.

sabitkondakci:
> I like "latest", it matches the general naming pattern for Docker image tags. And if everything else is labeled except "latest", what would an unlabeled release mean? Not immediately clear, so probably better to be explicit, IMO.

"latest" would inline well but I am more into "hot" label.

stevejgordon:
I'm definitely a fan of clarifying the labelling. While those of us on the bleeding edge are usually familiar with the `LTS` vs. `Current` distinction; most developers I've spoken to find them confusing, specifically `Current`. I like the proposal but would agree with @monoman that I'd like to see `STS` added. Take for example this comment...

>- .NET 7 will show a blank status after .NET 7 GA (but it will have a "Maintenance" status after .NET 8 GA)

A blank label could end up being as confusing as `Current` for the uninitiated. This release exists, but what status does it have? If it's not in `Maintenance` or `EOL` I think some could fairly assume it's supported and may not realise it is, but for a shorter period. I'd prefer `STS` or `LTS` against every release, with the other labels used as proposed. This makes the support model explicit post-release, moving to `Maintenance` nearer to `EOL`. I feel that labelling something with `STS` would help ensure most developers should feel compelled to understand what it means and its implications. `Current` is a problem because it can easily be assumed to mean something else, rather than a support lifecycle. No label, feels like it risks the same possibility.

While slightly out of scope for this proposal, I also agree with @pebezo and @poke that having STS and LTS expanded to their full name + adding the EOL date where practical (inside VS) would be useful. I've heard feedback from developers who misunderstood the support lifecycle and ended up on a `Current` release without being aware of the support trade-off. They then felt cornered into an earlier upgrade than planned to continue on a supported release.

SteveDesmond-ca:
Based on @monoman 's first comment, it seems clearer that we're trying to answer 2 different questions: what type of release is it, and what is the current support status? Maybe keeping these separate can help? Though if there are places where we can only display one, what's more important?

IEvangelist (approved):
I love this idea, and I believe it really helps to address the confusion.

IEvangelist:
@stevejgordon,

> While slightly out of scope for this proposal, I also agree with @pebezo and @poke that having STS and LTS expanded to their full name + adding the EOL date where practical (inside VS) would be useful. I've heard feedback from developers who misunderstood the support lifecycle and ended up on a Current release without being aware of the support trade-off. They then felt cornered into an earlier upgrade than planned to continue on a supported release.

+1 from me on this aspect of the proposal. Having the date in the EOL date on the label would save so much time, and I agree that spelling things out adds clarity.

monoman:
Perhaps adding a secondary label like TILL2311, TILL2411, ...

Rafael Teixeira
O..:.)oooo

joeloff:
Would this be considered a breaking change for releases.json data we publish since there are only one label per release?

terrajobst:
> Would this be considered a breaking change for releases.json data we publish since there are only one label per release?

I'd try to make this an additive schema change, i.e. have the existing property return the most significant label and add another property (both in JSON and the API) that returns all of them.

richlander:
I addressed the bulk of the remaining feedback. I think/hope most people will be happy with the updated text.

I realized -- based on the feedback -- that a few of my ideas were definitely wrong. Thanks for the help!

accepted/2022/dotnet-release-labels.md
```diff
@@ -153,16 +156,16 @@ A given presentation (like Visual Studio) can print long-form or alternate versi
 
 ## Affect on `releases.json`
```
joeloff, line 157: Should this be Effect instead of Affect?

accepted/2022/dotnet-release-labels.md
```diff
@@ -75,7 +78,7 @@ In some views, we can only show one label. In those cases, we will show the most
   - support status is (Maintenance or  EOL)
 - else show release type
 
-Note: "latest" will never be shown as a single most relevant label. Doing so might mislead users.
+For example, the `support-phase` propery in `releases-index.json` and `releases.json` will apply this algorithm.
```
joeloff, line 81: typo, "property" instead of "propery"

accepted/2022/dotnet-release-labels.md
richlander: Geez. That English degree was a waste!

```suggestion
## Effect on `releases.json`
```

accepted/2022/dotnet-release-labels.md
richlander:
```suggestion
For example, the `support-phase` property in `releases-index.json` and `releases.json` will apply this algorithm.
```

accepted/2022/dotnet-release-labels.md
```diff
@@ -0,0 +1,245 @@
...
+**Release-type (support duration) labels:**
+
+- **STS** -- Indicates a release is supported for the Short Term Support (STS) timeframe (18 months) and currently actively supported.
+- **LTS** -- Indicates a release is supported for the Long Term Support (STS) timeframe (3 years) and currently actively supported.
+
+**Release support status labels:**
+
+- **Preview** -- Indicates a release is in preview and (unless "Go-Live") is unsupported.
+- **Go-Live** -- Indicates a preview or release candidate is supported in production.
+- **Maintenance** -- Indicates that a release is within 6 months of remaining support, with security fixes only.
```
SIkebe, line 59: >with security fixes only

Does this mean "no bug fix" ?

accepted/2022/dotnet-release-labels.md
```diff
@@ -0,0 +1,245 @@
...
+The industry examples demonstrate that we are using the "LTS" label correctly and the "Current" label incorrectly. The Ubuntu/Canonical model is the closest to our practice.
+
+## Proposed labels
+
+New labels are proposed for a variety of purposes.
+
+**Release-type (support duration) labels:**
+
+- **STS** -- Indicates a release is supported for the Short Term Support (STS) timeframe (18 months) and currently actively supported.
+- **LTS** -- Indicates a release is supported for the Long Term Support (STS) timeframe (3 years) and currently actively supported.
```
danmoseley, line 53:
```suggestion
- **LTS** -- Indicates a release is supported for the Long Term Support (LTS) timeframe (3 years) and currently actively supported.
```

accepted/2022/dotnet-release-labels.md
richlander: It means security fixes only and only the most severe of non-security fixes. This has been our policy the whole time. This isn't a change in policy. Also, I think no one has ever complained about this policy. By the time we're at the very end of support, users value stability very highly. This policy delivers that.

SIkebe: Thank you for clarifying!

huoyaoyuan:
> The time-based label idea is interesting. Perhaps we can consider that as a virtual label for the website. In fact, we still need to consider the website changes. That's not covered in this doc.

I'd like it to be a column on the website, like now. Spelling a date like that isn't really readable.
image
The only complaint about the web site is dates aren't displayed in the same format (the left is correct for Chinese).

huoyaoyuan:
Maybe unrelated to this proposal, but we should clarify the term "support" too. For EOL releases, we should distinguish "runs as is" and "we will provide support".

https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-on-linux?view=powershell-7.2
I like the way PowerShell is doing. It helps tracking the version supporting history.

richlander:
I don't get the sense that anyone is confused by what support means. It hasn't come up before. Also, we don't remove links from our [release notes](https://github.com/dotnet/core/blob/main/releases.md) or [website](https://dotnet.microsoft.com/download/dotnet). This is very similar to what other projects do. We do remove old versions from docs, however. It's interesting that you referenced that PowerShell page. I just finished cleaning up our analogous page: https://github.com/dotnet/docs/pull/29462. The purpose of docs is to provide information that is relevant today not about (for example) workarounds required for .NET Core 2.2. Release notes are for historical information.

The website table isn't going to change much. 

image

We actually just discussed this a few days days ago. After the changes, "Preview" for .NET 7 would move to the "Support Phase" column and "STS" would replace it in the "Release Type" column. All releases would have an EOL date, since those are known ahead of time. No other changes.

Oh thats bad on the date format. Oops. Can you fix that @mairaw?

richlander:
I plan to merge this on Monday.

mairaw:
I opened a bug to track this. I'll talk to @danzhu54 this week about it. Thanks for reporting the issue @huoyaoyuan.

mairaw:
We merged the fix earlier today: https://dotnet.microsoft.com/zh-cn/download/dotnet

Let us know if this looks good now @huoyaoyuan!

huoyaoyuan:
@mairaw It looks good now. Thanks!

accepted/2022/dotnet-release-labels.md
```diff
@@ -0,0 +1,245 @@
...
+- .NET 5 (EOL)
+- .NET 6 (LTS, Latest)
+- .NET 7 (Preview, STS)
+
+Just before .NET 7 GA (October, 2022):
+
+- .NET Core 3.1 (Maintenance; LTS)
+- .NET 5 (EOL)
+- .NET 6 (LTS, Latest)
+- .NET 7 (Go-Live, Preview, STS)
```
mthemark, line 107: seems this labelling is confusing to me especially in this example...
Go-Live - _supported in production_
Preview - _not supported in production, but because Go-Live supported in production_
STS - _(18 months) and currently actively supported_

I'm more along the lines of only having one current label...

**Odd** year releases
Preview --> Go-Live --> **LTS** --> Maintenance -->EOL
**Even** year releases
Preview --> Go-Live --> **STS** --> Maintenance -->EOL

maybe STS-Preview/LTS-Preview and STS-Go-Live/LTS-Go-Live would be slightly useful but STS-Maintenance/LTS-Maintenance (both mean security only) or _STS-_EOL/_LTS-_EOL (both mean dead) does not really add any value.

IMHO I dislike Latest/Current/TILLyyyyMM from how much confusion was caused by the docker tag "latest"

richlander: There's really no value to having a single label. Also, compound labels are to avoided since they are objectively error prone and have bad performance characteristics. They require significant string handling beyond equality checks.

PR #271, opened 2022-08-10
Proposal on non-root container images

richlander:


accepted/2022/non-root-container-images.md
```diff
@@ -0,0 +1,162 @@
...
+```bash
+rich@kamloops:~/aspnetapp$ docker run --rm -d -p 8088:8080 -u app -e ASPNETCORE_URLS=http://+:8080 aspnetappwithuser
+cf0f115399059600fd933697dcb64cbf255e459b8a0e2d2b44b53aa71f2029de
+rich@kamloops:~/aspnetapp$ docker exec cf0f115399059600fd933697dcb64cbf255e459b8a0e2d2b44b53aa71f2029de whoami
+app
+```
+
+This little scenario also demonstrates why removing port `80` from the rootful images would be useful. We wouldn't have to redefine  `ASPNETCORE_URLS` in order to switch between `root` and `app` for images that are rootful by default. That would be very nice.
+
+For clarity, the approach that was used to add the user to the `aspnetapp` image was just a proof-of-concept. The intent is to add this user in the `runtime-deps` images (or `runtime` for Windows).
```
mthalman, line 162: This is the only mention of Windows in this doc. I think it deserves more clarity. Windows containers already have "root" and "non-root" users baked in. All the Windows container SKUs contain a ContainerAdministrator and ContainerUser user. By default, ContainerAdministrator is the default, except for Nano Server which sets ContainerUser as the default.

```
C:\>docker run --rm mcr.microsoft.com/windows/nanoserver:ltsc2022 cmd /c echo %"USERNAME%
ContainerUser

C:\>docker run --rm mcr.microsoft.com/windows/servercore:ltsc2022 cmd /c echo %"USERNAME%
ContainerAdministrator

C:\>docker run --rm mcr.microsoft.com/windows/server:ltsc2022 cmd /c echo %"USERNAME%
ContainerAdministrator
```

In addition, port 80 isn't restricted from being used when running as ContainerUser.

I think this deserved to be called out because the Linux/Windows environments aren't symmetrical in this aspect.

richlander:
https://github.com/dotnet/aspnetcore/pull/44194

accepted/2022/non-root-container-images.md
```diff
@@ -0,0 +1,162 @@
...
+We can satisfy those goals (not necessarily in that order) across multiple releases.
+
+- .NET 7
+  - Add same non-root user in both rootful and non-root images
+  - Expose same non-root ports -- `8080` and `8443` (for HTTPS) -- in both rootful and non-root images
+  - Continue to expose port `80` in rootful images, for compatibility.
+  - Transition `mcr.microsoft.com/dotnet/samples` to non-root images (which by definition means not exposing port `80`).
+- .NET 8
+  - Remove port `80` from rootful images.
+  - Announce that all .NET images are turnkey non-root capable.
```
mthalman, line 25: @richlander - Since none of these changes happened for .NET 7, what's the plan for .NET 8. Do we want to continue a phased approach between .NET 8 and 9 or do everything in .NET 8?

richlander: If we make these changes for Preview 1 or 2, then I think we can do everything at once and wait for feedback and then assess based on that.

mthalman: cc @lbussell

PR #272, opened 2022-09-20
Add a proposal which suggests updating the xarch baseline target

tannergooding:
As per the doc, I propose the minimum required hardware for x86/x64 on .NET should be changed from `x86-x64-v1` to `x86-x64-v2`.

### DIFF
--- a/proposed/xarch-baseline.md
+++ b/proposed/xarch-baseline.md
@@ -1,0 +1,59 @@
+# Revise the baseline machine for x86/x64 Hardware
+
+**Owner** [Tanner Gooding](https://github.com/tannergooding)
+
+## Proposal
+
+The minimum required hardware for x86/x64 on .NET should be changed from `x86-x64-v1` to `x86-x64-v2`
+
+## Historical Context
+
+x86 hardware is well-known for its backwards compatiblity and has been incrementally versioned on the same baseline since the introduction of the Intel 8086 in 1978 with its 16-bit instruction set. When the Intel 80386 was released in 1985, it was a direct extension of the original chips and you had to explicitly transition out of 16-bit mode and into 32-bit mode. The same was true for when the AMD Athlon was released in 2003 and exposed support for the 64-bit instruction set. In between these "key" milestones, various optional instruction sets have been added, most notably of which are typically the SIMD instruction sets starting with SSE in 1999 and SSE2 in 2000.
+
+With the introduction of the 64-bit instruction set, the minimum baseline changed from simply "base instructions" to also include `CMOV`, `CMPXCHG8B`, `SSE`, and `SSE2` (among other less important ones). The 3 major compilers (MSVC, Clang, and GCC) have also, over time, migrated their 32-bit baselines to match this and thus all native x86/x64 apps compiled today will target this "common demoninator". However, since the introduction of the 64-bit instruction set we have not had a major architectural revision that has forced apps to recompile and thus no newer baseline has been defined in 19 years.
+
+That being said, there have been newer "versions" of the 64-bit ISA defined. That is, much as ARM64 has defined `armv8.0-a`, `armv8.1-a`, ..., `armv8.6-a`, etc; x64 has defined `x86-x64`, `x86-x64-v2`, ..., `x86-x64-v3`. These versions are defined as part of the `x86-x64-ABI` for Linux and were created as a joint effort between `AMD`, `Inte`l, `Red Hat`, and `SUSE`. They are as follows (noting dates are approximate as "first introduced" to when both AMD and Intel provided hardware with the required support):
+* `x86-x64-v2` (~2007-2009+): `CMPXCHG16B`, `LAHF/SAHF`, `SSE4.2`, `POPCNT`
+* `x86-x64-v3` (~2013-2015+): `AVX2`, `BMI1`, `BMI2`, `F16C`, `FMA`, `LZCNT`, `MOVBE`
+* `x86-x64-v4` (~2017-????+): `AVX512F`, `AVX512BW`, `AVX512CD`, `AVX512DQ`, `AVX512VL`
+
+## .NET's instruction usage
+
+Since SIMD support was first introduced for .NET in 2014 customers have added varying levels of dependence on the functionality to accelerate core algorithms. With these types being moved in box in .NET Core 2.1 and the introduction of platform specific hardware intrinsics in .NET Core 3.0, the usage of these instructions to accelerate core BCL algorithms has become increasingly more prevalent and has been the backing factor for many of the performance wins each release. With the introduction of Crossgen 2, we have likewise updated the default R2R baseline to be a semi-custom set of instructions that are compatible with the base instruction encoding mechanism; this notably includes SSE4.1. However, .NET continues targeting `x86-x64-v1` as its minimum baseline (for 64-bit and an equivalent set for 32-bit).
+
+This has normally not been a large consideration since .NET has typically been expected to run in an environment where a JIT exists and therefore where code could be "re-jitted" for the current hardware, thus allowing it to take advantage of newer instruction sets, especially for known hot code. However, with the introduction of Native AOT we now have a higher consideration for scenarios where recompilation is not possible and therefore whatever the precompiled code targets is what all it has access to. There are some ways around this such as dynamic ISA checks or compiling a given method twice with dynamic dispatch selecting the appropriate implementation, but this comes with various downsides and often requires restructuring code in a way that can make it less maintainable.
+
+Likewise, SSE2 only provides a small subset of functionality primarily targeting `float`, `double`, and a handful of basic integer operations. Many of the instructions required to "complete" the integer support are provided in newer ISAs such as `SSE3`, `SSSE3`, `SSE4.1`, and `SSE4.2`. This is in contrast to a newer platform like `armv8.0-a` which was introduced in 2011 and therefore was able to provide more functionality as part of their baseline. What this means is that various SIMD acceleration techniques can easily become pessimized on the `x86-x64-v1` baseline because core instructions aren't available and alternatives must be substituted instead resulting in harder to maintain code.
+
+## Modern Hardware/Software Expectations
+
+Windows 8.1+ requires `CMPXCHG16B` and `LAHF/SAHF` for 64-bit and due to these being newer instruction sets, they imply a post `x86-64-v1` (but potentially pre `x86-x64-v2`) dependency. Windows Server 2008 R2+ and Windows 11+ both only support 64-bit processors. Windows 11 additionally only supports processors that are newer than Intel Coffee Lake or AMD Zen+. Windows versions prior to 8.1 are out of support for normal scenarios. Windows 7 remains in support only with a special license but was itself released in 2009 and only officially supported hardware from around the same timeframe. Notably the Operating System may run on other hardware, but doing so is technically unsupported.
+
+macOS is on an approximate 3 year support cycle and owns their entire supply chain. macOS 11 is currently the oldest supported version and only supports hardware dating back to ~2013 and therefore which provides `x86-x64-v3` support. One would have to go back to before `macOS 10.12` (that is to `OS X 10.11`) to find hardware that doesn't meet the `x86-x64-v2` baseline.
+
+Linux is an interesting scenario where they continue supporting older hardware by design and indeed the kernel purports to continue supporting the `80386` and therefore a pre `x86-x64-v1` baseline.
+
+For the cloud market, Azure, AWS, and Google Cloud all provide `x86-x64-v3` compatible or newer hardware as a minimum (to the best of my knowledge, based on publicly exposed hardware information).
+
+For customers reporting to `Steam` on their hardware data shows the following:
+* `SSE3` - 100%
+* `LAHF/SAHF` - 99.99%
+* `CMPXCHG16B` - 99.99%
+* `SSSE3` - 99.45%
+* `SSE4.1` - 99.15%
+* `SSE4.2` - 98.83%
+* `AVX` - 94.89%
+* `AVX2` - 88.15%
+* `AVX512F` - 8.3%
+
+From the above, we can see that only 1.17% users don't meet the `x86-x64-v2` baseline and by comparison only 11.85% of users don't meet the `x86-x64-v3` baseline. We can presume that not all users are reporting and that some of them may be on hardware that is more than 13-15 years old (or which otherwise doesn't meet the `x86-x64-v2` baseline). However, the expectation is that such hardware is rapidly decreasing in prevalence and that such users will have many more complications with trying to run modern versions of .NET or Operating Systems.
+
+## Benefits and Drawbacks
+
+The main drawback of updating the baseline will be that older hardware will no longer be able to target new versions of .NET and thus may be left behind if they have hardware without support for the `x86-x64-v2` instructions. Such users are already likely in a pessimized scenario however as this it not hardware we currently test against nor which we profile/benchmark for and thus it is increasingly likely that  perf or other regressions are introduced.
+
+However, there are many potential benefits for updating the baseline the primary of which is that it will provide an overall more consistent for users targeting AOT scenarios. Additionally, it would allow code to be simplified in the JIT as we could remove code paths designed to support such older hardware and would in turn make the product easier to maintain, version, and test.
+
+## Alternatives
+
+We could maintain the `x86-x64-v1` baseline for the JIT (optionally removing pre `v2` SIMD acceleration) while changing the default for AOT. This could emit a diagnostic by default elaborting to users that it won't support older hardware and indicate how they could explicitly retarget to `x86-64-v1` that is important for their domain.


tannergooding:
CC. @richlander, @jkotas, @MichalStrehovsky, @davidwrighton

masonwheeler:
Looks reasonable in general, though I'd like to touch on one specific point:

> However, with the introduction of Native AOT we now have a higher consideration for scenarios where recompilation is not possible and therefore whatever the precompiled code targets is what all it has access to. There are some ways around this such as dynamic ISA checks or compiling a given method twice with dynamic dispatch selecting the appropriate implementation, but this comes with various downsides and often requires restructuring code in a way that can make it less maintainable.

Why not go for the best-of-both-worlds approach?  Build and ship IL, and have AOT compilation occur on the destination machine at *installation* time, rather than JITting at runtime or AOTing at build time?  This is essentially what Android does and it works quite well there.

jkotas:
> This is essentially what Android does and it works quite well there.

This works when the runtime is part of the OS (or part of large app with complex installer) and the OS can manage the app lifecycle. 

It does not work well for runtimes that ship independently on OS, like what .NET runtime is today.

tannergooding:
>  Build and ship IL, and have AOT compilation occur on the destination machine at installation time, rather than JITting at runtime or AOTing at build time? This is essentially what Android does and it works quite well there.

One consideration is that Android owns the OS and so is able to guarantee the tools required to do that are available. They also don't support any concepts like "xcopy" of apps and centralize acquisition via the App Store.

I think doing the same for .NET would be pretty awesome, but it also comes with considerations like a larger deployment mechanism and potential other negative side effects.

Crossgen2 for a higher baseline is pretty much like this already, but without many of the drawbacks.

jkotas:
A few angles to consider:

- Microsoft official build vs. source build: We can bump the baseline for Microsoft official build, keep the code for x86 v1 around for the time being and tell anybody who really needs it to build their own bits from sources. (In other words, drop the x86 v1 support level to community supported.)

- 32-bit vs. 64-bit: We can consider keeping the baseline for 32-bit and raising it only on 64-bit.

proposed/xarch-baseline.md
```diff
@@ -0,0 +1,59 @@
...
+
+However, there are many potential benefits for updating the baseline the primary of which is that it will provide an overall more consistent for users targeting AOT scenarios. Additionally, it would allow code to be simplified in the JIT as we could remove code paths designed to support such older hardware and would in turn make the product easier to maintain, version, and test.
+
+## Alternatives
+
+We could maintain the `x86-x64-v1` baseline for the JIT (optionally removing pre `v2` SIMD acceleration) while changing the default for AOT. This could emit a diagnostic by default elaborting to users that it won't support older hardware and indicate how they could explicitly retarget to `x86-64-v1` that is important for their domain.
```
jkotas, line 59: Nit: Emitting diagnostic like this by default is just output clutter. If we were to do this, it would be just mentioned in the doc.

masonwheeler:
Sure, NativeAOT isn't built into the OS.  How much work would it be to integrate it with a standard installer-generator system like MSI, though?  It would never be The Standard, but it would at least be available for developers in the know.

jkotas:
> Sure, NativeAOT isn't built into the OS. How much work would it be to integrate it with a standard installer-generator system like MSI, though?

It depends on what your requirements are. You can certainly do some variant of it on your own.

I do not expect we (Microsoft .NET team) will provide or recommend a solution like this. It would not pass our security signoff.

masonwheeler:
Huh.  What are the security concerns here?

jkotas:
For example, the binaries cannot be signed.

masonwheeler:
Wasn't signing eliminated from .Core a few versions ago anyway?  I remember that one pretty clearly because there were breaking changes in .NET 6 that broke my compiler, and when I complained about it the team refused to make even the most inconsequential of changes to alleviate the compatibility break.

jkotas:
I am not talking about strong name signing. I am talking about Microsoft authenticode, Apple app code signing, and similar type of signatures.

masonwheeler:
All right.  So how does Android handle it?

jkotas:
I do not know the details on how Android handles this. I can tell you what it involved to make this scheme work with .NET Framework: NGen service process was recognized as a special process by the Windows OS that was allowed to vouch for authenticity of its output. It involved hardening like disallowing debugger attach to the NGen service process (again, another special service provided by the Windows OS) so that you cannot tamper with its execution.

masonwheeler:
Yeah, that makes sense.  The AOT compiler has to be in a position of high trust for a scheme like that to work.  Joe Duffy said something very similar about the Midori architecture.

proposed/xarch-baseline.md
```diff
@@ -0,0 +1,59 @@
...
+
+With the introduction of the 64-bit instruction set, the minimum baseline changed from simply "base instructions" to also include `CMOV`, `CMPXCHG8B`, `SSE`, and `SSE2` (among other less important ones). The 3 major compilers (MSVC, Clang, and GCC) have also, over time, migrated their 32-bit baselines to match this and thus all native x86/x64 apps compiled today will target this "common demoninator". However, since the introduction of the 64-bit instruction set we have not had a major architectural revision that has forced apps to recompile and thus no newer baseline has been defined in 19 years.
+
+That being said, there have been newer "versions" of the 64-bit ISA defined. That is, much as ARM64 has defined `armv8.0-a`, `armv8.1-a`, ..., `armv8.6-a`, etc; x64 has defined `x86-x64`, `x86-x64-v2`, ..., `x86-x64-v3`. These versions are defined as part of the `x86-x64-ABI` for Linux and were created as a joint effort between `AMD`, `Inte`l, `Red Hat`, and `SUSE`. They are as follows (noting dates are approximate as "first introduced" to when both AMD and Intel provided hardware with the required support):
```
gfoidl, line 15: Nit

```suggestion
That being said, there have been newer "versions" of the 64-bit ISA defined. That is, much as ARM64 has defined `armv8.0-a`, `armv8.1-a`, ..., `armv8.6-a`, etc; x64 has defined `x86-x64`, `x86-x64-v2`, ..., `x86-x64-v3`. These versions are defined as part of the `x86-x64-ABI` for Linux and were created as a joint effort between `AMD`, `Intel`, `Red Hat`, and `SUSE`. They are as follows (noting dates are approximate as "first introduced" to when both AMD and Intel provided hardware with the required support):
```

gfoidl:
Is related to https://github.com/dotnet/designs/pull/173?
Or is this one here more about AOT?

MichalStrehovsky:
Would this also cover how we compile the native parts of the (non-AOT) runtimes (GC, CoreCLR VM, etc.)?

My main concern would be about the user experience for the minority of users that don't meet this requirement - I'd like to avoid the user experience to be `STATUS_ILLEGAL_INSTRUCTION` with a crashdump. NativeAOT does a failfast with a message to stderr. It's not great. (It's obviously not visible for OutputType=WinExe, for example.)

MichalStrehovsky:
Do we have any motivating scenarios that we expect to meaningfully improve? I tried TechEmpower Json benchmark, but I'm seeing some very confusing results (`crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/platform.benchmarks.yml --scenario json --profile aspnet-citrine-lin --application.framework net7.0 --application.environmentVariables COMPlus_EnableSSE3=0` shows improved RPS with SSE3 disabled compared to the baseline which is the opposite of what I wanted/expected to measure).

EgorBo:
> shows improved RPS with SSE3 disabled compared to the baseline which is the opposite of what I wanted/expected to measure).

That's weird because SSE3 specifically doesn't bring any value (except, maybe, HADD for floats but it's unlikely to be touched in TE benchmarks). For shuffle it's SSSE3 that is interesting because it provides overloads we need.

MichalStrehovsky:
> > shows improved RPS with SSE3 disabled compared to the baseline which is the opposite of what I wanted/expected to measure).
> 
> That's weird because SSE3 specifically doesn't bring any value (except, maybe, HADD for floats but it's unlikely to be touched in TE benchmarks). For shuffle it's SSSE3 that is interesting because it provides overloads we need.

Wait does COMPlus_EnableSSE3=0 only disable SSE3? I thought it works similar to how we do detection in codeman.cpp - not detecting SSE3 means we also consider SSSE3/4/4.2/AVX etc. unavailable. Or do I need to COMPlus_EnableXXX everything one by one to get the measurement I wanted to measure?

tannergooding:
> Is related to https://github.com/dotnet/designs/pull/173?
Or is this one here more about AOT?

Its similar, but its about upgrading the baseline for AOT and therefore directly impacts all consumers of .NET

173 impacts the default for crossgen, which only provides worse startup performance on older hardware.

tannergooding:
> Would this also cover how we compile the native parts of the (non-AOT) runtimes (GC, CoreCLR VM, etc.)?

That would likely be up for debate. MSVC only provides `/arch:SSE2` and `/arch:AVX/AVX2`, there is no equivalent for `SSE3-SSE4.2`. Clang/GCC do support these intermediaries, but I don't expect them to be as big of wins for native given the typical dynamic linking and limited explicit vectorization.

> My main concern would be about the user experience for the minority of users that don't meet this requirement - I'd like to avoid the user experience to be STATUS_ILLEGAL_INSTRUCTION with a crashdump. NativeAOT does a failfast with a message to stderr. It's not great. (It's obviously not visible for OutputType=WinExe, for example.)

We have a similar message raised by the VM today as well, its just unlikely to ever be encountered since its just checking for SSE2.

tannergooding:
> Do we have any motivating scenarios that we expect to meaningfully improve? I tried TechEmpower Json benchmark, but I'm seeing some very confusing results (crank --config https://raw.githubusercontent.com/aspnet/Benchmarks/main/scenarios/platform.benchmarks.yml --scenario json --profile aspnet-citrine-lin --application.framework net7.0 --application.environmentVariables COMPlus_EnableSSE3=0 shows improved RPS with SSE3 disabled compared to the baseline which is the opposite of what I wanted/expected to measure).

Most codepaths that use `Vector128<T>`. Like @EgorBo called out there is pretty "core" functionality only available in these later ISAs.

Notably:
* SSE3 - Floating-Point `Alternating Add/Subtract`, `Horizontal Add`, `Horizontal Subtract`
* SSSE3 - Integer `Conditional Negate`, `Absolute Value`, `Bytewise Shuffle`, `Horizontal Add`, `Horizontal Subtract`
* SSE4.1 - `Dot Product` (FP only), `Blend`, `Round` (FP only), `Insert`, `Extract`, `Test`, Integer `Min/Max`
* SSE4.2 - `64-bit Compare Greater Than`
* POPCNT

For SSSE3, the most important is `Bytewise Shuffle`. Doing arbitrary vector reordering is very expensive otherwise and so this is key for doing things like `Reverse Endianness` or handling edge cases to emulate other functionality.

For SSE4.1, the most important are `Blend`, `Insert`, `Extract`, and `Test`. `Dot Product` and `Round` are both important for many scenarios including WPF and other image manipulation scenarios due to heavy use of floating-point. In the case of `Blend` it allows simplifying `(a & b) | (a & ~b)` to a single instruction, effectively giving you "vectorized ternary select". This is a key part of handling leading/trailing elements or operating only on matched data. `Insert`/`Extract` are key for getting data into and out of the vector registers efficiently, and `Test` is key for efficiently determining `(a & b) == 0` or `(a & b) != 0`, which is used for many paths to determine if any match exists and therefore whether more expensive computation has to be done.

Not having these means the codegen for many core algorithms, especially in the `string`/`span` handling can be significantly pessimized against newer hardware that a majority of customers are expected to have.

tannergooding:
> Wait does COMPlus_EnableSSE3=0 only disable SSE3? I thought it works similar to how we do detection in codeman.cpp - not detecting SSE3 means we also consider SSSE3/4/4.2/AVX etc. unavailable. Or do I need to COMPlus_EnableXXX everything one by one to get the measurement I wanted to measure?

They are hierarchical and so `EnableSSE3=0` will also disable `SSSE3`, `SSE4.1`, `SSE4.2`, `AVX`, `AVX2`, etc.

Could you provide more concrete numbers and possibly codegen? This sounds unexpected and doesn't match what I've seen in past benchmarking comparisons.

We could certainly get more concrete numbers by running all of `dotnet/performance` with `COMPlus_EnableSSE3=0` and see if anything exceptional pops out as having improved performance.

MichalStrehovsky:
I'd rather this decision wasn't made solely on microbenchmarks. I have no doubts it helps microbenchmarks. They're good supporting evidence, but something that impacts the users is better as the main evidence. That's why I'm trying TechEmpower (it's an E2E number we care about).

> Could you provide more concrete numbers and possibly codegen? This sounds unexpected and doesn't match what I've seen in past benchmarking comparisons.

I can give you what I did, but not much more than that. Hopefully it's enough to find what I'm doing wrong:

```
dotnet tool install -g Microsoft.Crank.Controller --version "0.2.0-*" 
```

And then just run the above crank command with/without the `--application.environmentVariables COMPlus_EnableSSE3=0`.

Without the EnableSSE3=0 argument:

```
| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 79         |
| Cores usage (%)        | 2,219      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 363        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 74         |
| Requests/sec           | 1,214,438  |
| Requests               | 18,336,914 |
| Mean latency (ms)      | 0.71       |
| Max latency (ms)       | 54.51      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 169.09     |
| Latency 50th (ms)      | 0.38       |
| Latency 75th (ms)      | 0.44       |
| Latency 90th (ms)      | 0.56       |
| Latency 99th (ms)      | 10.11      |
```

With the SSE3=0 argument:

```
| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 80         |
| Cores usage (%)        | 2,229      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 363        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 75         |
| Requests/sec           | 1,229,283  |
| Requests               | 18,560,989 |
| Mean latency (ms)      | 0.61       |
| Max latency (ms)       | 32.51      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 171.16     |
| Latency 50th (ms)      | 0.38       |
| Latency 75th (ms)      | 0.44       |
| Latency 90th (ms)      | 0.52       |
| Latency 99th (ms)      | 8.05       |
```

tannergooding:
> I'd rather this decision wasn't made solely on microbenchmarks. I have no doubts it helps microbenchmarks. They're good supporting evidence, but something that impacts the users is better as the main evidence.

I agree that it shouldn't be made solely based on microbenchmarks. However, we also know well how frequently span/string APIs are used and the cost of branches in hot loops, even predicted branches. We likewise know the importance of these operations in scenarios like ML, Image Processing, Games, etc. I imagine coming up with real world benchmarks showing improvements won't be difficult.

With that being said, we also really should not restrict ourselves to a 20 year old baseline regardless. Such hardware is all officially out of support and discontinued by the respective hardware manufacturers. Holding out for such a small minority of hardware that likely isn't even running on a supported OS is ultimately pretty silly. At some point, we need to have the freedom/flexibility to tell users that newer versions of .NET won't support hardware that old (at the very least "officially", Jan's suggestion of leaving the support in but making it community supported is reasonable; as would simply making it not the default).

> I can give you what I did, but not much more than that. Hopefully it's enough to find what I'm doing wrong:

Does this work on Windows, or is it Linux only? Do you also need `crank-agent` like the docs I found suggest?

On Windows, I see
```
The specified endpoint url 'http://asp-citrine-lin:5001' for 'application' is invalid or not responsive: "No such host is known. (asp-citrine-lin:5001)"
```

Likewise, how much variance is there here run-to-run?
How much of this is R2R compiled (doing `EnableSSE3=0` will throw out the corelib CG2 images/etc)?
Is this accounting for rejit/tiered compilation costs?

MichalStrehovsky:
> We likewise know the importance of these operations in scenarios like ML, Image Processing, Games, etc. I imagine coming up with real world benchmarks showing improvements won't be difficult.

I'd like us to have such E2E number - we're discussing making .NET-produced executables FailFast on 1 out of 100 machines in the wild by default - "`string.IndexOf` is a lot faster" is less convincing argument that it's the right choice than "X RPS improvement in web scenario Y", "X fps improvement in game Y", etc. That's the argument we'll give whenever someone complains about this choice (for me the important bit is that this is a requirement for where the code runs, not a requirement for the .NET developers machine - the .NET developers likely won't even know about this hardware floor until they hear from _their_ user).

> Does this work on Windows, or is it Linux only? Do you also need crank-agent like the docs I found suggest?

I run it on Windows. You need VPN on because asp-citrine-lin is a corpnet machine. AFAIK crank-agent is needed on the machine where you run the test (which is asp-citrine-lin in this case, so no need to worry about it).

> Likewise, how much variance is there here run-to-run (that is against separate attempts to profile using the same command line)?

I made 2 runs each and there was some noise but the difference in the two runs looked conclusive. I think crank does a warmup, but it's really a ASP.NET team tool that I don't have much experience with (only to the extent that we track it and it's part of our release criteria, and therefore looks relevant).

tannergooding:
> I'd like us to have such E2E number - we're discussing making .NET-produced executables FailFast on 1 out of 100 machines in the wild by default - "string.IndexOf is a lot faster" is less convincing argument that it's the right choice than "X RPS improvement in web scenario Y", "X fps improvement in game Y", etc. That's the argument we'll give whenever someone complains about this choice (for me the important bit is that this is a requirement for where the code runs, not a requirement for the .NET developers machine - the .NET developers likely won't even know about this hardware floor until they hear from their user).

I expect its much less than this in practice, especially when taking into account Enterprise/cloud hardware, the users that are likely to be running on a supported OS, and that are likely to be running/using .NET.

It's worth noting I opened https://github.com/dotnet/sdk/issues/28055 so that we can, longer term, get more definitive information on this and other important hardware characteristics.

> I run it on Windows. You need VPN on because asp-citrine-lin is a corpnet machine. AFAIK crank-agent is needed on the machine where you run the test (which is asp-citrine-lin in this case, so no need to worry about it).

👍. The below is the median of 5 results for each. I didn't notice any obvious outliers.

I notably ran both JSON and Plaintext to get two different comparisons. There is a clear disambiguator when SIMD is disabled entirely and there is a small but measurable difference between `-v1` and `-v2` with `-v2` winning. 

The default (which should be `-v3` assuming these machines have AVX2 support) tends to be a bit slower than `-v2` and this is likely because the payloads aren't large enough for `Vector256<T>` to benefit. Instead, the larger size coupled with the additional checks causes a small regression.

## Json

### .NET 7 - Default

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 66         |
| Cores usage (%)        | 1,847      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 115        |
| Requests/sec           | 985,938    |
| Requests               | 14,886,941 |
| Mean latency (ms)      | 0.42       |
| Max latency (ms)       | 51.21      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 142.92     |
| Latency 50th (ms)      | 0.24       |
| Latency 75th (ms)      | 0.27       |
| Latency 90th (ms)      | 0.32       |
| Latency 99th (ms)      | 7.16       |

### .NET 7 - EnableAVX=0 (effectively target `x86-64-v2`)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 66         |
| Cores usage (%)        | 1,847      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 113        |
| Requests/sec           | 990,885    |
| Requests               | 14,962,330 |
| Mean latency (ms)      | 0.46       |
| Max latency (ms)       | 39.10      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 143.64     |
| Latency 50th (ms)      | 0.23       |
| Latency 75th (ms)      | 0.27       |
| Latency 90th (ms)      | 0.32       |
| Latency 99th (ms)      | 8.18       |

### .NET 7 - EnableSSE3=0 (effectively target `x86-64-v1`)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 66         |
| Cores usage (%)        | 1,835      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 116        |
| Requests/sec           | 980,136    |
| Requests               | 14,799,763 |
| Mean latency (ms)      | 0.49       |
| Max latency (ms)       | 41.72      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 142.08     |
| Latency 50th (ms)      | 0.24       |
| Latency 75th (ms)      | 0.27       |
| Latency 90th (ms)      | 0.32       |
| Latency 99th (ms)      | 8.72       |

### .NET 7 - EnableHWIntrinsic=0 (effectively disable SIMD)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 64         |
| Cores usage (%)        | 1,783      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 211        |
| Requests/sec           | 944,005    |
| Requests               | 14,253,837 |
| Mean latency (ms)      | 0.50       |
| Max latency (ms)       | 48.26      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 136.84     |
| Latency 50th (ms)      | 0.25       |
| Latency 75th (ms)      | 0.29       |
| Latency 90th (ms)      | 0.34       |
| Latency 99th (ms)      | 9.27       |

## Plaintext

### .NET 7 - Default

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 44         |
| Cores usage (%)        | 1,221      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 90         |
| Requests/sec           | 4,625,118  |
| Requests               | 69,838,073 |
| Mean latency (ms)      | 0.60       |
| Max latency (ms)       | 29.17      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 582.23     |
| Latency 50th (ms)      | 0.52       |
| Latency 75th (ms)      | 0.76       |
| Latency 90th (ms)      | 1.05       |
| Latency 99th (ms)      | 0.00       |

### .NET 7 - EnableAVX=0 (effectively target `x86-64-v2`)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 44         |
| Cores usage (%)        | 1,232      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 93         |
| Requests/sec           | 4,679,347  |
| Requests               | 70,655,667 |
| Mean latency (ms)      | 0.58       |
| Max latency (ms)       | 35.71      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 589.06     |
| Latency 50th (ms)      | 0.51       |
| Latency 75th (ms)      | 0.75       |
| Latency 90th (ms)      | 1.03       |
| Latency 99th (ms)      | 0.00       |

### .NET 7 - EnableSSE3=0 (effectively target `x86-64-v1`)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 44         |
| Cores usage (%)        | 1,225      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 91         |
| Requests/sec           | 4,635,911  |
| Requests               | 69,999,632 |
| Mean latency (ms)      | 0.59       |
| Max latency (ms)       | 32.99      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 583.59     |
| Latency 50th (ms)      | 0.53       |
| Latency 75th (ms)      | 0.76       |
| Latency 90th (ms)      | 1.09       |
| Latency 99th (ms)      | 0.00       |

### .NET 7 - EnableHWIntrinsic=0 (effectively disable SIMD)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 42         |
| Cores usage (%)        | 1,178      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 358        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 158        |
| Requests/sec           | 4,370,389  |
| Requests               | 65,991,281 |
| Mean latency (ms)      | 0.63       |
| Max latency (ms)       | 32.96      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 550.17     |
| Latency 50th (ms)      | 0.55       |
| Latency 75th (ms)      | 0.80       |
| Latency 90th (ms)      | 1.10       |
| Latency 99th (ms)      | 0.00       |

masonwheeler:
I'm actually trying to do exactly this kind of real-world codebase benchmarking, to see if the .NET 7 performance benefits touted in the blog posts make a measurable difference in some performance-sensitive code.  Unfortunately, I've been stymied by [the inability to actually get anything to run in .NET 7.](https://github.com/dotnet/sdk/issues/28009)  Any help would be welcome, and I promise to report back with relevant numbers once I have some to share.

MichalStrehovsky:
> It's worth noting I opened https://github.com/dotnet/sdk/issues/28055 so that we can, longer term, get more definitive information on this and other important hardware characteristics.

I'm not sure if that one would help - it's the hardware the .NET developers use, not hardware where .NET code runs. Developers are more likely to skew towards the latest and greatest. Users are the "secretary's machine" and "school computer". Windows org is more likely to have the kind of telemetry.

> The below is the median of 5 results for each. I didn't notice any obvious outliers.

What command line arguments did you use for crank? The numbers for JSON are all a bit lower than I would expect (compare with mine above).

tannergooding:
> What command line arguments did you use for crank? The numbers for JSON are all a bit lower than I would expect (compare with mine above).

Ah, you know what I ran `json.benchmarks.yml` and `plaintext.benchmarks.yml` rather than `platform.benchmarks.yml`. That might've had something to do with it.

> I'm not sure if that one would help - it's the hardware the .NET developers use, not hardware where .NET code runs. Developers are more likely to skew towards the latest and greatest. Users are the "secretary's machine" and "school computer". Windows org is more likely to have the kind of telemetry.

Which again, doesn't really matter when you consider that most Operating Systesm don't support hardware that old.

In the case of macOS, it looks to be impossible for any OS we currently support to be running on pre-AVX2 hardware.

In the case of Windows, 8.1 is the oldest client SKU we still support. For 8.1, Windows themselves updated the baseline CPU required for x64 (must have CMPXCHG16B and LAHF/SAHF). Various articles quote a comment stating "the number of affected processors are extremely small since this instruction has been supported for greater than 10 years.". For 7, its only supported with an ESU subscription in which case other factors like the Windows Processor Requirements list comes into play and they are all post `-v3` processors. -- Even stricter requirements/expectations exist for Server

Linux is really the only interesting case where the kernel still officially supports running on an 80386 (older than we support) and where many distros intentionally keep their specs "low". This is also a case where many recommend using alternative GUIs or specialized distro-builds for such low-spec computers to help. Ubuntu's docs go so far as to describe 10 and 15 year old systems and the scenarios that will likely prevent their usage in a default configuration. The biggest of which is typically that they don't support and have no way of supporting an SSD.

In short, such hardware is simply too old to be meaningful and given our official OS support matrix, is already unlikely to have a good experience with the latest versions of .NET.

masonwheeler:
@tannergooding Agreed.  "Not supported" loses all meaning if no changes can be made based on disregarding the existence of things officially not supported.

EgorBo:
Notes about TechEmpower benchmarks:

They operate on extremely small data, e.g. here is what JSON benchmarks test: https://github.com/aspnet/Benchmarks/blob/e3095f4021fef7171bb3ae86616b9156df39b7bd/src/Benchmarks/Middleware/JsonMiddleware.cs#L51 its string representation probably doesn't even fit into AVX vector. And here is the Plaintext - https://github.com/aspnet/Benchmarks/blob/e3095f4021fef7171bb3ae86616b9156df39b7bd/src/Benchmarks/Middleware/PlaintextMiddleware.cs#L16-L41 even smaller. It's nowhere near to be called a "real" workload. I mean TE benchmarks are great to spot obvious regressions and they already helped us a lot to spot problems like in GC Regions, Crossgen, etc, measure internal aspnet overhead and threadpool scaling, but definitely not something we can use to make decisions around vector width IMO.

Same regarding HTTP headers, in TechEmpower benchmarks they're:
```
Accept: application/json,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7
Host: 10.0.0.102:5000
Connection: keep-alive
```
UTF8 so only Accept is "probably" worth using AVX for.

While normally your browser sends like 16 or more headers.

tannergooding:
> They operate on extremely small data, e.g. here is what JSON benchmarks test

It's worth noting its not just about the size of the data, its primarily about the amount of data required to be processed until the algorithm can exit.

Even for very large inputs, if you're just finding the first index of a common character then the number of iterations you execute is small and the overhead for the required checks slow things down. Where-as if its an uncommon character or if you have to process the whole input, then you can get a 1.5x or more perf improvement as payoff

> Ah, you know what I ran json.benchmarks.yml and plaintext.benchmarks.yml rather than platform.benchmarks.yml. That might've had something to do with it.

Reran ensuring I used `platform.benchmarks`. Similar results overall showing that `AVX2` is the primary reason why perf shows as "worse" and with the inputs being this small, that makes sense. In addition to what I ran before, I also ran with `EnableAVX2=0` which gives us effectively `x86-64-v2` but still allows the VEX encoding to be used and further helps cement that its 256-bit vectors that are the "problem" and that its not some `EnableSSE3=0` makes it faster (and indeed `EnableSSE3=0` is slower than `EnableAVX=0` as expected).

## Json

### .NET 7 - Default

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 79         |
| Cores usage (%)        | 2,214      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 363        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 77         |
| Requests/sec           | 1,196,802  |
| Requests               | 18,071,073 |
| Mean latency (ms)      | 0.86       |
| Max latency (ms)       | 46.53      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 166.64     |
| Latency 50th (ms)      | 0.38       |
| Latency 75th (ms)      | 0.45       |
| Latency 90th (ms)      | 0.69       |
| Latency 99th (ms)      | 11.81      |

### .NET 7 - EnableAVX2=0 (effectively target `x86-64-v2` but allowing VEX encoding)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 80         |
| Cores usage (%)        | 2,229      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 363        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 85         |
| Requests/sec           | 1,215,177  |
| Requests               | 18,348,339 |
| Mean latency (ms)      | 0.69       |
| Max latency (ms)       | 48.18      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 169.20     |
| Latency 50th (ms)      | 0.38       |
| Latency 75th (ms)      | 0.44       |
| Latency 90th (ms)      | 0.55       |
| Latency 99th (ms)      | 9.71       |


### .NET 7 - EnableAVX=0 (effectively target `x86-64-v2`)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 80         |
| Cores usage (%)        | 2,238      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 363        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 78         |
| Requests/sec           | 1,221,743  |
| Requests               | 18,447,377 |
| Mean latency (ms)      | 0.66       |
| Max latency (ms)       | 36.48      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 170.11     |
| Latency 50th (ms)      | 0.38       |
| Latency 75th (ms)      | 0.44       |
| Latency 90th (ms)      | 0.54       |
| Latency 99th (ms)      | 9.17       |

### .NET 7 - EnableSSE3=0 (effectively target `x86-64-v1`)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 79         |
| Cores usage (%)        | 2,206      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 363        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 79         |
| Requests/sec           | 1,205,550  |
| Requests               | 18,203,349 |
| Mean latency (ms)      | 0.75       |
| Max latency (ms)       | 46.42      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 167.86     |
| Latency 50th (ms)      | 0.38       |
| Latency 75th (ms)      | 0.45       |
| Latency 90th (ms)      | 0.60       |
| Latency 99th (ms)      | 10.20      |

### .NET 7 - EnableHWIntrinsic=0 (effectively disable SIMD)

| load                   |            |
| ---------------------- | ---------- |
| CPU Usage (%)          | 78         |
| Cores usage (%)        | 2,198      |
| Working Set (MB)       | 38         |
| Private Memory (MB)    | 363        |
| Start Time (ms)        | 0          |
| First Request (ms)     | 99         |
| Requests/sec           | 1,206,129  |
| Requests               | 18,211,020 |
| Mean latency (ms)      | 0.77       |
| Max latency (ms)       | 41.32      |
| Bad responses          | 0          |
| Socket errors          | 0          |
| Read throughput (MB/s) | 167.94     |
| Latency 50th (ms)      | 0.38       |
| Latency 75th (ms)      | 0.45       |
| Latency 90th (ms)      | 0.57       |
| Latency 99th (ms)      | 10.92      |

## Plaintext

### .NET 7 - Default

| load                   |             |
| ---------------------- | ----------- |
| CPU Usage (%)          | 93          |
| Cores usage (%)        | 2,602       |
| Working Set (MB)       | 38          |
| Private Memory (MB)    | 370         |
| Start Time (ms)        | 0           |
| First Request (ms)     | 78          |
| Requests/sec           | 10,914,328  |
| Requests               | 164,806,480 |
| Mean latency (ms)      | 1.28        |
| Max latency (ms)       | 61.41       |
| Bad responses          | 0           |
| Socket errors          | 0           |
| Read throughput (MB/s) | 1,310.72    |
| Latency 50th (ms)      | 0.76        |
| Latency 75th (ms)      | 1.14        |
| Latency 90th (ms)      | 1.88        |
| Latency 99th (ms)      | 14.39       |

### .NET 7 - EnableAVX2=0 (effectively target `x86-64-v2` but allowing VEX encoding)

| load                   |             |
| ---------------------- | ----------- |
| CPU Usage (%)          | 93          |
| Cores usage (%)        | 2,611       |
| Working Set (MB)       | 38          |
| Private Memory (MB)    | 370         |
| Start Time (ms)        | 0           |
| First Request (ms)     | 74          |
| Requests/sec           | 10,979,623  |
| Requests               | 165,786,328 |
| Mean latency (ms)      | 1.27        |
| Max latency (ms)       | 54.10       |
| Bad responses          | 0           |
| Socket errors          | 0           |
| Read throughput (MB/s) | 1,320.96    |
| Latency 50th (ms)      | 0.75        |
| Latency 75th (ms)      | 1.13        |
| Latency 90th (ms)      | 1.84        |
| Latency 99th (ms)      | 14.13       |

### .NET 7 - EnableAVX=0 (effectively target `x86-64-v2`)

| load                   |             |
| ---------------------- | ----------- |
| CPU Usage (%)          | 94          |
| Cores usage (%)        | 2,637       |
| Working Set (MB)       | 38          |
| Private Memory (MB)    | 370         |
| Start Time (ms)        | 0           |
| First Request (ms)     | 66          |
| Requests/sec           | 10,994,770  |
| Requests               | 165,985,771 |
| Mean latency (ms)      | 1.35        |
| Max latency (ms)       | 55.85       |
| Bad responses          | 0           |
| Socket errors          | 0           |
| Read throughput (MB/s) | 1,320.96    |
| Latency 50th (ms)      | 0.75        |
| Latency 75th (ms)      | 1.14        |
| Latency 90th (ms)      | 1.98        |
| Latency 99th (ms)      | 15.22       |

### .NET 7 - EnableSSE3=0 (effectively target `x86-64-v1`)

| load                   |             |
| ---------------------- | ----------- |
| CPU Usage (%)          | 92          |
| Cores usage (%)        | 2,585       |
| Working Set (MB)       | 38          |
| Private Memory (MB)    | 370         |
| Start Time (ms)        | 0           |
| First Request (ms)     | 71          |
| Requests/sec           | 10,916,742  |
| Requests               | 164,843,707 |
| Mean latency (ms)      | 1.20        |
| Max latency (ms)       | 51.97       |
| Bad responses          | 0           |
| Socket errors          | 0           |
| Read throughput (MB/s) | 1,310.72    |
| Latency 50th (ms)      | 0.76        |
| Latency 75th (ms)      | 1.13        |
| Latency 90th (ms)      | 1.78        |
| Latency 99th (ms)      | 13.05       |

### .NET 7 - EnableHWIntrinsic=0 (effectively disable SIMD)

| load                   |             |
| ---------------------- | ----------- |
| CPU Usage (%)          | 84          |
| Cores usage (%)        | 2,359       |
| Working Set (MB)       | 38          |
| Private Memory (MB)    | 370         |
| Start Time (ms)        | 0           |
| First Request (ms)     | 109         |
| Requests/sec           | 9,972,152   |
| Requests               | 150,576,179 |
| Mean latency (ms)      | 1.25        |
| Max latency (ms)       | 76.28       |
| Bad responses          | 0           |
| Socket errors          | 0           |
| Read throughput (MB/s) | 1,198.08    |
| Latency 50th (ms)      | 0.86        |
| Latency 75th (ms)      | 1.27        |
| Latency 90th (ms)      | 1.74        |
| Latency 99th (ms)      | 13.06       |

PR #273, opened 2022-09-25
Add proposal for out of the box fuzzing

samueleresca:
Adding a proposal for including fuzz testing as part of the .NET toolchain.

### DIFF
--- a/INDEX.md
+++ b/INDEX.md
@@ -81,6 +81,7 @@ Use update-index to regenerate it:
 | 2021 | [TFM for .NET nanoFramework](accepted/2021/nano-framework-tfm/nano-framework-tfm.md) | [Immo Landwerth](https://github.com/terrajobst), [Laurent Ellerbach](https://github.com/Ellerbach), [José Simões](https://github.com/josesimoes) |
 | 2021 | [Tracking Platform Dependencies](accepted/2021/platform-dependencies/platform-dependencies.md) | [Matt Thalman](https://github.com/mthalman) |
 | 2022 | [.NET 7 Version Selection Improvements](accepted/2022/version-selection.md) | [Rich Lander](https://github.com/richlander) |
+| 2022 | [Out of the box fuzzing](accepted/2022/fuzzing.md) | [Samuele Resca](https://github.com/samueleresca) |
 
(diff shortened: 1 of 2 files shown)


rstm-sf:
It seems to me that the property base tests should also be covered. [FsCheck](https://fscheck.github.io/FsCheck) is an example of a .NET library which is a port of [QuickCheck](http://www.cse.chalmers.se/~rjmh/QuickCheck)

danmoseley:
Cc @jeffhandley

svick:
This sounds useful to me, but I'm not clear on why it should be part of the .Net SDK, instead of a tool that can be installed using `dotnet tool install`.

samueleresca:
> This sounds useful to me, but I'm not clear why it should be part of the .Net SDK, instead of a tool that can be installed using `dotnet tool install`.

Instrumenting the target and running the fuzzing usually require some overhead.
having this process supported as part of the SDK would provide a unified experience. Using an external tool might mean that you need to tweak it depending on which platform you are on (for example, I think sharp fuzz runs only on Linux and macOS).

danmoseley:
Cc @krwq @GrabYourPitchforks

jkotas:
>having this process supported as part of the SDK would provide a unified experience. Using an external tool might mean that you need to tweak it depending on which platform you are on

You can bundle the external dependencies into the tool package to provide unified experience, just like you would bundle them into the SDK to provide uniform experience.

It is better to start by independent package for features like this. You can react to feedback and iterate quicky, and you have a lot of freedom for making breaking changes.

And then consider integrating with the SDK only once you have validated on real world use cases that the experience is right and stable. Once you integrate with SDK, the bar rises a lot: You need to worry about backward compatibility, aligning with SDK ship schedule, etc.

jeffhandley:
We have had some recent prototype efforts internally in this space. We are exploring some possible paths forward while we conduct our .NET 8 planning.

GrabYourPitchforks:
Took the conversation offline. Like Jeff said, we've done some internal prototyping, but nothing that's really in a position to show at the moment.

I propose we close this issue since it's not immediately actionable and it's already being tracked in a separate database.

PR #284, opened 2022-12-16
Clarify preview behavior for SDKs

terrajobst:
This adds a clarification that specifically calls out that preview features are only support for an SDK's current framework version.

### DIFF
--- a/accepted/2021/preview-features/preview-features.md
+++ b/accepted/2021/preview-features/preview-features.md
@@ -222,10 +222,12 @@ support of each other, you can't expect to use, say, the .NET 7 SDK to build for
 SDKs are usually backwards compatible, but it's really no different from the
 framework, which is also backwards compatible but not for preview features.
 
-This means:
+In order to ensure that users get actionable error message we decided to
+actively block using preview features in frameworks that aren't the SDK's
+current framework:
 
-> The .NET SDK will only support preview features for the current .NET
-> framework.
+> The .NET SDK only supports preview features for the current framework. More
+> specifically, `<EnablePreviewFeatures>` is ignored for all other frameworks.
 >
 > The current framework is defined as the framework whose major and minor
 > version match the SDK. For instance, the .NET 6 SDK will only support preview
--- a/accepted/2021/preview-features/preview-features.md
+++ b/accepted/2021/preview-features/preview-features.md
@@ -196,6 +196,44 @@ namespace System
 </Project>
 ```
 
+### SDK support of preview features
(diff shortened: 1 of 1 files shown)


accepted/2021/preview-features/preview-features.md
```diff
@@ -196,6 +196,44 @@ namespace System
...
+
+Just like in the case of runtime and framework, we don't plan to support preview
+features in the SDK across versions. For instance, a preview API in `net6.0`
+shouldn't be expected to have the same shape -- or even exists -- in `net7.0`.
+The same is true for the .NET SDK: .NET 6 SDK might have preview features in
+MSBuild and C# that don't exist or have a very different syntax in .NET 7.
+
+Since preview features between runtime, framework, and tooling are usually in
+support of each other, you can't expect to use, say, the .NET 7 SDK to build for
+.NET 6 when using preview features. This is likely counter intuitive because
```
jaredpar, line 221: This sentence makes it sound like it might work when in reality it will not work. There is a 100% chance using preview features will break when you move to a new .NET SDK right?

terrajobst: Well that depends. We have marked API as preview that had nothing to do with generic math/static interface members. And that's the are we're eager to expand.

jaredpar: Not sure why generic math / static has anything to do with this. The break seems irrespective of whether it's a language or runtime feature. 

The core issue is that `<EnablePreviewFeatures>` only takes effect when the target framework and .NET SDK major version are the same. When you upgrade the .NET SDK this invariant is broken so the analyzer will start flagging any preview feature usage, be it language feature or API. The only time this won't result in a warning is when you had the flag set but weren't actually using a preview feature. 

This at least is how I understand this to work. Am I missing something here?

terrajobst: No, you're correct and I was under a grave misapprehension: I didn't realize (or forgot) that the SDK actually enforces that `EnablePreviewFeatures` is only honored when the TFM matches the SDK's current framework. That certainly warrants a different language for this document.

jaredpar (approved):

accepted/2021/preview-features/preview-features.md
terrajobst: Addressed by [3831d17](https://github.com/dotnet/designs/pull/284/commits/3831d17ba2d312ba7ec522f634b3823d74864998)

PR #285, opened 2022-12-16
Experimental APIs

terrajobst:
In .NET 6, we've added a way to [ship preview features](https://github.com/dotnet/designs/blob/main/accepted/2021/preview-features/preview-features.md) in otherwise stable releases. It works by tagging APIs (and language features) as preview features and requiring the user's project to explicitly opt-into using them via the `<EnablePreviewFeatures>` property.

The feature was primarily designed to accommodate platform-level features, that is for features that span runtime, library, and language, such as generic math. However, we also believed that the feature would work for libraries that want to offer preview functionality in otherwise stable NuGet packages.

Since then, we've seen requests from partners who have tried using it and hit limitations that come from guardrails we put into place due to the complex nature of platform-level preview features.

The goal of this feature is to find a design that works for both, platform-level preview features and library-level preview features.

### DIFF
--- a/accepted/2023/preview-apis/preview-apis.md
+++ b/accepted/2023/preview-apis/preview-apis.md
@@ -189,6 +189,6 @@ them makes it easier to communicate that difference to our users.
 ### Why is this not an analyzer?
 
 The compiler team suggested to make this a compiler behavior, rather than an
-analyzer. This would allow the attribute to have a `DiagnosticId` which regular
-analyzers can't (because a given analyzer has to tell the compiler upfront which
-diagnostic IDs it will raise).
+analyzer. This would allow the attribute to specify custom `DiagnosticId` values
+which regular analyzers can't (because a given analyzer has to tell the compiler
+upfront which diagnostic IDs it will raise).
--- a/accepted/2023/preview-apis/preview-apis.md
+++ b/accepted/2023/preview-apis/preview-apis.md
@@ -7,10 +7,11 @@ otherwise stable releases. It works by tagging APIs (and language features) as
 preview features and requiring the user's project to explicitly opt-into using
 them via the `<EnablePreviewFeatures>` property.
 
-The feature was primarily designed to accommodate platform-level features, that
-is for features that span runtime, library, and language, such as generic math.
-However, we also believed that the feature would work for libraries that want to
-offer preview functionality in otherwise stable NuGet packages.
+The feature was primarily designed to accommodate platform-level preview
+features, that is for features that span runtime, library, and language, such as
+generic math. However, we also believed that the feature would work for
+libraries that want to offer preview functionality in otherwise stable NuGet
+packages.
 
 Since then, we've seen requests from partners who have tried using it and hit
 limitations that come from guardrails we put into place due to the complex
@@ -22,12 +23,12 @@ preview features and library-level preview features.
 Let's first define these two terms:
 
 * **platform-level preview feature**. These are features that rely on preview
- behaviors across runtime, core library, tooling, and languages. The nature of
- these features that they are in support of each other. For example, generic
- math requires new runtime behaviors for interfaces, which the language has to
- support, and of course new interfaces and methods in core library that
- leverages them. If you want to use generic math, you need a combination of
- runtime, framework, and language that were designed to work in conjunction.
+  behaviors across runtime, core library, tooling, and languages. The nature of
+  these features is that they are in support of each other. For example, generic
+  math requires new runtime behaviors for interfaces, which the language has to
+  support, and of course new interfaces and methods in core library that
+  leverages them. If you want to use generic math, you need a combination of
+  runtime, framework, and language that were designed to work in conjunction.
 
 * **library-level preview feature**. These are library APIs that don't require
   any platform-level preview features to be consumed. The only difference to
@@ -35,7 +36,7 @@ Let's first define these two terms:
   offers no backwards compatibility when upgrading to a later version of the
   library.
 
-Due to their nature, platform-level preview features can only be consumed in
+Due to their nature, platform-level preview features can only be consumed in a
 specific combination of framework and SDK: namely, the framework and SDK version
 must match. For example, you can only use preview features in `net6.0` when
 using the .NET 6 SDK and only use preview features in `net7.0` when using the
@@ -51,23 +52,40 @@ restriction not make sense, it's highly undesirable. Library authors simply want
 a way to express "these parts of my API surface might change". This statement
 isn't tied to any runtime-, framework-, or language features.
 
+Another concern we had for platform-level preview features is that we wanted to
+ensure that application developers are aware when their applications have a
+dependency on them. That is, we want to avoid a case where a library on NuGet
+uses platform-level preview features as an implementation detail that the
+application consumer isn't aware of. So we made the opt-in mechanism viral: a
+library that turns on platform-level preview features requires its consumer to
+turn them on as well.
+
+Since library-level preview features cannot change how the .NET runtime itself
+operates, we don't have that same level of concern. In fact, we believe the
+viral nature is also undesirable.
+
 ## Scenarios and User Experience
 
 ## Requirements
 
 ### Goals
 
-* Platform-level preview features need the same guardrails as before, specifically
+* Don't change the guardrails for platform-level preview features:
     - They require opt-in (without, the build fails with an error)
     - They require to be viral (i.e. opt-in is also required for indirect
       dependencies relying on preview features)
+    - The opt-in can be expressed centrally in the project file
     - They require SDK and framework to match
-* Library-level preview need to have same guard rails, except for the
-  requirement of SDK and framework to match
+
+* Offer library-level preview features with a different set of guardrails:
+    - They require opt-in (without, the build fails with an error)
+    - The opt-in isn't viral and can be an implementation detail
+    - The opt-in is centralized, but it's specific to each preview feature
+    - Using them poses no requirements on the SDK version
 
 ### Non-Goals
 
-* Extending this feature to allow for multiple independent preview features
+* Relaxing or extending the guardrails for platform-level preview features
 
 ## Stakeholders and Reviewers
 
@@ -86,51 +104,12 @@ isn't tied to any runtime-, framework-, or language features.
 
 ## Design
 
-### Project file
-
-We change the behavior of `EnablePreviewFeatures` to allow four values instead
-of the current `bool`:
-
-| Value                | Meaning                                                                                  |
-| -------------------- | ---------------------------------------------------------------------------------------- |
-| `False`              | Off. Neither platform- nor library-level preview features can be consumed.               |
-| `True`               | Same as `LibraryAndPlatform`                                                             |
-| `Library`            | You can consume library-level preview features, but not platform-level preview features. |
-| `LibraryAndPlatform` | You can consume both library- and platform-level preview features.                       |
-
-```xml
-<Project Sdk="Microsoft.NET.Sdk">
-
-  <PropertyGroup>
-    <TargetFramework>net8.0</TargetFramework>
-    <EnablePreviewFeatures>LibraryAndPlatform</EnablePreviewFeatures>
-  </PropertyGroup>
-
-</Project>
-```
-
-### SDK
-
-Currently, the SDK ignores `<EnablePreviewFeatures>` unless the target framework
-is the SDK's current version. This will change the behavior as follows:
-
-If `<EnablePreviewFeatures>` is set to `LibraryAndPlatform` and the target
-framework isn't the current version, it will be interpreted is if it was set to
-`Library`.
-
 ### Attribute
 
-We'll extend the `[RequiresPreviewFeatures]` to take in the scope. The existing
-constructors will behave as if the scope is `PreviewScope.LibraryAndPlatform`:
+We'll create a new attribute, separate from the existing `System.Runtime.Versioning.RequiresPreviewFeaturesAttribute`:
 
 ```C#
-namespace System.Runtime.Versioning;
-
-public enum PreviewScope
-{
-    LibraryAndPlatform,
-    Library
-}
+namespace System.Diagnostics.CodeAnalysis;
 
 [AttributeUsage(AttributeTargets.Assembly |
                 AttributeTargets.Module |
@@ -144,43 +123,25 @@ public enum PreviewScope
                 AttributeTargets.Event |
                 AttributeTargets.Interface |
                 AttributeTargets.Delegate, Inherited = false)]
-public sealed class RequiresPreviewFeaturesAttribute : Attribute
+public sealed class PreviewAttribute : Attribute
 {
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public RequiresPreviewFeaturesAttribute() {}
-        : this(PreviewScope.LibraryAndPlatform) {}
-
-    [EditorBrowsable(EditorBrowsableState.Never)]
-    public RequiresPreviewFeaturesAttribute(string? message)
-        : this(PreviewScope.LibraryAndPlatform, message) {}
-
-    // NEW
-    public RequiresPreviewFeaturesAttribute(PreviewScope scope) {}
-
-    // NEW
-    public RequiresPreviewFeaturesAttribute(PreviewScope scope, string? message) {}
-
-    // NEW
-    public PreviewScope scope { get; }
-
-    public string? Message { get; }
-    public string? Url { get; set; }
+    public PreviewAttribute(string diagnosticId);
+    public string DiagnosticId { get; }
+    public string? UrlFormat { get; set; }
 }
 ```
 
-### Analyzer
+### Compiler Behavior
 
-The analyzer will be changed as follows:
+The compiler will raise a diagnostic when a preview API is used, using supplied
+diagnostic ID. The severity is always error.
 
-* APIs marked with `PreviewScope.LibraryAndPlatform` can only be used if the
- consumer is also marked with `PreviewScope.LibraryAndPlatform`.
-
-* APIs marked with `PreviewScope.Library` can only be used if the consumer is
-  marked with either `PreviewScope.LibraryAndPlatform` or
-  `PreviewScope.Library`.
-
-We should add a new diagnostic ID when library-level preview APIs are being used
-without opt-in (i.e. the second case above).
+The semantics are identical to how obsolete is tracked, except that it will
+always raise a diagnostic regardless of whether the call site is marked as
+preview or not. There is also no special treatment when both caller and callee
+are in the same assembly -- any use will generate a diagnostic and the caller is
+expected to suppress it, with the usual means (i.e. `#pragma` or project-wide
+`NoWarn`).
 
 ## Tracking Items
 
@@ -194,3 +155,40 @@ without opt-in (i.e. the second case above).
     - [dotnet/roslyn#65915](https://github.com/dotnet/roslyn/issues/65915)
 
 [preview-features]: ../../2021/preview-features/preview-features.md
+
+## Q & A
+
+### Why didn't we evolve the existing RequiresPreviewFeaturesAttribute?
+
+We considered that initially, but there are major differences in the desired
+user experience for `RequiresPreviewFeaturesAttribute` and library preview
+features:
+
+1. `RequiresPreviewFeaturesAttribute` is designed to be viral. That is, by
+   default any library that turns it on also requires its consumers to turn it
+   on. This makes sense for runtime preview features but not so much for
+   libraries who simply offer some APIs that aren't stable yet.
+
+2. `RequiresPreviewFeaturesAttribute` is designed to be managed via a single
+   MSBuild property, specifically `EnablePreviewFeatures`. This is desirable for
+   library features where wanting to use a preview API in some library doesn't
+   necessarily mean that one wants to use any preview library of any library.
+
+3. `RequiresPreviewFeaturesAttribute` also covers tooling preview features and
+   generally requires the runtime and SDK to match. That is, you cannot use
+   preview features in `net6.0` unless you're using the .NET 6.0 SDK. For
+   example, if you try to use the .NET 7.0 SDK, you'll get a build error telling
+   you that this isn't supported. This alignment isn't necessary for library
+   preview APIs and in fact highly undesirable.
+
+We considered offering a different mode for `RequiresPreviewFeaturesAttribute`
+instead of using a different attribute but we believe this makes it harder to
+understand as these are really disjoint features. Having different names for
+them makes it easier to communicate that difference to our users.
+
+### Why is this not an analyzer?
+
+The compiler team suggested to make this a compiler behavior, rather than an
+analyzer. This would allow the attribute to have a `DiagnosticId` which regular
+analyzers can't (because a given analyzer has to tell the compiler upfront which
+diagnostic IDs it will raise).
--- a/accepted/2023/preview-apis/preview-apis.md
+++ b/accepted/2023/preview-apis/preview-apis.md
@@ -1,0 +1,196 @@
+# Library Preview Features
+
+**Owner** [Immo Landwerth](https://github.com/terrjobst)
+
+In .NET 6, we've added a way to [ship preview features][preview-features] in
+otherwise stable releases. It works by tagging APIs (and language features) as
+preview features and requiring the user's project to explicitly opt-into using
+them via the `<EnablePreviewFeatures>` property.
+
+The feature was primarily designed to accommodate platform-level features, that
+is for features that span runtime, library, and language, such as generic math.
+However, we also believed that the feature would work for libraries that want to
+offer preview functionality in otherwise stable NuGet packages.
+
+Since then, we've seen requests from partners who have tried using it and hit
+limitations that come from guardrails we put into place due to the complex
+nature of platform-level preview features.
+
+The goal of this feature is to find a design that works for both, platform-level
+preview features and library-level preview features.
+
+Let's first define these two terms:
+
+* **platform-level preview feature**. These are features that rely on preview
+ behaviors across runtime, core library, tooling, and languages. The nature of
+ these features that they are in support of each other. For example, generic
+ math requires new runtime behaviors for interfaces, which the language has to
+ support, and of course new interfaces and methods in core library that
+ leverages them. If you want to use generic math, you need a combination of
+ runtime, framework, and language that were designed to work in conjunction.
+
+* **library-level preview feature**. These are library APIs that don't require
+  any platform-level preview features to be consumed. The only difference to
+  regular APIs is that these are in preview so their shape isn't stable and thus
+  offers no backwards compatibility when upgrading to a later version of the
+  library.
+
+Due to their nature, platform-level preview features can only be consumed in
+specific combination of framework and SDK: namely, the framework and SDK version
+must match. For example, you can only use preview features in `net6.0` when
+using the .NET 6 SDK and only use preview features in `net7.0` when using the
+.NET 7 SDK. Specifically, you can't use the .NET 7 SDK to use preview features
+in `net6.0`. That's because preview features aren't backwards compatible and
+platform-level preview features span both framework and tooling. In
+multi-targeting scenarios this means you can only use preview features for the
+latest framework version.
+
+For platform-level preview features this restriction makes sense because it
+models reality. However, for library-level preview features not only does this
+restriction not make sense, it's highly undesirable. Library authors simply want
+a way to express "these parts of my API surface might change". This statement
+isn't tied to any runtime-, framework-, or language features.
+
+## Scenarios and User Experience
+
+## Requirements
+
+### Goals
+
+* Platform-level preview features need the same guardrails as before, specifically
+    - They require opt-in (without, the build fails with an error)
+    - They require to be viral (i.e. opt-in is also required for indirect
+      dependencies relying on preview features)
+    - They require SDK and framework to match
+* Library-level preview need to have same guard rails, except for the
+  requirement of SDK and framework to match
+
+### Non-Goals
+
+* Extending this feature to allow for multiple independent preview features
+
+## Stakeholders and Reviewers
+
+* Runtime teams
+* C#, VB, F# compiler and IDE teams
+* Library teams
+    - Core libraries
+    - ASP.NET
+    - EF
+    - Azure SDK
+    - R9
+* Project system team
+* SDK team
+* MSBuild team
+* NuGet team
+
+## Design
+
+### Project file
+
+We change the behavior of `EnablePreviewFeatures` to allow four values instead
+of the current `bool`:
+
+| Value                | Meaning                                                                                  |
+| -------------------- | ---------------------------------------------------------------------------------------- |
+| `False`              | Off. Neither platform- nor library-level preview features can be consumed.               |
+| `True`               | Same as `LibraryAndPlatform`                                                             |
+| `Library`            | You can consume library-level preview features, but not platform-level preview features. |
+| `LibraryAndPlatform` | You can consume both library- and platform-level preview features.                       |
+
+```xml
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <EnablePreviewFeatures>LibraryAndPlatform</EnablePreviewFeatures>
+  </PropertyGroup>
+
+</Project>
+```
+
+### SDK
+
+Currently, the SDK ignores `<EnablePreviewFeatures>` unless the target framework
+is the SDK's current version. This will change the behavior as follows:
+
+If `<EnablePreviewFeatures>` is set to `LibraryAndPlatform` and the target
+framework isn't the current version, it will be interpreted is if it was set to
+`Library`.
+
+### Attribute
+
+We'll extend the `[RequiresPreviewFeatures]` to take in the scope. The existing
+constructors will behave as if the scope is `PreviewScope.LibraryAndPlatform`:
+
+```C#
+namespace System.Runtime.Versioning;
+
+public enum PreviewScope
+{
+    LibraryAndPlatform,
+    Library
+}
+
+[AttributeUsage(AttributeTargets.Assembly |
+                AttributeTargets.Module |
+                AttributeTargets.Class |
+                AttributeTargets.Struct |
+                AttributeTargets.Enum |
+                AttributeTargets.Constructor |
+                AttributeTargets.Method |
+                AttributeTargets.Property |
+                AttributeTargets.Field |
+                AttributeTargets.Event |
+                AttributeTargets.Interface |
+                AttributeTargets.Delegate, Inherited = false)]
+public sealed class RequiresPreviewFeaturesAttribute : Attribute
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public RequiresPreviewFeaturesAttribute() {}
+        : this(PreviewScope.LibraryAndPlatform, message) {}
+
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public RequiresPreviewFeaturesAttribute(string? message)
+        : this(PreviewScope.LibraryAndPlatform, message) {}
+
+    // NEW
+    public RequiresPreviewFeaturesAttribute(PreviewScope scope) {}
+
+    // NEW
+    public RequiresPreviewFeaturesAttribute(PreviewScope scope, string? message) {}
+
+    // NEW
+    public PreviewScope scope { get; }
+
+    public string? Message { get; }
+    public string? Url { get; set; }
+}
+```
+
+### Analyzer
+
+The analyzer will be changed as follows:
+
+* APIs marked with `PreviewScope.LibraryAndPlatform` can only be used if the
+ consumer is also marked with `PreviewScope.LibraryAndPlatform`.
+
+* APIs marked with `PreviewScope.Library` can only be used if the consumer is
+  marked with either `PreviewScope.LibraryAndPlatform` or
+  `PreviewScope.Library`.
+
+We should add a new diagnostic ID when library-level preview APIs are being used
+without opt-in (i.e. the second case above).
+
+## Tracking Items
+
+* The overall work of enabling this feature for libraries
+    - [dotnet/runtime#77869](https://github.com/dotnet/runtime/issues/77869)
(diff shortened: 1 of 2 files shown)


333fred (approved):

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,196 @@
...
+                AttributeTargets.Property |
+                AttributeTargets.Field |
+                AttributeTargets.Event |
+                AttributeTargets.Interface |
+                AttributeTargets.Delegate, Inherited = false)]
+public sealed class RequiresPreviewFeaturesAttribute : Attribute
+{
+    [EditorBrowsable(EditorBrowsableState.Never)]
+    public RequiresPreviewFeaturesAttribute() {}
+        : this(PreviewScope.LibraryAndPlatform, message) {}
```
buyaa-n, line 151:
```suggestion
        : this(PreviewScope.LibraryAndPlatform) {}
```

jeffhandley:
When we first introduced preview features, we considered adding some sort of a name so that consumers could enable/disable preview features by name; we decided it was overkill at that time. Should we reconsider that approach here instead of introducing a differentiation between "platform" and "library"?

We could have the nameless (`null` name) attributes denote _platform_, but then allow the name to be viral the way is described here. That would allow libraries to even have different sets of preview features within their libraries. Imagine:

1. Library Foo introduces preview feature set "FooBar" in version 34.0.0
    - There are many APIs annotated as "FooBar"
2. Consumers consume Foo version 34.1.0 and enable "FooBar" features to be used
3. Library Foo introduces preview feature set "QuxBaz" in version 35.0.0
    - There are many APIs annotated with "QuxBaz"
    - The "FooBar" APIs are refined based on consumer feedback, but they remain in preview
4. A consumer wants to upgrade to 35.0.0, compensate for "FooBar" breaking changes, but disallow "QuxBaz" usage

Of course, even just allowing or disallowing preview feature usage at the library level also could benefit from a name. But, ugh, would we want it to be a fully-qualified name? I submit that usability is more important than name collision avoidance, and therefore just passing through whatever name the library author applies to the attribute is fine. In fact, unqualified names make the viral nature much easier to manage anyway.

In summary though, I propose that instead of differentiating between library and platform, we allow the attribute to include a name--and then consumers can specify the list of names that are enabled. A value of `true` indicates that platform-level preview features are enabled and _all_ named preview features are enabled too. And only when set to `true` do we require the SDK version match as we do today.

terrajobst:
> When we first introduced preview features, we considered adding some sort of a name so that consumers could enable/disable preview features by name; we decided it was overkill at that time.

I'm not opposed to that (shocker, since I proposed the name-based approach before); the primary pushback about this design was that it creates combinatorial complexity. For example, what happens if these features depend on each other (explicitly or inadvertently)? Are we going to test combinations for stuff we as the .NET platform ship (runtime, core libraries, compilers, ASP.NET, EF etc)? Do we need to track the list of these names to avoid conflicts?

I have a negative knee-jerk reaction whenever we build facilities that differentiate between platform- and ecosystem use but I think in this case this would be acceptable because there is a stark difference between the features that require runtime/language support and the ones that don't, which are very visible (and as proven by the conversations we had partner teams) surprising in its limitations.

By having the feature distinguish between them, I think we can deliver a better UX than if we build a generic name-based system because we then can no longer track which features lock you into a matching SDK and which ones don't -- which in my opinion is the biggest issue in the UX.

333fred:
I also don't think we're locking ourselves out of providing a more granular approach in the future: we could retconn `platform` and `library` as just two different names.

marcpopMSFT:
The original trigger for this update was customers who were using preview features with the 6.0 SDK and targeting net6.0 were upgrading their SDK and being broken without any other changes (ie still targeting net6.0). This proposal sounds like those customers would be working if they were only using library level features as preview features would still be enabled at the library level but still be broken if they were using net6.0 platform level preview features. Am I reading that correctly?  

The customer would still be targeting net6.0 and using the 7.0 SDK so the platform level preview features would be disabled. Question though, breaking them in this config makes sense if they wanted to use 7.0 platform preview features (as they aren't targeting 7.0 yet) but what about 6.0 platform preview features when using 7.0?

akoeplinger:
Regarding the name-based approach: I guess this would also be handy for eventually extending beyond preview features to regular features (e.g. "threading" for the wasm case)

terrajobst:
@marcpopMSFT 

> This proposal sounds like those customers would be working if they were only using library level features as preview features would still be enabled at the library level but still be broken if they were using net6.0 platform level preview features. Am I reading that correctly?

Yes, that is precisly the intent.

akoeplinger:
Because it wouldn't be a preview feature. We discussed tagging APIs back when we did the platform annotations in the sense of having capability tags rather than simple OS tags (think "threading", "file i/o", "simd", ...). I realize this might be a bit too far fetched for this proposal 😆

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,194 @@
...
+                AttributeTargets.Struct |
+                AttributeTargets.Enum |
+                AttributeTargets.Constructor |
+                AttributeTargets.Method |
+                AttributeTargets.Property |
+                AttributeTargets.Field |
+                AttributeTargets.Event |
+                AttributeTargets.Interface |
+                AttributeTargets.Delegate, Inherited = false)]
+public sealed class PreviewAttribute : Attribute
```
terrajobst, line 126: I'm open to using a the term *experimental* in order to differentiate from the existing `RequiresPreviewFeaturesAttribute`. The downside is that we generally don't use that terminology because and instead use *preview*.

Having two different attributes in two different namespaces should be enough to differentiate their use.

* `System.Runtime.Versioning` is basically not used by anyone; I don't think people are like to discover it.
* The name `RequiresPreviewFeatures` is a bit of a mouthful, so I don't think it's likely that people type it out and accidentally complete the using suggestion from the IDE either.
* The namespace `System.Diagnostics.CodeAnalysis` has become the standard namespace for API-specific annotations, like for nullable reference types.
* The name `PreviewAttribute` is quite short and attractive, so is more likely to be discovered by library authors first.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,194 @@
...
+
+The compiler will raise a diagnostic when a preview API is used, using supplied
+diagnostic ID. The severity is always error.
+
+The semantics are identical to how obsolete is tracked, except that it will
+always raise a diagnostic regardless of whether the call site is marked as
+preview or not. There is also no special treatment when both caller and callee
+are in the same assembly -- any use will generate a diagnostic and the caller is
+expected to suppress it, with the usual means (i.e. `#pragma` or project-wide
+`NoWarn`).
```
terrajobst, line 144: @jaredpar @333fred If this becomes a compiler feature, what would be our downlevel story? For analzyers, downlevel consumers can just reference the analyzer as a package. AFAIK that's possible for the compiler too but I don't think this will play nicely in all the cases, right (like the IDE won't use that compiler).

Does this mean that for downlevel consumers we'd still need an analyzer? If so, then that's unfortunate because it means the suppression model for preview APIs would be either on/off for the entire project or use-site suppression (basically how obsoletion worked before we added the diagnostic id).

Any ideas?

accepted/2023/preview-apis/preview-apis.md
geeknoid: Hmm.

```
[Preview]
void Foo();

[Experimental]
void Foo();
```

When I see [Preview], it doesn't strike me as being descriptive and feels more imperative. "Show me a preview" kind of thing. [Experimental] doesn't have that problem. I think the equivalent form using preview-based terminology would be things like:

```
[InPreview]
void Foo();

[PreviewFeature]
void Foo();

[PreviewApi]
void Foo();
```

But I still prefer [Experimental] :-)

terrajobst: Is it because `Preview` isn't an adjective?

danmoseley: Is it possible some may read preview as its common sense of "first look at something unfinished that will be stable later"? if so, Experimental may have the advantage that it doesn't imply the feature is going to stay.

accepted/2023/preview-apis/preview-apis.md
333fred: If this is a compiler feature, it will necessarily prevent downlevel SDKs from benefiting from the feature, _unless_ the user referenced `Microsoft.NET.Compilers.Toolset` from their project. We could poison-pill the apis with a `CompilerFeatureRequired` attribute on it as well, but that is also a fairly new compiler and I wouldn't expect the .NET 6 SDK to block access to the API.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,194 @@
...
+
+## Tracking Items
+
+* The overall work of enabling this feature for libraries
+    - [dotnet/runtime#77869](https://github.com/dotnet/runtime/issues/77869)
+* Shipping `RequiresPreviewFeaturesAttribute` for downlevel
+    - [dotnet/runtime#79479](https://github.com/dotnet/runtime/issues/79479)
```
jeffhandley, line 151: Do we still need to ship `RequiresPreviewFeaturesAttribute` down-level, or would it be the new attribute that gets shipped down-level?

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,194 @@
...
+instead of using a different attribute but we believe this makes it harder to
+understand as these are really disjoint features. Having different names for
+them makes it easier to communicate that difference to our users.
+
+### Why is this not an analyzer?
+
+The compiler team suggested to make this a compiler behavior, rather than an
+analyzer. This would allow the attribute to have a `DiagnosticId` which regular
+analyzers can't (because a given analyzer has to tell the compiler upfront which
+diagnostic IDs it will raise).
```
jeffhandley, line 194:
```suggestion
The compiler team suggested to make this a compiler behavior, rather than an
analyzer. This would allow the attribute to specify custom `DiagnosticId` values which regular
analyzers can't (because a given analyzer has to tell the compiler upfront which
diagnostic IDs it will raise).
```

accepted/2023/preview-apis/preview-apis.md
geeknoid: I think only the new attribute is needed on down-level platforms.

accepted/2023/preview-apis/preview-apis.md
danmoseley: > experimental in order to differentiate from the existing RequiresPreviewFeaturesAttribute

I think differentiation would be valuable here given the conceptual overlap

> . The downside is that we generally don't use that terminology and instead use preview.

Microsoft use that term, but library authors will be applying this attribute. In my experience it is typical for OSS libraries to refer to features that may change as “experimental.”

accepted/2023/preview-apis/preview-apis.md
jaredpar: > For analzyers, downlevel consumers can just reference the analyzer as a package. 

Only so far back though that the compiler supports the API used by the analyzer. That is a way that we already put limits on how old of a compiler can be used.

accepted/2023/preview-apis/preview-apis.md
jkoritzinsky: WinRT uses the name [`ExperimentalAttribute`](https://learn.microsoft.com/en-us/uwp/api/Windows.Foundation.Metadata.ExperimentalAttribute?view=winrt-22621) for this scenario.

terrajobst: Looks we're leaning towards `Experimental` then.

accepted/2023/preview-apis/preview-apis.md
terrajobst: Yep

accepted/2023/preview-apis/preview-apis.md
terrajobst: Resolved with e8797e507eb45df346ca66c112f570bf61a90782.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+  behaviors across runtime, core library, tooling, and languages. The nature of
+  these features is that they are in support of each other. For example, generic
+  math requires new runtime behaviors for interfaces, which the language has to
+  support, and of course new interfaces and methods in core library that
+  leverages them. If you want to use generic math, you need a combination of
+  runtime, framework, and language that were designed to work in conjunction.
+
+* **library-level preview feature**. These are library APIs that don't require
+  any platform-level preview features to be consumed. The only difference to
+  regular APIs is that these are in preview so their shape isn't stable and thus
```
danmoseley, line 35: not just not stable, but may be entirely removed, even in a minor version.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+}
+```
+
+### Consuming an experimental API
+
+David builds a new version of Jabber.NET and consumes Martin's library. He very
+much wants to play with the new fancy logging Martin added. When David calls
+`AddFancyLogging` he gets a compilation error:
+
+> error ML123: FancyLoggerExtensions.AddFancyLogging() is not a stable API. In
```
danmoseley, line 111: I assume we can't offer any guidance on the diagnostic ID format, since folks need to make their own attempts to avoid overlap with others, if they care.
assuming so, when we ultimately doc this, we should not consistently use ML\d+ as you do here.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+   library features where wanting to use a preview API in some library doesn't
+   necessarily mean that one wants to use any preview library of any library.
+
+3. `RequiresPreviewFeaturesAttribute` also covers tooling preview features and
+   generally requires the runtime and SDK to match. That is, you cannot use
+   preview features in `net6.0` unless you're using the .NET 6.0 SDK. For
+   example, if you try to use the .NET 7.0 SDK, you'll get a build error telling
+   you that this isn't supported. This alignment isn't necessary for library
+   preview APIs and in fact highly undesirable.
+
```
danmoseley, line 256: Further, we implicitly (or explicitly?) reserve RequiresPreviewFeaturesAttribute for ourselves to apply, right? Especially if we accept this proposal. 

If that's the case we should probably document that, so that when folks apply `<EnablePreviewFeatures>` they can have confidence that it's only enabling the platform features they've read about in the blog, and not modifying the behavior of any well behaved libraries in their app.

danmoseley (approved):

KrzysztofCwalina:
I like the idea of having the features named. It will help with the following problem: a user adds "allow preview features" (APF) to be able to call feature A. After a couple of releases feature A ships stable, but nobody ever cleans the project file, the AFP stays in the project system (unnecessary). In the future another developer starts caling experimental feature B without realizing what's happening.

accepted/2023/preview-apis/preview-apis.md
terrajobst: Hm, that's a good point. AFAIK diagnostic IDs are just strings. Maybe should have a convention like `Product.Feature`? Could be more readable than a numeric literal.

@dotnet/roslyn-analysis what do you think?

accepted/2023/preview-apis/preview-apis.md
terrajobst: @KrzysztofCwalina [also thinks](https://github.com/dotnet/designs/pull/285#issuecomment-1409387130) this would be better

accepted/2023/preview-apis/preview-apis.md
mconnew: Being explicit that minor versions may remove api's would be really useful for CoreWCF. We have a hard commitment not to remove api's between major version releases, but our major version lifetime is a lot longer than .NET, and the time between our minor releases is relatively short. We've struggled with how to manage introducing new features which might need to have their api changed based on feedback. There might be high confidence in the code quality, but some uncertainty of the new public api's.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+ensure that application developers are aware when their applications have a
+dependency on them. That is, we want to avoid a case where a library on NuGet
+uses platform-level preview features as an implementation detail that the
+application consumer isn't aware of. So we made the opt-in mechanism viral: a
+library that turns on platform-level preview features requires its consumer to
+turn them on as well.
+
+Since library-level preview features cannot change how the .NET runtime itself
+operates, we don't have that same level of concern. In fact, we believe the
+viral nature is also undesirable.
```
Pilchie, line 65: A concern here from the Azure SDK side is that having it not be viral can lead to diamond dependency problems, where Library A exposes an experimental API, and then Libraries B and C both consume it, but at different versions where there are breaking changes.  Consumers who want to use both libraries B and C run into issues in that case.

accepted/2023/preview-apis/preview-apis.md
danmoseley: Do we also want to offer any guidance about servicing? Should consumers assume that even security patches might break experimental API?
A different way of asking this is -- should consumers assume they are in an unsupported state if their app uses experimental API (because a patch might break them anytime by removing the API) or is it a reasonable configuration to deploy in production, so long as they've tested (because only a minor version update would break the API)
I'm guessing any guidance woudl be -- if you deploy to production with an experimental API, you should make no assumptions about even a security patch breaking you.

jkoritzinsky: I feel like our guidance should be as lax as possible, and we should direct libraries to make stronger guarantees if their scenario allows for it. For example, our guidance should be "Any experimental API has no shape or stability guarantees and no support guarantees", but CoreWCF could choose to guarantee "Any experimental API will not change shape in a security patch" for any such-attributed APIs in CoreWCF. This gives the most flexibility for developers to provide experimental features in their libraries without ending up back in a situation where we need to introduce a "SuperExperimental attribute" that signifies something looser than we define here.

accepted/2023/preview-apis/preview-apis.md
jeffhandley: If we were to introduce experimental APIs within the shared framework, I presume we'd use `SYSLIB` as our prefix there too, perhaps starting with `SYSLIB5000` or something.

accepted/2023/preview-apis/preview-apis.md
terrajobst: How does virality help with that? It would just meant that the consumer would also need to suppress `A.F1` and `B.F2`. There is logically nothing that would help with compilation or runtime errors around mismatched version of the shared diamond.

accepted/2023/preview-apis/preview-apis.md
terrajobst: @KrzysztofCwalina [also thinks](https://github.com/dotnet/designs/pull/285#issuecomment-1409387130) this would be better

accepted/2023/preview-apis/preview-apis.md
terrajobst: > Further, we implicitly (or explicitly?) reserve `RequiresPreviewFeaturesAttribute` for ourselves to apply, right?

It's basically impractical for library developers today. However, with `ExperimentalAttribute` we could make that statement more explicitly and warn when code outside of `System.*` uses the `RequiresPreviewFeaturesAttribute` and suggest `ExperimentalAttribute` instead.

accepted/2023/preview-apis/preview-apis.md
terrajobst: The reason we don't think we need to make it viral goes like this:

1. If it bleeds into a consumer's public API (i.e. it's a type) then its consumer will need to suppress it anyway.

2. If it's not part of the API, then logically the code signs up for shielding its consumers from any breaking changes. If the code doesn't want that, they can apply the attribute to their entry points as well.

Pilchie: It doesn't directly help once in that situation, but it helps a consumer of B and C to know that they are getting themselves into this danger zone, which they normally wouldn't expect with stable versioned dependencies.

This *could* be a guidance thing where we recommend that any API that suppresses the diagnostic for a call it makes, re-apply the attribute.  Alternatively, you could treat it like `Obsolete` where you have the choice to *either* suppress, or to make viral.

accepted/2023/preview-apis/preview-apis.md
danmoseley: Sounds reasonable.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+                AttributeTargets.Class |
+                AttributeTargets.Struct |
+                AttributeTargets.Enum |
+                AttributeTargets.Constructor |
+                AttributeTargets.Method |
+                AttributeTargets.Property |
+                AttributeTargets.Field |
+                AttributeTargets.Event |
+                AttributeTargets.Interface |
+                AttributeTargets.Delegate, Inherited = false)]
```
marek-safar, line 198: Is this missing `AttributeTargets.Parameter` intentionally ?

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+## Stakeholders and Reviewers
+
+* Runtime teams
+* C#, VB, F# compiler and IDE teams
+* Library teams
+    - Core libraries
+    - ASP.NET
+    - EF
+    - Azure SDK
+    - R9
```
RussKie, line 172:
```suggestion
* Library teams
    - Core libraries
    - ASP.NET
    - EF
    - Azure SDK
    - R9
    - Windows Forms
    - WPF
```

@merriemcgaw @singhashish-wpf

accepted/2023/preview-apis/preview-apis.md
kirankumarkolli: +1 

Also may be all experimental features are not same as well. 

For example, Azure services might flight many features but may decide to follow-up different path with each feature/API in the same library.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+models reality. However, for library-level preview features not only does this
+restriction not make sense, it's highly undesirable. Library authors simply want
+a way to express "these parts of my API surface might change". This statement
+isn't tied to any runtime-, framework-, or language features.
+
+Another concern we had for platform-level preview features is that we wanted to
+ensure that application developers are aware when their applications have a
+dependency on them. That is, we want to avoid a case where a library on NuGet
+uses platform-level preview features as an implementation detail that the
+application consumer isn't aware of. So we made the opt-in mechanism viral: a
```
kirankumarkolli, line 59: Awesome!

accepted/2023/preview-apis/preview-apis.md
ericstj: I'm worried about how the error surfaces when using B and upgrading A to a version that changes or no longer supports the preview feature.  This will likely show up as missing method / type exceptions only at runtime.  How does the customer realize this is a problem with B and not A?  How can we get the customer to realize this is a problem at design time instead of runtime?

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,279 @@
...
+
+* **platform-level preview feature**. These are features that rely on preview
+  behaviors across runtime, core library, tooling, and languages. The nature of
+  these features is that they are in support of each other. For example, generic
+  math requires new runtime behaviors for interfaces, which the language has to
+  support, and of course new interfaces and methods in core library that
+  leverages them. If you want to use generic math, you need a combination of
+  runtime, framework, and language that were designed to work in conjunction.
+
+* **library-level preview feature**. These are library APIs that don't require
```
ericstj, line 33: Might also interesting to scope in source generators to this.

accepted/2023/preview-apis/preview-apis.md
terrajobst: @danmoseley

> should consumers assume they are in an unsupported state if their app uses experimental API 

Yes. I'd argue that's the definition of using a preview/experimental/beta API. They aren't stable and thus unsupported. RCs are an exception to this rule but I think we can safely ignore that nuance.

> if you deploy to production with an experimental API, you should make no assumptions about even a security patch breaking you.

Agreed.

@jkoritzinsky 

> I feel like our guidance should be as lax as possible

I think you mean the opposite: we want the most amount of flexibility, so I think the guidance should be as definitive as possible, along the lines of what Dan was suggesting.

accepted/2023/preview-apis/preview-apis.md
terrajobst: @Pilchie @ericstj 

Thanks for providing more detail; I think you convinced me that virality is more desirable, because NuGet packages can be independently updated and there is no way for NuGet or the build system to provide actionable errors in case there is a mismatch in experimental API surface. I like @Pilchie suggestion of following `[Obsolete]` and `[RequiresRuntimeFeatures]` approach saying you either suppress (shield) or forward (mark yourself). That feels intuitive.

@geeknoid?

accepted/2023/preview-apis/preview-apis.md
terrajobst: Yes. I don't know what it would mean to only mark a parameter as experimental as removing/changing a parameter is equivalent of removing the method and offering a new method with a different set of parameters.

accepted/2023/preview-apis/preview-apis.md
terrajobst: Tell me more about the source generator case. How are they different from manually written code?

accepted/2023/preview-apis/preview-apis.md
jkoritzinsky: Whatever the right term is, I think the guidance should allow the most flexibility possible.

geeknoid: I think the semantics need to be clear, and keep the guidance to a minimum:


APIs marked as experimental indicate that the author of the API is not committing to supporting this API over time, and the quality of the API's implementation may be lower than the rest of the API surface. If you take a dependency on an experimental API, you may find that the API signature changes or the API is completely removed the next time you upgrade to a new version of the assembly containing the API.

And then the Experimental attribute should have a note property that lets the author provide more detailed guidance if needed. "We're pretty confident about this API, we expect to remove this attribute in the 1.2 release", "Please don't use this right now, we're still in active development". Stuff like that.

accepted/2023/preview-apis/preview-apis.md
geeknoid: How would viral behavior work then?

If my assembly calls any experimental APIs, then my assembly gets marked as an experimental assembly? Then when a customer tries to bind to my assembly, they get a warning about the fact my assembly uses experimental APIs in its implementation, and so the customer would then need to suppress that warning in order to allow the binding to happen. And then this would cause their assembly to be marked experimental as well.

Basically, we'd need ab assembly-level annotation to indicate "consumes experimental APIs", and then perhaps reflect this in the nuget package metadata as well somehow.

Is that about right?

accepted/2023/preview-apis/preview-apis.md
geeknoid: Hey, could we just do magic with the API signature instead? You want to allow use of API X.Y? Then use X.Y as the ID.

> error CS1234-X.Y: This API is still in active development.

So CS1234 would be a stable error id defined by the C# compiler, and the X.Y part is the FQN of the API. So the compiler emits unique error codes for each experimental API. A customer can then suppress warnings for the specific methods.

danmoseley: It needs to match tooling expectations. Those may be fairly lax eg

https://github.com/dotnet/msbuild/blob/7cfb36cb90d1c9cc34bc4e0910d0c9ef42ee47b6/src/Shared/CanonicalError.cs#L70

VS has its own logic for passing non MSBuild tools; it hopefully matches more or less.

Also to consider is the UX eg width of the code column in the Error List. Users will expect a short code.

A variation of your suggestion could be to use the hashcode of the signature but either way this would require compiler work.

I recommend the conventional scheme used by every tool today: library owner picks any prefix they like, and then increments their suffix.

accepted/2023/preview-apis/preview-apis.md
ericstj: We could do better than a bool.  What if we had produces/consumes.  A produces experimental feature X.  B consumes experimental feature X.  If things line up everything is happy, no warnings.  If they get out of sync then an analyzer can flag B that it requires an experimental feature that's no longer present in A?

timheuer:
I think this will be okay, but just throwing this out there -- assuming you would want the initial experimental usage in tools to be shown as 'squiggles' in things like editors.

Also, how would this reflect things where APIs are reflected like objects in UI frameworks? e.g., XAML or Blazor tag helpers?

terrajobst:
@timheuer 

> I think this will be okay, but just throwing this out there -- assuming you would want the initial experimental usage in tools to be shown as 'squiggles' in things like editors.

Yes, I have [linked a work item](https://github.com/dotnet/designs/pull/285/files#diff-79d19c08a0b4e05e6971d4d4a1a52d298e64bb4192471d75dbd0e03e28f052baR227-R230) where this is being discussed with the Roslyn folks as well as the docs team.

> Also, how would this reflect things where APIs are reflected like objects in UI frameworks? e.g., XAML or Blazor tag helpers?

That's a good question; I didn't think of that yet. XAML and Tag helpers could work the same way as code, given that there is a text editor rendition of the API use. But there are other places (toolbox and property grid) that could also benefit from this. I'll add a note to the document.

accepted/2023/preview-apis/preview-apis.md
terrajobst: That's basically how the original spec for named preview features worked. We could revive that.

accepted/2023/preview-apis/preview-apis.md
terrajobst: @geeknoid not sure that scheme is desirable. Imagine an experimental feature that is a new type + members on existing types. You want to mark the type plus those members as being the same feature. That scheme would force each instance to have separate IDs.

geeknoid:
I think we need to say something about the transitive nature of the attribute. Presumably:

* If the experimental attribute is applied to a type, then all of the type's members and nested types are also considered experiemental.

* If the experimental attribute is applied to an assembly, then all of the assembly's types and members are considered experimental.

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,291 @@
...
+## Scenarios and User Experience
+
+### Shipping an experimental API in a stable package
+
+Martin builds a library that integrates with ASP.NET Core and provides features
+to make it easier to build services. The library is distributed as a NuGet
+package. The package is already in a version well past 1.0 and is considered
+stable.
+
+He ships the package often and doesn't want to maintain multiple flavors of the
```
jmarolf, line 79: > doesn't want to maintain multiple flavors of the package, one marked as prerelease and one marked as stable.

I don't quite understand this. It feels like this proposal is a way to escape semantic versioning. If Martin does ship an experimental package it should be clear to consumers that things will change. If someone takes a dependency on a pre-release package and tries to ship a release package that depends on it, nuget will warn the user and they will need to explicitly opt out of this behavior. I don't see how shipping pre-release apis in release packages is the better solution from a user-confusion standpoint.

jmarolf: Is the problem that we are solving for these nuget package authors that they want a better way to advertise their new functionality and asking people to consume pre-release packages is too big a burden? Why should we ever ship pre-release packages with this proposal?

Pilchie: As we discussed this with the Azure SDK team, we came to realize that there are two dimensions:

1. General updates/refactorings/etc to a package, that may thus be of uncertain quality across the entire surface area.
2. Specific updates to one set of APIs that are in preview forms, without the stability risks implied to the rest of the package.

Today pre-release packages in NuGet do not have a way to distinguish between these two.  This is a hope to allow us give people a chance to use and give feedback on preview apis/services without taking on all of the risk inherent in 1.  We see that many of our customers can't use the new service features we build for them when they are in preview because they have company wide central package management, and policies forbidding the use of unstable packages.

jmarolf: It feels like it would be cheaper to build this concept into nuget (`1.0.0-experimental` and `1.0.0-beta`) instead of taking compiler and other changes? I don't quite understand why we need and entirely new concept that is orthogonal to our existing ones here.

jmarolf: Or without any platform changes, you could ship a new package `MyPackage.Experimental` like what ASP.NET did. I assume there was a lot of customer pain here from that approach and that's why it's not a viable solution?

danmoseley: @jmarolf I think what has emerged here is evidence that there's a need (demonstrated eg by they're doing it already) for package authors to be able to include non-stable features within stable packages. Ie., they don't strictly sequence preview and stable work, because that isn't agile enough for them, and they and their customers don't want to maintain parallel package sequences either (`*Experimental*`..) as it's harder to reason about and doubles their costs. Second it is understood that in many cases those non stable features map closely to a new API (ie., if you use other API we can say you're using stable API in a strict sense). So this is not about getting around corporate policies that forbid preview packages but rather to make it possible for customers to work with the above behaviors in a safer way.

danmoseley: On the producer side, of course it is a new build flavor and package series. On the consumer side -- @geeknoid can give a better answer perhaps. I'm imagining transitive dependencies may be part of the challenge.

jkoritzinsky: One of the big challenges with using separate packages is introducing experimental APIs on existing types. There’s no good solution that doesn’t lead to binary compatibility breaks in the future for APIs that don’t change shape (aka non-source-breaking changes), which is what the “separate package” model is generally used to support.

jmarolf: > On the producer side, of course it is a new build flavor and package series

Unsure if the additional MSBuild scripts that would need to be added are more expensive than this feature, but I see your point.

> I'm imagining transitive dependencies may be part of the challenge.
> One of the big challenges with using separate packages is introducing experimental APIs on existing types. 

Thanks for clarifying, these are good examples of things that can't really be worked out via existing mechanisms.

accepted/2023/preview-apis/preview-apis.md
ericstj: Could require experimental compiler features, are also not (always) addressed with API in reference assemblies.  Probably there are other wrinkles.  I didn't have any major problems in mind, but felt they were different enough that we should walk through what an experimental source generator would look like.  I can also imagine we have cases of these that we'd want to represent as experimental.

accepted/2023/preview-apis/preview-apis.md
geeknoid: Can we embed my wording suggestion into the doc? I think it hits all the salient points.

geeknoid (approved):

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -0,0 +1,319 @@
...
+
+## Requirements
+
+### Goals
+
+* Don't change the behavior for platform-level preview features
+
+* Offer library-level preview features with a different set of guardrails:
+    - Modelled after `[Obsolete]`, except the severity is always an error
+    - The opting-ing is expressed similar to obsolete, that is the call site is
```
jskeet, line 138: Should this have been opting-in?

jskeet (commented):
Overall, I like this very much. I like the "per-feature" opt-in, and I'm fine with the distinction between platform and library in this situation.

One aspect which may be worth thinking about is that this is a per-member opt-in: it behaves quite differently to there being a beta version of a library and a stable version of the same library, where the library owner "knows" the context of whether it's beta or stable. If one piece of code needs to behave in different ways in the different contexts with the experimental attribute, the library author will need to have some sort of "turn on this behavior" method (protected by the attribute). That's entirely feasible of course, and could even be desirable in many cases (e.g. one instance of a type opting in to the new behavior, and another not) but for *some* contexts it won't be quite what's wanted.

This is in no way a criticism of the feature - it's more that if we publish documentation for library authors, it may be something to highlight.

accepted/2023/preview-apis/preview-apis.md
terrajobst: Yep, but 

image

terrajobst: Fixed with [342c288](https://github.com/dotnet/designs/pull/290/commits/342c28871a785880bc114146809128e33d7d1bd6).

terrajobst:
@jskeet 

> One aspect which may be worth thinking about is that this is a per-member opt-in: it behaves quite differently to there being a beta version of a library and a stable version of the same library, where the library owner "knows" the context of whether it's beta or stable. If one piece of code needs to behave in different ways in the different contexts with the experimental attribute, the library author will need to have some sort of "turn on this behavior" method (protected by the attribute).

I'm not I fully understand. The intent is that supressing the warning is that mechanism. The consumer has two options: they can decide that the buck stops at them and they intent to shieled their consumers from any breaking changes. The other option is that they forward that supression to their callers by marking themselves with the same attribute.

It's not quite true that we intent to have a member-level opt-in, that's why the diagnostic ID is required. The intent is that string can be shared across many types/members so that they can all be turned on/off as a single step.

Does this make sense?

jskeet:
I think I expressed myself poorly in the previous comment. Let me try again :)
The attribute is API-surface-oriented, rather than behaviorally-oriented. There may well be many features which you'd like users to be able to try out which don't involve any API changes. Examples might be:

- Small changes to value handling ([example in Noda Time for rounding changes](https://github.com/nodatime/nodatime/issues/1695))
- Changes to validation (e.g. prohibiting values which were only accidentally allowed previously)
- Optimizations

These *could* use this attribute via an API surface addition of "EnableXyzFeature" somewhere - but that would be a bit clunky.

I *suspect* that [AppContext.SetSwitch](https://learn.microsoft.com/en-us/dotnet/api/system.appcontext.setswitch?view=net-8.0) is a better solution for things like that. And that's fine - I'm just suggesting that if that is indeed the recommendation, it would be worth mentioning in the same documentation that the experiment attribute is described.

Does that make more sense? Happy to hop on a call if that would help more.

PR #286, opened 2022-12-29
New reflection invoke APIs [draft]

steveharter:
Adding to get some feedback, especially around `TypedReference` support by Roslyn.

### DIFF
--- a/accepted/2022/ReflectionInvoke.md
+++ b/accepted/2022/ReflectionInvoke.md
@@ -1,0 +1,354 @@
+# Reflection Invoke for 8.0 (draft \ in progress)
+#### Steve Harter
+#### December 15, 2022
+
+# Background
+For additional context, see:
+- [Developers using reflection invoke should be able to use ref struct](https://github.com/dotnet/runtime/issues/45152)
+- [Originally library issue; needs refreshing\updating](https://github.com/dotnet/runtime/issues/10057)
+
+The invoke APIs and capabilities have essentially remained the same since inception of .NET Framework. Basically, [`MethodBase.Invoke`](https://learn.microsoft.com/dotnet/api/system.reflection.methodbase.invoke):
+```cs
+public object? Invoke(object? obj, object?[]? parameters);
+
+public object? Invoke(
+    object? obj,
+    BindingFlags invokeAttr,
+    Binder? binder,
+    object?[]? parameters,
+    CultureInfo? culture);
+```
+which is implemented by [MethodInfo](https://learn.microsoft.com/dotnet/api/system.reflection.methodinfo) and [ConstructorInfo](https://learn.microsoft.com/dotnet/api/system.reflection.constructorinfo). Fields do not have an `Invoke()` and instead have `GetValue()` and `SetValue()` methods since there is no invokable code around field access.
+
+The `Invoke()` APIs are easy to use and flexible:
+- Based on `System.Object` to support both reference types (as a base type) and value types (through boxing).
+    - Note that boxing does an automatic `Nullable<T>` to `null`.
+- Automatic conversions:
+    - Implicit casts between primitives such as from `int` to `long`.
+    - `Enum` to\from its underlying type.
+    - Pointer (*) types to\from `IntPtr`.
+
+However the object-based Invoke() is not very performant:
+- Boxing is required for value types.
+  - Requires a heap allocation and associated GC overhead.
+  - A cast is required during unbox.
+  - Value types require a copy during box and unbox.
+- An `object[]` must be allocated (or manually cached by the caller) to contain parameter values.
+- The automatic conversions add overhead.
+- `ref` and `out` parameters require overhead after the invoke due to re-assignment (or "copy back") to the `parameters` argument.
+  - `Nullable<T>` is particularly expensive due to having to convert the boxed `null` to `Nullable<T>` in order to invoke methods with `Nullable<T>`, and when used with `ref` \ `out`, having to box `Nullable<T>` after invoke to "copy back" to the `parameters` argument.
+- The additional arguments (`BindingFlags`, `Binder`, `CultureInfo`) add overhead even when not used. Plus using those are actually quite rare with the exception of `BindingFlags.DoNotWrapExceptions` which is proposed to be the default going forward.
+  
+and has limitations and issues:
+- Cannot be used with byref-like types like `Span<T>` either has the target or as an argument. This is because by-ref like types cannot be boxed. **This is the key limitation expressed in this document.**
+- `ref` and `out` parameters are retrieved after `Invoke()` through the `parameters` argument. This is a manual mechanism performed by the user and means there is no argument or return value "aliasing" to the original variable.
+- Boxing of value types makes it not possible to invoke a mutable method and have the target `obj` parameter updated.
+- [`System.Reflection.Pointer.Box()`](https://learn.microsoft.com/dotnet/api/system.reflection.pointer.box?view=net-7.0) and `UnBox()` must be used to manually box and unbox a pointer (`*`) type.
+- When an exception occurs in the target method, the exception is wrapped with a `TargetInvocationException` and re-thrown. This approach is not desired in most cases and somewhat recently, `BindingFlags.DoNotWrapExceptions` was added to change this behavior as an opt-in. Not having a `try\catch` would help a bit with performance as well.
+
+Due to the performance issues and usability issues, workarounds and alternatives are used including:
+- [MethodInfo.CreateDelegate()](https://learn.microsoft.com/dotnet/api/system.reflection.methodinfo.createdelegate) which supports a direct method invocation and thus is strongly-typed and not appropriate for loosely-typed invoke scenarios.
+- [Dynamic methods](https://learn.microsoft.com/dotnet/framework/reflection-and-codedom/how-to-define-and-execute-dynamic-methods) which are IL-emit based and include a non-trivial implementation.
+- Compiled [expression trees](https://learn.microsoft.com/dotnet/csharp/programming-guide/concepts/expression-trees/) which use dynamic method if IL Emit is available but also conveniently fallback to standard reflection when IL Emit is not available.
+
+# 8.0 Goals
+The 7.0 release had a 3-4x perf improvement for `Invoke()` by using IL Emit when available and falling back to standard reflection when not available, keeping the same object-based APIs. Although this improvement is significant, it doesn't replace the need to use the IL-Emit based alternatives for the highly-performance-sensitive scenarios such as the `System.Text.Json` serializer. New APIs are required.
+
+For 8.0, there are two primary goals:
+1) Support byref-like types both for invoking and passing as arguments. An unsafe approach may need to be used for V8. This unblocks scenarios.
+2) Support "fast invoke" so that dynamic methods no longer have to be used for performance. Today both `System.Text.Json` and [dependency injection](https://learn.microsoft.com/dotnet/core/extensions/dependency-injection) use IL emit for performance. As a litmus test, these areas will be changed to use the new APIs proposed here.
+
+# Design
+In order to address the limitations and issues:
+- A least-common-denominator approach of using stack-based "interior pointers" for the invoke parameters, target and return value. This supports `in\ref\out` aliasing and the various classifications of types including value types, reference types, pointer types and byref-like types (byref-like types may require additional Roslyn support). Essentially this is a different way to achieve "boxing" in order to have a common representation of any parameter type.
+- Optionally, depending on final performance of the "interior pointer" approach, add support for creating an object-based delegate for properties and fields, like `MethodInfo.CreateDelegate()` but taking `object` as the target. This would be used for `System.Text.Json` and other areas that have strongly-typed `parameters` but a weakly-typed `target` and need maximum performance from that.
+
+## Interior pointers
+An interior pointer differs from a managed reference:
+- Can only exist on the stack (not the heap)
+- References an existing storage location, either a method parameter, a field, a variable, or an array element.
+- To be GC-safe, either requires pinning the pointer or letting the runtime track it by various means (more on this below). This may require overhead during a GC when physical memory locations change, but that is somewhat rare due to the nature of it being stack-only.
+
+See also [C++\CLI documentation](https://learn.microsoft.com/en-us/cpp/extensions/interior-ptr-cpp-cli?view=msvc-170).
+
+An interior pointer is created in several ways:
+- [System.TypedReference](https://learn.microsoft.com/en-us/dotnet/api/system.typedreference). Since this a byref-like type, it is only stack-allocated. Internally, it uses the `ref byte` approach below.
+- Using `ref byte`. This is possible in 7.0 due to the new "ref field" support. Previously, there was an internal `ByReference<T>` class that was used. This `ref byte` approach is used today in reflection invoke when there are <=4 parameters.
+- Using `IntPtr` with tracking or pinning. Currently, tracking is only supported internally through the use of a "RegisterForGCReporting()" mechanism. This approach is used today in reflection invoke when there are >=5 parameters.
+
+This design proposes the use of `TypedReference` for the safe version of APIs, and either `ref byte` or `IntPtr` approaches for the unsafe version. The `ref byte` and `IntPtr` approaches require the use of pointers, through the internal `object.GetRawData()` and other means.
+
+## TypedReference
+The internal reflection implementation will not likely be based on `TypedReference` -- instead it will take, from public APIs, `TypedReference` instances and "peel off" the interior pointer as an `IntPtr` along with tracking to support GC. The use of `TypedReference` is essentially to not require the use of unsafe code in the public APIs. Unsafe pubic APIs will likely be added as well that should have slightly better performance characteristics.
+
+### Existing usages
+`TypedReference` is used in `FieldInfo` via [`SetValueDirect(TypeReference obj, object? value)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvaluedirect) and [`object? GetValueDirect(TypeReference obj)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.getvaluedirect). The approach taken by `FieldInfo` will be expanded upon in the design here. The existing approach supports the ability to get\set a field on a target value type without boxing. Boxing the target through through [`FieldInfo.SetValue()`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvalue) instead is useless since the changes made to the boxed value type would not be reflected back to the original value type instance.
+
+`TypedReference` is also used by the undocumented `__arglist` along with `System.ArgIterator`. The approach taken by `__arglist` will not be leveraged or expanded upon in this design. It would, however, allow a pseudo-strongly-typed approach like
+```cs
+string s = "";
+int i = 1;
+object o = null;
+// This is kind of nice, but not proposed for 8.0:
+methodInfo.Invoke(__arglist(s, ref i, o));
+```
+
+### API additions
+Currently `TypedReference` is not designed for general use, although it has been part of the CLR since inception. It must be created through the undocumented `__makeref` C# keyword. There is also a `__refvalue` to obtain the value and a `__reftype` to obtain the underlying `Type`. The design here will add new APIs that are expected to be used instead of these keywords to make the APIs more mainstream and accessible from other languages.
+
+### Byref-like (or `ref struct`) support
+Currently, C# allows one to construct a `TypedReference` to a reference type or value type, but not a byref-like type:
+```cs
+int i = 42;
+TypedReference tr1 = __makeref(i); // okay
+
+Span<int> span = stackalloc int[42];
+TypedReference tr2 = __makeref(span); // Error CS1601: Cannot make reference to variable of type 'Span<int>'
+```
+
+An 8.0 ask from Roslyn is to allow this to compile. See [C# ask for supporting TypedReference + byref-like types](https://github.com/dotnet/roslyn/issues/65255).
+
+Since `TypedReference` internally stores a reference to the **storage location**, and not the actual value or managed-reference-passed-by-value, it effectively supports the `ref\out\in` modifiers without using the `ref` keyword. Attempting to use the `ref` keyword is not allowed:
+```cs
+int i = 42;
+TypedReference tr = __makeref(ref i); // error CS1525: Invalid expression term 'ref'
+```
+and not necessary due to `__refvalue`:
+```cs
+int i = 42;
+TypedReference tr = __makeref(i);
+__refvalue(tr, int) = 100;
+Console.WriteLine(i); // 100
+```
+or by using `ref` along with `__refvalue`:
+```cs
+int i = 42;
+TypedReference tr = __makeref(i);
+ChangeIt(tr);
+Console.WriteLine(i); //100
+
+public static void ChangeIt(TypedReference tr)
+{
+    ref int i = ref __refvalue(tr, int);
+    i = 100;
+}
+```
+
+### Byref-like support + simplification constraints
+Below are some simplfication constraints that may help with any Roslyn implementation around lifetime rules.
+
+**A `TypedReference` can't reference another `TypedReference`**
+```cs
+TypedReference tr = ...
+TypedReference tr2 = __makeref(tr); //  error CS1601: Cannot make reference to variable of type 'TypedReference'
+```
+
+This is similar to trying to make a `TypedReference` to a `Span<int>` above -- both are byref-like types.
+
+However, this limitation for `TypedReference-to-TypedReference` does **not** need to be removed for the goals in this design if it helps with simplifying lifetime rules.
+
+**Just support what is allowed today.** No new capabilities are expected; reflection invokes existing methods which must have already been compiled according to existing rules:
+```cs
+internal class Program
+{
+    static void Main()
+    {
+        // Case 1
+        MyRefStruct rs = default;
+        TypedReference tr = __makeref(rs._span); // Assume we allow instead of CS1601
+        ChangeIt(tr);
+
+        // Case 2
+        Span<int> heap = new int[42];
+        Span<int> stack = stackalloc int[42];
+        CallMe(ref stack, heap); // okay
+        CallMe(ref heap, stack); // CS8350 as expected
+    }
+
+    public static void ChangeIt(TypedReference tr)
+    {
+        // This compiles today:
+        ref Span<int> s = ref __refvalue(tr, Span<int>);
+
+        // And can be assigned to default
+        s = default;
+
+        Span<int> newspan = stackalloc int[42];
+
+        // But this causes CS8352 as expected:
+        s = ref newspan;
+    }
+
+    public static void CallMe(ref Span<int> span1, Span<int> span2) { }
+}
+
+public ref struct MyRefStruct
+{
+    public Span<int> _span;
+}
+```
+
+If allowed today in strongly-typed manner, then reflection should also allow:
+```cs
+internal class Program
+{
+    static void Main()
+    {
+
+        Span<int> span = stackalloc int[42];
+        TypedReference tr = __makeref(span); // Assume we allow instead of CS1601
+
+        // Using a proposed invoke API; calling should be supported passing byvalue
+        MethodInfo mi1 = typeof(Program).GetMethod(nameof(ChangeIt1));
+        mi1.InvokeDirect(target: default, arg1: tr);
+
+        // and supported passing byref
+        MethodInfo mi2 = typeof(Program).GetMethod(nameof(ChangeIt2));
+        mi2.InvokeDirect(target: default, arg1: tr);
+
+        // Just like these methods can be called today:
+        ChangeIt1(span);
+        ChangeIt2(ref span);
+    }
+
+    public static void ChangeIt1(Span<int> span) { }
+    public static void ChangeIt2(ref Span<int> span) { }
+}
+```
+
+FWIW `__arglist` currently compiles with byref-like types:
+```cs
+Span<byte> s1 = stackalloc byte[1];
+Span<byte> s2 = stackalloc byte[11];
+Span<byte> s3 = stackalloc byte[111];
+CallMe(__arglist(s1, s2, s3));
+
+static unsafe void CallMe(__arglist)
+{
+    // However, when enumerating __arglist here, the runtime throws when accessing a byref-like
+    // type although that limitation could be removed (just a runtime limitation; not compiler)
+```
+
+## Variable-length, safe collections
+(todo; see https://github.com/dotnet/runtime/issues/75349. _I have a local prototype that does a callback that doesn't require any new language or runtime features however it is not a clean programming model. The callback is necessary to interop with hide our internal GC "tracking" feature via "RegisterForGCReporting()"._)
+
+## Exception handling
+Todo; discuss using `BindingFlags.DoNotWrapExceptions` semantics only (no wrapping of exceptions).
+
+# Proposed APIs
+(work in progress; various prototypes exist here)
+## TypedReference
+```diff
+namespace System
+{
+    public ref struct TypedReference
+    {
+        // Equivalent of __makeref (except for a byref-like type since they can't be a generic parameter)
++       public static TypedReference Make<T>(ref T? value);
+        // Helper used for boxed or loosely-typed cases
++       public static TypedReference Make(ref object value, Type type);
+
+        // Equivalent of __refvalue
++       public ref T GetValue<T>();
+        // Used for byref-like types or loosely-typed cases with >4 parameters
++       public readonly unsafe ref byte TargetRef { get; }
+    }
+}
+```
+
+## MethodInfo
+(these may be extension methods instead)
+
+```diff
+namespace System.Reflection
+{
+    public abstract class MethodBase
+    {
+        // Helpers for <= 4 parameters
+
+        // Note that 'default(TypedReference)' can be specified for any "not used" arguments, such as for
+        // static methods ('target' is 'default'), 'void'-returning methods ('result' is 'default') or
+        // for any trailing unused arguments (e.g. 'arg4' can be 'default').
+
+        // We may also want to consider using 'ref byte' instead of or in addition to 'TypedReference'
+        // along with new helper methods like 'public static ref byte GetRawData(object o)'
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference arg1,
++           TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference arg1,
++           TypedReference arg2,
++           TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference arg1,
++           TypedReference arg2,
++           TypedReference arg3,
++           TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference arg1,
++           TypedReference arg2,
++           TypedReference arg3,
++           TypedReference arg4,
++           TypedReference result);
+
+        // Unsafe (todo: more on this; samples)
++       public virtual unsafe void InvokeDirect(
++           TypedReference target,
++           TypedReference* parameters,
++           TypedReference result);
+
+       // Also consider a safe variable-length callback that does tracking internally
+       // since we don't support a safe variable-length collection mechanism
+       // or the ability to use a Span<TypedReference>
+       // (todo: more on this?; prototype exists)
+    }
+}
+```
+
+## PropertyInfo \ FieldInfo
+```diff
+namespace System.Reflection
+{
+    public abstract class PropertyInfo
+    {
++       [System.CLSCompliantAttribute(false)]
+        public virtual void GetValueDirect(TypedReference target, TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
+        public virtual void SetValueDirect(TypedReference target, TypedReference value);
+
+        // Possible for performance in System.Text.Json:
++       public virtual Func<object, TValue> CreateGetterDelegate<TValue>();
++       public virtual Action<object, TValue> CreateSetterDelegate<TValue>();
+    }
+
+    public abstract class FieldInfo
+    {
++       [System.CLSCompliantAttribute(false)]
+        public virtual void GetValueDirect(TypedReference target, TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
+        public virtual void SetValueDirect(TypedReference target, TypedReference value);
+
+        // Possible adds to get max performance in System.Text.Json:
++       public virtual Func<object, TValue> CreateGetterDelegate<TValue>();
++       public virtual Action<object, TValue> CreateSetterDelegate<TValue>();
+    }
+}
+```


accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+- Compiled [expression trees](https://learn.microsoft.com/dotnet/csharp/programming-guide/concepts/expression-trees/) which use dynamic method if IL Emit is available but also conveniently fallback to standard reflection when IL Emit is not available.
+
+# 8.0 Goals
+The 7.0 release had a 3-4x perf improvement for `Invoke()` by using IL Emit when available and falling back to standard reflection when not available, keeping the same object-based APIs. Although this improvement is significant, it doesn't replace the need to use the IL-Emit based alternatives for the highly-performance-sensitive scenarios such as the `System.Text.Json` serializer. New APIs are required.
+
+For 8.0, there are two primary goals:
+1) Support byref-like types both for invoking and passing as arguments. An unsafe approach may need to be used for V8. This unblocks scenarios.
+2) Support "fast invoke" so that dynamic methods no longer have to be used for performance. Today both `System.Text.Json` and [dependency injection](https://learn.microsoft.com/dotnet/core/extensions/dependency-injection) use IL emit for performance. As a litmus test, these areas will be changed to use the new APIs proposed here.
```
jkotas, line 59: Replacing dynamic methods in `Microsoft.Extensions.DependencyInjection` is more important than replacing dynamic methods in `System.Text.Json`.

Our .NET 8 plan is to keep `Microsoft.Extensions.DependencyInjection` dependency on reflection for trimming and native AOT compatibility. It means that we really need a fast reflection invoke that does not depend on dynamic code generation.

The plan for System.Text.Json is to use source generators for trimming and native AOT compatibility. It means that we do need to get rid of the dynamic code generation in (non-source generated) System.Text.Json to achieve our .NET 8 goals.

cc @eerhardt @davidfowl

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+
+public static void ChangeIt(TypedReference tr)
+{
+    ref int i = ref __refvalue(tr, int);
+    i = 100;
+}
+```
+
+### Byref-like support + simplification constraints
+Below are some simplfication constraints that may help with any Roslyn implementation around lifetime rules.
```
jkotas, line 138:
```suggestion
Below are some simplification constraints that may help with any Roslyn implementation around lifetime rules.
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+        // Note that 'default(TypedReference)' can be specified for any "not used" arguments, such as for
+        // static methods ('target' is 'default'), 'void'-returning methods ('result' is 'default') or
+        // for any trailing unused arguments (e.g. 'arg4' can be 'default').
+
+        // We may also want to consider using 'ref byte' instead of or in addition to 'TypedReference'
+        // along with new helper methods like 'public static ref byte GetRawData(object o)'
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference result);
```
MichalStrehovsky, line 278: Since this is a bigger struct, it's going to be passed by reference on many platforms anyway. Should we make it explicit?

```suggestion
+           in TypedReference target,
+           out TypedReference result);
```

jkotas: `TypedReference` is same size as `Span<T>` (2*IntPtr.Size). We do not pass `Span<T>` via `in` and it would be de-optimization in typical case. Also, the .NET Framework Design Guidelines are advising against `in` in public APIs (look for "DO NOT pass value types by read-only reference (in).").

> `out TypedReference result`

This would have to be `in TypedReference result`. The `TypedReference` is meant to point to a location to store the result.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+- Compiled [expression trees](https://learn.microsoft.com/dotnet/csharp/programming-guide/concepts/expression-trees/) which use dynamic method if IL Emit is available but also conveniently fallback to standard reflection when IL Emit is not available.
+
+# 8.0 Goals
+The 7.0 release had a 3-4x perf improvement for `Invoke()` by using IL Emit when available and falling back to standard reflection when not available, keeping the same object-based APIs. Although this improvement is significant, it doesn't replace the need to use the IL-Emit based alternatives for the highly-performance-sensitive scenarios such as the `System.Text.Json` serializer. New APIs are required.
+
+For 8.0, there are two primary goals:
+1) Support byref-like types both for invoking and passing as arguments. An unsafe approach may need to be used for V8. This unblocks scenarios.
+2) Support "fast invoke" so that dynamic methods no longer have to be used for performance. Today both `System.Text.Json` and [dependency injection](https://learn.microsoft.com/dotnet/core/extensions/dependency-injection) use IL emit for performance. As a litmus test, these areas will be changed to use the new APIs proposed here.
+
```
jkotas, line 60: I would be useful to get a list of reflection APIs that we believe we would need to eliminate dynamic codegen from Microsoft.Extensions.DependencyInjection and System.Text.Json without incurring unacceptable perf regressions (two lists - one for each library).

accepted/2022/ReflectionInvoke.md
davidfowl: Right. Today, it's hard to remove reflection from dependency injection because the resolution of instances relies on the knowledge that only exists at runtime (the list of services registered in the container). We could do partial codegen for each module but that could end up being slower (it's similar to partial crossgen).

This is the implementation that gets used at runtime when resolving instances in AOT mode using DI:

https://github.com/dotnet/runtime/blob/5b8ebeabb32f7f4118d0cc8b8db28705b62469ee/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/CallSiteRuntimeResolver.cs

accepted/2022/ReflectionInvoke.md
davidfowl: - [Array.CreateInstance](https://github.com/dotnet/runtime/blob/5b8ebeabb32f7f4118d0cc8b8db28705b62469ee/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/CallSiteRuntimeResolver.cs#L165)
- [ConstructorInfo.Invoke](https://github.com/dotnet/runtime/blob/5b8ebeabb32f7f4118d0cc8b8db28705b62469ee/src/libraries/Microsoft.Extensions.DependencyInjection/src/ServiceLookup/CallSiteRuntimeResolver.cs#L69)

Some of the overhead will be in DI itself, but these are the 2 methods that are "reflection invoke based" in the hot path. The other uses of reflection are metadata based (ConstructorInfo, ParamterInfo etc), but I'm assuming we're not talking about those calls.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+### Existing usages
+`TypedReference` is used in `FieldInfo` via [`SetValueDirect(TypeReference obj, object? value)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvaluedirect) and [`object? GetValueDirect(TypeReference obj)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.getvaluedirect). The approach taken by `FieldInfo` will be expanded upon in the design here. The existing approach supports the ability to get\set a field on a target value type without boxing. Boxing the target through through [`FieldInfo.SetValue()`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvalue) instead is useless since the changes made to the boxed value type would not be reflected back to the original value type instance.
```
AaronRobinsonMSFT, line 85:
```suggestion
`TypedReference` is used in `FieldInfo` via [`SetValueDirect(TypeReference obj, object? value)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvaluedirect) and [`object? GetValueDirect(TypeReference obj)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.getvaluedirect). The approach taken by `FieldInfo` will be expanded upon in the design here. The existing approach supports the ability to get\set a field on a target value type without boxing. Boxing the target through [`FieldInfo.SetValue()`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvalue) instead is useless since the changes made to the boxed value type would not be reflected back to the original value type instance.
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+FWIW `__arglist` currently compiles with byref-like types:
+```cs
+Span<byte> s1 = stackalloc byte[1];
+Span<byte> s2 = stackalloc byte[11];
+Span<byte> s3 = stackalloc byte[111];
+CallMe(__arglist(s1, s2, s3));
+
+static unsafe void CallMe(__arglist)
+{
+    // However, when enumerating __arglist here, the runtime throws when accessing a byref-like
```
AaronRobinsonMSFT, line 228: Is the runtime tracking lifetime properly in this case?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+
+# Proposed APIs
+(work in progress; various prototypes exist here)
+## TypedReference
+```diff
+namespace System
+{
+    public ref struct TypedReference
+    {
+        // Equivalent of __makeref (except for a byref-like type since they can't be a generic parameter)
```
AaronRobinsonMSFT, line 246: Reference https://github.com/dotnet/runtime/issues/65112

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
++       public static TypedReference Make<T>(ref T? value);
+        // Helper used for boxed or loosely-typed cases
++       public static TypedReference Make(ref object value, Type type);
+
+        // Equivalent of __refvalue
++       public ref T GetValue<T>();
+        // Used for byref-like types or loosely-typed cases with >4 parameters
++       public readonly unsafe ref byte TargetRef { get; }
```
AaronRobinsonMSFT, line 254: This is for an on stack array, right?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+- Using `IntPtr` with tracking or pinning. Currently, tracking is only supported internally through the use of a "RegisterForGCReporting()" mechanism. This approach is used today in reflection invoke when there are >=5 parameters.
+
+This design proposes the use of `TypedReference` for the safe version of APIs, and either `ref byte` or `IntPtr` approaches for the unsafe version. The `ref byte` and `IntPtr` approaches require the use of pointers, through the internal `object.GetRawData()` and other means.
+
+## TypedReference
+The internal reflection implementation will not likely be based on `TypedReference` -- instead it will take, from public APIs, `TypedReference` instances and "peel off" the interior pointer as an `IntPtr` along with tracking to support GC. The use of `TypedReference` is essentially to not require the use of unsafe code in the public APIs. Unsafe pubic APIs will likely be added as well that should have slightly better performance characteristics.
```
jnm2, line 82:
```suggestion
The internal reflection implementation will not likely be based on `TypedReference` -- instead it will take, from public APIs, `TypedReference` instances and "peel off" the interior pointer as an `IntPtr` along with tracking to support GC. The use of `TypedReference` is essentially to not require the use of unsafe code in the public APIs. Unsafe public APIs will likely be added as well that should have slightly better performance characteristics.
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+- [Dynamic methods](https://learn.microsoft.com/dotnet/framework/reflection-and-codedom/how-to-define-and-execute-dynamic-methods) which are IL-emit based and include a non-trivial implementation.
+- Compiled [expression trees](https://learn.microsoft.com/dotnet/csharp/programming-guide/concepts/expression-trees/) which use dynamic method if IL Emit is available but also conveniently fallback to standard reflection when IL Emit is not available.
+
+# 8.0 Goals
+The 7.0 release had a 3-4x perf improvement for `Invoke()` by using IL Emit when available and falling back to standard reflection when not available, keeping the same object-based APIs. Although this improvement is significant, it doesn't replace the need to use the IL-Emit based alternatives for the highly-performance-sensitive scenarios such as the `System.Text.Json` serializer. New APIs are required.
+
+For 8.0, there are two primary goals:
+1) Support byref-like types both for invoking and passing as arguments. An unsafe approach may need to be used for V8. This unblocks scenarios.
```
marek-safar, line 58: Using V8 is confusing, could you use common .NET8 instead?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
++       public virtual void InvokeDirect(
++           TypedReference target,
++           TypedReference arg1,
++           TypedReference result);
```
marek-safar, line 284: What will be TypedReference value of method return a void?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+public object? Invoke(object? obj, object?[]? parameters);
+
+public object? Invoke(
+    object? obj,
+    BindingFlags invokeAttr,
+    Binder? binder,
+    object?[]? parameters,
+    CultureInfo? culture);
+```
+which is implemented by [MethodInfo](https://learn.microsoft.com/dotnet/api/system.reflection.methodinfo) and [ConstructorInfo](https://learn.microsoft.com/dotnet/api/system.reflection.constructorinfo). Fields do not have an `Invoke()` and instead have `GetValue()` and `SetValue()` methods since there is no invokable code around field access.
```
jaredpar, line 21: Assume that properties are implemented by `MethodInfo` on the accessors?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+
+### Existing usages
+`TypedReference` is used in `FieldInfo` via [`SetValueDirect(TypeReference obj, object? value)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvaluedirect) and [`object? GetValueDirect(TypeReference obj)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.getvaluedirect). The approach taken by `FieldInfo` will be expanded upon in the design here. The existing approach supports the ability to get\set a field on a target value type without boxing. Boxing the target through through [`FieldInfo.SetValue()`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvalue) instead is useless since the changes made to the boxed value type would not be reflected back to the original value type instance.
+
+`TypedReference` is also used by the undocumented `__arglist` along with `System.ArgIterator`. The approach taken by `__arglist` will not be leveraged or expanded upon in this design. It would, however, allow a pseudo-strongly-typed approach like
```
jaredpar, line 87: I thought `__arglist` was Windows only. Is that not the case or do we plan on extending it?

accepted/2022/ReflectionInvoke.md
jaredpar: FWIW the compiler does do lifetime tracking here.

accepted/2022/ReflectionInvoke.md
steveharter: > It means that we really need a fast reflection invoke that does not depend on dynamic code generation and that supports everything that Microsoft.Extensions.DependencyInjection needs.

To clarify: for "fast invoke", the new APIs still require "dynamic code generation" through IL Emit but of course use the "slow invoke" interpreted fallback approach when not available. So DI would use the new reflection APIs instead instead of expressions. An unstated goal is to support https://github.com/dotnet/runtime/issues/66153 which doesn't want to use expressions in Blazor client due to having to bring along the large System.Linq.Expressions.dll.

accepted/2022/ReflectionInvoke.md
steveharter: Yes it would be used to help "peel off" the reference so it can be passed as a non-Typed referenced. See this [prototype ](https://github.com/dotnet/runtime/compare/main...steveharter:runtime:InvokeWithStj#diff-a6a6bf9694f968751e223700b17bb8588bddc4a9ed62e0a9d8534bf390e60891R68) for the simple case of a single parameter (property value).

accepted/2022/ReflectionInvoke.md
steveharter: I'm thinking just `default(TypedReference)` would be used. I believe it mentions this just above.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,354 @@
...
+# Proposed APIs
+(work in progress; various prototypes exist here)
+## TypedReference
+```diff
+namespace System
+{
+    public ref struct TypedReference
+    {
+        // Equivalent of __makeref (except for a byref-like type since they can't be a generic parameter)
++       public static TypedReference Make<T>(ref T? value);
```
jaredpar, line 247: This is strictly less powerful than `__makeref` though because it can't support `ref struct`. Until we get support for `ref struct` generic parameters my recommendation would be to use `__makeref` everywhere.

accepted/2022/ReflectionInvoke.md
steveharter: Yep. Just clarified that.

accepted/2022/ReflectionInvoke.md
steveharter: Sure. Thanks

accepted/2022/ReflectionInvoke.md
steveharter: I added the sections for DI\STJ listing their usage of reflection.

accepted/2022/ReflectionInvoke.md
steveharter: Yes that's the case -- Windows only and no plans to change that for V8 although it has been requested.

I was just pointing out here that Roslyn and the runtime already support `TypedReference` here.

accepted/2022/ReflectionInvoke.md
steveharter: Thanks. Added

accepted/2022/ReflectionInvoke.md
steveharter: FWIW Roslyn actually doesn't allow `TypedReference` to be passed today via `in\ref\out` -- CS1601. I don't think we need to change this. Also that would only work if bound to a ref field in a ref struct since it is be stack-allocated.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,376 @@
...
+
+### Existing usages
+`TypedReference` is used in `FieldInfo` via [`SetValueDirect(TypeReference obj, object? value)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvaluedirect) and [`object? GetValueDirect(TypeReference obj)`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.getvaluedirect). The approach taken by `FieldInfo` will be expanded upon in the design here. The existing approach supports the ability to get\set a field on a target value type without boxing. Boxing the target through [`FieldInfo.SetValue()`](https://learn.microsoft.com/dotnet/api/system.reflection.fieldinfo.setvalue) is useless since the changes made to the boxed value type would not be reflected back to the original value type instance.
+
+`TypedReference` is also used by the undocumented `__arglist` along with `System.ArgIterator` although `__arglist` is Windows-only. The approach taken by `__arglist` will not be leveraged or expanded upon in this design. It would, however, allow a pseudo-strongly-typed approach like
```
jaredpar, line 106: Should we add a top level item to track expanding `__arglist` to non-windows platforms? Most of other non-trivial work items have a top level item and my guess is this is non-trivial work.

AaronRobinsonMSFT: The idea of native varargs is tracked at https://github.com/dotnet/runtime/issues/48796.

accepted/2022/ReflectionInvoke.md
AaronRobinsonMSFT: > my recommendation would be to use __makeref everywhere.

Would we then advocate and evangelize it usage? I thought it was considered bad form to use any `__` keywords nowadays.

jaredpar: One of the explicit goals of this work is to enable `ref struct` to participate here. Rather than say use `TypedReference.Make<T>` for most types but `__makeref` for `ref struct` I prefer the simpler, context-free, advice of use `__makeref`. 

I agree that `__` is explicitly unsupported now but part of this work is basically us defining these operators, adding docs and telling people when we support them.

accepted/2022/ReflectionInvoke.md
jaredpar: We should link to that here then.

accepted/2022/ReflectionInvoke.md
```diff
@@ -76,15 +95,15 @@ An interior pointer is created in several ways:
 - Using `ref byte`. This is possible in 7.0 due to the new "ref field" support. Previously, there was an internal `ByReference<T>` class that was used. This `ref byte` approach is used today in reflection invoke when there are <=4 parameters.
 - Using `IntPtr` with tracking or pinning. Currently, tracking is only supported internally through the use of a "RegisterForGCReporting()" mechanism. This approach is used today in reflection invoke when there are >=5 parameters.
```
jkotas, line 96:
```suggestion
- Using `void*` with tracking or pinning. Currently, tracking is only supported internally through the use of a "RegisterForGCReporting()" mechanism. This approach is used today in reflection invoke when there are >=5 parameters.
```
We are using `void*` for pointers in new APIs.

accepted/2022/ReflectionInvoke.md
```diff
@@ -76,15 +95,15 @@ An interior pointer is created in several ways:
...
 
-This design proposes the use of `TypedReference` for the safe version of APIs, and either `ref byte` or `IntPtr` approaches for the unsafe version. The `ref byte` and `IntPtr` approaches require the use of pointers, through the internal `object.GetRawData()` and other means.
+This design proposes the use of `TypedReference` for the safe version of APIs, and either `ref byte`, `IntPtr` or `TypeReference*` approaches for the unsafe version. The `ref byte` and `IntPtr` approaches require the use of pointers through the internal `object.GetRawData()` and other means which would need to be exposed publicaly or wrapped in some manner.
```
jkotas, line 98:
```suggestion
This design proposes the use of `TypedReference` for the safe version of APIs, and either `ref byte`, `void*` or `TypeReference*` approaches for the unsafe version. The `ref byte` and `IntPtr` approaches require the use of pointers through the internal `object.GetRawData()` and other means which would need to be exposed publicaly or wrapped in some manner.
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -76,15 +95,15 @@ An interior pointer is created in several ways:
...
 
 ## TypedReference
-The internal reflection implementation will not likely be based on `TypedReference` -- instead it will take, from public APIs, `TypedReference` instances and "peel off" the interior pointer as an `IntPtr` along with tracking to support GC. The use of `TypedReference` is essentially to not require the use of unsafe code in the public APIs. Unsafe pubic APIs will likely be added as well that should have slightly better performance characteristics.
+The internal reflection implementation will not likely be based on `TypedReference` -- instead it will take, from public APIs, `TypedReference` instances and "peel off" the interior pointer either as a `ref byte` or as an `IntPtr` along with tracking to support GC. The use of `TypedReference` is essentially to not require the use of unsafe code in the public APIs. Unsafe public APIs will likely be added as well that should have slightly better performance characteristics.
```
jkotas, line 101:
```suggestion
The internal reflection implementation will not likely be based on `TypedReference` -- instead it will take, from public APIs, `TypedReference` instances and "peel off" the interior pointer either as a `ref byte` or as an `void*` along with tracking to support GC. The use of `TypedReference` is essentially to not require the use of unsafe code in the public APIs. Unsafe public APIs will likely be added as well that should have slightly better performance characteristics.
```

accepted/2022/ReflectionInvoke.md
jkotas: >  for "fast invoke", the new APIs still require "dynamic code generation" through IL Emit but of course use the "slow invoke" interpreted fallback approach when not available. 

For native AOT config, the invocation should be done via a stub generated at build time. This stub is similar to the stub generated via IL Emit and it has similar performance. It is how reflection invoke works in native AOT today. The slow interpreted fallback does not exist in native AOT.

accepted/2022/ReflectionInvoke.md
RikkiGibson: Based on separate discussion, my feeling here is: let's use `__makeref` etc. for the time being, but eventually make sure we introduce the ability to express `TypedReference.Make<T>` in the language in subsequent releases, and steer people towards that once it's available.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,376 @@
...
+       // (todo: more on this?; prototype exists)
+    }
+}
+```
+
+## PropertyInfo \ FieldInfo
+```diff
+namespace System.Reflection
+{
+    public abstract class PropertyInfo
```
davidfowl, line 350: We should make sure these APIs allow ASP.NET to delete this:

https://github.com/dotnet/aspnetcore/blob/4535ea1263e9a24ca8d37b7266797fe1563b8b12/src/Shared/PropertyHelper/PropertyHelper.cs

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,376 @@
...
+    {
++       [System.CLSCompliantAttribute(false)]
+        public virtual void GetValueDirect(TypedReference target, TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
+        public virtual void SetValueDirect(TypedReference target, TypedReference value);
+
+        // Possible for performance in System.Text.Json:
++       public virtual Func<object, TValue> CreateGetterDelegate<TValue>();
++       public virtual Action<object, TValue> CreateSetterDelegate<TValue>();
```
davidfowl, line 360: https://github.com/dotnet/aspnetcore/blob/4535ea1263e9a24ca8d37b7266797fe1563b8b12/src/Shared/PropertyHelper/PropertyHelper.cs#L73-L92

accepted/2022/ReflectionInvoke.md
davidfowl: it seems like we could delete a large amount of code that uses expression trees in some cases if we could have the CoreCLR version of these APIs use reflection emit under the covers. Then we could just use the built in methods and have them be "best in class" on each of the available platforms.

davidfowl:
The examples in this proposal feel unrealistic for the cases where reflection need to be used. Typically, there's an assembly boundary between the code doing the reflection and the code being reflected on. 

App.dll (depends on Library.dll)

```C#
app.RegisterController<HomeController>();

public class AlbumController : ControllerBase
{
    [HttpGet("/albums")]
    public Task<IEnumerable<Album>> GetAlbums(Db db) => db.Albums.ToListAsync();
    
    [HttpGet("/albums/{id}")]
    public IActionResult GetAlbum(int id, Db db) => db.Albums.Find(id) switch
    {
        Album a =>  Ok(a),
        _ => NotFound();
    };
    
    [HttpPost("/albums")]
    public async Task<Person> GetPersonSlow(Album album, Db db)
    {
        db.Albums.Add(album);
        return Created(album);
    }
}
```

Library.dll

```C#
static void RegisterController<TController>(this WebApplication app)
{
    foreach (MethodInfo method in typeof(TController).GetMethods())
    {
        var routeAttribute = method.GetCustomAttribute<HttpMethodAttribute>();

        if (routeAttribute is null) continue;

        app.MapMethods(routeAttribute.Template!, routeAttribute.HttpMethods, async context =>
        {
            TController instance = Activator.CreateInstance<TController>();
            object[] args = PrepareArguments(method, context);
            object? result = method.Invoke(instance, args);
            if (result is IResult iResult)
            {
                await iResult.ExecuteAsync(context);
            }
            if (result is Task<IResult> taskOfResult)
            {
                await (await taskOfResult).ExecuteAsync(context);
            }
            // Etc
        });
    }
}
```

The interesting part here is the `PrepareArguments` method. This method needs to create the arguments for passing into the method. Let's say we're preparing the arguments for `GetAlbum`, we need to produce an int and a Db (reference type owned by the user that we cannot reference directly). Here's some sample code for how the binding for this method could look:

```C#
static object?[] PrepareArguments(MethodInfo methodInfo, HttpContext context)
{
    var parameters = methodInfo.GetParameters();
    var args = new object?[parameters.Length];
    for (int i = 0; i < args.Length; i++)
    {
        args[i] = Resolve(context, parameters[i]);
    }
    return args;
}

static object? Resolve(HttpContext context, ParameterInfo parameterInfo)
{
    // Assume query string
    if (parameterInfo.ParameterType == typeof(int))
    {
        return int.TryParse(context.Request.Query[parameterInfo.Name!], out var val) ?
            val : null;
    }

    // Assume service
    return context.RequestServices.GetRequiredService(parameterInfo.ParameterType);
}
```

In the above, I don't know how to avoid boxing the integer even with these new APIs. This is the missing piece from the examples. 

Using the above example, here are 3 versions of what "dynamic code generation" could look like for the following thunk:

```C#
IActionResult GetAlbum(int id, Db db);
```

1. For clarity, we start with a potentially source generated piece of code:

```C#
Task SourceGeneratedThunk(HttpContext context)
{
    int id = int.Parse(context.Request.Query["id"]);
    Db db = context.RequestServices.GetRequiredService<Db>();
    GetAlbum(id, )
}
```

This code can only be generated in user code, it cannot be generated as part of framework logic since it directly references user code.

2. Source Expression trees (implementation left out for brevity and sanity)

3. ILEmit (I don't have all night 😅)

4. Generics (what this model is pushing)

```C#
Task PrecompiledThunk<T0, T1>(HttpContext context) where T0 : IParseable<T0>
{
    T0 arg0 = T0.Parse(context.Request.Query["id"]);
    T1 arg1 = context.RequestServices.GetRequiredService<T0>();
    GetAlbum(arg0, arg1);
}
```

I can't figure out how I would use TypedReference in these situations other than if I know the type at compile time, what am I missing?

MichalStrehovsky:
If I understand this correctly, the problem David writes about in https://github.com/dotnet/designs/pull/286#issuecomment-1382708887 is with getting a storage location where we could prepare the parameters in if we're in dynamic code that doesn't have any T's representing method parameters.

Does this assume we'll be able to dynamically create storage locations? Just throwing a thing here because I know very little about the GC - could we have something that allows to stackalloc a byte buffer that we mark as conservatively reported to the GC that could then be used to partition into TypedReferences for each parameter? I.e. the size of the buffer would be the sum of all arguments (plus padding).

AaronRobinsonMSFT:
> could we have something that allows to stackalloc a byte buffer that we mark as conservatively reported to the GC that could then be used to partition into TypedReferences for each parameter?

Yes. This "feature" does exist internally. It was added for other Reflection scenarios. How to express this and what sort API would be provided is the more complicated question.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,376 @@
...
+
+For .NET 8, there are two primary goals:
+1) Support byref-like types both for invoking and passing as arguments; this unblocks various scenarios. An unsafe approach may used for .NET 8 if support for `TypedReference` isn't addressed by Roslyn (covered later).
+2) Support "fast invoke" so that dynamic methods no longer have to be used for performance. Today both `STJ (System.Text.Json)` and [DI (dependency injection)](https://learn.microsoft.com/dotnet/core/extensions/dependency-injection) use IL Emit for performance although DI uses emit through expressions while STJ uses IL Emit directly.
+
+## STJ and DI
+As a litmus test, STJ and DI will be changed to use the new APIs proposed here. This is more important to DI since, unlike STJ which has a source generator that can avoid reflection, DI must continue to use reflection. See also https://github.com/dotnet/runtime/issues/66153 which should be addressed by having a fast constructor invoke that can be used by DI.
```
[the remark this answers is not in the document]
steveharter: That is the thinking - if it's not fast enough for internal use, it may not be fast enough for others. However, I have a prototype for STJ using the proposed APIs and has a 5% regression in deserialization (note that without any Emit, deserialization is ~2x slower, so getting it to just 5% is pretty significant). That was the reason I also added the `PropertyInfo.CreateGetter\SetterDelegate` to match how STJ uses delegates today. There may also be some gains in going down an unsafe approach, but I haven't measured the perf of that yet.

For DI, I hope to use the new APIs directly.

accepted/2022/ReflectionInvoke.md
steveharter: The proposed TypedReference APIs will be able to take a boxed `object` and use the internal reference, so object is supported that way.

However, if there are a variable number of parameters that need to be passed, you'll still need to pass them down somehow (e.g. `object[]` or an unsafe approach (e.g. `TypedReference*[] \ void*`) until an Invoke can be done.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+- Compiled [expression trees](https://learn.microsoft.com/dotnet/csharp/programming-guide/concepts/expression-trees/) which use dynamic methods if IL Emit is available but also conveniently falls back to standard reflection when not available. Using an expression to invoke a member isn't intuitive and brings along the large `System.Linq.Expressions.dll` assembly.
+
+# .NET 8 Goals
+The .NET 7 release had a 3-4x perf improvement for the existing `Invoke()` APIs by using IL Emit when available and falling back to standard reflection when IL Emit is not available. Although this improvement is significant, it still doesn't replace the need to use the IL-Emit based alternatives (dynamic methods and expression trees) for highly-performance-sensitive scenarios including the `System.Text.Json` serializer. New APIs are required that don't have the overhead of the existing `Invoke()` APIs.
+
+For .NET 8, there are two primary goals:
+1) Support byref-like types both for invoking and passing as arguments; this unblocks various scenarios. An unsafe approach may used for .NET 8 if support for `TypedReference` isn't addressed by Roslyn (covered later).
```
jkotas, line 62: I believe that we know by know that we are not going to get safe TypedReferences from Roslyn in .NET 8.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+        // Return
+        public object? GetReturn()
+        public ref T GetReturn<T>()
+
+        public void SetReturn(object value)
+        public void SetReturn(TypedReference value)
+        public unsafe void SetReturn(void* value, Type type)
+        public void SetReturn<T>(ref T value)
+
+        // Invoke direct (limited validation and defaulting)
```
jkotas, line 131: What does limited validation mean precisely?

I think it would be useful to have fully unsafe API that does not do any validation and just assumes that you have passed in arguments of the right types. Safety and performance wise, such API would be closest to Reflection.Emit where we do not perform any validation either.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
+# Reflection Invoke for 8.0 (draft \ in progress)
+#### Steve Harter
+#### April 13, 2023
+
+# Background
+For additional context, see:
+- [Developers using reflection invoke should be able to use ref struct](https://github.com/dotnet/runtime/issues/45152)
+- [Originally library issue; needs refreshing\updating](https://github.com/dotnet/runtime/issues/10057)
```
RussKie, line 8: Good read overall. Though I found the use of "\\" instead of "/" jarring
```suggestion
- [Originally library issue; needs refreshing/updating](https://github.com/dotnet/runtime/issues/10057)
```

Ditto other places.

accepted/2022/ReflectionInvoke.md
steveharter: Sure - I'll clarify that and move to the "future" section.

accepted/2022/ReflectionInvoke.md
steveharter: What I was thinking was to at least verify values type are the correct type to prevent destabilizing the system in a nondeterministic manner.  This can be done most efficiently in IL with castclass with minimal impact on perf. Verification on reference types should also happen, but I believe this comes for free at least.

However, the exceptions thrown will be CastClassException I believe, not a nice "ArgumentException" unless we do extra work ahead of time to throw ArgumentException or do extra work with a try\catch and back-walk the validation if CastClassException is thrown.

I was also planning on doing "defaulting" for `out` values for value types if no default values were provided by the caller. This is especially common for return values. Unfortunately, it is not easy\fast to see check if a managed reference for an argument is `in` vs `ref` vs `out` so perhaps that should be deferred for now except for the return value which is always `out`.

jkotas: I think it would be useful to have a completely unsafe API as one of the options. The completely unsafe API would not do any argument validation. It would be closest to what one can get with RefEmit where the user is responsible for emitting the right IL and where mistake can destabilize the system too.

> Verification on reference types should also happen, but I believe this comes for free at least.

Validation of reference types is not free. It is more expensive than validation of value types.

davidfowl: I agree, we have the unsafe API as well.

accepted/2022/ReflectionInvoke.md
jkotas: > I'll assume for now "InvokeDirect()` is as fast as possible with perhaps only argument number validated

The set\get parameters have to store the additional type information so that it can maybe validated later. Is that correct? `InvokeDirect` maybe as fast as possible, but the rest is not.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+        public object? GetReturn()
+        public ref T GetReturn<T>()
+
+        public void SetReturn(object value)
+        public void SetReturn(TypedReference value)
+        public unsafe void SetReturn(void* value, Type type)
+        public void SetReturn<T>(ref T value)
+
+        // Invoke direct (limited validation and defaulting)
+        public unsafe void InvokeDirect(MethodBase method)
```
jkotas, line 132: Nit: The unsafe API with non-validation should have `Unsafe` in the name.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+This ref struct is the mechanism to specify the target + arguments (including return value) and supports these mechanisms:
+- `object` (including boxing). Supports loose coupling scenarios are supported, like reflection today.
+- `ref <T>`. Supports new scenarios as mentioned earlier; type must be known ahead-of-time and due to no language support, cannot be a byref-like type like `Span<T>`.
+- `void*`. Unsafe cases used to support byref-like types in an unsafe manner.
+- `TypedReference`. Optional for now; pending language asks, it may make supporting byref-like types a safe operation.
+
+```cs
+namespace System.Reflection
+{
+     public ref struct MethodInvoker
```
jaredpar, line 90: Is this actually a mutable `struct` or was `readonly` omitted from the doc by mistake?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+
+        // Fixed length (say up to 8)
+        public MethodInvoker(ref ArgumentValuesFixed values)
+
+        // Dispose needs to be called with variable-length case
+        public void Dispose()
+
+        // Target
+        public object? GetTarget()
+        public ref T GetTarget<T>()
```
jaredpar, line 106: Did you consider a `GetTarget` method that can return the `TypedReference`? The current design has a lack of symmetry with the `SetTarget` methods.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+     // Internally it is based on 3 IntPtrs that are for 'ref', 'object value' and 'Type':
+     // - 'ref' points to either the 'value' location or a user-provided location with "void*", "ref <T>" or TypedReference.
+     // - 'object value' captures any user-provided object value in a GC-safe manner.
+     // - 'Type' is used in "void*", "ref <T>" and TypedReference cases for validation and in rare cases to prevent
+     //    Types from being GC'd.
+     public struct ArgumentValue { }
+```
+
+## ArgumentValuesFixed
+This class is used for cases where the known arguments are small.
```
jaredpar, line 148:
```suggestion
This struct is used for cases where the known arguments are small.
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+
+```cs
+namespace System.Reflection
+{
+    public ref partial struct ArgumentValuesFixed
```
jaredpar, line 153: Why are you using this struct vs. a `Span<object>` which is backed by an `[InlineArray(8)]` struct that you slice to the appropriate length?

accepted/2022/ReflectionInvoke.md
jaredpar: Think there is a simple design where we have safe APIs for constructing `TypedReference` instances and getting the values back from them. The only type `unsafe` would come into the mix is when you attempted to extract the value of a `ref struct` by `ref`. That would put some burden on the implementation here to be careful with the values but consumers shouldn't need to. 

The conversation on whether or not this will work for the reflection APIs keeps lagging though so it's hard to tell if we should be investing in it.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+        Span<int> heap = new int[42];
+        Span<int> stack = stackalloc int[42];
+        CallMe(ref stack, heap); // okay
+        CallMe(ref heap, stack); // CS8350 as expected
+    }
+
+    public static void ChangeIt(TypedReference tr)
+    {
+        // This compiles today:
+        ref Span<int> s = ref __refvalue(tr, Span<int>);
```
jaredpar, line 398: I think the easiest path forward for `TypedReference` safely holding a `Span<int>` is for this to turn into an `unsafe` operation. This is where the danger really sets in, unless we significantly complicate the existing lifetime rules, hence this is where `unsafe` should go. Essentially this is when the rules break down if we allow capturing `Span<int>` safely. 

Note: specifically I mean when `ref __refvalue(expr, T)` and `T` is a `ref struct`.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+```
+
+## Examples
+### Fixed-length arguments
+```cs
+MethodInfo method = ... // Some method to call
+ArgumentValuesFixed values = new(4); // 4 parameters
+InvokeContext context = new InvokeContext(ref values);
```
cston, line 237: Should `InvokeContext` be `MethodInvoker` in these examples?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+context.InvokeDirect(method);
+```
+
+### Variable-length object arguments
+Unsafe and slightly slower than fixed-length plus requires `using` or `try\finally\Dispose()`.
+```cs
+unsafe
+{
+    ArgumentValue* args = stackalloc ArgumentValue[4];
+    using (InvokeContext context = new InvokeContext(ref args))
```
cston, line 266: Should this be `= new MethodInvoker(args, 4)` instead?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+```cs
+namespace System.Reflection
+{
+     public ref struct MethodInvoker
+     {
+        // Zero-arg case:
+        public MethodInvoker()
+
+        // Variable-length number of arguments:
+        public unsafe MethodInvoker(ArgumentValue* argumentStorage, int argCount)
```
cston, line 96: Would it be possible to use a `Span<ArgumentValue>` instead?
```csharp
public MethodInvoker(Span<ArgumentValue> values)
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+    {
++       [System.CLSCompliantAttribute(false)]
+        public virtual void GetValueDirect(TypedReference obj, TypedReference result);
+
++       [System.CLSCompliantAttribute(false)]
+        public virtual void SetValueDirect(TypedReference obj, TypedReference value);
+
+        // Possible for performance in System.Text.Json:
++       public virtual Func<object, TValue> CreateGetterDelegate<TValue>();
++       public virtual Action<object, TValue> CreateSetterDelegate<TValue>();
```
cston, line 214: Perhaps `CreateGetValueDelegate` and `CreateSetValueDelegate`, to match `GetValue`, `GetValueDirect`, etc.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
```
cston, line 227: Perhaps `CreateGetValueDelegate` and `CreateSetValueDelegate`, to match `GetValue`, `GetValueDirect`, etc.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+        public MethodInvoker()
+
+        // Variable-length number of arguments:
+        public unsafe MethodInvoker(ArgumentValue* argumentStorage, int argCount)
+
+        // Fixed length (say up to 8)
+        public MethodInvoker(ref ArgumentValuesFixed values)
+
+        // Dispose needs to be called with variable-length case
+        public void Dispose()
```
cston, line 102: Why is `Dispose()` needed for the variable-length case but not the fixed-length case?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+#pragma warning disable CS8500    
+    void* ptr = (void*)new IntPtr(&span);
+#pragma warning restore CS8500    
+    // Ideally we can use __makeref(span) instead of the above.
+
+    context.SetArgument(0, ptr, typeof(Span<int>));
+    context.InvokeDirect(method);
+}
+```
+# Design ext
```
cston, line 310: "extension"?

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+2) Support "fast invoke" so that using IL Emit with dynamic methods has little to no performance advantage. Today both `STJ (System.Text.Json)` and [DI (dependency injection)](https://learn.microsoft.com/dotnet/core/extensions/dependency-injection) use IL Emit for performance although DI uses emit through expressions while STJ uses IL Emit directly.
+    - Although the new APIs will be zero alloc (no `object` and boxing required) and performace faster than standard reflection, it will not necessarily be as fast as hand-coded IL Emit for specific scenarios that can optimize for any constraints such as fixed-size list of parameters or by not doing full defaulting and validation of values. Note that property get\set is a subset of this case since there is either a return value (for _get_) or a single parameter for _set_ and since property get\set is such a common case with serialization, this design does propose a separate API for this common case to enable maximum performance.
```
cston, line 64:
```suggestion
    - Although the new APIs will be zero alloc (no `object` and boxing required) and performance faster than standard reflection, it will not necessarily be as fast as hand-coded IL Emit for specific scenarios that can optimize for any constraints such as fixed-size list of parameters or by not doing full defaulting and validation of values. Note that property get\set is a subset of this case since there is either a return value (for _get_) or a single parameter for _set_ and since property get\set is such a common case with serialization, this design does propose a separate API for this common case to enable maximum performance.
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+
+Today, a managed pointer is obtained safely through the `ref` keyword in C#. It references a storage location of an object or value type which can be a stack variable, static variable, parameter, field or array element. If the storage location is a field or array element, the managed pointer is referred to as an "interior pointer" which is supported by GC meaning that GC won't collect the owning object even if there are only interior pointers to it.
+
+A managed pointer can becreated in several ways:
+- [System.TypedReference](https://learn.microsoft.com/en-us/dotnet/api/system.typedreference). Since this a byref-like type, it is only stack-allocated. Internally, it uses a `ref byte` approach along with a reference to a `Type`. Note `TypeReference` is a special type and has its own opcodes (mkrefany, refanytype, refanyval) which translate to C# keyworkds (`__makeref`, `__reftype`, `__refvalue`).
```
cston, line 74:
```suggestion
A managed pointer can be created in several ways:
- [System.TypedReference](https://learn.microsoft.com/en-us/dotnet/api/system.typedreference). Since this a byref-like type, it is only stack-allocated. Internally, it uses a `ref byte` approach along with a reference to a `Type`. Note `TypeReference` is a special type and has its own opcodes (mkrefany, refanytype, refanyval) which translate to C# keywords (`__makeref`, `__reftype`, `__refvalue`).
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+    // However, when enumerating __arglist here, the runtime throws when accessing a byref-like
+    // type although that limitation is easily fixable on Windows (just a runtime limitation; not compiler)
+```
+
+
+# Future
+Holding area of features discussed but not planned yet.
+
+## Variable-length, safe collections
+The API proposal below does have a variable-lenth stack-only approach that uses an internal GC tracking mechanism. A easier-to-pass or callback version is not expected in 8.0; see https://github.com/dotnet/runtime/issues/75349.
```
cston, line 464:
```suggestion
The API proposal below does have a variable-length stack-only approach that uses an internal GC tracking mechanism. A easier-to-pass or callback version is not expected in 8.0; see https://github.com/dotnet/runtime/issues/75349.
```

accepted/2022/ReflectionInvoke.md
steveharter: > The set\get parameters have to store the additional type information so that it can maybe validated later. Is that correct?

Not when using the `object` set overloads. The invoker maintains a tuple **(byreference,** **object,** **type)** for each arg and depending on the set method used and whether the target method has byref parameters. For the simple `object`-only case used by DI, for example, we don't need the `byreference` or the `type` for Core at least.

For perf, Core will have both the byref- and object-based emit. NativeAOT may elect to force everything to byrefs to have just one stub.

steveharter: For DI I'll provide more detail, but given this perf test:
- DI factory logic which is `object`-based with no byref args
- The perf measures calling the factory method. This was earlier obtained from `ActivatorUtilities.CreateFactory(Type, Type[])` but that is not included in the benchmark.
- That factory method was changed to use the new `MethodInvoker`.
- The factory method has 3 parameters, 2 of which are "matched" and 1 is "unmatched" so there's some logic involved in calling the IServiceProvider for the unmatched case.

The breakdown of time spent:
- ~12% for calling the invoke delegate which call the ctor of the target type passing the 3 args.
- ~5% for MethodInvoker.Invoke() overhead
- ~80% for setting the args and preparing to call MethodInvoker.Invoke(). This can vary depending on fixed vs. variable MethodInvoker APIs, but not much. This also includes the DI logic for default values and mapping.

That 80% dwarfs the actual invoke of 12%.

The test ran in ~55ns compared to ~66ns for today's reflection (which uses emit since 7.0). This ~17% gain also includes some other misc optimizations. In summary, this is not getting us much except there is no more alloc for the `object[]` for the args.

Compare this to LINQ with compiled expressions which is ~2.5x faster (~22ns).

Based on this I have a new prototype:
- Added additional Invoke() APIs for fixed-length cases (say up to 8). Added a static MethodInvoker.Invoke(...) or MethodInvoker.GetInvokeDelegate(...) that requires no allocs or setting of values. It will only take `object` args and be focused around existing `MethodBase.Invoke` cases which can only take `object`.
- Perform a manual "AOT" generation of the DI factory lambda code depending how the matching of args occurs. This is based on whether default values are required and the mapping logic. It avoids capturing unnecessary state by the factory's lambda and avoids various logic to check for default values etc.

This new prototype looks promising but will never be as fast as LINQ expressions since that can hard-code the default values and mapping information in the expression. For the happy path, I think I can get it to within 30-50% (so will still be 1.3-1.5x slower, but not 2.5x).

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+
+# Design with managed pointers
+In order to address the limitations and issues, a least-common-denominator approach of using _managed pointers_ for the parameters, target and return value. This supports `in\ref\out` aliasing and the various classifications of types including value types, reference types, pointer types and byref-like types. Essentially this is a different way to achieve "boxing" in order to have a single representation of any parameter.
+
+However, since managed pointers require a reference to a storage location, they does not directly support the same loosely-coupled scenarios as using `object` + boxing for value types. The proposed APIs, however, do make interacting with `object` possible with the new API for the cases that do not require `in\ref\out` variable aliasing for example.
```
jjonescz, line 69:
```suggestion
However, since managed pointers require a reference to a storage location, they do not directly support the same loosely-coupled scenarios as using `object` + boxing for value types. The proposed APIs, however, do make interacting with `object` possible with the new API for the cases that do not require `in\ref\out` variable aliasing for example.
```

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+- Using `ref <T>` for the strongly typed case or `ref byte` for the loosely-typed case. This is expanded in 7.0 due to the new "ref field" support. Previously, there was an internal `ByReference<T>` class that was used and in 7.0 this was changed to `ByReference` in 8.0 which no longer maintains the `<T>` type and internally just contains `ref byte`. This `ByReference` type is used today in reflection invoke when there are <=4 parameters.
+- Using unsafe `void*` (or `IntPtr`) with GC tracking or pinning to make the use GC-safe. Tracking is supported internally through the use of a newer "RegisterForGCReporting()" mechanism. This approach is used today in reflection invoke when there are >=5 parameters.
+
+# Proposed APIs
+
+## MethodInvoker
+This ref struct is the mechanism to specify the target + arguments (including return value) and supports these mechanisms:
+- `object` (including boxing). Supports loose coupling scenarios are supported, like reflection today.
```
jjonescz, line 82:
```suggestion
- `object` (including boxing). Loose coupling scenarios are supported, like reflection today.
```

accepted/2022/ReflectionInvoke.md
jjonescz, line 102: I'm wondering if having Dispose which is only required in some cases won't trigger false positives in disposable analyzers.

accepted/2022/ReflectionInvoke.md
steveharter: In most other areas we don't add "Unsafe" since it is a bit redundant since it's in the signature and requires unsafe context to call. I assume then "Unsafe" makes it clearer that this case is more dangerous than the others.

accepted/2022/ReflectionInvoke.md
```diff
@@ -0,0 +1,474 @@
...
+        public ArgumentValuesFixed(object? obj1, object? o2, object? o3, object? o4, object? o5, object? o6, object? o7, object? o8)
+    }
+}
+```
+
+## TypedReference
+This is currently optional and being discussed. If `TypedReference` ends up supporting references to byref-like types like `Span<T>` then it will be much more useful otherwise just the existing `ref <T>` API can be used instead. The advantage of `TypedReference` is that it does not require generics so it can be made to work with `Span<T>` easier than adding a feature that would allowing generic parameters to be a byref-like type.
+
+To avoid the use of C#-only "undocumented" keywords, wrappers for `__makeref`, `__reftype`, `__refvalue` which also enable other languages.
+```diff
```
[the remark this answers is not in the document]
steveharter: Because TypedReference is an existing class with existing members.

accepted/2022/ReflectionInvoke.md
steveharter: Yes it is mutable.

accepted/2022/ReflectionInvoke.md
steveharter: Yes it should be added; currently that is a CS1599 (can't return a TypedReference) and I didn't call it out.

However, I'm pushing these TypedReference cases to the "future" section for now.

accepted/2022/ReflectionInvoke.md
steveharter: That struct actually has 3 different types of storage (object, ByReference, and Type) and they need to be grouped something like this in order to register efficiently with the GC tracking support:
```cs
public unsafe ref struct ArgumentValuesFixed
  {
      internal object? _obj0;
      internal object? _obj1-to-lengthl

      internal RuntimeType? _type0;
      internal RuntimeType? _type1-to-lengthl

      internal ref byte _ref0;
      internal ref byte _ref1-to-lengthl
}
```

Also, I moved this type to the "future" section and will add it back, if necessary, based on perf.

accepted/2022/ReflectionInvoke.md
steveharter: Yes thanks.

accepted/2022/ReflectionInvoke.md
steveharter: Perhaps; for now I moved this to the "future" section since it is somewhat speculative based on perf.

accepted/2022/ReflectionInvoke.md
steveharter: The variable-length case hooks up in a low level to the runtime to track these references to support the GC and needs to be unregistered once finished.

accepted/2022/ReflectionInvoke.md
steveharter: Will change to "addendum"; just placing less-important details and findings in an area that isn't up-front and cluttering.

accepted/2022/ReflectionInvoke.md
steveharter: Good question.

accepted/2022/ReflectionInvoke.md
steveharter: Updated: the `InvokeDirectUnsafe()` APIs won't do any validation.

accepted/2022/ReflectionInvoke.md
steveharter: Thanks. Make sense.

accepted/2022/ReflectionInvoke.md
steveharter: (dup of above)

PR #289, opened 2023-02-16
net8.0-browser TFM

javiercn:
Adds a proposal for creating a browser specific TFM.

[Rendered](https://github.com/dotnet/designs/blob/40b628e100a8b82bdc92dcdf210a8743e30128c0/accepted/2023/net8.0-browser-tfm.md)

### DIFF
--- a/INDEX.md
+++ b/INDEX.md
@@ -81,7 +81,7 @@ Use update-index to regenerate it:
 | 2021 | [TFM for .NET nanoFramework](accepted/2021/nano-framework-tfm/nano-framework-tfm.md) | [Immo Landwerth](https://github.com/terrajobst), [Laurent Ellerbach](https://github.com/Ellerbach), [José Simões](https://github.com/josesimoes) |
 | 2021 | [Tracking Platform Dependencies](accepted/2021/platform-dependencies/platform-dependencies.md) | [Matt Thalman](https://github.com/mthalman) |
 | 2022 | [.NET 7 Version Selection Improvements](accepted/2022/version-selection.md) | [Rich Lander](https://github.com/richlander) |
-| 2023 | [net8.0-browser TFM for applications running in the browser](accepted/2023/net8.0-browser-tfm.md) | [Javier Calvarro](https://github.com/javiercn), [Javier Calvarro](https://github.com/javiercn) |
+| 2023 | [net8.0-browser TFM for applications running in the browser](accepted/2023/net8.0-browser-tfm.md) | [Javier Calvarro](https://github.com/javiercn) |
 
 ## Drafts
 
--- a/INDEX.md
+++ b/INDEX.md
@@ -81,6 +81,7 @@ Use update-index to regenerate it:
 | 2021 | [TFM for .NET nanoFramework](accepted/2021/nano-framework-tfm/nano-framework-tfm.md) | [Immo Landwerth](https://github.com/terrajobst), [Laurent Ellerbach](https://github.com/Ellerbach), [José Simões](https://github.com/josesimoes) |
 | 2021 | [Tracking Platform Dependencies](accepted/2021/platform-dependencies/platform-dependencies.md) | [Matt Thalman](https://github.com/mthalman) |
 | 2022 | [.NET 7 Version Selection Improvements](accepted/2022/version-selection.md) | [Rich Lander](https://github.com/richlander) |
+| 2023 | [net8.0-browser TFM for applications running in the browser](accepted/2023/net8.0-browser-tfm.md) | [Javier Calvarro](https://github.com/javiercn), [Javier Calvarro](https://github.com/javiercn) |
 
 ## Drafts
 
--- a/accepted/2023/net8.0-browser-tfm.md
+++ b/accepted/2023/net8.0-browser-tfm.md
@@ -5,8 +5,8 @@
 **Contact** [Artak Mkrtchyan](https://github.com/mkartakmsft) | [Artak Mkrtchyan](https://github.com/mkartakmsft)
 
 For a video introduction to what we are trying to achieve, see:
-https://www.youtube.com/watch?v=48G_CEGXZZM
 https://www.youtube.com/live/kIH_Py8ytlY?feature=share&t=473
+https://www.youtube.com/watch?v=48G_CEGXZZM
 
 We have been shipping .NET in the browser via webassembly since the introduction of Blazor 3.2.0. During that initial release we decided to avoid adding a TFM for the browser as we were not clear on whether we needed it and we knew it was a one way operation. Once we introduced a TFM, we would never be able to take it back.
 
@@ -27,9 +27,9 @@ Over time, we have learned that the lack of a TFM is limiting for us and our cus
     class IJSInProcessRuntime {
       <<interface>>
       Invoke<&zwj;TResult>(string identifier, params object?[]? args) TResult
-    }    
+    }
   ```
-* **Varying framework/library code contents/behavior based on whether it's running in the browser or not:** For example, having DI dependencies or other library code references varying by platform. Doing it based on TFM (not on RuntimeInformation) helps compilation and trimming. 
+* **Varying framework/library code contents/behavior based on whether it's running in the browser or not:** For example, having DI dependencies or other library code references varying by platform. Doing it based on TFM (not on RuntimeInformation) helps compilation and trimming.
   * A library that targets Blazor Server and Webassembly must provide at a minimum an additional package for the server to avoid bringing the server dependencies to webassembly.
   * A webassembly app with a companion server backend must also have, a project to host the webassembly app, a project to share common abstractions and a project to host the server backend. This exact scheme is reflected in the default Blazor Webassembly Hosted template.
 
@@ -65,7 +65,7 @@ Finally, we are looking at two new core experiences for which a TFM is heavily d
   ```csharp
   public class RecipesStore
   {
-  #if BROWSER  
+  #if BROWSER
       private readonly HttpClient _http;
   #else
       IDictionary<string, Recipe> recipes;
@@ -76,7 +76,7 @@ Finally, we are looking at two new core experiences for which a TFM is heavily d
   #if !BROWSER
       public RecipesStore()
       {
-        ...        
+        ...
       }
   #else
       public RecipesStore(HttpClient http)
@@ -85,7 +85,7 @@ Finally, we are looking at two new core experiences for which a TFM is heavily d
       }
   #endif
 
-  #if BROWSER        
+  #if BROWSER
       public async Task<string> AddImage(Stream imageData)
       {
           var response = await _http.PostAsync("api/images", new StreamContent(imageData));
@@ -126,7 +126,7 @@ Finally, we are looking at two new core experiences for which a TFM is heavily d
       public Task<byte[]> GetImage(string filename)
           => ...;
   #endif
-  }    
+  }
   ```
 
 ## Requirements
--- a/accepted/2023/net8.0-browser-tfm.md
+++ b/accepted/2023/net8.0-browser-tfm.md
@@ -4,9 +4,7 @@
 **Contact** [Daniel Roth](https://github.com/danroth27) | [Daniel Roth](https://github.com/danroth27)
 **Contact** [Artak Mkrtchyan](https://github.com/mkartakmsft) | [Artak Mkrtchyan](https://github.com/mkartakmsft)
 
-For a video introduction to what we are trying to achieve, see:
-https://www.youtube.com/live/kIH_Py8ytlY?feature=share&t=473
-https://www.youtube.com/watch?v=48G_CEGXZZM
+For a video introduction to what we are trying to achieve, see [this demo from the community standup](https://www.youtube.com/live/kIH_Py8ytlY?feature=share&t=473) as well as the [Blazor United prototype](https://www.youtube.com/watch?v=48G_CEGXZZM).
 
 We have been shipping .NET in the browser via webassembly since the introduction of Blazor 3.2.0. During that initial release we decided to avoid adding a TFM for the browser as we were not clear on whether we needed it and we knew it was a one way operation. Once we introduced a TFM, we would never be able to take it back.
 
--- a/accepted/2023/net8.0-browser-tfm.md
+++ b/accepted/2023/net8.0-browser-tfm.md
@@ -184,6 +184,23 @@ The general design guidelines for this feature follow the same principles as the
 
 The ASP.NET Core team will build a similar experience on top of the TFM with similar conventions.
 
+The work that needs to happen in NuGet is ensuring that the friendly parses into a target platform of `browser`. The version is assumed to be non-existing but should follow what we have been doing for the (already existing) runtime identifier (RID).
+
+| TFM Component | Value            |
+| ------------- | ---------------- |
+| Friendly name | `net8.0-browser` |
+| TPI           | `browser`        |
+
+### Open design challenges
+
+* Is there any TFM work on the IDE side?
+  - For example, there are drop downs that show frameworks (editor as well as project properties). Do they need a display name?
+* Are there any existing APIs that should be moved from `net7.0` to `net8.0-browser`?
+* This document focuses on the TFM, but we need to think about the multi-targeting experience in the IDE
+  - Logically the server needs to reference the client (because the client is downloaded from the server). With this proposal the intent is to have a single multi-targeted project where the server output (`net8.0`) depends on the client output (`net8.0-browser`)
+  - Seems we want some amount of "streamlining" in the IDE/build experience, for instance by having a single output where the browser output is nested under the server output, making F5 work without being able to accidentally launch the wrong side. There are also other aspects where the IDE prunes the TFM graph, for instance, when running unit tests. We'd have to make sure that these features work, but all of these are multi-targeting IDE experience issues not issues with having a TFM.
+  - The sample code above also used conditions in the build file which are limited and intrinsic functions are verbose. We should consider making this more natural in MSBuild, not just for this experience but for multi-targeting in general. One idea was to make certain properties intrinsic and have the build engine parse their values from the `TargetFramework` property on demand.
+
 ## Q & A
 
 ## Can `net8.0` reference `net8.0-browser`?
--- a/accepted/2023/net8.0-browser-tfm.md
+++ b/accepted/2023/net8.0-browser-tfm.md
@@ -145,7 +145,7 @@ The general rule will be that this TFM will work in the same way as net8.0-andro
 **TL;DR**
 * Per browser TFM
 * Per browser version TFM
-* Browser specific API sets (although might be expanded in the future to cover this)
+* In-box browser specific API sets (although might be expanded in the future to cover this)
 
 #### Details
 
--- a/accepted/2023/net8.0-browser-tfm.md
+++ b/accepted/2023/net8.0-browser-tfm.md
@@ -1,16 +1,16 @@
 # net8.0-browser TFM for applications running in the browser
 
-**Owner** [Javier Calvarro](https://github.com/javiercn) | [Javier Calvarro](https://github.com/javiercn)
-**Contact** [Daniel Roth](https://github.com/danroth27) | [Daniel Roth](https://github.com/danroth27)
-**Contact** [Artak Mkrtchyan](https://github.com/mkartakmsft) | [Artak Mkrtchyan](https://github.com/mkartakmsft)
+**Owner** [Javier Calvarro](https://github.com/javiercn) |
+**Contact** [Daniel Roth](https://github.com/danroth27) |
+**Contact** [Artak Mkrtchyan](https://github.com/mkartakmsft) |
 
 For a video introduction to what we are trying to achieve, see [this demo from the community standup](https://www.youtube.com/live/kIH_Py8ytlY?feature=share&t=473) as well as the [Blazor United prototype](https://www.youtube.com/watch?v=48G_CEGXZZM).
 
 We have been shipping .NET in the browser via webassembly since the introduction of Blazor 3.2.0. During that initial release we decided to avoid adding a TFM for the browser as we were not clear on whether we needed it and we knew it was a one way operation. Once we introduced a TFM, we would never be able to take it back.
 
-Our reasoning was that we could likely get away with a RID as that would allow third-parties to provide alternative pre-compiled webassembly assets (including native dependencies compiled to wasm).
+Our reasoning was that we could likely get away with a RID alone as that would allow third-parties to provide alternative pre-compiled webassembly assets (including native dependencies compiled to wasm).
 
-We reasoned that we did not need a TFM as we could always provide a PAL over any functionality we needed and we would annotate the APIs with the `[SupportedOSPlatform]` attribute and an analyzer that would warn against incorrect usage when targeting the browser.
+We reasoned that we did not need a TFM as we could always provide a PAL over any functionality we needed and we would annotate the APIs with the `[SupportedOSPlatform]` attribute which would warn against incorrect usage when targeting the browser.
 
 Over time, we have learned that the lack of a TFM is limiting for us and our customers. The lack of a TFM introduces friction in two ways:
 * **Exposing additional customer-facing APIs when running in browser (and not on desktop)**: For example, Blazor includes many platform agnostic abstractions like IJSRuntime, IJSInProcessRuntime and different implementations to be able to reflect the different capabilities of the different platforms. This makes taking advantage of some of the webassembly unique capabilities more challenging as this kind of functionality needs to be carefully designed, and creates other problems like lack of discoverability (you need to know the existence of IJSInProcessRuntime and downcast to it).
@@ -172,7 +172,11 @@ So in general, any API that has been supported by browsers for at least 1 year i
 
 ## Stakeholders and Reviewers
 
-@lewing, @pavelsavara, @marek-safar, @terrajobst, more to be added here.
+- Runtime
+- ASP.NET Core
+- Web Tools 
+- SDK
+- NuGet
 
 ## Design
 
@@ -182,4 +186,6 @@ The ASP.NET Core team will build a similar experience on top of the TFM with sim
 
 ## Q & A
 
-What will be the outcome of a net8.0 project referencing a net8.0-browser? (No multi-targeting)
+## Can `net8.0` reference `net8.0-browser`?
+
+No. The compatibility rules for platform-based TFMs are syntax based. `net8.0-browser` will be able to reference `net8.0` but not vice versa.
--- a/accepted/2023/net8.0-browser-tfm.md
+++ b/accepted/2023/net8.0-browser-tfm.md
@@ -1,0 +1,187 @@
+# net8.0-browser TFM for applications running in the browser
+
+**Owner** [Javier Calvarro](https://github.com/javiercn) | [Javier Calvarro](https://github.com/javiercn)
+**Contact** [Daniel Roth](https://github.com/danroth27) | [Daniel Roth](https://github.com/danroth27)
+**Contact** [Artak Mkrtchyan](https://github.com/mkartakmsft) | [Artak Mkrtchyan](https://github.com/mkartakmsft)
+
+For a video introduction to what we are trying to achieve, see:
+https://www.youtube.com/watch?v=48G_CEGXZZM
+https://www.youtube.com/live/kIH_Py8ytlY?feature=share&t=473
+
+We have been shipping .NET in the browser via webassembly since the introduction of Blazor 3.2.0. During that initial release we decided to avoid adding a TFM for the browser as we were not clear on whether we needed it and we knew it was a one way operation. Once we introduced a TFM, we would never be able to take it back.
+
+Our reasoning was that we could likely get away with a RID as that would allow third-parties to provide alternative pre-compiled webassembly assets (including native dependencies compiled to wasm).
+
+We reasoned that we did not need a TFM as we could always provide a PAL over any functionality we needed and we would annotate the APIs with the `[SupportedOSPlatform]` attribute and an analyzer that would warn against incorrect usage when targeting the browser.
+
+Over time, we have learned that the lack of a TFM is limiting for us and our customers. The lack of a TFM introduces friction in two ways:
+* **Exposing additional customer-facing APIs when running in browser (and not on desktop)**: For example, Blazor includes many platform agnostic abstractions like IJSRuntime, IJSInProcessRuntime and different implementations to be able to reflect the different capabilities of the different platforms. This makes taking advantage of some of the webassembly unique capabilities more challenging as this kind of functionality needs to be carefully designed, and creates other problems like lack of discoverability (you need to know the existence of IJSInProcessRuntime and downcast to it).
+  ```mermaid
+  classDiagram
+    IJSRuntime <|-- IJSInProcessRuntime
+    class IJSRuntime {
+      <<interface>>
+      InvokeAsync<&zwj;TValue>(string identifier, object[] args) ValueTask~TValue~
+      InvokeAsync<&zwj;TValue>(string identifier, CancellationToken cancellationToken, object?[]? args) ValueTask~TValue~
+    }
+    class IJSInProcessRuntime {
+      <<interface>>
+      Invoke<&zwj;TResult>(string identifier, params object?[]? args) TResult
+    }    
+  ```
+* **Varying framework/library code contents/behavior based on whether it's running in the browser or not:** For example, having DI dependencies or other library code references varying by platform. Doing it based on TFM (not on RuntimeInformation) helps compilation and trimming. 
+  * A library that targets Blazor Server and Webassembly must provide at a minimum an additional package for the server to avoid bringing the server dependencies to webassembly.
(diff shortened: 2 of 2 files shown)


accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+  * A library that targets Blazor Server and Webassembly must provide at a minimum an additional package for the server to avoid bringing the server dependencies to webassembly.
+  * A webassembly app with a companion server backend must also have, a project to host the webassembly app, a project to share common abstractions and a project to host the server backend. This exact scheme is reflected in the default Blazor Webassembly Hosted template.
+
+One of the reasons we were hesitant to ask for a TFM in the initial release of Blazor was that we were concerned about people writing .NET webassembly specific code that did not work with Blazor Server. Over time, we have determined that it is not a real risk for two reasons:
+* Blazor Server has a healthy user base, which makes it desirable for library authors to target net8.0.
+* A core piece of Blazor functionality (prerendering) depends on being able to run the application on net8.0.
+
+Given this, we consider unlikely that developers will author wasm specific code (net8.0-wasm) without also including a net8.0 version, as it will make it incompatible with any other Blazor flavor.
```
pavelsavara, line 40: I think that people will create non-blazor-rendering client side C# components which have strong JavaScript dependency, which would not be able to run on server side. Should they use `net8.0-wasm` ?

Example would be C# wrapper for WebGL.

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+
+* Enable single project experiences for Blazor Webassembly applications that target ASP.NET Core environments and the browser with the same codebase.
+* Enable library authors to have dependencies thar are tailored for the browser.
+
+### Non-Goals
+
+**TL;DR**
+* Per browser TFM
+* Per browser version TFM
+* Browser specific API sets (although might be expanded in the future to cover this)
```
pavelsavara, line 150: I guess NodeJS is just another "browser" flavor worth mentioning ?

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+
+#### Details
+
+It is well-known that the web ecosystem is a very large, diverse and changing landscape. There are multiple browsers with different sets of supported APIs and APIs are added and removed on a regular basis, which introduces risks for the stability of the platform.
+
+For these reasons, we are going to be very conservative about shipping additional APIs as part of the runtime and define a very restrictive criteria for when we think it is acceptable to add a new API. One such criteria can be:
+* The API is part of the current web standards.
+* All major vendors (Chrome, Firefox, Safari) ship an implementation of the API.
```
pavelsavara, line 158: NodeJS too ?

accepted/2023/net8.0-browser-tfm.md
javiercn: If it can only run in the browser that would be fine. For running on regular .NET they could provide an alternative version that targets .NET 8.0 in that case.

accepted/2023/net8.0-browser-tfm.md
javiercn: I would not bundle node with other browsers, it has a different set of APIs and capabilities, we should handle that separately in the future

accepted/2023/net8.0-browser-tfm.md
javiercn: Same comment as above, node is a different beast.

accepted/2023/net8.0-browser-tfm.md
pavelsavara: They could or they have to, to keep it Blazor compatible ?

javiercn: Only if they wanted to run on Blazor Server or do Server Side Rendering, but it would be reasonable for a standalone Blazor app

accepted/2023/net8.0-browser-tfm.md
pavelsavara: That's not how runtime treats it today. Do we need yet another TFM for it ?
@lewing what's your opinion ?

accepted/2023/net8.0-browser-tfm.md
javiercn: > Right now, we don't have that limitation. We could have the same assembly to run on both client and server side and just have conditional code.

You can still have that decision made at runtime if you use net8.0, won't you? The only requirement is that the APIs you need are available as a net8.0 target.

The difference is that the decision gets made at runtime, not compile time, and you do not get to differentiate the set of dependencies.

pavelsavara: If you use `net8.0` you would not be able to use the APIs of other assemblies which target `net8.0-browser`, right ?

INDEX.md
```diff
@@ -81,6 +81,7 @@ Use update-index to regenerate it:
 | 2021 | [TFM for .NET nanoFramework](accepted/2021/nano-framework-tfm/nano-framework-tfm.md) | [Immo Landwerth](https://github.com/terrajobst), [Laurent Ellerbach](https://github.com/Ellerbach), [José Simões](https://github.com/josesimoes) |
 | 2021 | [Tracking Platform Dependencies](accepted/2021/platform-dependencies/platform-dependencies.md) | [Matt Thalman](https://github.com/mthalman) |
 | 2022 | [.NET 7 Version Selection Improvements](accepted/2022/version-selection.md) | [Rich Lander](https://github.com/richlander) |
+| 2023 | [net8.0-browser TFM for applications running in the browser](accepted/2023/net8.0-browser-tfm.md) | [Javier Calvarro](https://github.com/javiercn), [Javier Calvarro](https://github.com/javiercn) |
```
terrajobst, line 84:
```suggestion
| 2023 | [net8.0-browser TFM for applications running in the browser](accepted/2023/net8.0-browser-tfm.md) | [Javier Calvarro](https://github.com/javiercn) |
```

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
+# net8.0-browser TFM for applications running in the browser
+
+**Owner** [Javier Calvarro](https://github.com/javiercn) | [Javier Calvarro](https://github.com/javiercn)
```
terrajobst, line 3:
```suggestion
**Owner** [Javier Calvarro](https://github.com/javiercn) |
```

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+**Contact** [Daniel Roth](https://github.com/danroth27) | [Daniel Roth](https://github.com/danroth27)
```
terrajobst, line 4:
```suggestion
**Contact** [Daniel Roth](https://github.com/danroth27) |
```

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+**Contact** [Artak Mkrtchyan](https://github.com/mkartakmsft) | [Artak Mkrtchyan](https://github.com/mkartakmsft)
```
terrajobst, line 5:
```suggestion
**Contact** [Artak Mkrtchyan](https://github.com/mkartakmsft) |
```

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+
+Our reasoning was that we could likely get away with a RID as that would allow third-parties to provide alternative pre-compiled webassembly assets (including native dependencies compiled to wasm).
+
+We reasoned that we did not need a TFM as we could always provide a PAL over any functionality we needed and we would annotate the APIs with the `[SupportedOSPlatform]` attribute and an analyzer that would warn against incorrect usage when targeting the browser.
+
+Over time, we have learned that the lack of a TFM is limiting for us and our customers. The lack of a TFM introduces friction in two ways:
+* **Exposing additional customer-facing APIs when running in browser (and not on desktop)**: For example, Blazor includes many platform agnostic abstractions like IJSRuntime, IJSInProcessRuntime and different implementations to be able to reflect the different capabilities of the different platforms. This makes taking advantage of some of the webassembly unique capabilities more challenging as this kind of functionality needs to be carefully designed, and creates other problems like lack of discoverability (you need to know the existence of IJSInProcessRuntime and downcast to it).
```
terrajobst, line 18: I don't understand how having a single TFM would help with disjoint APIs sets here?

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+      <<interface>>
+      InvokeAsync<&zwj;TValue>(string identifier, object[] args) ValueTask~TValue~
+      InvokeAsync<&zwj;TValue>(string identifier, CancellationToken cancellationToken, object?[]? args) ValueTask~TValue~
+    }
+    class IJSInProcessRuntime {
+      <<interface>>
+      Invoke<&zwj;TResult>(string identifier, params object?[]? args) TResult
+    }
+  ```
+* **Varying framework/library code contents/behavior based on whether it's running in the browser or not:** For example, having DI dependencies or other library code references varying by platform. Doing it based on TFM (not on RuntimeInformation) helps compilation and trimming.
```
terrajobst, line 32: I thought our linker is able to tell that the return values for `OSPlatform.IsBrowser()` is a constant. Is that not true?

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+
+* Projects can have conventions based on the TFM (like MAUI)
+  ![image](https://user-images.githubusercontent.com/6995051/219371569-0699ae29-8bbf-4e93-b1e1-904dc6e3616d.png)
+
+* Projects can have specific dependencies for the browser (like MAUI)
+  ```xml
+  <ItemGroup Condition="'$(Browser)' == 'true'">
+    <Reference Include="Microsoft.AspNetCore.Components.WebAssembly" />
+    <Reference Include="System.Net.Http.Json" />
+  </ItemGroup>
```
terrajobst, line 62: Does this syntax actually work? I thought we can't do that due to MSBuild evaluation order. /cc @dsplaisted

You can vary dependencies on the platform, but I think you'll need to use the MSBuild intrinsic functions for parsing the platform name out of the TFM.

accepted/2023/net8.0-browser-tfm.md
javiercn: @pavelsavara but you do not offer different APIs tailored for node either, isn't it?

I would be conservative here about introducing a TFM for the same reasons we were before in the initial wasm versions. We can always add it later, but we can't take it back.

I would not expect us to have any sort of relationship between `net8.0-browser` and `net8.0-(xxx)` even if some APIs might be similar (like fetch). As they are in general different environments and evolve at different speeds for different needs.

It would come down to how many users you have that want to target node, and potentially also provide a different API shape and dependencies for the browser.

This proposal does not change what the runtime will do today. The runtime can continue shipping as net8.0 (as it does today for Maui) but the TFM gives the ecosystem the ability to choose dependencies, API shapes, and provide different implementations in a natural way. 

Apps can target net8.0-browser and get access to browser specific APIs and implementations provided by packages (similar to Maui).

If later on we want to provide a set of out of the box APIs, that becomes a different story.

accepted/2023/net8.0-browser-tfm.md
dsplaisted: This will likely work, though I don't know how the `Browser` property is set.  It works because item evaluation happens in a later phase than property evaluation.  So if this was properties in a `PropertyGroup` it wouldn't work.

So yes it works in this case, but it's also confusing and if people see this pattern they will probably try to apply it to properties where it won't work.

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+
+## Stakeholders and Reviewers
+
+@lewing, @pavelsavara, @marek-safar, @terrajobst, more to be added here.
+
+## Design
+
+The general design guidelines for this feature follow the same principles as the TFMs for other environments like ios, maccatalyst, android and windows.
+
+The ASP.NET Core team will build a similar experience on top of the TFM with similar conventions.
```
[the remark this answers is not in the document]
dsplaisted: Just a thought, maybe we should go with `1.0` for the `TargetPlatformVersion`.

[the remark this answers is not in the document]
terrajobst: I think either one works; but since we have zero plans to add more versions, `0.0` felt a tad more natural to me.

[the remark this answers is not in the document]
JonDouglas: Forgive my completely ignorant take here.

Why name it browser? Why not wasm or webassembly given it is more straightforward to match versions.

Is the v1.0 not related to the 1.0 of wasm?

I'd make the general assumption that when 2.0 comes out, we might ship a newer supported version too?

I.e. https://webassembly.github.io/spec/core/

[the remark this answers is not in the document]
SteveSandersonMS: > Why name it browser? Why not wasm or webassembly given it is more straightforward to match versions.

We've found it useful to think of WebAssembly as being more akin to a CPU architecture than to an API surface. This especially makes sense when including non-browser WebAssembly environments (e.g., WASI). There are at least 3 different major scenarios for WebAssembly:

 * Pure WebAssembly as per its MVP spec alone. This doesn't entail *any* API support at all. The only APIs available are custom ones that the developer provides.
 * WASI. This implies a certain set of server-oriented APIs.
 * .NET in the browser (WebAssembly via Emscripten). This provides the widest range of APIs and is what we use for Blazor WebAssembly. **The proposed net8.0-browser TFM is for this scenario in particular, and not the two above**.

There will be yet more scenarios over time, especially as WebAssembly itself evolves, the component model reaches critical mass, and so on. But in general we can't assume there's any useful single lowest-common-denominator.

accepted/2023/net8.0-browser-tfm.md
SteveSandersonMS: Could you clarify what disjoint API sets you mean? In case you mean "some things are server-only, and some things are browser-only, and neither is a subset of the other" then that's true.

Here's an example of how things are in practice today:

 * `net7.0`
   * `System.Net.Sockets.Socket.Listen()` - works if you use it on server, throws at runtime if you try to use it in browser
   * `Microsoft.JSInterop.IJSInProcessRuntime` - works if you use it in browser, throws at runtime if you try to use it in server (i.e., when you try to downcast `IJSRuntime` to it)

... whereas here's how things *could* have been with a browser TFM:

 * `net7.0`
   * `System.Net.Sockets.Socket.Listen()` - works if you use it on server, throws at runtime if you try to use it in browser
 * `net7.0-browser`
   * `Microsoft.JSInterop.IJSInProcessRuntime` - works if you use it in browser. You can't accidentally try to use it in server because you're not targeting this TFM.

So, adding a browser TFM is not a 100% solution because TFMs just don't model this kind of scenario exactly. Adding the TFM solves *some* API misuse issues but not all of them. It's the best we can do with a TFM, and is an improvement over where we are today.

But as a recap, solving API misuse issues is not the main driving reason to do this - it's all the application model stuff Javier mentioned in this doc.

I hope this is what you're talking about. Apologies if I've completely misinterpreted your comment/question.

accepted/2023/net8.0-browser-tfm.md
SteveSandersonMS: In certain *very carefully balanced* cases it's true that the linker may resolve `OSPlatform.IsBrowser()`, discard some `if`/`else` branches, and consequently discard entire assembly references. That's great if it works out.

However in practice it's really difficult to make this happen reliably. If you assign `OSPlatform.IsBrowser()` to a variable then check based on that, or if you have some other field referencing a type that's only in one of these dependencies, or if you simply try to publish an app without trimming (which app developers are always within their rights to do) then this level of reliance on the linker will get you into trouble.

If we were going to rely on this, we would need better infrastructure to assert at publish time that certain assemblies must not end up in the output or that only a specified set can do so.

And even then, it wouldn't help guide application developers to avoid referencing types that aren't usable on their target platform. For example, we may reference EF conditionally based on your TFM, and we want developers to know at compile time that they can only use those EF APIs in the `net8.0` build and not in the `net8.0-browser` build.

accepted/2023/net8.0-browser-tfm.md
SteveSandersonMS: Thanks for bringing this up.

If there are concerns about the correctness or reliability of this technique it would be great to dig into that, since we took this approach in our prototype (where it appears to work perfectly) and are taking a reliance on it as a core part of how these projects will be structured. I *thought* this was equivalent to what MAUI does but if this is mistaken we should try to clarify.

Is there a different/better way to vary things based on TFM?

javiercn: I read up a bit on this last night, the thing that always works is the TFM if I am not mistaken, so we won't use the `$(Browser) == 'true'` pattern but the `$(TargetFramework) == 'net8.0-browser'` pattern, and that should work.

accepted/2023/net8.0-browser-tfm.md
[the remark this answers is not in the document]
JonDouglas: Makes sense. Thanks for the context. Maybe that should also go in the proposal for future archival purposes? The only follow up id have is whether the tfm name is intuitive to the developer and what they think it does. But that's more of a UX research / survey thing.

accepted/2023/net8.0-browser-tfm.md
dsplaisted: The problem with `$(TargetFramework) == 'net8.0-browser'` is that it stops working when you retarget to .NET 9 (`net9.0-browser`).

The "right" way to do it is `'$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework)))' == 'browser'`, but that's quite ugly to put in a project file.

So there's no great solution.  If the main scenario you see for this is to modify items and not properties, it might be OK to go with `'$(Browser)' == 'true'` or `'$(TargetPlatformIdentifier)' == 'browser'`.

javiercn: @dsplaisted fair enough, I don't remember the exact details of how we do it in Maui, but in any case, I believe we can have the "cleaner" pattern for dependencies if we want to.

Maui does the "ugly" suggestion in the template, for reference the Blazor Maui template (although it's the same on the maui one)
https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-blazor/MauiApp.1.csproj#L25-L30

javiercn: I suspect people in their apps, very likely use the target framework directly and do a round of find and replace when they update.

accepted/2023/net8.0-browser-tfm.md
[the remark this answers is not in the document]
javiercn: Yes, that's a good clarification.

```suggestion
* In-box browser specific API sets (although might be expanded in the future to cover this)
```

accepted/2023/net8.0-browser-tfm.md
dsplaisted: Keep in mind that some people may retarget using the Visual Studio project properties UI, and not realize that they need to  open the project file and update other places manually.

accepted/2023/net8.0-browser-tfm.md
javiercn: > If you use net8.0 you would not be able to use the APIs of other assemblies which target net8.0-browser, right ?

I think "use" needs clarification here. Strictly speaking you won't be able to reference a package/project that targets `net8.0-wasm` from a `net8.0` project. Similarly speaking, if you author a library for `net8.0-wasm` only, it can be consumed by anyone targeting `net8.0`.

Now, if your app targets net8.0-wasm, it can consume libraries that target either net8.0 or net8.0-wasm, and if your app targets net8.0, then it won't be able to consume anything that exclusively targets net8.0-wasm.

From the application developer perspective, if they want to consume bits that are only available in net8.0-wasm, they should be able to change their TFM to target net8.0-wasm and everything should continue to compile since they would be targeting a superset of net8.0.

accepted/2023/net8.0-browser-tfm.md
javiercn: > Keep in mind that some people may retarget using the Visual Studio project properties UI, and not realize that they need to open the project file and update other places manually.

That's fair, we'll have to check how Maui deals with these types of things, we might follow the same approach and use the MSBuild expression or provide a helper property. We'll have to make a call when we figure out the details for the template.

@dsplaisted won't the property work against other properties if we define it earlier in the props file for the SDK? AFAIK, properties still get evaluated in the order in which they appear in the file, isn't it? I know there are like 6 stages of evaluation, so I understand that it might be more nuanced than what I suggested.

accepted/2023/net8.0-browser-tfm.md
```diff
@@ -0,0 +1,187 @@
...
+Browsers vendors have adopted a very aggressive support policy and as such, it is fair to say that new framework versions do not necesarily need to worry about older browser versions. As an example:
+* Edge support policy for enterprise customers [here](https://learn.microsoft.com/en-us/deployedge/microsoft-edge-support-lifecycle#service-and-assisted-support-timeline)
+* Firefox support policy [here](https://support.mozilla.org/en-US/kb/choosing-firefox-update-channel?_gl=1*12w45u6*_ga*MTM1MTA0NzU1MS4xNjcxMTk4MDQz*_ga_MQ7767QQQW*MTY3MTE5OTUyMy4xLjAuMTY3MTE5OTUyMy4wLjAuMA..)
+* Safari support policy is typically linked to the OS support policy and reaches the two last major OS versions.
+
+So in general, any API that has been supported by browsers for at least 1 year is a candidate to be added if we choose to. With that said, for .NET 8.0 we do not necessarily want to add additional browser APIs in the runtime itself, just enable developers to leverage this capability.
+
+## Stakeholders and Reviewers
+
+@lewing, @pavelsavara, @marek-safar, @terrajobst, more to be added here.
```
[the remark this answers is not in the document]
javiercn:
```suggestion
- Runtime
- ASP.NET Core
- Web Tools 
- SDK
- NuGet
```

accepted/2023/net8.0-browser-tfm.md
zivkan: >  it might be OK to go with '$(Browser)' == 'true' or '$(TargetPlatformIdentifier)' == 'browser'.

I thought that ` '$(TargetPlatformIdentifier)' == 'browser' ` should already work today, as the .NET SDK already populates that property: https://github.com/dotnet/sdk/blob/9f43a4a55078a22daac527c99e98d26e827a0ee2/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.TargetFrameworkInference.targets#L62

However, I see that it's in a targets file. But I'm 99% sure that ` '$(TargetFrameworkIdentifier)' == '.NETFramework' ` works in a csproj, as I use it when I need to multi-target netfx. Maybe there's some inner-build magic going on.

Anyway, I just wanted to point out that I think it will work already (in a csproj), I just don't understand how.

accepted/2023/net8.0-browser-tfm.md
pavelsavara: This comment helped me understand that the main goal is to **allow** multiple DLLs for the same project.
Perhaps with `#if` sprinkled in the code.

And that would assure that client-side-only references would not be pulled into to server side deployment.

It would not prevent server-side-only dependencies to be referenced from the client side. Because `net8-browser` is superset of `net8`.

accepted/2023/net8.0-browser-tfm.md
pavelsavara: We do not offer browser-only features either. And that's how it is the same API surface with NodeJS too.

accepted/2023/net8.0-browser-tfm.md
dsplaisted: Comparing against the `$(TargetFrameworkIdentifier)` works for item / ItemGroup conditions.  It doesn't work for properties / PropertyGroups.

> won't the property work against other properties if we define it earlier in the props file for the SDK? AFAIK, properties still get evaluated in the order in which they appear in the file, isn't it? I know there are like 6 stages of evaluation, so I understand that it might be more nuanced than what I suggested.

If the property depends on the `TargetFramework`, then you can't set it in a .props file, since the `TargetFramewrok` won't be set yet.

nkolev92: Given that we wouldn't expect the browser tfm to have versions we can be a bit more creative.

```xml
<TargetFramework>browser</TargetFramework>
<TargetFrameworkVersion>8.0</TargetFrameworkVersion>
```

Then the SDK logic that can "special case" `browser` and not populate the TargetFrameworkVersion, but use whatever is provided. 

Then you'd be able to use `'(TargetFramework)' == 'browser'` like constructs and it'll always work. 

The disadvantage is that this cannot be extended to a general solution.

accepted/2023/net8.0-browser-tfm.md
SteveSandersonMS: > It would not prevent server-side-only dependencies to be referenced from the client side. Because net8-browser is superset of net8.

It wouldn't *prevent* them but it would give app developers and library developers the power to control that. They can mark a `PackageReference` or `ProjectReference` on `$(TargetFramework)` being exactly equal to `net8.0` or `net8.0-browser`, and then it won't be present in the other case.

marek-safar: We should decorate `Microsoft.JSInterop.IJSInProcessRuntime` with `[SupportedOSPlatform("browser")]` either way. I also think that moving such type to to TFM only assembly at this point would be noticeable breaking change.

accepted/2023/net8.0-browser-tfm.md
marek-safar: > In certain very carefully balanced cases it's true that the linker may resolve OSPlatform.IsBrowser(), discard some if/else branches, and consequently discard entire assembly references. That's great if it works out.

Not only that but linker, JIT, AOT compiler and also many of our C# analyzers can do that as well (making the code effectively disabled during coding) so using ifdefs is not really giving you any perf/size benefits.

SteveSandersonMS: The claim I'm making is that relying on what symbols happen to be statically referenced is too fragile in reality. There's nothing stopping a developer from accidentally/ignorantly referencing symbols from one of the assemblies that you really wanted to get linked out. Compiler enforcement is required, which is what the TFM enables.

accepted/2023/net8.0-browser-tfm.md
terrajobst: @SteveSandersonMS 

> Could you clarify what disjoint API sets you mean?

The paragraph in spec says a TFM would help in offering WASM-specific APIs. A TFM is an API set. Multiple TFMs, disjoint API sets.

Historically, the problem of .NET was fragmentation. For library authors it's much easier to call APIs conditionally ("light-up") the building for multiple TFMs. The downside is discovery which APIs work on which platforms and bloat (you carry code that doesn't run). The platform analyzer addresses the former. The latter is a judgement call on our side which is why I asked earlier how many APIs we think will be WASM specific.

IOW, my question was why you believe that having multiple TFMs is making the UX easier.

accepted/2023/net8.0-browser-tfm.md
terrajobst: I assumed our linker is smart enough, but if you find that this is too fragile do deliver what you need, then I'll take your word for it.

accepted/2023/net8.0-browser-tfm.md
terrajobst: @dsplaisted have we considered using a naming convention for `.props` and `.targets` that are automatically imported, like:

* `NETCoreApp.props`
* `NETFramework.props`
* `macOS.props`
* `ios.props`
* `windows.props`
* `windows10.0`.props`
* `NETCoreApp-windows.props`
...

Like naively, this could work in this exists somewhere in the SDK:

```xml
<Project>
    <PropertyGroup>
        <FrameworkProps>$([MSBuild]::GetTargetFrameworkIdentifier($(TargetFramework))).props</FrameworkProps>
        <PlatformProps>$([MSBuild]::GetTargetPlatformIdentifier($(TargetFramework))).props</PlatformProps>
    </PropertyGroup>

    <Import Project="$(FrameworkProps)" Condition="Exists('$(FrameworkProps)')" />
    <Import Project="$(PlatformProps)" Condition="Exists('$(PlatformProps)')" />
</Project>
```

accepted/2023/net8.0-browser-tfm.md
SteveSandersonMS: > I think what you're talking about is that by having multiple TFMs we avoid cases where folks accidentally call browser-specific APIs from the server. I was hoping that the built-in analyzer would have solved that, but it seems it didn't. Is that because customers turned it off or they don't understand it?

Thanks for clarifying.

If you're making a class library that is meant to support both server-side and WebAssembly execution modes, I'm not sure the platform compatibility analyzer really helps. You have to declare that you support `browser` (because you do), and at that point you lose all guarantees about what happens when you're running on the server. The PCA really only helps detect simple cases (you try to use an API that you definitely-can't-ever use, not when you unconditionally use an API that you sometimes-can-and-sometimes-can't).

Multitargeting over two TFMs correctly represents the scenario where you sometimes-can and sometimes-can't and forces you to account for this difference.

SteveSandersonMS:
There's been a lot of confusion about what the goals are here, and why/whether a TFM would be the right way to solve them. As a matter of clarity and honesty, I want to break down the goals and address the fact that this is genuinely a different primary reason for using a TFM than we see in most other cases.

1. Our **primary goal** here is to have two different compilations from a single project - one for server, one for WebAssembly. And to be able to vary which files/lines/dependencies are included for each of these two compilations. At runtime, Blazor United needs access to the outputs from both of these compilations, so it can run the server-output on the server, and can serve the WebAssembly-output to the client.
2. Having distinct API sets for the two compilations is a **secondary goal** that Blazor WebAssembly has lived without so far and nothing is *truly* changing that forces us to have distinct API sets right now. The platform compatibility analyzer helps to a limited extent to guide people to do the right thing, so if `SupportsBrowser` could be toggled based on which of the two compilations you're in, that would still help.

This proposal is about using TFMs to solve both the primary and secondary goals. A TFM can indeed solve both of these pretty comprehensively. However it's not the only possible solution for the primary goal. It's possible to imagine some other way of hooking into the build so we produce multiple compilation outputs, and vary some `IsWebAssembly` MSBuild property across the two so we can condition the itemgroups/references/defines/etc.

The key advantage of using a TFM versus some other more custom/hacky mechanism is:

 * It automatically flows through the whole chain of dependencies in a compilation, so anything anywhere in the transitive closure of dependencies can reliably vary its API/dependency usage based on the selected TFM
 * VS and other tooling already knows about TFMs. For example, it will let the developer pick between `net8.0` and `net8.0-browser` in a nice dropdown and updates the code editor to highlight which code will and won't get used in each of the compilations. If you've conditionally excluded some of your own methods based on TFM, it will correctly highlight any cases where you're incorrectly calling those methods from the wrong TFM. In fact when hovering any method it will tell you which TFMs it's callable from.

If there was a good non-TFM way to achieve these goals I think that would be great too, but I'm not aware of one.

accepted/2023/net8.0-browser-tfm.md
terrajobst: That makes sense, thanks for explaining. I consider this resolved then.

accepted/2023/net8.0-browser-tfm.md
terrajobst: We discussed this today and added a note to the document.

accepted/2023/net8.0-browser-tfm.md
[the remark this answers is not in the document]
terrajobst: I have added the name to the design section.

accepted/2023/net8.0-browser-tfm.md
terrajobst: In the discussion it became clear that this isn't about the linker; it's that the developer wants a single application project with the ability to depend on different components between client and server while also having the tools that avoids accidentally pulling server component to the client, which is what multi-targeting would solve, but it requires a TFM.

trungnt2910:
With `net8.0-browser` accepted, will there be a (standalone, non-Blazor) workload that enables this TFM? Or will this only be a part of Blazor WASM?

I want to use the browser TFM for browser .NET builds on top of the `wasm-experimental` workload only, similar to what I'm doing with [this](https://github.com/trungnt2910/DotnetBrowser) repo, and not reference any extended libraries/build tools/workloads that Blazor provides.

SteveSandersonMS:
@trungnt2910 `net8.0-browser` is not proposed to be Blazor-specific. It's a TFM that produces browser-compatible build output. Whether or not your app depends on the Blazor UI framework libraries is an independent choice.

trungnt2910:
Will there be a workload that enables this TFM then, as I can't see one in the proposal.

SteveSandersonMS:
I expect it would become an aspect of the `wasm-experimental` workload which already exists.

PR #292, opened 2023-03-28
.NET 8.0 Polyfills

terrajobst:


accepted/2023/net8.0-polyfills/net8.0-polyfills.md
```diff
@@ -0,0 +1,29 @@
...
+
+This document covers which APIs we intend to ship for older versions of .NET,
+which includes .NET Standard and .NET Framework.
+
+## Polyfills
+
+| Polyfill     | Assembly                   | Package                    | Existing? | API                    | Contacts                   |
+| ------------ | -------------------------- | -------------------------- | --------- | ---------------------- | -------------------------- |
+| TimeProvider | Microsoft.Bcl.TimeProvider | Microsoft.Bcl.TimeProvider | No        | [dotnet/runtime#36617] | [@tarekgh] [@geeknoid]     |
+| IPNetwork    | Microsoft.Bcl.IPNetwork    | Microsoft.Bcl.IPNetwork    | No        | [dotnet/runtime#79946] | [@antonfirsov] [@geeknoid] |
```
bartonjs, line 13: IPNetwork seems like an over-scoped name.  When we did the SP800 compat package for crypto we found a general name (Microsoft.Bcl.Cryptography).  It seems like an equivalent here would be Microsoft.Bcl.Networking; and that any similarly needed networking types (that don't introduce a layering/dependency problem) would live in that package.

TimeProvider doesn't really have a more general bucketing that I can think of, so I don't have feedback on that line.

accepted/2023/net8.0-polyfills/net8.0-polyfills.md
```diff
@@ -0,0 +1,29 @@
+# .NET 8.0 Polyfill
+
+**Owner** [Immo Landwerth](https://github.com/terrajobst)
+
+This document covers which APIs we intend to ship for older versions of .NET,
+which includes .NET Standard and .NET Framework.
+
+## Polyfills
```
geeknoid, line 8: Is it worth considering CompositeFormat? We have much of the code in hand for this, it would just be a matter of packaging it appropriately.

geeknoid: Also, there are a bunch of attributes that the compiler depends on which need to be implemented by hand when using the new compiler on an older runtime. Those really should be provided down-level. Here's the set we're having to maintain ourselves, and that I've seen replicated in other code bases:

d----           2/24/2023     7:26                CallerAttributes
d----           2/24/2023     7:26                DiagnosticAttributes
d----           2/24/2023     7:26                ExperimentalAttribute
d----           2/24/2023     7:26                IsExternalInit
d----           2/24/2023     7:26                SkipLocalsInitAttribute
d----           2/24/2023     7:26                StringSyntaxAttribute
d----           2/24/2023     7:26                TrimAttributes

accepted/2023/net8.0-polyfills/net8.0-polyfills.md
stephentoub: We need a really strong reason to create an OOB.  For TimeProvider, it's an abstraction at the bottom of the stack that we want to plumb through types that are already OOB, e.g. Microsoft.Extensions, which also pulls it towards that path.  I don't see that for IPNetwork.  It's "just" a helper type.  What's the significant reason we need to break from our standard approach?

accepted/2023/net8.0-polyfills/net8.0-polyfills.md
stephentoub: > Is it worth considering CompositeFormat? We have much of the code in hand for this, it would just be a matter of packaging it appropriately.

I don't think so.  It's not an exchange type, it's not an abstraction, etc.  It's new in .NET 8, integrated with overloads of existing methods like string.Format.  And its implementation is heavily based on types new in .NET 6, e.g. the interpolated string handlers, ISpanFormattable, etc.

accepted/2023/net8.0-polyfills/net8.0-polyfills.md
```diff
@@ -0,0 +1,29 @@
...
+
+| Polyfill     | Assembly                   | Package                    | Existing? | API                    | Contacts                   |
+| ------------ | -------------------------- | -------------------------- | --------- | ---------------------- | -------------------------- |
+| TimeProvider | Microsoft.Bcl.TimeProvider | Microsoft.Bcl.TimeProvider | No        | [dotnet/runtime#36617] | [@tarekgh] [@geeknoid]     |
```
danmoseley, line 12: Can we avoid perpetuating the BCL acronym? I don't think an existing package is a good enough reason. BCL is only familiar to .NET aficionados and a descriptive term like libraries, core libraries, platform, core, etc would be fine. Note we chose to use SYSLIBnnnn for codes not BCLnnnn

jkotas: We have shipped number of poly-fill packages with this naming pattern: `Microsoft.Bcl.AsyncInterfaces`, `Microsoft.Bcl.HashCode`, `Microsoft.Bcl.Compression`, `Microsoft.Bcl.Cryptography`, `Microsoft.Bcl.Json.Sources`, `Microsoft.Bcl.Async`.

jkotas: IIRC, the naming pattern is hideous on purpose.

accepted/2023/net8.0-polyfills/net8.0-polyfills.md
huoyaoyuan: For non-exchange types, user-local or third-party definition won't lose their functionality. Exchange types would need compatible definition and proper forwarding logic in order to be used in libraries.

What about `DateOnly` and `TimeOnly` then? The inability to polyfill members on existing `DateTime` type?

accepted/2023/net8.0-polyfills/net8.0-polyfills.md
terrajobst: Alright, looks we agreed not to do this one.

terrajobst:
Merging as we completed this work.

PR #293, opened 2023-03-31
Update BinaryFormatter obsoletion plan for net8

GrabYourPitchforks:
Updating the list of obsoletions for .NET 8, bringing a few more in and pushing some out to .NET 9.

The primary reason for doing this is to provide a better experience for developers. Compared to the original plan, this should result in fewer overall warnings being observed, and the warnings which are observed should be more directly actionable.

### DIFF
--- a/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md
+++ b/accepted/2020/better-obsoletion/binaryformatter-obsoletion.md
@@ -50,11 +50,12 @@ Encouraging migration away from `BinaryFormatter` has the additional benefit of 
 - All first-party dotnet org code bases complete migration away from `BinaryFormatter`
 - `BinaryFormatter` disabled by default across all project types
 - All not-yet-obsolete `BinaryFormatter` APIs marked obsolete _as warning_
-- Entirety of legacy serialization infrastructure marked obsolete _as warning_
+- Additional legacy serialization infrastructure marked obsolete _as warning_
 - No new `[Serializable]` types introduced (all target frameworks)
 
 ### .NET 9 (Nov 2024)
 
+- Remainder of legacy serialization infrastructure marked obsolete _as warning_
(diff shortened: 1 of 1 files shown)


yaakov-h:
Would this include corresponding changes for roslyn-analyzers and the `Exception` code snippet in Visual Studio?

IIRC serialization code is still included in the snippet, and omitting those functions can trigger CA2237 and CA2229 (which would be conflicting with SYSLIB0050).

GrabYourPitchforks:
@yaakov-h CA2237 and CA2229 are still useful. If you do still need to implement `ISerializable` for some reason (e.g., you need to interoperate with legacy code), then CA2237 and CA2229 ensure you're following the standard way of doing it.

The `Exception` code snippet will not be changed. I asked the owners about this last week before posting the plan, including that it will begin to produce compiler warnings, and they stated there are no plans to update the snippet in the coming version. I was going to include in the release notes that you can simply delete that constructor to make the warning go away.

ReubenBond:
I see that `FormatterServices.GetUninitializedObject` is being made obsolete. Can we safely assume `RuntimeHelpers.GetUninitializedObject` will remain?

Orleans respects `NonSerializedAttribute`, so we will need to add some other, equivalent attribute if it's deprecated. That's not a big issue.

GrabYourPitchforks:
> Can we safely assume `RuntimeHelpers.GetUninitializedObject` will remain?

Yes. I'm not aware of any plans to deprecate or hide this API.

huoyaoyuan:
> The `Exception` code snippet will not be changed.

The whole code snippet infrastructure need an update. A lot portion of them are only applicable to C# 2.0-4.0. Please push a feature request to the owners!

jeffhandley (approved):
The updated plan looks good to me. I just have one suggestion for getting rid of an "I" statement.

PR #294, opened 2023-03-31
Add design for simplifying workload versioning

dsplaisted:
This proposal simplifies how workload versions are expressed and understood.

Currently, the design mainly covers the proposed changes to the user experience for workloads.  That is the primary thing we are looking for feedback on now: Would this be a better user experience?

There will also be significant changes needed to implement this.  Those still need to be written up and I plan to put those in a separate document, probably as part of this same pull request.

### DIFF
--- a/accepted/2023/simplify-workload-versioning.md
+++ b/accepted/2023/simplify-workload-versioning.md
@@ -1,0 +1,140 @@
+# Simplify Workload Versioning
+
+.NET SDK Workloads are optional components of the .NET SDK.  When designing workloads, one of the goals was to be able to update workloads separately from the .NET SDK.  As a result, workloads version and update separately from the .NET SDK.  This has made it hard to understand what versions of workloads are installed or available.  Complicating this is the fact that workloads themselves don't directly have a version.  Workload manifests, which can have a many-to-many relationship with workloads, have versions.  These versions can be displayed with the `dotnet workload update --print-rollback` command, which displays output in the following format:
+
+```
+==workloadRollbackDefinitionJsonOutputStart==
+{
+  "microsoft.net.sdk.android": "33.0.26/7.0.100",
+  "microsoft.net.sdk.ios": "16.2.1030/7.0.100",
+  "microsoft.net.sdk.maccatalyst": "16.2.1030/7.0.100",
+  "microsoft.net.sdk.macos": "13.1.1030/7.0.100",
+  "microsoft.net.sdk.maui": "7.0.59/7.0.100",
+  "microsoft.net.sdk.tvos": "16.1.1527/7.0.100",
+  "microsoft.net.workload.mono.toolchain.net6": "7.0.3/7.0.100",
+  "microsoft.net.workload.mono.toolchain.net7": "7.0.3/7.0.100",
+  "microsoft.net.workload.emscripten.net6": "7.0.3/7.0.100",
+  "microsoft.net.workload.emscripten.net7": "7.0.3/7.0.100"
+}
+==workloadRollbackDefinitionJsonOutputEnd==
+```
+
+This is very complicated and doesn't make it easy to understand workload versions.
+
+## Goals
+
+- Introduce a single version number that represents all workload versions for the the .NET SDK
+- Workloads can continue to release outside the normal .NET SDK release schedule
+- Deadlines for workload completion and insertion do not change from current processes
+- Workload updates should require minimal (or no) additional validation of non-workload scenarios
+
+## Workload patch versions
+
+Currently, we use a 3 part version number for the .NET SDK, for example 8.0.100.  Typically we release a patch each month, for example 8.0.101, 8.0.102, etc.
+
+We will create a workload patch version number that encapsulates all the workload manifest versions that are released together.  We will represent the workload patch version by adding an additional component to the .NET SDK version.  The versions of the workloads that release on the same release date as the .NET SDK will be number `.0`.  Subsequent workload updates before the next .NET SDK patch will be numbered `.1`, `.2`, etc.  For example, the full workload patch version for the initial 8.0 release would be `8.0.100.0`, and if there was a workload update before the first .NET SDK patch, the updated workload patch version would be `8.0.100.1`.
+
+A workload patch is essentially a mapping from the workload patch version to the different versions for each workload manifest.  Creating a workload patch should be a lightweight process, which involves creating a NuGet packages and corresponding installers to deliver this mapping.  The .NET SDK assets and installers, workload manifests, and workload packs should already be built and should not need to be created when building a workload patch.
+
+## Experience
+
+`dotnet --version` would continue to print the SDK version, as it does today:
+
+```
+> dotnet --version
+8.0.201
+```
+
+We will add a new `dotnet workload --version` command to print the workload patch version:
+
+```
+> dotnet workload --version
+8.0.201.2
+```
+
+We will also update `dotnet --info` and `dotnet workload --info` to display the workload patch version.  For example:
+
+```
+> dotnet --info
+.NET SDK:
+ Version:              8.0.201
+ Commit:               <commit>
+ Workloads version:    8.0.201.2
+
+<further information>
+
+> dotnet workload --info
+ Workloads version:     8.0.201.2
+ [wasm-tools]
+   Installation Source: SDK 8.0.100-preview.4
+   Manifest Version:    8.0.0-preview.4.23181.9/8.0.100-preview.4
+   Manifest Path:       C:\git\dotnet-sdk-main\.dotnet\sdk-manifests\8.0.100-preview.4\microsoft.net.workload.mono.toolchain.current\WorkloadManifest.json
+   Install Type:        FileBased
+```
+
+When updating workloads, console output will include the old and new workload patch versions:
+
+```
+> dotnet workload install wasm-tools
+Checking for updated workloads version...
+Updating workloads version from 8.0.201.0 to 8.0.201.2...
+Installing workload manifest microsoft.net.sdk.android version 34.0.0-preview.4.230...
+<Further workload manifest update messages>
+Installing pack Microsoft.NET.Runtime.WebAssembly.Sdk version 8.0.0-preview.4.23181.9...
+<Further workload pack installation manifests>
+Garbage collecting for SDK feature band(s) 8.0.200...
+
+Successfully updated workloads version from 8.0.201.0 to 8.0.201.2.
+Successfully installed workload(s) wasm-tools.
+```
+
+The proposed [workload history](https://github.com/dotnet/sdk/pull/30486) command will also display the workload patch versions.
+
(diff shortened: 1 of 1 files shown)


accepted/2023/simplify-workload-versioning.md
```diff
@@ -0,0 +1,140 @@
...
+```
+> dotnet workload update --version 8.0.201.0
+Updating workloads version from 8.0.201.2 to 8.0.201.0...
+<Manifest and pack installation / garbage collection messages>
+Successfully updated workloads version from 8.0.201.2 to 8.0.201.0.
+```
+
+## Specifying workload versions with global.json
+
+We will add support for specifying the workloads version in global.json.  For example:
```
joeloff, line 122: Wouldn't this break global.json? It would look for an SDK with that version, which would never exist, or is the plan to look for an SDK version of 8.0.201 and then a workload patch version of 8.0.201.0?

How would someone specify a patch version when the SDK version is prerelease? We cannot have a 4-part version number as that breaks semver 2.0

Should we consider adding a separate entry instead, e.g. `"workload-patch": "0"? That still won't solve the problem with preview SDKs I think

accepted/2023/simplify-workload-versioning.md
```diff
@@ -0,0 +1,140 @@
...
+  "sdk": {
+    "version": "8.0.201.0",
+    "rollForward": "disable"
+  }
+}
+```
+
+This would force the SDK to use workloads version `8.0.201.0`, and would error if that version was not installed.
+
+We will support side-by-side workload version installations.  If 8.0.201.2 is installed, we would support running `dotnet workload install --version 8.0.201.0` to install that version of the workloads.  After installing the earlier version, the .NET SDK would still by default use the latest installed workloads version.  An earlier workloads version would only be used if it was specified in a global.json file.
```
joeloff, line 135: This would require having SxS manifests instead of requiring them to upgrade within a band across SDK patches? It would help with the downgrade functionality as well I think.

accepted/2023/simplify-workload-versioning.md
```diff
@@ -0,0 +1,140 @@
...
+Updates are available for the following workload(s): wasm-tools. Run `dotnet workload update` to get the latest.
+```
+
+We will modify both commands to print the same message:
+
+```
+A workload update (version 8.0.201.2) is available. Run `dotnet workload update` to install the latest workloads.
+```
+
+We will add a new `--version` option to `dotnet workload update` to allow installing (or downgrading to) a specific workloads version:
```
joeloff, line 111: Would downgrading remove the previous workload version? What happens if a user installs VS and they get 8.0.203.6 of the workloads, but then drop to the CLI and try to downgrade. Would we block this scenario?

accepted/2023/simplify-workload-versioning.md
dsplaisted: The idea was that only the first three parts of the version would be used for SDK selection.  You're right though, it's probably better to have a separate value for the workloads version though.

accepted/2023/simplify-workload-versioning.md
dsplaisted: I think it would remove the previous version if possible, but if it was installed by VS it would leave it and print a message saying you'd need to use global.json to select the earlier version.

accepted/2023/simplify-workload-versioning.md
rolfbjarne, line 135: How do we test workloads before they're released? Is there any way to install a locally built workload?

accepted/2023/simplify-workload-versioning.md
```diff
@@ -0,0 +1,153 @@
...
+
+## Specifying workload versions with global.json
+
+We will add support for specifying the workloads version in global.json.  For example:
+
+```json
+{
+  "sdk": {
+    "workloadsVersion": "8.0.201"
+  }
```
jonathanpeppers, line 144: Right now, `AutoImport.props` is imported into every project for .NET 6, 7, 8, etc. If you have 8.0.201 and 8.0.202 workloads installed, is `AutoImport.props` only imported for the pinned version?

If so, we might want to add some more robust logging to understand why the workload resolver imported one file over another.

jonathanpeppers (commented):
The most common error message I see when workloads are broken for customers is something like: `TargetPlatformIdentifier 'android' not found`.

Usually what happened, is the `WorkloadManifest.targets` is missing for the `android` workload, and so the build did not understand the value for `$(TargetPlatformIdentifier)`.

Will these changes be able to give a bit more information in this case? We might have an error message like "Workload Version 8.0.202 is not installed"?

dsplaisted:
> The most common error message I see when workloads are broken for customers is something like: `TargetPlatformIdentifier 'android' not found`.
> 
> Usually what happened, is the `WorkloadManifest.targets` is missing for the `android` workload, and so the build did not understand the value for `$(TargetPlatformIdentifier)`.
> 
> Will these changes be able to give a bit more information in this case? We might have an error message like "Workload Version 8.0.202 is not installed"?

Do you know why the workload manifest targets aren't installed in these cases?  They should always be there.

The new design might be able to help with this but without understanding how people get into this state, it's hard to tell.

jonathanpeppers:
I don't think anyone knows exactly how people are getting into this state. I think it is the combination of installing from VS and command-line, then some number of attempted upgrades/downgrades.

Usually the registry query + `msiexec` I mention here is what solves the problem:

https://gist.github.com/jonathanpeppers/63beb491a9185ac06710261536cc35c9#windows

MiYanni (commented):
I only looked at the first document so far. Will look at the second one tomorrow.

accepted/2023/simplify-workload-versioning-implementation.md
```diff
@@ -0,0 +1,77 @@
...
+
+Non Visual Studio installs of the .NET SDK do not include workloads themselves in the .NET SDK package or installer.  In this case the .NET SDK only hase workload manifests.  Because the final versions of workloads are not inserted into the .NET SDK, these manifests (which we call "baseline manifests") are not the final versions and do not refer to the workloads that are actually released.  When installing a workload, the default behavior of the .NET SDK is to first look for the latest workload manifests and install them before installing the workload.  So as long as the workload packages are released publicly (on NuGet.org) on release day, installing a workload for a non-VS install of the .NET SDK will also install the right workload versions, even though the baseline manifest versions built into that SDK are not the final release versions.
```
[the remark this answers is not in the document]
MiYanni: Ah, I found this answer in the other document.

accepted/2023/simplify-workload-versioning-implementation.md
```diff
@@ -0,0 +1,77 @@
...
+
+### Release schedule updates
+
+Workload sets will be created in a separate  repo (possibly dotnet/workloadversions).  This repo will have the versions for all the workload manifests, either via Arcade dependency flow (for workloads that build as part of .NET) or via manual updates.  The build will create a json file in the rollback file format from these versions, and create NuGet packages, MSIs, and (in the future) other installers to deliver this json file as a workload set.
```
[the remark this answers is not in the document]
dsplaisted: To me, `dotnet/workloads` implies that the workloads are built in that repo.  This repo wouldn't include any contents of the workloads, it just keeps track of the versions of the manifests and what workload set we assign to them.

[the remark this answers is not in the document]
joeloff: What about `dotnet/workload-sets`?

joeloff:
What are the rules around the versioning scheme inside the manifest? 4-part Windows version or a semantic version?

accepted/2023/simplify-workload-versioning-implementation.md
```diff
@@ -0,0 +1,129 @@
...
+The workload resolver should first look for a workload set version from a global.json file.  If not specified via global.json, then it should look in the workload install state folder for a workload set or workload manifest versions to use.  If not specified via the install state folder, then the resolver should use the latest workload set version installed for the feature band.  Finally, if no workload set is installed, the resolver should use the latest version of each manifest for the feature band.
+
+### Installing workload set from global.json
+
+`dotnet workload update` - If workload set is specified in global.json, this should install that workload set.  Also should register this global.json path in workload install state folder as a GC root.
+
+- `dotnet workload restore` - global.json aware, same as update
+- `dotnet workload install` - global.json aware
+- `dotnet workload clean` - should look through all records of global.json, see if they are still there and up-to-date, delete as necessary before GC
+- `dotnet workload roots` - (not sure about command name) List all registered global.json files and versions of workload sets they pin
```
jonathanpeppers, line 103: Does this mean in some future, it's possible to have old workload packs around on your system. And then you'd have to delete a `global.json` to successfully remove those packs?

That seems OK: as the behavior now is you'd lose old packs that some project might be using.

accepted/2023/simplify-workload-versioning-implementation.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Release schedule updates
+
+Workload sets will be created in a separate  repo (possibly dotnet/workloadversions).  This repo will have the versions for all the workload manifests, either via Arcade dependency flow (for workloads that build as part of .NET) or via manual updates.  The build will create a json file in the rollback file format from these versions, and create NuGet packages, MSIs, and (in the future) other installers to deliver this json file as a workload set.
```
rolfbjarne, line 15:
```suggestion
Workload sets will be created in a separate repo (possibly dotnet/workloadversions).  This repo will have the versions for all the workload manifests, either via Arcade dependency flow (for workloads that build as part of .NET) or via manual updates.  The build will create a json file in the rollback file format from these versions, and create NuGet packages, MSIs, and (in the future) other installers to deliver this json file as a workload set.
```

accepted/2023/simplify-workload-versioning.md
```diff
@@ -0,0 +1,194 @@
...
+}
+```
+
+This would force the SDK to use workload set version `8.0.201`, and would error if that version was not installed.
+
+NOTE: We may not implement the global.json workload set version support in the same release as the rest of the changes described in this design.
+
+## Side by side workload sets
+
+The workload `restore`, `update`, and `install` commands will be updated to check if there is a global.json file specifying a workload set version applying to the current folder.  If so, these commands will install that workload set version side-by-side with any other installed workload set.  They will also record the path to the global.json file and the workload set version it specified.  This record will be used as a root during garbage collection to prevent the side-by-side workload set (and its packs) from being removed.
```
lambdageek, line 174: Does this mean some file in `/usr/local/share/dotnet` will include a path to `/Users/aklig/work/fooProject/global.json` ?

What if I rename the `fooProject` folder to `barProject`?

lambdageek: Maybe it would be better to have an explicit `sudo dotnet workload pin <workload version set>` and `sudo dotnet workload unpin <workload version set>` command that could protect a workload set from being removed by `dotnet workload clean` ?   And then `dotnet workload --info` would display if a certain workload set is pinned or not.

This is what Homebrew does when you need to prevent some out of date version from being removed.

joeloff (approved):

PR #296, opened 2023-05-18
Add Simple C# Programs

WeihanLi:
> Let's see if we can make C# a little simpler.

Continues of https://github.com/dotnet/designs/pull/213

since the original PR closed, migrated from https://github.com/cartermp/designs/blob/a27fd42a68370f315bc2293dc748145f4075697f/proposed/simple-csharp-programs.md

### DIFF
--- a/proposed/simple-csharp-programs.md
+++ b/proposed/simple-csharp-programs.md
@@ -1,0 +1,499 @@
+# Simple C# Programs
+
+This speclet defines "Simple C# Programs" and covers what they can do from the user's standpoint. It does **not** get into detailed design points, a deep dive of a particular mechanisms, etc. The goal is to drive consensus on the user experience and then walk back towards the right design and implementation.
+
+It is also designed to be incremental. Not everything written here must be implemented. For example, we could ship a version without support for packages/frameworks.
+
+## Motivation
+
+C# has lots of great things going for it, but it falls short in a key area: getting started with the least possible friction.
+
+Lots of developers in the world have experience with languages like JavaScript, Python, and Ruby. In these languages, you typically get started in one of these three ways:
+
+1. Launch a REPL and start typing code and getting results
+2. Create a file with the right extension and run the code in that file
+3. Use build tooling to create an application and then run that application
+
+If you look at online guides for the above languages, they tend to start people off with (1) and (2) and rarely with (3).
+
+In the C# world, we start people off with (3) with the base tooling, the .NET CLI.
+
+This is problematic, because these people **do not expect that they need to do (3) to get started**. The tools don't support their instincts and there is very little information online to support them.
+
+Of course, REPLs are another great way to get started. But they are outside the scope of this document.
+
+## What is a Simple C# Program
+
+A Simple C# Program is this:
+
+* A C# file
+* A set of global `using`s inferred by the build environment
+* A tool that can run that file / produce a working program
+* A way to incorporate libraries and frameworks/sdks into that file and run it
+* A way to "graduate" to using a project file
+
+A Simple C# Program is **not** a replacement for project-based C# code or .NET Interactive.
+
+## Simple C# is just C# code
+
+Simple C# Programs are not a dialect of the C# language, like the interactive dialect, CSX. Any code in a Simple C# Program can be built and run in a project-based program.
+
+## User experience walkthrough
+
+Given the following code in a singular `hello.cs` in a directory:
+
+```csharp
+WriteLine("Hello World");
+```
+
+You can just run it:
+
+```bash
+$ dotnet run
+Hello World
+```
+
+Or with a `#!` directive,
+
+```csharp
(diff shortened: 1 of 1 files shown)


proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,499 @@
...
+
+This is, frankly, too complex and probably shouldn't be done. But it's worth listing.
+
+## Prior art
+
+This isn't the first proposal that deals with trying to cut down the number of concepts involved in writing a small C# program. There's some prior art to consider in this space.
+
+### A note on .NET Interactive
+
+.NET Interactive _may_ be able to service these goals, but since it is currently focused on notebooks-based programming it doesn't really address the same scenario like JavaScript, Python, and Ruby. It is highly relevant for people who are used to notebook environments, and since notebooks are breaking into educational spaces, it may also play a big role in getting people started there.
```
jonsequitur, line 456: .NET Interactive isn't intended just for notebooks. For example, the [`dotnet-repl`](https://github.com/jonsequitur/dotnet-repl) project allows you to run code files (including but not limited to notebook file formats) directly from the command line. I believe many of your samples will work right now using `dotnet-repl`.

One important difference currently, though, is that the semantics of C# Script differ somewhat from "classic" C#. Support for classic C# is in progress, though.

proposed/simple-csharp-programs.md
```diff
@@ -0,0 +1,499 @@
...
+* A set of global `using`s inferred by the build environment
+* A tool that can run that file / produce a working program
+* A way to incorporate libraries and frameworks/sdks into that file and run it
+* A way to "graduate" to using a project file
+
+A Simple C# Program is **not** a replacement for project-based C# code or .NET Interactive.
+
+## Simple C# is just C# code
+
+Simple C# Programs are not a dialect of the C# language, like the interactive dialect, CSX. Any code in a Simple C# Program can be built and run in a project-based program.
```
zivkan, line 39: Here you say that Simple C# programs will not a dialect of the C# language, but then you propose that the `#r "nuget:PackageId"` syntax be copied from .NET Interactive.

Therefore, are you proposing that directories with a csproj allow individual `.cs` files to also use `#r` to reference packages?

Zeioth:
I'm interested. Please ping me in the future if you need testers or any kind of help.

AraHaan:
These are my 2 cents on this:

Support this only if all of the files are ``*.csx``

All ``*csx`` files in the same directory as well as any nested directories are assumed to be in the same compilation at runtime in ``dotnet run``

``dotnet run`` will only fallback to this only if:
- directory does not have any csproj files.
- no project.json files.

``#r `` reference not needed to reference other ``*.csx`` files, dotnet run would need a special way to support this one though.
``#r "nuget:PackageId"`` is only required in a single ``*.csx`` file, after that it is usable in all of them. The user can also add an ``GlobalUsings.csx`` file for any and all global usings they want to use for the entire thing, however it must be all done by hand and not by any special tooling.

With all of this I think it would be a good compromise for easier ways of running .NET Program code in a way that is similar in nature to python, JavaScript, etc.

AraHaan:
I just thought of this as well but an alternative to using the ``#r "nuget:PackageId"`` preprocessor on csx files:
- a packages.json that defines the nuget packages required to build and run the scripts, along with the versions.
  - this has a benefit as well of being able to explicitly specify a specific version as well which for some might be a good idea as well.

AraHaan:
After thinking about this more, the processing of this for ``dotnet run`` could be made on the MSBuild side in the ``Build`` target for when they use only csx + frameworks.json + packages.json files (no cs, csproj, sln, and no project.json).

And due to ``dotnet run`` invoking an implicit build by default, that means a binary would be cached that it can then run.

proposed/simple-csharp-programs.md
WeihanLi: Yeah, wondering if it's possible to treat the `#r` syntax as a comment when building with a project file?
We should be able to tell if the build is built with a project/solution file, right?
If yes, think then we could make it different when in different modes

WeihanLi:
> Support this only if all of the files are *.csx.

Hoping for `*.cs` file also

> All *.csx files in the same directory as well as any nested directories are assumed to be in the same compilation at runtime in dotnet run

I think this could be considered as an option, like enable default items or not in the MS Build

And I'm trying on the [`dotnet-exec`](https://github.com/WeihanLi/dotnet-exec) project, trying to make it possible to simpler C# programs

WeihanLi:
Closed since .NET SDK is implementing the `dotnet run file` feature, which is ready to play since .NET 10 SDK Preview 4, see detail on
https://github.com/dotnet/sdk/blob/main/documentation/general/dotnet-run-file.md

Hopefully, there's a detailed design document for it, created a ticket for it
https://github.com/dotnet/sdk/issues/49157

PR #300, opened 2023-08-21
Clarify behavior of severity

terrajobst:
This supersedes #297.

### DIFF
--- a/accepted/2023/preview-apis/preview-apis.md
+++ b/accepted/2023/preview-apis/preview-apis.md
@@ -232,7 +232,7 @@ supplied diagnostic ID.
 >
 > However, there will also be a generic compiler diagnostic ID that applies to
 > all warnings raised for using experimental APIs (like nullable). The built-in
-> `editor.config` that we ship with the .NET SDK will elevate these warnings to
+> `.editorconfig` that we ship with the .NET SDK will elevate these warnings to
 > errors. From a user's standpoint this will result in these diagnostics to
 > appear as errors that they are expected to suppress, which is the UX we
 > desire.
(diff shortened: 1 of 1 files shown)


accepted/2023/preview-apis/preview-apis.md
```diff
@@ -225,7 +225,17 @@ both options are being presented to the consumer.
 ### Compiler Behavior
 
 The compiler will raise a diagnostic when an experimental API is used, using the
-supplied diagnostic ID. The severity is always error.
+supplied diagnostic ID.
+
+> [!NOTE]
+> The severity is warning, because errors cannot be suppressed.
```
jcouv, line 231: Per email thread, we ended up landing on unconditional promotion/reporting of this warning as an error. We could add the `/warnAsError:experimental` flag, but we'll still need special knowledge of the error code, so this is just more complicated for not much benefit...
I'm not sure how to word this yet :-/ 
The options I see are: 
1. this is a suppressible error
2. that is a promoted warning
I think we should stick with option 2 (keeps the rule that errors can't be suppressed).
I'll make a corresponding update to the language speclet: https://github.com/dotnet/csharplang/pull/7472

accepted/2023/preview-apis/preview-apis.md
```diff
@@ -225,7 +225,17 @@ both options are being presented to the consumer.
...
+>
+> However, there will also be a generic compiler diagnostic ID that applies to
+> all warnings raised for using experimental APIs (like nullable). The built-in
+> `editor.config` that we ship with the .NET SDK will elevate these warnings to
```
akoeplinger, line 235:
```suggestion
> `.editorconfig` that we ship with the .NET SDK will elevate these warnings to
```

PR #301, opened 2023-09-27
[browser][wasm] threads and JS interop

pavelsavara:
Contributes to https://github.com/dotnet/runtime/issues/85592 and https://github.com/dotnet/runtime/issues/68162


accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,268 @@
...
+     - there are managed and un-managed locks
+     - in single-threaded build of the runtime, all of this is NOOP. That's why it works on UI thread.
+
+**2)** UI thread in the browser can't synchronously block
+    - you can spin-lock but it's bad idea.
+        - Deadlock: when you spin-block, the JS timer loop and any messages are not pumping. But code in other threads may be waiting for some such event to resolve.
+        - It eats your battery
+        - Browser will kill your tab at random point (Aw, snap).
+        - It's not deterministic and you can't really test your app to prove it harmless.
+    - Firefox (still) has synchronous XHR which could be captured by async code in service worker
```
jkotas, line 34: What is XHR?

NinoFloris: >What is XHR?


https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest

accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,268 @@
...
+        - but other browsers don't and it's unlikely they will implement it
+        - there are deployment and security challenges with it
+    - all the other threads/workers could synchronously block
+    - if we will have managed thread on the UI thread, any `lock` or Mono GC barrier could cause spin-wait
+        - in case of Mono code,  we at least know it's short duration
+        - we should prevent it from blocking in user code
+
+**3)** JavaScript engine APIs and objects have thread affinity. 
+    - The DOM and few other browser APIs are only available on the main UI "thread"
+        - and so, you need to have C# interop with UI, but you can't block there.
```
jkotas, line 45: Every UI stack has this problem. SynchronizationContext was designed to help with it.

accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,268 @@
...
+     - If people want to use existing MT code-bases, most of the time, the code is full of locks. 
+     - People want to use existing desktop/server multi-threaded code as is.
+ - allow HTTP and WS C# APIs to be used from any thread
+     - Underlying JS object have thread affinity
+ - JSImport/JSExport interop in maximum possible extent
+ - don't change/break single threaded build. †
+
+## Lower priority goals
+ - try to make it debugging friendly
+ - implement crypto via `subtle` browser API
```
kg, line 17: we should expand on how this connects, i.e. why can't we use the API right now, how will this change allow us to

accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,268 @@
+# Multi-threading on a browser
+
+## Goals
+ - CPU intensive workloads on dotnet thread pool
+ - enable blocking .Wait APIs from C# user code on all threads
+     - Current public API throws PNSE for it
+     - This is core part on MT value proposition.
+     - If people want to use existing MT code-bases, most of the time, the code is full of locks. 
+     - People want to use existing desktop/server multi-threaded code as is.
+ - allow HTTP and WS C# APIs to be used from any thread
```
kg, line 10: this is currently true, should we rephrase this goal to capture the new property we want? i.e. allow any HTTP/WS object to be used from any thread despite affinity

accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,268 @@
...
+        - there are deployment and security challenges with it
+    - all the other threads/workers could synchronously block
+    - if we will have managed thread on the UI thread, any `lock` or Mono GC barrier could cause spin-wait
+        - in case of Mono code,  we at least know it's short duration
+        - we should prevent it from blocking in user code
+
+**3)** JavaScript engine APIs and objects have thread affinity. 
+    - The DOM and few other browser APIs are only available on the main UI "thread"
+        - and so, you need to have C# interop with UI, but you can't block there.
+    - HTTP & WS objects have affinity, but we would like to consume them (via Streams) from any managed thread
```
kg, line 46: note that it may actually be a bad idea to support this. If someone is calling Stream.Read(byte[256]) from the wrong thread repeatedly the performance is going to be really bad to an extent that is not true in any other environment

accepted/2023/wasm-browser-threads.md
pavelsavara: There is difference between 
- "you should not do it" on normal OS and 
- "you don't have ways how to do it without eating battery"  in the browser

accepted/2023/wasm-browser-threads.md
pavelsavara: This is requirements from Blazor team.

accepted/2023/wasm-browser-threads.md
pavelsavara: Updated the section, 
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Synchronous_and_Asynchronous_Requests#synchronous_request

accepted/2023/wasm-browser-threads.md
pavelsavara: I added section for it and also for lazy DLLImport

accepted/2023/wasm-browser-threads.md
pavelsavara: updated

accepted/2023/wasm-browser-threads.md
```diff
@@ -236,6 +394,10 @@
 - `JSImport` used for logging: `globalThis.console.debug`, `globalThis.console.error`, `globalThis.console.info`, `globalThis.console.warn`, `Blazor._internal.dotNetCriticalError`
     - probably could be any JS context
 
+# WebPack, Rollup friendly
+- it's not clear how to make this single-file
+- because web workers need to start separate script(s)
```
kg, line 403: single file that behaves differently depending on whether the current context is `window` or not?

accepted/2023/wasm-browser-threads.md
jkotas: Is Blazor team aware of the fact that satisfying this requirement will come with performance and reliability cliffs?

accepted/2023/wasm-browser-threads.md
curiousdannii: You can start a WebWorker with a Blob.

pavelsavara: > You can start a WebWorker with a Blob.

But that's `eval` not CSP friendly. Typically, the people who want single-file also want CSP, because they are doing browser plugin.

pavelsavara: > single file that behaves differently depending on whether the current context is `window` or not?

For starting the sidecar, we could re-use `dotnet.js` loader. But if both of those scripts were bundled into user code "my-react-application.js" we don't have way how to `new Worker('./dotnet.js')` anymore.
Also, emscripten uses `dotnet.native.worker.js` script and at the moment we don't have nice way how to customize what is inside of it.

I will update the section with these details.

accepted/2023/wasm-browser-threads.md
pavelsavara: I added @SteveSandersonMS to review here and will add more details.

accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,506 @@
...
+
+## (13) Deputy + emscripten dispatch to UI + JSWebWorker + with sync JSExport
+- **B,F,J,N,R,T,U,W,a/b/c,d+f,i,l,n,s,v**
+- is variation of **(12)**
+    - with emscripten dispatch and marshaling in UI thread
+- this uses `emscripten_dispatch_to_thread_async` for `call_entry_point`, `complete_task`, `cwraps.mono_wasm_invoke_method_bound`, `mono_wasm_invoke_bound_function`, `mono_wasm_invoke_import`, `call_delegate_method` to get to the UI thread.
+- it uses other `cwraps` locally on UI thread, like `mono_wasm_new_root`, `stringToMonoStringRoot`, `malloc`, `free`, `create_task_callback_method`
+    - it means that interop related managed runtime code is running on the UI thread, but not the user code.
+    - it means that parameter marshalling is fast (compared to sidecar)
+    - it still needs to enter GC barrier and so it could block UI for GC run shortly
```
lambdageek, line 240: An interesting idea here is to create a GC-aware "transactions" API between the interop layer and the runtime.  Essentially: allow the marshaling code to succeed at calls to the runtime even if the GC is currently in a STW pause.  But instead of actually doing a GC Unsafe transition and executing some operation, just create a transaction log entry and immediately return back to marshaling.  Then after all the operations are queued up, the marshaling could commit the transaction and this will either block until GC STW is finished, or immediately do the operation.
(Another variant of this: if the marshaling is just interacting with handles or proxy objects and doesn't actually care about the final locations of things in the managed memory - just allocate into some side buffer and queue some work to be done at the end of STW to promote the values from the side buffer to the real managed heap).

This could possibly be done as a later optimization...

accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,506 @@
...
+- Solves **1,2** for managed code
+    - unless there is sync `JSImport`->`JSExport` call
+- Ignores **1,2** for JS callback
+    - emscripten main loop stays responsive
+- Solves **3,4,5**
+
+## (12) Deputy + managed dispatch to UI + JSWebWorker + with sync JSExport
+- **B,F,K,N,Q,S/T,U,W,a/b/c,d+f,h,l,n,s/z,v**
+- this uses `JSSynchronizationContext` to dispatch calls to UI thread
+    - this is "dirty" as compared to sidecar because some managed code is actually running on UI thread
```
lambdageek, line 222: I think this approach is also harder to explain than (13) - There is some C# code that is "special" or "knows what it is doing" that gets to do something that normal user C# code can't do.  I think it's only a matter of time before we get requests to allow some way for sophisticated users to be allowed to run on the UI thread too.  With (13) (and with all the sidecar proposals) I think this is a very different conversation.

If we believe that deputy proposals represent a point of stability (in the sense that user C# code is unlikely to cause the runtime to deadlock itself), I think (13) is better than (12).

On the other hand, writing a pile of C code feels more fragile than keeping the implementation in C#.

lambdageek (commented):
Pavel and I had a long offline discussion today and I'm coming around on the "deputy" proposal, provided we have a good way to explain for why some C# code is ok on the UI thread but not other C# code.

accepted/2023/wasm-browser-threads.md
pavelsavara: > If we believe that deputy proposals represent a point of stability (in the sense that user C# code is unlikely to cause the runtime to deadlock itself), I think (13) is better than (12).

I plan to make prototype next week, which will show if more MT unit tests pass the CI.

accepted/2023/wasm-browser-threads.md
pavelsavara: This is good idea.

I think that proxy disposal could be scheduled to later time.

I hope we are able to schedule also proxy **creation** via pre-existing handle, it would make sidecar approach more performant. 
Because it would not have to do per parameter call cross-threads. The difficult one I can think of would be 
`GCHandle` of a `TaskCompletionSource`. We could pre-allocate some TCSs and then just use them as needed.

If we are also able to schedule allocation (for arrays), perhaps we could serialize it all to the point that we could avoid double proxy of "comlink".

pavelsavara: Well, not so easy we can't pre-create arbitrary `TaskCompletionSource` because we are strongly typed `TaskCompletionSource<T>` in many cases.

pavelsavara: I added this as design **(16)**

accepted/2023/wasm-browser-threads.md
```diff
@@ -0,0 +1,682 @@
...
+ - we can configure pthread to keep running after synchronous thread_main finished. That's necessary to run any async tasks involving JavaScript interop.
+ - legacy interop has problems with GC boundaries.
+ - JSImport & JSExport work
+ - There is private JSSynchronizationContext implementation which is too synchronous
+ - There is draft of public C# API for creating JSWebWorker with JS interop. It must be dedicated un-managed resource, because we could not cleanup JS state created by user code.
+ - There is MT version of HTTP & WS clients, which could be called from any thread but it's also too synchronous implementation.
+ - Many unit tests fail on MT https://github.com/dotnet/runtime/pull/91536
+ - there are MT C# ref assemblies, which don't throw PNSE for MT build of the runtime for blocking APIs.
+
+# Alternatives - as considered 2023 Sep
```
pavelsavara, line 255: I'm not sure if the alternative designs we considered last year are useful to stay in this document for posterity.
I'm happy to delete anything below.
We experimented with many of those options.

lewing: Maybe move them to a public gist and link to it from here?

lewing (approved):

PR #302, opened 2023-10-19
Add initial Swift Interop design doc

jkoritzinsky:
Add a design doc for Swift/.NET interop, primarily focused on the runtime layer initially.

cc:
@dotnet/interop-contrib 
@dotnet/macios
@lambdageek
@SamMonoRT
@kotlarmilos
@JulieLeeMSFT
@amanasifkhalid
@davidortinau
@jkotas


proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+The first option seems like a natural fit, but there is one significant limitation: Attributes cannot be used in function pointers. Function pointers are a vital scenario for us to support as we want to support virtual method tables as they are used in scenarios like protocol witness tables. The second option is a natural fit as the `MemberFunction` calling convention combined with the various C-based calling conventions specifies that there is a `this` argument as the first argument. Defining `Swift` + `MemberFunction` to imply/require the `self` argument is a perfect conceptual extension of the existing model.
+
+###### Error register
+
+We have many options for handling the error register in the Swift calling convention:
+
+1. Use an attribute like `[SwiftError]` on a by-ref or `out` parameter to indicate the error parameter.
+2. Use a special type like `SwiftError<T>` on a by-ref or `out` parameter to indicate the error parameter.
+3. Use a special return type `SwiftReturn<T, TError>` to indicate the error parameter and combine it with the return value.
```
jkotas, line 60: Why do we need `TError` to be generic? I would expect that the low-level error to be a fixed pointer-sized type.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+Swift has two calling conventions at its core, the "Swift" calling convention, and the "SwiftAsync" calling convention. We will begin by focusing on the "Swift" calling convention, as it is the most common. The Swift calling convention has a few elements that we need to contend with. Primarily, it allows passing arguments in more registers. Additionally, it has two dedicated registers for specific components of the calling convention, the self register and the error register. The runtime support for this calling convention must support all of these features. The additional registers per argument is relatively easy to support, as each of our compilers must already have some support for this concept to run on Linux or Apple platforms today. We have a few options for the remaining two features we need to support.
+
+The "SwiftAsync" calling convention has an additional "async context" register and specific behaviors expected around how the Swift async state machine is called. Full investigation into the differences needed to implement it are still TBD.
+
+##### Self register
+
+We have two options for supporting the Self register in the calling convention:
```
AaronRobinsonMSFT, line 47: Another option is for a type of generic wrapper types that express the concept in a direct way. For example

```csharp
struct SwiftSelf<T>
{
}
```

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+Option 4 provides an alternative approach. As the "Error Register" is always register-sized, we can use runtime-supported helper functions to stash away and retrieve the error register. Unlike options 2 or 3, we don't need to do any signature manipulation as the concept of "this Swift call can return an error" is not represented in the signature. Responsibility to convert the returned error value into a .NET type, such as an exception, would fall to the projection tooling. Since option 4 does not express whether or not the target function throws at the signature level, the JIT/AOT compilers would always need to emit the calls to the helpers when compiling calls to and from Swift code. If the projection of Swift types into .NET will always use exception and not pass Swift errors as the error types directly, then Option 4 reduces the design burden on the runtime teams by removing the type that would only be used at the lowest level of the Swift/.NET interop.
```
jkotas, line 67: Also, this option leaves performance on the table.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+- [@dotnet/interop-contrib](https://github.com/orgs/dotnet/teams/interop-contrib)
+- [@dotnet/macios](https://github.com/orgs/dotnet/teams/macios)
+- [@lambdageek](https://github.com/lambdageek)
+- [@SamMonoRT](https://github.com/SamMonoRT)
+- [@kotlarmilos](https://github.com/kotlarmilos)
+- [@JulieLeeMSFT](https://github.com/JulieLeeMSFT)
+- [@amanasifkhalid](https://github.com/amanasifkhalid)
+- [@davidortinau](https://github.com/davidortinau)
+- [@jkotas](https://github.com/jkotas)
```
AaronRobinsonMSFT, line 33:
```suggestion
- [@jkotas](https://github.com/jkotas)
- [@stephen-hawley](https://github.com/stephen-hawley)
```

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+##### Self register
+
+We have two options for supporting the Self register in the calling convention:
+
+1. Use an attribute like `[SwiftSelf]` on a parameter to specify it should go in the self register.
+2. Specify that combining the `Swift` calling convention with the `MemberFunction` calling convention means that the first argument is the `self` argument.
+
+The first option seems like a natural fit, but there is one significant limitation: Attributes cannot be used in function pointers. Function pointers are a vital scenario for us to support as we want to support virtual method tables as they are used in scenarios like protocol witness tables. The second option is a natural fit as the `MemberFunction` calling convention combined with the various C-based calling conventions specifies that there is a `this` argument as the first argument. Defining `Swift` + `MemberFunction` to imply/require the `self` argument is a perfect conceptual extension of the existing model.
```
AaronRobinsonMSFT, line 52: >there is a `this` argument as the first argument.

Is this **always** the case in Swift? I would like to ensure we don't make odd constructs just because. @stephen-hawley Do you know?

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+##### Tuples
+
+In the Swift language, tuples are "exploded" out in the calling convention. Each element of a tuple is passed as a separate parameter. C# has its own language-level tuple feature in the ValueTuple family of types. The Swift/.NET interop story could choose to automatically handle tuple types in a similar way for a `CallConvSwift` function call. However, processing the tuple types in the JIT/AOT compilers is complicated and expensive. It would be much cheaper to defer this work upstack to the language projection, which can split out the tuple into individual parameters to pass to the underlying function. The runtime and libraries teams could add analyzers to detect value-tuples being passed to `CallConvSwift` functions at compile time to help avoid pits of failure.
```
AaronRobinsonMSFT, line 71:
```suggestion
In the Swift language, tuples are "unpacked" out in the calling convention. Each element of a tuple is passed as a separate parameter. C# has its own language-level tuple feature in the ValueTuple family of types. The Swift/.NET interop story could choose to automatically handle tuple types in a similar way for a `CallConvSwift` function call. However, processing the tuple types in the JIT/AOT compilers is complicated and expensive. It would be much cheaper to defer this work upstack to the language projection, which can split out the tuple into individual parameters to pass to the underlying function. The runtime and libraries teams could add analyzers to detect value-tuples being passed to `CallConvSwift` functions at compile time to help avoid pits of failure.
```

jkotas, line 71: > The runtime and libraries teams could add analyzers to detect value-tuples being passed to `CallConvSwift` functions at compile time to help avoid pits of failure.

I do not think that this is necessary. I assume that the swift bindings are going to be auto-generated, manually written `CallConvSwift` methods are going to rare, written by expert only.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+Like .NET, Swift has both struct types and class types. Unlike .NET, Swift's struct types have strong lifetime semantics more similar to C++ types than .NET structs. At the Swift ABI layer, there are broadly three types of structs: "POD/Trivial" structs, "Bitwise Takable/Movable" structs, and non-bitwise movable structs. The [Swift documentation](https://github.com/apple/swift/blob/main/docs/ABIStabilityManifesto.md#layout-and-properties-of-types) covers these different kinds of structs. Let's look at how we could map each of these categories of structs into .NET.
+
+"POD/Trivial" structs have no memory management required and no special logic for copying/moving/deleting the struct instance. Structs of this category can be represented as C# structs with the same field layout.
+
+"Bitwise Takable/Movable" structs have some memory management logic and require calls to Swift's ref-counting machinery to maintain expected lifetimes. Structs of this category can be projected into C# as a struct. When creating this C# struct, we would semantically treat each field as a separate local, create the C# projection of it, and save this "local" value into a field in the C# struct.
```
jkotas, line 83: Should this rather be in "Swift to .NET Language Feature Projections" section? I do not think we would want to do this as part of the runtime support.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+We plan to interop with Swift's Library Evolution mode, which brings an additional wrinkle into the Swift struct story. Swift's Library Evolution mode abstracts away all type layout and semantic information unless a type is explicitly marked as `@frozen`. In the Library Evolution case, all structs have "opaque" layout, meaning that their exact layout and category cannot be determined until runtime. As a result, we need to treat all "opaque" layout structs as possibly non-bitwise-movable at compile time as we will not know until runtime what the exact layout is. Swift/C++ interop does not use the Library Evolution mode, so it is not limited by opaque struct layouts. On the bright side, all "opaque" layout structs are passed as pointers in the calling convention, which helps simplify signature construction in the higher-level case. The size and layout information of a struct is available in its [Value Witness Table](https://github.com/apple/swift/blob/main/docs/ABIStabilityManifesto.md#value-witness-table), so we can look up this information at runtime for allocating struct instances and manipulating struct memory correctly.
```
jkotas, line 87: > Swift/C++ interop does not use the Library Evolution mode

How can they get away with it? I would expect that all calls into OS libraries must be done using the evolution mode.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+- [@dotnet/macios](https://github.com/orgs/dotnet/teams/macios)
+- [@lambdageek](https://github.com/lambdageek)
+- [@SamMonoRT](https://github.com/SamMonoRT)
+- [@kotlarmilos](https://github.com/kotlarmilos)
+- [@JulieLeeMSFT](https://github.com/JulieLeeMSFT)
+- [@amanasifkhalid](https://github.com/amanasifkhalid)
+- [@davidortinau](https://github.com/davidortinau)
+- [@jkotas](https://github.com/jkotas)
+
+## Design
```
jkotas, line 35: This can use a few sentences about the overall plan and major components of the design.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+4. (optional) A .NET shared framework package when targeting macOS, Mac Catalyst, iOS, or tvOS platforms that exposes the platform APIs or NuGet packages for each `.framework` that is exposed from Swift to .NET.
+    - This would be required to provide a single source of truth for Swift types so they can be exposed across an assembly boundary.
+
+#####  Exporting .NET to Swift
+
+There are two components to exporting .NET to Swift: Implementing existing Swift types in .NET and passing instances of those types to Swift, and exposing novel types from .NET code to Swift code to be created from Swift. Exposing novel types from .NET code to Swift code is considered out of scope at this time.
+
+For implementing existing Swift types in .NET, we will require the following tooling:
+
+1. A Roslyn source generator to generate any supporting code needed to produce any required metadata to pass instances of Swift-type-implementing .NET types to Swift.
```
AaronRobinsonMSFT, line 117: This is one way of doing this, but not the only way. One could also use the Swift details as the source of truth, emit source from a small tool and then include those during the compilation step. There is no requirement for a Roslyn source generator here. In fact, doing everything in a small tool ensures that both sides are the same.

I agree this could be done in a Roslyn source generator, but I do not think it is a requirement. Determining the UX needs for the Swift interop will help focus this decision point.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+## Scenarios and User Experience
+
+We would expect that users would be able to write C# code that can do simple Swift interop. Additionally, we would expect that for cases where the interop system does not support seamless interop, developers could write a shim in Swift that could be called from C# code. Developers should not need to write a shim in C or Assembly to interact with Swift APIs.
+
+## Requirements
+
+### Goals
+
+In short, we should completely eliminate the required C/assembly sandwich that's currently required to call Swift from C# code, and potentially vice versa. In particular, neither C# nor Swift users should have to deal with lower-level system state, like registers or stack state, to call Swift from C#.
```
AaronRobinsonMSFT, line 17: You might want to enumerate the runtimes that will be supported (i.e., mono, coreclr, NAOT). Is R2R support a goal or non-goal? If it is a goal, where is the line?

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+The projection tooling should be split into these components:
+
+##### Importing Swift into .NET
+
+1. A tool that takes in a `.swiftinterface` file or Swift sources and produces C# code.
+2. A library that provides the basic support for Swift interop that the generated code builds on.
+3. User tooling to easily generate Swift projections for a given set of `.framework`s.
+4. (optional) A .NET shared framework package when targeting macOS, Mac Catalyst, iOS, or tvOS platforms that exposes the platform APIs or NuGet packages for each `.framework` that is exposed from Swift to .NET.
+    - This would be required to provide a single source of truth for Swift types so they can be exposed across an assembly boundary.
```
jkotas, line 109: You do not need shared framework for single source of truth. A regular package is enough.

We have not done shared framework for WinRT and Objective-C for good reasons (versioning and size), so I do not see why we would want to do it for Swift.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+- [@davidortinau](https://github.com/davidortinau)
+- [@jkotas](https://github.com/jkotas)
+
+## Design
+
+### Runtime
+
+#### Calling Conventions
+
+Swift has two calling conventions at its core, the "Swift" calling convention, and the "SwiftAsync" calling convention. We will begin by focusing on the "Swift" calling convention, as it is the most common. The Swift calling convention has a few elements that we need to contend with. Primarily, it allows passing arguments in more registers. Additionally, it has two dedicated registers for specific components of the calling convention, the self register and the error register. The runtime support for this calling convention must support all of these features. The additional registers per argument is relatively easy to support, as each of our compilers must already have some support for this concept to run on Linux or Apple platforms today. We have a few options for the remaining two features we need to support.
```
AaronRobinsonMSFT, line 41: Are these two calling conventions or just an optional register usage when involving "async"? Not trying to quibble or terms, but if it is simply a register being used when async is involved, then I wouldn't phrase this as two and instead call out that as another thing to consider like the self and error semantics below.

AaronRobinsonMSFT (commented):
Great start!

lemonmojo:
Very much looking forward to this! 😎

I understand that Swift -> C# interop is not a big focus here but allow me to add my vote for that here as well.
While we found ways to make this work for us ([Beyond.NET](https://github.com/royalapplications/beyondnet)), first party support would obviously be very much appreciated.

proposed/swift-interop.md
rolfbjarne: This seems natural to me (the concept of certain types living in specific registers is already a common concept: for instance floats and doubles often do).

proposed/swift-interop.md
rolfbjarne: While I don't think this is a concern today, it would break if one day there's an architecture where the presence of the error return changes how other registers are used.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+The "SwiftAsync" calling convention has an additional "async context" register and specific behaviors expected around how the Swift async state machine is called. Full investigation into the differences needed to implement it are still TBD.
+
+##### Self register
+
+We have two options for supporting the Self register in the calling convention:
+
+1. Use an attribute like `[SwiftSelf]` on a parameter to specify it should go in the self register.
+2. Specify that combining the `Swift` calling convention with the `MemberFunction` calling convention means that the first argument is the `self` argument.
```
kotlarmilos, line 50: Does the `MemberFunction` always assume a self reference?

In Swift calling convention, in static methods on class types, self is a reference to the class metadata. Also, in non-mutating instance methods on value types, self is a value which may require multiple registers or may need to be be passed indirectly.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+We have a prototype that uses the first option; however, using an attribute has the same limitations as the attribute option for the self register. As a result, we should consider an alternative option.
+
+Options 2 and 3 have similar characteristics: They both express that the to-be-called Swift function uses the Error Register in the signature and they both require signature manipulation in the JIT/AOT compilers. Option 2 is likely cheaper, as the argument with `ref/out SwiftError<T>` type can easily be associated with the error register. With Option 2 though, we'd have to determine the behavior if multiple parameters used the `SwiftError` type (likely throw an exception). Option 3 provides a more functional-programming-style API that is more intuitive, but likely more expensive to implement.
```
kotlarmilos, line 65: > Option 3 provides a more functional-programming-style API that is more intuitive, but likely more expensive to implement.

@stephen-hawley Is there a practical use case for the `SwiftReturn<T, TError>`? This may be responsibility of the projection tools.

kotlarmilos, line 65: > With Option 2 though, we'd have to determine the behavior if multiple parameters used the `SwiftError` type (likely throw an exception).

What are implications if multiple parameters used the `SwiftError` type?

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+##### Automatic Reference Counting and Lifetime Management
+
+Swift has a very strongly-defined lifetime and ownership model. This model is specified in the Swift ABI and is similar to Objective-C's ARC (Automatic Reference Counting) system. The Binding Tools for Swift tooling handles these explicit lifetime semantics with some generated Swift code. In the new Swift/.NET interop, management of these lifetime semantics will be done by the Swift projection and not by the raw calling-convention support.
```
kotlarmilos, line 75: How is this implemented in Objective-C interop? Is it the responsibility of projection tools and not the runtime? If we consider future extensibility, it may be beneficial to handle it at the runtime level.

proposed/swift-interop.md
```diff
@@ -0,0 +1,132 @@
...
+
+#### Projection Tooling Components
+
+The projection tooling should be split into these components:
+
+##### Importing Swift into .NET
+
+1. A tool that takes in a `.swiftinterface` file or Swift sources and produces C# code.
+2. A library that provides the basic support for Swift interop that the generated code builds on.
+3. User tooling to easily generate Swift projections for a given set of `.framework`s.
```
kotlarmilos, line 107: Is this a combination of 1. and 2.?

proposed/swift-interop.md
kotlarmilos: I believe there are both pros and cons; If it is implemented as a source generator, it is closer to the runtime and is platform-agnostic. On the other hand, if it is integrated into projection tools, it is probably faster but is more platform-specific.

AaronRobinsonMSFT: > If it is implemented as a source generator, it is closer to the runtime and is platform-agnostic.

Any small stand alone tool can be pure C# and ship just like a source generator. This is merely about packaging logic, nothing more. There is no requirement that the tooling for this live in the Roslyn process.

proposed/swift-interop.md
AaronRobinsonMSFT: > Does the MemberFunction always assume this reference as the first argument?

Yes. See my comment at https://github.com/dotnet/designs/pull/302#discussion_r1366184584.

proposed/swift-interop.md
AaronRobinsonMSFT: > How is this implemented in Objective-C interop? 

There is a complex dance between the runtime/GC and the code shipped out of macios. See https://github.com/dotnet/designs/blob/main/accepted/2021/objectivec-interop.md and https://learn.microsoft.com/dotnet/api/system.runtime.interopservices.objectivec.

proposed/swift-interop.md
stephen-hawley: Stashing away and retrieving the error later needs to be thread safe.

proposed/swift-interop.md
stephen-hawley: As an FYI, in the early days of Binding Tools for Swift, I tried to do exactly this, making a distinction between blitable and non-blittable structs. It doubles the number of special cases and was a factory for bugs all along the line. In addition, this should not be limited to structs, but should be extended to enums, which are also value types. In addition to blitability issues, this brings up the packing of cases in swift enums which is not only non-trivial, it's not documented, subject to change and fragile (adding a case can change the packing). Further, value types in swift are semantically different from value types in C# in that code may get executed when they go out of scope. For all of these reasons, I chose to implement them as `IDisposable` classes with an opaque payload. In terms of passing them more efficiently, we have more options at our disposal. We could take a pinvoke like this:
```
internal extern void someCallWithAStruct (SwiftStruct st);
````
and reimplement it as this:
```
internal extern void someCallWithAStruct_exploded4w (ulong w0, ulong w1, ulong w2, ulong w4)
```
As a final however, this may be moot if we have access to the self pointer. This is because if we require `enable-library-evolution` then the swift compiler generates thunks that use pass by reference for the self pointer (if any) and IIRC, passes all structs by reference (making a copy on the stack).

File: proposed/swift-interop.md (hunk 5 above)
Comment (stephen-hawley) [reply proposed/swift-interop.md:52]: We can make this moot pretty easily by noting that the stable witness table layouts are fragile, and undefined. We should never make use of them directly. However, with `enable-library-evolution` all virtual calls and protocol calls are available via thunks that use the `self` pointer. In terms "is the self pointer always first" that depends. When BTfS has to wrap a swift function to make it callable, the general approach is to:
1.  insert a pointer to the instance before the first argument if there is an instance pointer
2. insert a pointer to memory allocated for the return value before the first argument (which may already be self)


So, no, it is not always the case. However, I have not looked into detail how the swift ABI handles larger value types as return values in `enable-swift-evolution`.

Comment (AaronRobinsonMSFT) [reply proposed/swift-interop.md:52]: I think relying on `MemberFunction` might be a bit too much magic then. I think being more declarative via something akin to `SwiftThis<T>` might be more appropriate.

File: proposed/swift-interop.md (hunk 13 above)
Comment (stephen-hawley) [reply proposed/swift-interop.md:41]: Technically, there are more calling conventions. In addition to "Swift" and "SwiftAsync",  there is `enable-library-evolution` which uses a similar calling convention in order to manage value type layouts that may change. In addition, you can annotate any function or function type with `@convention (foo)` where today foo is a member of the set {c, block, objc}. Finally, if a class inherits from `NSObject` and has the `@objc` attribute, the compiler will writer an objective C dispatcher for every method. This dispatcher ends up calling the vtable methods.

File: proposed/swift-interop.md (hunk 15 above)
Comment (stephen-hawley) [reply proposed/swift-interop.md:65]: I wouldn't a method like that. I would write a C# method that takes does something like this:
```
public nint SomeMethodThatMightThrow ()
{
    var errReturn = SomeMethodThatMightThrow_Pinvoke ();
    if (errReturn.HasError) {
        throw SwiftException.FromError (errReturn.Error);
    }
    return errReturn.Value;
}
```

File: proposed/swift-interop.md (hunk 6 above)
Comment (stephen-hawley) [reply proposed/swift-interop.md:71]: I have no problem packing and unpacking tuples from swift, but in the name of expedience, tuples as arguments are passed as a pointer to the tuple representation in a wrapper. This will present issues in closures.

File: proposed/swift-interop.md (hunk 15 above)
Comment (stephen-hawley) [reply proposed/swift-interop.md:65]: But also important to consider is the other side of the coin: C# overrides an `open` class from swift and overrides and `open` method that throws. Today, BTfS does this for all `open` classes:
1. write a swift class that overrides ever overridable method which includes a simulated vtable for virtual methods
2. write a C# class named after the original class which in its static constructor sends a pile of function pointers into the swift class which are used to populate the vtable.  Each of the function pointers points to a static C# method which I call a receiver. The receiver gets (among other things) the swift instance which it uses to look up the C# instance to handle the call and a pointer to a Medusa tuple for the return and exception.
3. In the case of an exception, BTfS injects a try catch, projects the exception onto a swift exception. This in turns gets set into the Medusa tuple and sent back to swift where the invokes unmarshals it.
This is a lot of work and I would like to find ways for it to be more betterer.

Comment (stephen-hawley) [reply proposed/swift-interop.md:65]: I should define this - a Medusa tuple is of the form `(Bool, T, Error)` where T is the return type of a function. A method that may throw gets passed a pointer to memory allocated for a Medusa tuple, but leaves it uninitialized. If it has thrown, it sets the `Bool` portion to true and sets the `Error` portion leaving `T` uninitialized. In the case of not throwing, it sets the `Bool` portion to false and sets the `T` portion. This violates swift semantics in that Swift *really* hates having uninitialized members, and if you ever try to look at the tuple, you will die. Why not use a Swift enum for this and crack it open? Because getting at the payload of an enum in C# is roughly 5-10x the cost of getting at the elements of an tuple and the means of accessing a tuple's elements can be memoized whereas getting the payload out of an enum is fixed and can't be memoized in a useful way.
When I started looking at exceptions in Swift, I was surprised that they didn't just define them like this:
```
public enum ErrorReturn<T> {
    case value (T)
    case error(Error)
}
```
and make the call syntactic sugar. If I had to speculate, it was to make interoperation with ObjC errors easier/fast/smoother/more bettererer.

Comment (agocke) [reply proposed/swift-interop.md:65]: One thought -- this is very much expressing exceptions in Swift. The syntax is even throw/catch. Should we simply expose this as a C# exception from the runtime? Maybe of a specific type, like a new SwiftException class?

Comment (jkoritzinsky) [reply proposed/swift-interop.md:65]: Exposing the error directly as an exception from the runtime layer is what we did in Objective-C. It introduces additional complexity in the interop layer (we always generate IL stubs for P/Invokes into the Objective-C selector functions to enable this feature) and it doesn't enable a projection to provide a nice error-mapping experience (ie to translate a swift error for "file not found" into FileNotFoundException would require a try-catch-throw instead of just an "if error then throw" flow).

File: proposed/swift-interop.md (hunk 5 above)
Comment (jkoritzinsky) [reply proposed/swift-interop.md:52]: > We can make this moot pretty easily by noting that the stable witness table layouts are fragile, and undefined. We should never make use of them directly. However, with `enable-library-evolution` all virtual calls and protocol calls are available via thunks that use the `self` pointer. In terms "is the self pointer always first" that depends. When BTfS has to wrap a swift function to make it callable, the general approach is to:
> 
> 1. insert a pointer to the instance before the first argument if there is an instance pointer
> 2. insert a pointer to memory allocated for the return value before the first argument (which may already be self)
> 
> So, no, it is not always the case. However, I have not looked into detail how the swift ABI handles larger value types as return values in `enable-swift-evolution`.

Re item 1: Is that argument that was introduced a replacement for the `self` argument as the functions emitted by BTfS were static and did not have a `self` argument?

Re item 2: For large return types that are passed by value, that is an implementation detail that our JIT/AOT compilers should handle automatically (the primary use case of `MemberFunction` initially was to help ensure this determination was correct in some MSVC corner cases). The user should never need to declare the "return buffer" argument manually.

File: proposed/swift-interop.md (hunk 13 above)
Comment (jkoritzinsky) [reply proposed/swift-interop.md:41]: @stephen-hawley Technically, `enable-library-evolution` is not a separate calling convention, but it does change the ABI.

Here's my definition of "calling convention" for this section (I'll add it do the doc): "A distinct calling convention is represented by a specific calling convention name in LLVM IR".

The Swift compiler doesn't use a separate calling convention for Library Evolution. Instead it changes how the arguments are passed to the call (changing parameters from their struct types to pointers) while still using the Swift or SwiftAsync calling convention.

I'll make sure to mention the `@convention` option and `@objc` as well.

File: proposed/swift-interop.md (hunk 14 above)
Comment (jkoritzinsky) [reply proposed/swift-interop.md:50]: In these cases, I would expect the class metadata to be passed as the first argument (to represent the `self`). By passing the argument as the value type instead of a pointer, the JIT/AOT compilers should handle lowering the call to the correct register usage.

File: proposed/swift-interop.md (hunk 15 above)
Comment (jkoritzinsky) [reply proposed/swift-interop.md:65]: We'd likely want to throw an `InvalidProgramException` as that would not be a valid signature.

Comment (agocke) [reply proposed/swift-interop.md:65]: Two follow-up questions: the Swift ABI describes that throwing is a part of the signature, but there's effectively a covariance relationship in that non-throwing functions can convert to throwing versions. If we go the extra-parameter route, presumably it would be up to either the user or the translation layer to provide a conversion function which never sets the error.

And second, what happens for exception thrown in C# code that is called from Swift? Kind of the same question of what happens in C++, but I could see us wanting to provide a slightly better experience for Swift than for C++.

Comment (AaronRobinsonMSFT) [reply proposed/swift-interop.md:65]: > I could see us wanting to provide a slightly better experience for Swift than for C++.

We have some mechanism for Objective-C on this front, but it is global and less than ideal. It also requires the ownership platform (in this case, Objective-C) to provide a mechanism for throwing an exception in their desired manner. I would be very reluctant to try and handle the throwing of a Swift exception from the runtime ourselves or define it in managed code.

All that aside, if all that is needed to "throw" is creating a Swift projection and setting the error pointer, then that should be relatively simple.

File: proposed/swift-interop.md (hunk 12 above)
Comment (jkoritzinsky) [reply proposed/swift-interop.md:109]: I was intending for this to refer to a "FrameworkReference" that is automatically referenced as part of a TPM in a TFM, as we have done with WinRT and Objective-C projections on their relevant platforms. I'll re-word this to better explain my intent.

File: proposed/swift-interop.md (hunk 1 above)
Comment (jkoritzinsky) [reply proposed/swift-interop.md:60]: I wasn't sure if the Swift error value could be passed by value and split across multiple registers (as can be done with the `self` register). If it can be passed by value, then we'd need this value to be generic to represent that. It looks like it can only be passed by pointer, so we can remove the generic argument.

File: proposed/swift-interop.md (hunk 6 above)
Comment (jkoritzinsky) [reply proposed/swift-interop.md]: Our goal is to avoid emitting any Swift code for simple cases, so the plan here would be to have the signature unpack tuples as the ABI expects. That way we avoid any issues with callbacks.

File: proposed/swift-interop.md [hunk 18]
```diff
@@ -31,60 +33,75 @@ C# and Swift are different languages with different language semantics. It is no
 - [@amanasifkhalid](https://github.com/amanasifkhalid)
 - [@davidortinau](https://github.com/davidortinau)
 - [@jkotas](https://github.com/jkotas)
+- [@stephen-hawley](https://github.com/stephen-hawley)
 
 ## Design
 
+We plan to split the work into at least three separate components. There will be work at the runtime layer to handle the basic calling-convention and register-allocation work to enable calling basic Swift functions without needing to write custom C or Assembly code or wrapping every Swift function with a C-style shape. Up-stack, there will be suite of code-generation tools to provide a higher-level projection of Swift concepts into .NET. Finally, the product will include some mechanism to easily reference Swift libraries from .NET on platforms that natively provide a Swift runtime and libraries, Apple platforms in particular.
```
AaronRobinsonMSFT, line 40:
```suggestion
We plan to split the work into at least three separate components. There will be work at the runtime layer to handle the basic calling-convention and register-allocation work to enable calling basic Swift functions without needing to write custom C or Assembly code or wrapping every Swift function with a C-style shape. Upstack, there will be suite of code-generation tools to provide a higher-level projection of Swift concepts into .NET. Finally, the product will include some mechanism to easily reference Swift libraries from .NET on platforms that natively provide a Swift runtime and libraries, Apple platforms in particular. 
```

proposed/swift-interop.md
```diff
@@ -94,7 +111,21 @@ All designs in this section should be designed such that they are trimming and A
...
+
+We plan to interop with Swift's Library Evolution mode, which brings an additional wrinkle into the Swift struct story. Swift's Library Evolution mode abstracts away all type layout and semantic information unless a type is explicitly marked as `@frozen`. In the Library Evolution case, all structs have "opaque" layout, meaning that their exact layout and category cannot be determined until runtime. As a result, we need to treat all "opaque" layout structs as possibly non-bitwise-movable at compile time as we will not know until runtime what the exact layout is. Swift/C++ interop is not required to use the Library Evolution mode in all cases as it can statically link against Swift libraries, so it is not limited by opaque struct layouts in every case. The size and layout information of a struct is available in its [Value Witness Table](https://github.com/apple/swift/blob/main/docs/ABIStabilityManifesto.md#value-witness-table), so we can look up this information at runtime for allocating struct instances and manipulating struct memory correctly.
```
AaronRobinsonMSFT, line 124: > We plan to interop with Swift's Library Evolution mode

Is this needed for the narrow set of libraries/frameworks we want to support in v1?

jkoritzinsky: Yes. We need to use LibraryEvolution mode for our Swift interop as that is the only ABI-stable story for Swift. Otherwise we'd need to re-compile the Swift code per-OS-version for each possible target OS, instead of building the managed code once for all target iOS or macOS versions (which is how .NET generally works today). Also, we wouldn't be able to use the documented `.swiftinterface` files. We'd need to use the compiler-specific `.swiftmodule` files or parse Swift code directly, both of which are much more expensive. Also, I'm not sure if it's even possible to compile against "Library Evolution" libraries using the non-stable ABI (testing in Godbolt looks like it's one or the other).

proposed/swift-interop.md
```diff
@@ -0,0 +1,170 @@
...
+In Swift, sometimes the `self` register is used for non-instance state. For example, in static functions, the type metadata is passed as the `self` argument. Since static functions are not member functions, we may want to not use the `MemberFunction` calling convention. Alternatively, we could provide a `SwiftSelf<T>` type to specify "this argument goes in the self register". Specifying the type twice in a signature would generate an `InvalidProgramException`. This would allow the `self` argument to be specified anywhere in the argument list.
+
+For reference, explicitly declaring a function with the Swift or SwiftAsync calling conventions in Clang requires the "context" argument, the value that goes in the "self" register, as the last parameter or the penultimate parameter followed by the error parameter.
+
+###### Error register
+
+We have many options for handling the error register in the Swift calling convention:
+
+1. Use an attribute like `[SwiftError]` on a by-ref or `out` parameter to indicate the error parameter.
+2. Use a special type like `SwiftError` on a by-ref or `out` parameter to indicate the error parameter.
```
rolfbjarne, line 73: ref/out parameters aren't blittable, so maybe should this be a `SwiftError*` instead?

proposed/swift-interop.md
```diff
@@ -0,0 +1,170 @@
...
+
+##### Structs/Enums
+
+Like .NET, Swift as both value types and class types. The value types in Swift include both structs and enums. When passed at the ABI layer, they are generally treated as their composite structure and passed by value in registers. However, when Library Evolution mode is enabled, struct layouts are considered "opaque" and their size, alignment, and layout can vary between compile-time and runtime. As a result, all structs and enums in Library Evolution mode are passed by a pointer instead of in registers. Frozen structs and enums (annotated with the `@frozen` attribute) are not considered opaque and will be enregistered. We plan to interoperate with Swift through the Library Evolution mode, so we will generally be able to pass structures using opaque layout.
```
rolfbjarne, line 102:
```suggestion
Like .NET, Swift has both value types and class types. The value types in Swift include both structs and enums. When passed at the ABI layer, they are generally treated as their composite structure and passed by value in registers. However, when Library Evolution mode is enabled, struct layouts are considered "opaque" and their size, alignment, and layout can vary between compile-time and runtime. As a result, all structs and enums in Library Evolution mode are passed by a pointer instead of in registers. Frozen structs and enums (annotated with the `@frozen` attribute) are not considered opaque and will be enregistered. We plan to interoperate with Swift through the Library Evolution mode, so we will generally be able to pass structures using opaque layout.
```

proposed/swift-interop.md
jkoritzinsky: Since we don't actually pass the pointer value in the call (and instead use it basically as a storage location for the result), we can make the `ref/out` behavior not cause a problem due to blittability. We also may be able to optimize it better by using a managed ref instead of a pointer (as the address-exposure of the managed ref could be optimized away, but we can't do that for an unmanaged pointer). If we don't want to special case it this much though, an unmanaged pointer is also a reasonable option.

jkoritzinsky: Looks like C# disallows `ref/out` in `UnmanagedCallersOnly` signatures. I'll update the proposal with some alternative designs.

proposed/swift-interop.md
```diff
@@ -0,0 +1,175 @@
...
+
+Swift also provides a strong Objective-C interop story through the `@objc` attribute. Types with the `@objc` attribute are exposed through the Swift ABI as well as through Objective-C selectors. We will not call into Swift APIs using the Objective-C interop in this interop story. However, the upstack projection tooling may need to do additional work for types that may need to have Objective-C interop support exposed (in the case that a .NET user type derives from a Swift type that had the `@objc` attribute).
+
+##### Self register
+
+We have a few options for supporting the Self register in the calling convention:
```
jkotas, line 56: What is going to the process to pick the option that we are going with? Do you plan to do that as part of this PR?

Once we have the design decided, this should be restructured to mention the final design first and then discuss the rejected alternatives.

jkoritzinsky: I wanted to make sure we had options for open discussion here first, and then I'll pick the options that we will go with. I think we're closing in on one of a few options for the SwiftError (and async context) types, but I wanted to give a little more time for discussion on the Self register as I added conceptually new options (the `SwiftSelf` type in addition to the calling-convention and attribute ideas) since the last set of reviews.

proposed/swift-interop.md
```diff
@@ -0,0 +1,183 @@
...
+
+1. Use a special type like `SwiftError` on a `ref` or `out` parameter to indicate the error parameter.
+2. Use a special type like `SwiftError*` to indicate the error parameter;
+
+Both options characteristics: They both express that the to-be-called Swift function uses the Error Register in the signature and they both require signature manipulation in the JIT/AOT compilers. The first option represents the "error return" as a managed `ref` or `out` parameter. Like with `SwiftSelf`, we would throw an `InvalidProgramException` for a signature with multiple `SwiftError` parameters.
+
+Option 2 provides an alternative to Option 2 as `ref` and `out` parameters are not supported in `UnmanagedCallersOnly` methods. Option 1 likely provides more possible JIT optimizations as managed pointers are more easily reasoned about in some of our JIT/AOT intermediate representations, but supporting it in all of our goal scenarios (P/Invoke, UnmanagedCallersOnly, function pointers) would require some minor language changes to allow the managed pointers in the `UnmanagedCallersOnly` signature for these special types. Option 2 (`SwiftError*`) does not require any language changes.
```
jkotas, line 84: For the `UnmanagedCallersOnly` case, the JIT would need to:
- Introduce a dummy local
- Pretend that the address of this local is method argument to allow user code to set it
- Read the value of the local at the end of the method and store it into the error register

I think it is about at least as complicated to implement as the rejected option 2, for RyuJIT at least. The rejected option 2 may be actually simpler - the JIT just needs to take the SwiftReturn value at the end of the method and store it into the error register and the return value register or return buffer.

proposed/swift-interop.md
```diff
@@ -0,0 +1,183 @@
...
+
+##### Structs/Enums
+
+Like .NET, Swift has both value types and class types. The value types in Swift include both structs and enums. When passed at the ABI layer, they are generally treated as their composite structure and passed by value in registers. However, when Library Evolution mode is enabled, struct layouts are considered "opaque" and their size, alignment, and layout can vary between compile-time and runtime. As a result, all structs and enums in Library Evolution mode are passed by a pointer instead of in registers. Frozen structs and enums (annotated with the `@frozen` attribute) are not considered opaque and will be enregistered. We plan to interoperate with Swift through the Library Evolution mode, so we will generally be able to pass structures using opaque layout.
```
jkotas, line 115: How are the structs returned using opaque layout? Is the caller required to allocate big-enough buffer and pass it as an argument?

proposed/swift-interop.md
jkotas, line 84: > Option 1 likely provides more possible JIT optimizations as managed pointers are more easily reasoned about in some of our JIT/AOT intermediate representations

This may be true in general. For this case, the JIT should be able to see all places that the address flown into and optimize it well. For example, JIT is able to optimize this fine:

https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunykGArgDtc2AGYwGASyEYGGGLgwAKAJRcenajx3TZDBAwC8DcgG4NuvRgBUDAA7GGAMgQXtVpgHYGN++50AXxpAoA=

proposed/swift-interop.md
```diff
@@ -0,0 +1,183 @@
...
+
+In Swift, sometimes the `self` register is used for non-instance state. For example, in static functions, the type metadata is passed as the `self` argument. Since static functions are not member functions, we may want to not use the `MemberFunction` calling convention. Alternatively, we could provide a `SwiftSelf` type to specify "this argument goes in the self register". Specifying the type twice in a signature would generate an `InvalidProgramException`. This would allow the `self` argument to be specified anywhere in the argument list. Alternatively, many sections of the Swift ABI documentation, as well as the Clang docs refer to this parameter as the "context" argument instead of the "self" argument, so an alternative name could be `SwiftContext`.
+
+For reference, explicitly declaring a function with the Swift or SwiftAsync calling conventions in Clang requires the "context" argument, the value that goes in the "self" register, as the last parameter or the penultimate parameter followed by the error parameter.
```
AaronRobinsonMSFT, line 65: This statement would seem to imply that (2) is the best option here.

proposed/swift-interop.md
```diff
@@ -0,0 +1,183 @@
...
+Rejected Option 3 provides an alternative approach. As the "Error Register" is always register-sized, we can use runtime-supported helper functions to stash away and retrieve the error register. Unlike options 2 or 3, we don't need to do any signature manipulation as the concept of "this Swift call can return an error" is not represented in the signature. Responsibility to convert the returned error value into a .NET type, such as an exception, would fall to the projection tooling. Since option 4 does not express whether or not the target function throws at the signature level, the JIT/AOT compilers would always need to emit the calls to the helpers when compiling calls to and from Swift code. If the projection of Swift types into .NET will always use exception and not pass Swift errors as the error types directly, then Option 4 reduces the design burden on the runtime teams by removing the type that would only be used at the lowest level of the Swift/.NET interop. However, Option 4 would leave some performance on the table as it would effectively require us to store and read the error value into thread-local storage instead of reading the error value from the register directly.
```
AaronRobinsonMSFT, line 97: I don't see the rational for rejection here.

proposed/swift-interop.md
```diff
@@ -0,0 +1,183 @@
...
+
+##### Tuples
+
+In the Swift language, tuples are "unpacked" in the calling convention. Each element of a tuple is passed as a separate parameter. C# has its own language-level tuple feature in the ValueTuple family of types. The Swift/.NET interop story could choose to automatically handle tuple types in a similar way for a `CallConvSwift` function call. However, processing the tuple types in the JIT/AOT compilers is complicated and expensive. It would be much cheaper to defer this work upstack to the language projection, which can split out the tuple into individual parameters to pass to the underlying function. The runtime and libraries teams could add analyzers to detect value-tuples being passed to `CallConvSwift` functions at compile time to help avoid pits of failure. As we expect most developers to use the higher-level tooling and to not use `CallConvSwift` directly, we would likely defer any analyzer work until we have a suitable use case.
```
AaronRobinsonMSFT, line 107: I'm not following the UX considerations here. For Tuples it seems we are okay with projection generation tooling to unpack potentially complex Tuples, but for the Error register asking the projection tooling to insert specific calls to `Marshal.Get/SetLastSwiftError` is too much? How are we reconciling this?

jkoritzinsky: The issue with `Marshal.Get/SetLastSwiftError` is more about the cost of reading and writing to thread-local storage than the cost in the projection tooling of inserting the calls.

proposed/swift-interop.md
```diff
@@ -0,0 +1,183 @@
...
+
+All designs in this section should be designed such that they are trimming and AOT-compatible by construction. We should work to ensure that no feature requires whole-program analysis (such as custom steps in the IL Linker) to be trim or AOT compatible.
+
+#### Swift to .NET Language Feature Projections
+
+##### Structs/Value Types
+
+Unlike .NET, Swift's struct types have strong lifetime semantics more similar to C++ types than .NET structs. At the Swift ABI layer, there are broadly three types of structs/enums: "POD/Trivial" structs, "Bitwise Takable/Movable" structs, and non-bitwise movable structs. The [Swift documentation](https://github.com/apple/swift/blob/main/docs/ABIStabilityManifesto.md#layout-and-properties-of-types) covers these different kinds of structs. Let's look at how we could map each of these categories of structs into .NET.
```
AaronRobinsonMSFT, line 129: > how we could map each of these categories of structs into .NET.

I assume we will need some indication of when a `struct` is "POD/Trivial" or "Bitwise Takable/Movable"? The last one appears to be akin to implementing the type as a `class`, if not will we need another flag/inidication?

proposed/swift-interop.md
```diff
@@ -0,0 +1,183 @@
...
+#####  Exporting .NET to Swift
+
+There are two components to exporting .NET to Swift: Implementing existing Swift types in .NET and passing instances of those types to Swift, and exposing novel types from .NET code to Swift code to be created from Swift. Exposing novel types from .NET code to Swift code is considered out of scope at this time.
+
+For implementing existing Swift types in .NET, we will require one of the following tooling options:
+
+1. A Roslyn source generator to generate any supporting code needed to produce any required metadata, such as type metadata and witness tables, to pass instances of Swift-type-implementing .NET types defined in the current project to Swift.
+2. An IL-post-processing tool to generate the supporting code and metadata from the compiled assembly.
+
+An IL-post-processing tool would not integrate well with Hot-Reload support, so a Roslyn source-generator is likely a better option.
```
AaronRobinsonMSFT, line 165: > integrate well with Hot-Reload support

This is an assumption I would like to confirm. Is there latitude with how Swift operates where users can reload types that Swift may already know about?

AaronRobinsonMSFT: What I am getting at, is HotReload a tenable feature for autogenerated interop code? Where are the boundaries?

proposed/swift-interop.md
jkoritzinsky: I believe we can treat "Bitwise Takable/Movable" structs as a set of individual values (which we wrap into a .NET struct) instead of having to treat it as a Swift struct (with Swift lifetime rules) when marshalled into .NET. I believe we could design the projection tooling to not need an indication at the managed code layer for "POD/Trivial" vs "Bitwise Takable/Movable". The non-bitwise-movable structs definitely need to be represented with a `class`.

In any case, the distinction here only matters for `@frozen` structs where their layout can be known at compile time. For all opaque structs, we'll need to treat them as non-bitwise-movable until proven otherwise. Based on @stephen-hawley's comments, trying to determine the difference here may be very expensive and bug-prone, especially at runtime. Having one solution here for all structs may be the best option.

proposed/swift-interop.md
jkoritzinsky: Swift has a Hot Reload feature for SwiftUI, but Swift in general doesn't look to have a general Hot Reload feature. Changing the contents of a method body (a la Edit-and-Continue style edits) on the .NET side should be something we can make work (especially as .NET on iOS and .NET on macOS currently supports Hot Reload), but adding members that make a type adopt new protocols likely wouldn't work well.

The biggest problem with using a post-processing tool is not that it would break Hot-Reload for the Swift interop code, but that it would break Hot-Reload for the assembly as a whole, including in .NET-only portions of the code, due to the IL post-processing step.

AaronRobinsonMSFT: > but adding members that make a type adopt new protocols likely wouldn't work well.

Or fields.

> The biggest problem with using a post-processing tool is not that it would break Hot-Reload for the Swift interop code, but that it would break Hot-Reload for the assembly as a whole, including in .NET-only portions of the code, due to the IL post-processing step.

Yep, that is the clarity we need here.

proposed/swift-interop.md
jkoritzinsky: The proposal for `Get/SetLastSwiftError` is to cache the result when coming in from a Swift call/returning from a Swift call like how we handle `Get/SetLastSystemError` in LibraryImport.

Making it directly operate on the register itself would mean that the JITs/AOT compilers/interpreters would need to ensure that the error register is not touched by any other code, user or otherwise from the Swift call to the `GetLastSwiftError` call. Modelling the error as a special type in the signature makes "ensure that we return the value from this register after the call" easier to model.

proposed/swift-interop.md
jkoritzinsky: Yes, the caller is required to allocate a big-enough buffer (using the Value Witness Table to retrieve the required size) and pass it as an argument.

jkoritzinsky: I'll add a blurb talking about this to this section.

yaakov-h:
I can see roughly how value types and class types would map over, but what about Swift Extensions?

rolfbjarne (approved):

jkoritzinsky:
> I can see roughly how value types and class types would map over, but what about Swift Extensions?

We're still working on the language projections for features like Extensions and protocols. We'll update this proposal with those designs as we start to work on them.

proposed/swift-interop.md
```diff
@@ -0,0 +1,189 @@
...
+2. Use a special return type `SwiftReturn<T>` to indicate the error parameter and combine it with the return value.
+3. Use special helper functions `Marshal.Get/SetLastSwiftError` to get and set the last Swift error. Our various compilers will automatically emit a call to `SetLastSwiftError` from Swift functions and emit a call to `GetLastSwiftError` in the epilog of `UnmanagedCallersOnly` methods. The projection generation will emit calls to `Marshal.Get/SetLastSwiftError` to get and set these values to and from exceptions.
+4. Implicitly transform the error register into an exception at the interop boundary.
+5. Use `SwiftError` on a `ref` or `out` parameter to indicate the error parameter.
+
+We have a prototype that uses Rejected Option 1; however, using an attribute has the same limitations as the attribute option for the self register.
+
+Rejected Option 2 provides a more functional-programming-style API that is more intuitive than the `SwiftError` options above, but likely more expensive to implement. As a result, Option 2 would be better as a higher-level option and not the option implemented by the JIT/AOT compilers.
```
jkotas, line 97: > but likely more expensive to implement.

I do not think that this is measurably more expensive to implement.

Having said that, I am ok the currently selected `SwiftError*` design.

jkotas (approved):

AaronRobinsonMSFT (approved):

kotlarmilos (approved):

dalexsoto (approved):

proposed/swift-interop.md
jakobbotsch: Sorry for the late response, I just only noticed the design document today. My response here is mainly about RyuJIT, and I realize that UX is a large factor too.

For RyuJIT, there are definitely many separate complexities around each of these. For UnmanagedCallersOnly, with SwiftReturn:
1. For return buffers, the JIT very early transforms these to regular stores. We would still need a separate representation internally to represent the return of the error.
2. For structs returned in registers the JIT can keep struct values in the more natural representation, but this is tied into multi-reg handling and promotion. Some platforms (e.g. win-x64) does not have multi-reg call returns today, so we would need to do the work to support this for the Swift case if we wanted to keep things enregistered. 

I think there are similar complexities in the other direction (when calling a function returning SwiftReturn).

For SwiftError* in the signature you run into another set of problems -- for one, it is not so natural that the parameter/argument differs in a level of indirection to the actual thing we are trying to represent. It also promotes a pattern to the user where they end up using the address of this local in ways that will cause inefficiencies in the codegen.

In the UnmanagedCallersOnly case the suggested implementation by @jkotas sounds reasonable to me. It requires us to special case this dummy local in a number of places because its uses do not appear explicitly in the IR, or otherwise to handle it conservatively (no enregistration). One unfortunate thing here is that it conflates error returns with normal returns -- it will be hard for the JIT to separate these.

Calling something with the SwiftError* is probably the most difficult of all of these cases to represent. If we want to handle this pattern efficiently (without stack spilling) then I do not think this can be represented like this very long within the JIT -- likely we are going to transform something like
```csharp
void* error;
var value = CallToSwiftFunction(x, y, &error);
```
into the RyuJIT IR equivalent of
```csharp
void* error;
var value = CallToSwiftFunction(x, y);
error = SwiftErrorRegister();
```
immediately during importation. Some work will be needed to make sure the CallToSwiftFunction/SwiftErrorRegister pairs do not get separated during JIT optimizations, but I do not think this will be hard. LSRA work will be needed to make sure that the error register is considered busy until it has been read after CallToSwiftFunction.

I might not understand what exactly option 4 entails here, but if it just means having the JIT generate a callout to a VM function when the error register is non-zero after a Swift call, then I think it would be by far the simplest to implement in RyuJIT.

I think option 3 is implementable in a way that does not leave any performance on the table, by considered the operations as intrinsics. I expect that this effectively amounts to directly exposing the mechanisms that RyuJIT will end up with under the hood to the user. I would probably switch "Set" to a direct "return" intrinsic instead, so that we would end up with intrinsics like:
```csharp
[Intrinsic]
void* SwiftErrorReturn() => throw null;
[Intrinsic]
T ReturnSwiftError<T>(void* error) => throw null;
[Intrinsic]
void ReturnSwiftError<T>(void* error) => throw null;
```

SwiftErrorReturn() would return the error returned by the last Swift call. in RyuJIT, it would be transformed directly into a use of the "swift error local" that is defined (by the transformation above) after every call to a Swift function.
The user visible behavior of ReturnSwiftError is an immediate return from the calling function with that error in the error register. A benefit of exposing it like this is that it models that no "normal" return value is necessary (though I am not sure whether or not that is actually how it works in Swift -- can you still look at the value when there was an error return?). Also, it means that we know which returns are error paths, and ensure that these paths to not introduce phantom uses of unrelated values. But the UX of this may not be ideal.

jkoritzinsky: Honestly, the shape you mention for the generated code when calling a method is fine with me. Clang uses the same representation.

In any case, we want to use the same representation for calling Swift via P/Invoke and function pointers and UnmanagedCallersOnly.

Our primary focus is on Apple platforms, so the concerns about win-x64 suggests that we should push Swift on Windows support out from v1 if it will have extra costs.

jakobbotsch: > In any case, we want to use the same representation for calling Swift via P/Invoke and function pointers and UnmanagedCallersOnly.

I would say that my main concern for the `SwiftError*` design is around returns in `UnmanagedCallersOnly` functions. In addition to being harder to implement than an intrinsic it promotes writing code that makes it hard for the JIT to separate normal and error returns. For example, returning an error from a function with a large struct return is likely going to incur cost that the same function implemented in Swift wouldn't incur. I am not sure whether it matters or not.

> Our primary focus is on Apple platforms, so the concerns about win-x64 suggests that we should push Swift on Windows support out from v1 if it will have extra costs.

I think with the `SwiftError*` design there shouldn't be much difference in handling between the platforms.

kotlarmilos: > can you still look at the value when there was an error return?

Yes, caller gets a return value and should check for an error immediately after the call. 

> For example, returning an error from a function with a large struct return is likely going to incur cost that the same function implemented in Swift wouldn't incur.

I believe it depends on how heavily `UnmanagedCallersOnly` is utilized by projection tools.

jakobbotsch: > Yes, caller gets a return value and should check for an error immediately after the call.

Is it possible in a Swift function to return both an error and a valid value? Is it possible for the caller to act on both pieces of information, even if there was an error?
For example, when a return buffer is used, I am curious whether the Swift compiler is clearing it out or not on return of an error.

kotlarmilos: Good point. I _think_ the Swift compiler is not clearing out the return buffer if an error is thrown. The caller could still access the return value even though the error register is set. Consider the following example from https://github.com/kotlarmilos/runtime/blob/7bcf61f381fb97edfdf493a4359af89cba8aa4f0/src/tests/Interop/Swift/ErrorHandling.swift#L7-L10.

```swift
public func someFuncThatMightThrow (dummy: UnsafeRawPointer) throws -> Int {
    throw MyError.runtimeError ("Catch me if you can!");
    return 42;
}
```

If the function throws and returns a value, C# caller can still access the return value.

jkotas: CLang/LLVM models swift error the same way as the currently proposed model (pretend that the value is passed via addressable memory): https://clang.llvm.org/docs/AttributeReference.html#swift-error-result , with additional twist that it must be the last parameter (shoud we do the same?). It suggests that the proposed model is a decent choice.

We are going to pay interop transition cost on each of the swift calls. Also, larger structs are typically going to be passed using opaque convention that is expensive compared to passing things in registers. It is ok if the current JIT is not able to 100% optimize some of cases of error results.

proposed/swift-interop.md
```diff
@@ -0,0 +1,189 @@
...
+
+When calling a function that returns an opaque struct, the Swift ABI always requires the struct to be passed by a return buffer. Since we cannot know the size at compile time, we'll need to handle this case manually in the calling convention signature as well by manually introduce the struct return buffer parameter. Alternatively, we could enlighten the JIT/AOT compilers about how to look up the correct buffer size for an opaque struct. This would likely be quite expensive, so we should avoid this option and just manually handle the return buffer. We can determine the return buffer size by using the value witness table for the struct type. We could always move to the JIT-intrinsic model in the future if we desired.
```
jakobbotsch, line 121: The return buffer in the Swift calling convention is passed in a special register, so we will need some way to establish in the JIT which argument is meant to be a return buffer pointer if it appears as a normal parameter in the signature.

proposed/swift-interop.md
jakobbotsch: > Good point. How can we verify that? Probably we can try to set the return register before the call and inspect if it has changed after the call.

I verified it in https://godbolt.org/z/58srMbfK6. `throws_does_not_clear` does not write to the return buffer pointed to by `rax`.

> CLang/LLVM models swift error the same way as the currently proposed model (pretend that the value is passed via addressable memory): https://clang.llvm.org/docs/AttributeReference.html#swift-error-result , with additional twist that it must be the last parameter (shoud we do the same?). It suggests that the proposed model is a decent choice.

As far as I can read in https://github.com/apple/swift/blob/main/docs/ABI/CallingConvention.rst, the Swift compiler lowers Swift signatures into simpler signatures before they are presented to LLVM. It means the Swift compiler can just avoid creating any IR that writes to the return buffer pointer when it sees `throw` expression. I do not think this is an optimization done at the LLVM level.

jakobbotsch: With that said, the design is fine and looks implementable to me, just wanted to point out this difference between `throw` and using a pseudo local.

I am curious whether we are going to do the same kind of lowering as the Swift compiler appears to be doing for signatures. In particular the frozen struct/tuple lowering looks very complex and not described with a lot of detail in https://github.com/apple/swift/blob/main/docs/ABI/CallingConvention.rst. Will the JIT be expected to handle the lowering or can we rely on seeing simplified signatures for swift calls/`UnmanagedCallersOnly`?

kotlarmilos: > CLang/LLVM models swift error the same way as the currently proposed model (pretend that the value is passed via addressable memory): https://clang.llvm.org/docs/AttributeReference.html#swift-error-result , with additional twist that it must be the last parameter (shoud we do the same?)

When calling from C# into Swift, if `error` or `self` are the last parameters, there is no need to handle additional  arguments at the runtime level. These arguments will be automatically ignored by the Swift callee function.

PR #303, opened 2023-11-08
Supporting local SDK deployment in global.json

jaredpar:
This proposal extends global.json such that it supports locally deployed instances of the .NET SDK. This will significantly reduce friction our local deployment has with developer tools like Visual Studio, VS Code and make our own developer story much simpler.


xoofx:
Thanks for writing this proposal! Looks like it's covering our use case for using a custom dotnet SDK with Unity + MSBuild.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Motivation
+There is currently a disconnect between the ways the .NET SDK is deployed in practice and what the host resolver can discover when searching for compatible SDKs. By default the host resolver is only going to search for SDKS in machine wide locations: `C:\Program Files\dotnet`, `%PATH%`, etc ...  The .NET SDK though is commonly deployed to local locations: `%LocalAppData%\Microsoft\dotnet`, `$HOME/.dotnet`. Many repos embrace this and restore the correct .NET for their builds into a local `.dotnet` directory.
+
+The behavior of the host resolver is incompatible with local based deployments. It simply will not find these deployments and instead search in machine wide locations. That means tools like Visual Studio and VS Code simply do not work with local deployment by default. Developers must take additional steps like manipulating `%PATH%` before launching these editors. That reduces the usefulness of items like the quick launch bar, short cuts, etc ...
```
rainersigwald, line 22:
```suggestion
The behavior of the host resolver is incompatible with local based deployments. It will not find these deployments without additional environment variable configuration and instead search in machine wide locations. That means tools like Visual Studio and VS Code simply do not work with local deployment by default. Developers must take additional steps like manipulating `%PATH%` before launching these editors. That reduces the usefulness of items like the quick launch bar, short cuts, etc ...
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+
+- Machine wide
+- User wide: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
+- Repo: `.dotnet`
+
+Our installation tooling tends to avoid redundant installations. For example if restoring a repository that requires 7.0.400, the tooling will not install it locally if 7.0.400 is installed machine wide. It also will not necessarily delete the local `.dotnet` folder or the user wide folder. That means developers end up with .NET SDK installs in all three locations but only the machine wide install has the correct ones.
+
+As a result solutions like "just use .dotnet if it exists" fall short. It will work in a lot of casse but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
+
+### Do we need additionalLocationsOnly?
```
sandyarmstrong, line 80:
```suggestion
### Do we need additionalPathsOnly?
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+
+In this configuration the host resolver would find a compatible SDK if it exists in `.dotnet` or a machine wide location.
+
+The values in the `additionalLocations` property can be a relative or absolute path. When a relative path is used it will be resolved relative to the location of global.json. These values also support the following substitutions:
+
+- `"$local$"`: this matches the local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
+- `"$machine$"`: this matches the machine installation point of .NET for the current operating system.
+- `%VARIABLE%/$VARIABLE`: environment variables will be substituted. Either the Windows or Unix format can be used here and will be normalized for the operating system the host resolver executes on.
+
+The host resolver will consider the `additionalLocations` in the order they are defined and will stop at the first match. If none of the locations have a matching SDK then it it will fall back to considering machine wide installations (unless `additionalLocationsOnly` is `true`).
```
sandyarmstrong, line 64: Maybe mention this ordering earlier. The first thing I wanted to know when reading this was whether `additionalLocations` would take priority over default locations.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+
+As a result solutions like "just use .dotnet if it exists" fall short. It will work in a lot of casse but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
+
+### Do we need additionalLocationsOnly?
+The necessity of this property is questionable. The design includes it for completeness and understanding that the goal of some developers is complete isolation from machine state. As long as we're considering designs that embrace local deployment, it seemed sensible to extend the design to embrace _only_ local deployment.
+
+At the same time the motivation for this is much smaller. It would be reasonable to cut this from the design and consider it at a future time when the motivation is higher.
```
sandyarmstrong, line 83: I believe we can do without it, but at the same time, being able to set this makes silly onboarding mistakes much easier to catch.

sandyarmstrong: When .NET workloads are involved, I definitely don't want my teammates accidentally polluting their "global" SDK locations with workloads they don't need. It's nicer to keep those isolated.

But still, I agree, not a requirement.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+}
+```
+
+In this configuration the host resolver would find a compatible SDK if it exists in `.dotnet` or a machine wide location.
+
+The values in the `additionalLocations` property can be a relative or absolute path. When a relative path is used it will be resolved relative to the location of global.json. These values also support the following substitutions:
+
+- `"$local$"`: this matches the local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
+- `"$machine$"`: this matches the machine installation point of .NET for the current operating system.
+- `%VARIABLE%/$VARIABLE`: environment variables will be substituted. Either the Windows or Unix format can be used here and will be normalized for the operating system the host resolver executes on.
```
baronfel, line 62: What's motivating needing to parse a DSL in these tokens? Do some of the usages you linked to above use these idioms? Is there any reason why we couldn't start only be using relative path-based mechanisms?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+
+This is not a problem specific to the .NET team. Indeed this problem is felt sharply there due to arcade infrastructure using xcopy style deployment into `.dotnet`. Our customers feel this problem as well. Consider the following examples:
+
+- This [issue](https://github.com/dotnet/sdk/issues/8254) from 2017 attempting to solve this problsem. It gets several hits a year from customers who are similarly struggling with our toolings inability to handle local deployment.
+- This [internal discussion](https://teams.microsoft.com/l/message/19:ed7a508bf00c4b088a7760359f0d0308@thread.skype/1698341652961?tenantId=72f988bf-86f1-41af-91ab-2d7cd011db47&groupId=4ba7372f-2799-4677-89f0-7a1aaea3706c&parentMessageId=1698341652961&teamName=.NET%20Developer%20Experience&channelName=InfraSwat&createdTime=1698341652961) from a C# team member. They wanted to use VS as the product is shipped to customers and got blocked when we shipped an SDK that didn't have a corresponding MSI and hence VS couldn't load Roslyn anymore.
+- [VS code](https://github.com/dotnet/vscode-csharp/issues/6471) having to adjust to consider local directories for SDK because our resolver can't find them.
```
rainersigwald, line 38: These two are internal .NET teams, are they not? "Customer impact" is still the sticking point for me on this.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+        "additionalPaths": [ ".dotnet" ],
+    }
+}
+```
+
+In this configuration the host resolver would find a compatible SDK if it exists in `.dotnet` or a machine wide location.
+
+The values in the `additionalLocations` property can be a relative or absolute path. When a relative path is used it will be resolved relative to the location of global.json. These values also support the following substitutions:
+
+- `"$local$"`: this matches the local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
```
rainersigwald, line 60:
```suggestion
- `"$user$"`: this matches the user-local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
```

?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+- Machine wide
+- User wide: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
+- Repo: `.dotnet`
+
+Our installation tooling tends to avoid redundant installations. For example if restoring a repository that requires 7.0.400, the tooling will not install it locally if 7.0.400 is installed machine wide. It also will not necessarily delete the local `.dotnet` folder or the user wide folder. That means developers end up with .NET SDK installs in all three locations but only the machine wide install has the correct ones.
+
+As a result solutions like "just use .dotnet if it exists" fall short. It will work in a lot of casse but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
+
+### Do we need additionalLocationsOnly?
+The necessity of this property is questionable. The design includes it for completeness and understanding that the goal of some developers is complete isolation from machine state. As long as we're considering designs that embrace local deployment, it seemed sensible to extend the design to embrace _only_ local deployment.
```
marcpopMSFT, line 81: There are things that can change a particular SDK state (namely workload installs). So in the event a user has the same SDK in both the additional paths and globally, this may be needed for that type of situation but it's also likely fairly niche.

proposed/local-sdk-global-json.md
marcpopMSFT, line 64: Will it prioritize the PATH first before any of the additional locations? If rollforward is enabled in global.json (or via environment), would it pick the first version that matched even if it wasn't the best match? Example: if the global.json has a 7.0 SDK specified and rollforward is Major and it first finds an 8.0, should it stop or keep looking for a closer match?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+
+- `"additionalLocations"`: this is a list of paths that the host resolver should consider when looking for compatible SDKs. Relative paths will be interpreted relative to the global.json file.
+- `"additionalLocationOnly"`: when true the resolver will _only_ consider paths in `additionalLocations`. It will not consider any machine wide locations (unless they are specified in `additionalLocations`). The default for this property is `false`.
+
+The `additionalLocations` works similar to how multi-level lookup works. It adds additional locations that the host resolver should consider when trying to resolve a compatible .NET SDK. For example:
+
+```json
+{
+    "sdk": {
+        "additionalPaths": [ ".dotnet" ],
```
marcpopMSFT, line 51: Likely it's marginal cost to support more than one path but I'm curious if there are cases where it'd be useful to have 2+ paths here.

rainersigwald (commented):
Are you proposing a change to the failure case? For instance "I need `9.0.100-preview.2` exactly and it must be in `.dotnet` but the user hasn't run the script to make that happen".

proposed/local-sdk-global-json.md
jaredpar: When I was looking through the docs and how `dotnet-install.sh/ps1` works it seemed like a valid scenario would be `additionalPaths: [ ".dotnet", "$local$"]`. Those are the two most common cases where the SDK is installed locally and seemed reasonable to support searching them both.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+- [VS code](https://github.com/dotnet/vscode-csharp/issues/6471) having to adjust to consider local directories for SDK because our resolver can't find them.
+
+## Detailed Design
+The global.json file will support two new properties under the `sdk` object:
+
+- `"additionalLocations"`: this is a list of paths that the host resolver should consider when looking for compatible SDKs. Relative paths will be interpreted relative to the global.json file.
+- `"additionalLocationOnly"`: when true the resolver will _only_ consider paths in `additionalLocations`. It will not consider any machine wide locations (unless they are specified in `additionalLocations`). The default for this property is `false`.
```
sharwell, line 44: 📝 All references in text should match the examples

proposed/local-sdk-global-json.md
jaredpar: I was viewing it more as people just trying to use the product and running into local deployment issues. It's not strictly related to their positions on the .NET team. Can be frustrating when you're trying to use the prooduct as a customer but get labeled as "it's part of .NET so think about it differently". Was trying to thread that needle in my wording here. 

I can rephrase this as customre impact though. It's a better term.

rainersigwald: My position remains that if you're using a global.json with no rollforward you're not just "trying to use the product". Someone opted you into specific pain to lock down behavior. Is that justified? Maybe! But it's not required for everyone.

proposed/local-sdk-global-json.md
jaredpar: You think in a single doc I could keep terminology straight in my head ... guess not.

proposed/local-sdk-global-json.md
jaredpar: I like that much better.

proposed/local-sdk-global-json.md
jaredpar: The motivating scenarios I had in mind don't require environment variable support. At the same time this proposal is about closing gaps in the host resolver for how we build, deploy and develop. Environment variables are very ingrained into how we build today. It's a key piece of data currency in CI scenarios. It seemed very likely that once we support local paths, then CI scenarios would popup to say "this would be even better with environment variables support too". That was my thought in adding it now: basically it's inevitable. 

If we wanted to push this into if we get more data section I wouldn't be too opposed to that.

proposed/local-sdk-global-json.md
jaredpar: The order would be `additionalPaths` then how the resolver works today. That is `%PATH%` followed by machine wide installs. I will make that clearer in the doc. 

The intent of the design is that the resolver will stop when it finds the first compatible SDK. For example if `.dotnet` had a compatible SDK but machine wide had a _better_ one then it would pick `.dotnet` because it searched that first. Will make this clearer as well.

proposed/local-sdk-global-json.md
jaredpar: > My position remains that if you're using a global.json with no rollforward you're not just "trying to use the product".

I don't agree with this point. There are plenty of customers out there who desire a very locked down build environment. As compatible as the .NET SDK is, upgrading can cause problems, particularly with complex projects. Also for people who are eagerly dogfooding preview versions of the SDK, disabling roll forward seems helpful if they're playing with different previews in different projects. These all seem like valid customer scenarios. 

At the same time, a large part of the motivation for disabling `rollForward` would actually go away with this feature. Generally this is done for CI scenarios where we want to avoid machine state making CI unpredictable. Once we have `additionalPaths` that problem kinda goes away because we know that will be matched first. Machine state doesn't matter as much.

rainersigwald: > These all seem like valid customer scenarios.

I think they are, but I don't think the evidence you've presented is convincing that they're common.

> At the same time, a large part of the motivation for disabling `rollForward` would actually go away with this feature. Generally this is done for CI scenarios where we want to avoid machine state making CI unpredictable.

This is surprising to me, since I thought it was the inverse: CI is fine and VS is where this feature would benefit. Why not solve this with "use whatever magic you used to create the `.dotnet` folder to execute the build"?

proposed/local-sdk-global-json.md
jaredpar: Ha ... no one caught my "it it" usage. You all are slacking in your reviews!

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,83 @@
...
+        "additionalPaths": [ ".dotnet" ],
+    }
+}
+```
+
+In this configuration the host resolver would find a compatible SDK if it exists in `.dotnet` or a machine wide location.
+
+The values in the `additionalLocations` property can be a relative or absolute path. When a relative path is used it will be resolved relative to the location of global.json. These values also support the following substitutions:
+
+- `"$user$"`: this matches the user-local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
```
RussKie, line 60:
```suggestion
- `"$user$"`: this matches the user-local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet` on Windows and `$HOME/.dotnet` on Linux/macOS.
```

baronfel:
Just a note that if we take this we should update the official schema for global.json files that's over in [SchemaStore](https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/global.json) to declare the new properties (and link to docs for them).

proposed/local-sdk-global-json.md
jaredpar: Okay tried to clean that up a bit.

proposed/local-sdk-global-json.md
jaredpar: Moved ordering earlier and added some examples.

jaredpar:
> Are you proposing a change to the failure case? For instance "I need 9.0.100-preview.2 exactly and it must be in .dotnet but the user hasn't run the script to make that happen".

This proposal is narrowly focused on changing the host resolver algorithm only. This means it should be effectively a transparent change to the rest of the tooling stack. If the correct SDK was not found it would emit an error just as it does today. It would not attempt anything like auto-acquiring the correct SDK.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,105 @@
...
+- `%VARIABLE%/$VARIABLE`: environment variables will be substituted. Either the Windows or Unix format can be used here and will be normalized for the operating system the host resolver executes on.
+
+This design requires us to only change the host resolver. That means other tooling like Visual Studio, VS Code, MSBuild, etc ... would transparently benefit from this change. Repositories could update global.json to have `additionalPaths` support `.dotnet` and Visual Studio would automatically find it without any design changes.
+
+## Considerations
+### Installation Points
+One item to keep in mind when considering this area is the .NET SDK can be installed in many locations. The most common are:
+
+- Machine wide
+- User wide: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
```
RussKie, line 85:
```suggestion
- User wide: `%LocalAppData%\Microsoft\dotnet` on Windows and `$HOME/.dotnet` on Linux/macOS.
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,105 @@
...
+- Repo: `.dotnet`
+
+Our installation tooling tends to avoid redundant installations. For example if restoring a repository that requires 7.0.400, the tooling will not install it locally if 7.0.400 is installed machine wide. It also will not necessarily delete the local `.dotnet` folder or the user wide folder. That means developers end up with .NET SDK installs in all three locations but only the machine wide install has the correct ones.
+
+As a result solutions like "just use .dotnet if it exists" fall short. It will work in a lot of casse but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
```
RussKie, line 90:
```suggestion
As a result solutions like "just use .dotnet, if it exists" fall short. It will work in a lot of casse but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,105 @@
...
+
+```json
+{
+    "sdk": {
+        "additionalPaths": [ ".dotnet" ],
+    }
+}
+```
+
+In this configuration the host resolver would find a compatible SDK if it exists in `.dotnet` or a machine wide location. The host resolver will consider the `additionalPaths` in the order they are defined and will stop at the first match. If none of the locations have a matching SDK then
```
RussKie, line 55:
```suggestion
In this configuration the host resolver would find a compatible SDK, if it exists in `.dotnet` or a machine wide location. The host resolver will consider the `additionalPaths` in the order they are defined and will stop at the first match. If none of the locations have a matching SDK then
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,105 @@
...
+- [VS code](https://github.com/dotnet/vscode-csharp/issues/6471) having to adjust to consider local directories for SDK because our resolver can't find them.
+
+## Detailed Design
+The global.json file will support two new properties under the `sdk` object:
+
+- `"additionalPaths"`: this is a list of paths that the host resolver should consider when looking for compatible SDKs. Relative paths will be interpreted relative to the global.json file.
+- `"additionalPathsOnly"`: when true the resolver will _only_ consider paths in `additionalPaths`. It will not consider any machine wide locations (unless they are specified in `additionalPaths`). The default for this property is `false`.
```
RussKie, line 43:
```suggestion
- `"additionalPathsOnly"`: when `true` the resolver will _only_ consider paths in `additionalPaths`. It will not consider any machine wide locations (unless they are specified in `additionalPaths`). The default for this property is `false`.
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Motivation
+There is currently a disconnect between the ways the .NET SDK is deployed in practice and what the host resolver can discover when searching for compatible SDKs. By default the host resolver is only going to search for SDKS in machine wide locations: `C:\Program Files\dotnet`, `%PATH%`, etc ...  The .NET SDK though is commonly deployed to local locations: `%LocalAppData%\Microsoft\dotnet`, `$HOME/.dotnet`. Many repos embrace this and restore the correct .NET for their builds into a local `.dotnet` directory.
+
+The behavior of the host resolver is incompatible with local based deployments. It will not find these deployments without additional environment variable configuration and instead search in machine wide locations. That means tools like Visual Studio and VS Code simply do not work with local deployment by default. Developers must take additional steps like manipulating `%PATH%` before launching these editors. That reduces the usefulness of items like the quick launch bar, short cuts, etc ...
```
RussKie, line 21:
```suggestion
The behavior of the host resolver is incompatible with local based deployments. It will not find these deployments without additional environment variable configuration and instead search in machine wide locations. That means tools like Visual Studio and VS Code simply do not work with local deployment by default. Developers must take additional steps like manipulating `%PATH%` before launching these editors. That reduces the usefulness of items like the quick launch bar, short cuts, etc.
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Considerations
+### Installation Points
+One item to keep in mind when considering this area is the .NET SDK can be installed in many locations. The most common are:
+
+- Machine wide
+- User wide: `%LocalAppData%\Microsoft\dotnet`` on Windows and `$HOME/.dotnet`` on Linux/macOS.
+- Repo: `.dotnet`
+
+Our installation tooling tends to avoid redundant installations. For example if restoring a repository that requires 7.0.400, the tooling will not install it locally if 7.0.400 is installed machine wide. It also will not necessarily delete the local `.dotnet` folder or the user wide folder. That means developers end up with .NET SDK installs in all three locations but only the machine wide install has the correct ones.
```
RussKie, line 88:
```suggestion
Our installation tooling tends to avoid redundant installations. For example, if restoring a repository that requires 7.0.400, the tooling will not install it locally if 7.0.400 is installed machine wide. It also will not necessarily delete the local `.dotnet` folder or the user wide folder. That means developers end up with .NET SDK installs in all three locations but only the machine wide install has the correct ones.
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,105 @@
...
+At the same time the motivation for this is much smaller. It would be reasonable to cut this from the design and consider it at a future time when the motivation is higher.
+
+### Best match or first match?
+This proposal is designed at giving global.json more control over how SDKs are found. If the global.json asked for a specific path to be considered and it has a matching SDK but a different SDK was chosen, that seems counter intuitive. Even in the case where the chosen SDK was _better_. This is a motivating scenario for CI where certainty around SDK is often more desirable than _better_. This is why the host discovery stops at first match vs. looking at all location and choosing the best match.
+
+Best match is a valid approach though. Can certainly see the argument for some customers wanting that. Feel like it cuts against the proposal a bit because it devalues `additionalPaths` a bit. If the resolver is switched to best match then feel like the need for `additionalPathsOnly` is much stronger. There would certainly be a customer segment that wanted to isolate from machine state in that case.
+
+
+
+The host resolver search stops at the first matching SDK. This proposal
```
RussKie, line 104: Ummm, what? :)

RussKie (commented):
Few nits

proposed/local-sdk-global-json.md
jaredpar: That is completely clear? What's the confusion ... 

;)

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+```json
+{
+    "sdk": {
+        "additionalPaths": [ ".dotnet" ],
+    }
+}
+```
+
+In this configuration the host resolver would find a compatible SDK, if it exists in `.dotnet` or a machine wide location. The host resolver will consider the `additionalPaths` in the order they are defined and will stop at the first match. If none of the locations have a matching SDK then
+ will fall back to the existing SDK resolution strategy: `%PATH%` followed by machine wide installations.
```
333fred, line 56:
```suggestion
the host resolver will fall back to the existing SDK resolution strategy: `%PATH%` followed by machine wide installations.
```

proposed/local-sdk-global-json.md
RussKie: Do I understand correctly this in effect equals to `DOTNET_MULTILEVEL_LOOKUP=0`?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+}
+```
+
+In a scenario where the `.dotnet` directory had 7.0.200 SDK but there was a machine wide install of 7.0.300 SDK, the host resolver would pick 7.0.200 out of `.dotnet`. That location is considered first, it has a matching SDK and hence discovery stops there.
+
+The values in the `additionalPaths` property can be a relative or absolute path. When a relative path is used it will be resolved relative to the location of global.json. These values also support the following substitutions:
+
+- `"$user$"`: this matches the user-local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet` on Windows and `$HOME/.dotnet` on Linux/macOS.
+- `"$machine$"`: this matches the machine installation point of .NET for the current operating system.
+- `%VARIABLE%/$VARIABLE`: environment variables will be substituted. Either the Windows or Unix format can be used here and will be normalized for the operating system the host resolver executes on.
```
rolfbjarne, line 76: What happens if the environment variable contains multiple paths (such as PATH)?

Would it support the `${VARIABLE}` syntax as well?

What happens if the path / environment variable doesn't exist? (One scenario would be if someone wanted to use an environment variable that's different between platforms - HOME vs USERPROFILE for instance).

Would either path separator (`\`, `/`) work on all platforms?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+### Installation Points
+One item to keep in mind when considering this area is the .NET SDK can be installed in many locations. The most common are:
+
+- Machine wide
+- User wide: `%LocalAppData%\Microsoft\dotnet` on Windows and `$HOME/.dotnet` on Linux/macOS.
+- Repo: `.dotnet`
+
+Our installation tooling tends to avoid redundant installations. For example, if restoring a repository that requires 7.0.400, the tooling will not install it locally if 7.0.400 is installed machine wide. It also will not necessarily delete the local `.dotnet` folder or the user wide folder. That means developers end up with .NET SDK installs in all three locations but only the machine wide install has the correct ones.
+
+As a result solutions like "just use .dotnet, if it exists" fall short. It will work in a lot of casse but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
```
rolfbjarne, line 90:
```suggestion
As a result solutions like "just use .dotnet, if it exists" fall short. It will work in a lot of cases but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Do we need additionalPathsOnly?
+The necessity of this property is questionable. The design includes it for completeness and understanding that the goal of some developers is complete isolation from machine state. As long as we're considering designs that embrace local deployment, it seemed sensible to extend the design to embrace _only_ local deployment.
+
+At the same time the motivation for this is much smaller. It would be reasonable to cut this from the design and consider it at a future time when the motivation is higher.
```
rolfbjarne, line 95: I would very much like this, it's not fun when random machine state changes behavior.

proposed/local-sdk-global-json.md
akoeplinger: My first thought was "does this support env vars?" so yeah I'll agree that we'd regret it pretty soon if we don't support them right away.

proposed/local-sdk-global-json.md
vitek-karas: Overall I feel that supporting env. variable expansion is unnecessary. What would be the typical use case for it?
I also agree with basically everything Rolf points out - especially the fact that the env. variables are unlikely to match between platforms, and as such it will be effectively impossible to use them for anything in a repo, only useful for local dev work where the dev authors `global.json` themselves. Not sure how common that would be, and it's a case where asking the developer to use other env. variables (`DOTNET_ROOT`, `PATH`) doesn't seem too bad.

proposed/local-sdk-global-json.md
vitek-karas: That is my understanding as well. With the difference that now it would work the same across all platforms (the env. variable only ever worked on windows, and now it's off by default anyway).

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+
+- Machine wide
+- User wide: `%LocalAppData%\Microsoft\dotnet` on Windows and `$HOME/.dotnet` on Linux/macOS.
+- Repo: `.dotnet`
+
+Our installation tooling tends to avoid redundant installations. For example, if restoring a repository that requires 7.0.400, the tooling will not install it locally if 7.0.400 is installed machine wide. It also will not necessarily delete the local `.dotnet` folder or the user wide folder. That means developers end up with .NET SDK installs in all three locations but only the machine wide install has the correct ones.
+
+As a result solutions like "just use .dotnet, if it exists" fall short. It will work in a lot of casse but will fail in more complex scenarios. To completely close the disconnect here we need to consider all the possible locations.
+
+### Do we need additionalPathsOnly?
```
vitek-karas, line 92: Did you consider that having the `$machine$` means that:

```json
"additionalPaths": [ ".dotnet", "$machine$" ],
"additionalPathsOnly": "true"
```

is equivalent to
```json
"additionalPaths": [ ".dotnet" ],
"additionalPathsOnly": "false"
```

Would it make sense to only use `additionalPaths` - and maybe rename it to something like `installPaths`. And if specified it would the only list searched?

agocke:
Confirming: this is an SDK-only feature, correct? That is, the behavior of the app host or of `dotnet blah.dll` would not be affected? That appears to be true, as written. However, I would change the "host resolver" to simply "SDK resolver" to remove confusion.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+
+At the same time the motivation for this is much smaller. It would be reasonable to cut this from the design and consider it at a future time when the motivation is higher.
+
+### Best match or first match?
+This proposal is designed at giving global.json more control over how SDKs are found. If the global.json asked for a specific path to be considered and it has a matching SDK but a different SDK was chosen, that seems counter intuitive. Even in the case where the chosen SDK was _better_. This is a motivating scenario for CI where certainty around SDK is often more desirable than _better_. This is why the host discovery stops at first match vs. looking at all location and choosing the best match.
+
+Best match is a valid approach though. Can certainly see the argument for some customers wanting that. Feel like it cuts against the proposal a bit because it devalues `additionalPaths` a bit. If the resolver is switched to best match then feel like the need for `additionalPathsOnly` is much stronger. There would certainly be a customer segment that wanted to isolate from machine state in that case.
+
```
vitek-karas, line 101: Additional things worth mentioning:
* This will not affect the runtime resolution when executing `dotnet exec app.dll`
  * Maybe not a big deal - depends on the use cases - potentially we could introduce something like `dotnet exec --use-sdk-locations app.dll` or something like that to ask runtime resolution to also consider global.json

vitek-karas: * Compatibility considerations - this relies on the `dotnet.exe` which is on the user's path to be able to run the SDK specified in the `global.json`. This is not a big concern for the "Fallback to machine wide" since that will typically be the `dotnet.exe` used, but it is a concern for repo-local installs. This can lead to things like:
  * 8.0 `dotnet.exe` is used to run `9.0.0-preview...` SDKs (so we have to be really strict about forward compatibility)
  * 8.0 `dotnet.exe` is used to run `3.1.0` SDK (so we have to be really strict about backward compatibility)
None of this is exactly new, but specifically the forward compatibility doesn't have a common use case currently (it can happen, but it's very rare).

proposed/local-sdk-global-json.md
vitek-karas, line 56: I think this will need more detailed specification. Currently the SDK resolution has two "modes":
* Multi-level lookup disabled (the default for .NET 8) - in this case only the location of the `dotnet.exe` used to invoke the command.
* Multi-level lookup enabled (opt-in via `DOTNET_MULTILEVEL_LOOKUP=1`, which is considered "obsolete" really) - in this case the location of the `dotnet.exe` is considered first, followed by the global location

Note that the resolver doesn't look at `PATH`. Frequently the `dotnet.exe` will come from `PATH`, but if I run `C:\myfolder\dotnet.exe build` then `PATH` is actually completely ignored.

We need to define what fallback is considered - if it means the location of the `dotnet.exe` used to invoke the command, or the actual "machine-wide" install, or both.

sharwell:
We (Roslyn team) are desperately in need of this fix.

proposed/local-sdk-global-json.md
agocke: The forward compatibility is somewhat concerning to me. Do we have a list of API dependencies that the muxer needs? And a strategy to ensure that those APIs don't change in the future?

vitek-karas: I could get you the list - it's VERY short (I think 2 APIs or maybe even just 1).
We are very aware of this requirement and we never touched those APIs. So I'm not super concern about this, but it does raise the importance of keeping it compatible since so far all the cases where it can happen are pretty rare, this would be basically the first real feature which makes use of it.

proposed/local-sdk-global-json.md
mmitche: Agreed. I generally think the design already generally feels like it's encoding local dev machine setup into the global.json. While not ideal, it's a decent compromise since a lot of "common" practices would put an SDK locally into the repo layout at predictable locations. Adding environment variable expansion on top of that seems unnecessary. That said, why not add the flexibility? What's the harm?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+### Do we need additionalPathsOnly?
+The necessity of this property is questionable. The design includes it for completeness and understanding that the goal of some developers is complete isolation from machine state. As long as we're considering designs that embrace local deployment, it seemed sensible to extend the design to embrace _only_ local deployment.
+
+At the same time the motivation for this is much smaller. It would be reasonable to cut this from the design and consider it at a future time when the motivation is higher.
+
+### Best match or first match?
+This proposal is designed at giving global.json more control over how SDKs are found. If the global.json asked for a specific path to be considered and it has a matching SDK but a different SDK was chosen, that seems counter intuitive. Even in the case where the chosen SDK was _better_. This is a motivating scenario for CI where certainty around SDK is often more desirable than _better_. This is why the host discovery stops at first match vs. looking at all location and choosing the best match.
```
mmitche, line 98: On the other hand, a repo workflow that installs the desired SDK at `.dotnet` unless it is available globally already could end up with a `.dotnet` folder locally while actually wanting to use the global version. This means confusing behavior if first match is taken, unless the repo is cleaned.

Although, I suppose in this case, the global.json should better be encoding the needs of the repo, so that the first match is the global install...

I guess all this points to is that the behavior is bound to be a little confusing, and ways to help the user out are necesasry

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+    "rollForward": "latestFeature"
+  }
+}
+```
+
+In a scenario where the `.dotnet` directory had 7.0.200 SDK but there was a machine wide install of 7.0.300 SDK, the host resolver would pick 7.0.200 out of `.dotnet`. That location is considered first, it has a matching SDK and hence discovery stops there.
+
+The values in the `additionalPaths` property can be a relative or absolute path. When a relative path is used it will be resolved relative to the location of global.json. These values also support the following substitutions:
+
+- `"$user$"`: this matches the user-local installation point of .NET SDK for the current operating system: `%LocalAppData%\Microsoft\dotnet` on Windows and `$HOME/.dotnet` on Linux/macOS.
```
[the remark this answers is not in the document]
vitek-karas: Honestly don't know, could only guess. I would start with the fact that I didn't know where the "user location" was until this doc came about. And also, that `dotnet-install` script wasn't that popular back in the day all of this was designed. Nowadays we seem to have the opposite problem (too many people use the script for things it was not meant to do).

proposed/local-sdk-global-json.md
vitek-karas: The harm is complexity of implementation - in a component which is not versioned in the normal sense (we always use the latest version on the machine), thus requiring effectively 100% backward compatibility all the time (or close to it). Additionally, this would require some additional string parsing logic (like ability escape the special characters, how to deal with errors and so on). Basically, introducing another parser, which is always a concern. Personally I think that if we can make this useful without env. var support, I think we should do that.

There's also the potential unintended side-effects:
* "My SDK doesn't work in one window, but works in the other"
* "What's your global.json"
* <file with env. vars>
* Oh, that's why :-(
Basically reintroducing parts of the problem which we have today with PATH and DOTNET_ROOT.

proposed/local-sdk-global-json.md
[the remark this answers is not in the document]
jkotas: > Why aren't user locations included in the default behavior of the resolver already?

I think it would be a security hole. In general, it is not ok for programs installed in admin managed locations to load components from user-controlled location.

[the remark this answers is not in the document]
mmitche: Good point. Does that potentially points to a hole in this design? If I added `$machine`, `$user` and `.dotnet` to my additional paths, could the machine wide host end up loading a user-installed SDK?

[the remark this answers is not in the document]
jkotas: If you opt-in into unsecure configuration, it is your problem. You can do that via DOTNET_ROOT today. I think it is important that the defaults are secure.

proposed/local-sdk-global-json.md
jaredpar: > I don't like this property as much. If the goal of some developers would be isolation from the machine state, then why not just set DOTNET_ROOT after a repo-specific install?

How can tooling like say Visual Studio pick up such an environment variable? Yes you can do tricks like set the variable then start VS but that impacts the entire dev workflow. Actions like using the quick  launch menu don't work anymore cause VS can't process your repo without extra information. Any external process that launches VS (like say a debugger attach event) don't work in the repo cause they don't set the environment variable. It's an endless series of gotchas.

proposed/local-sdk-global-json.md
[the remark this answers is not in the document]
jaredpar: > In general, it is not ok for programs installed in admin managed locations to load components from user-controlled location.

MSBuild does this today via tasks, analyzers, generators, etc ... Is that _okay_ because the user has effectively opted into it by specifying references to the NuPkg that brought down the binaries?

proposed/local-sdk-global-json.md
jaredpar: That's a good argument. Leaning towards pulling env var support out.

proposed/local-sdk-global-json.md
jaredpar: I did think about that and I agree it's a valid approach to this. I've frankly waffled on which I think is the better approach here. 

I think in most scenarios customers would want to consider machine based installations. No data here, just intuition. The `$machine$` value is a bit of a _magic_ value. In the `installPaths` approach that would mean that `$machine$` would end up in the majority of uses of `installPaths`. Had a hard time convincing myself that was the best outcome which is why I ended up with the current approach.

rolfbjarne: I'm not advocating for either way, but having `$machine$` is slightly more flexible, because you can switch the order:

```json
"additionalPaths": [ "$machine$", ".dotnet" ],
```

Not sure why anyone would want to do that, but I'm sure someone will come up with a use-case at some point.

baronfel:
One thing to consider: how does this impact CI/CD setups like GitHub Actions' actions/setup-dotnet? I'd hope that those actions could learn to check for this new field and install the on-demand SDK to that location?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+- `"$machine$"`: this matches the machine installation point of .NET for the current operating system.
```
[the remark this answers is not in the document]
jkotas: > Linux systems with the source-built packages installed since those have made different decisions about where to install to (and they set DOTNET_ROOT so that the rest of the system is informed about this change).

The distro provided source-built packages should not be setting DOTNET_ROOT env variable. They should be writing the system wide dotnet location under `/etc/dotnet`.

If there are any distro provided source-built packages that set DOTNET_ROOT env variables, we should get them to fix that and write under `/etc/dotnet` instead.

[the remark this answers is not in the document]
baronfel: I'll set up a new install on Ubuntu today and verify if this is still the case. If not then I think there's a clear semantic: on Linux/MacOS, `$machine$` should be the location written to `/etc/dotnet`. Sound good?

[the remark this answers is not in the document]
vitek-karas: The `$machine` should be equivalent to running the algorithm described in https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md (and in the previous https://github.com/dotnet/designs/blob/main/accepted/2020/install-locations.md). The algorithm is a bit more complicated than just looking into `/etc/dotnet` although it's true that on most systems that's what it should end up doing.

RussKie:
Something has been bothering me about the proposal, and I think I finally figured out what that was. The developer experience, or the lack of, when the required SDK is absent.

In my opinion there are two gaps in the current functionality:
1. The ability to specify _where_ the developer wants the tooling (such as VS) to look for the SDK - and this proposal is tacking this.
2. The ability to inform a contributor what to do when the SDK is missing. 

If we consider the typical use case - a contributor clones a repo, which declares in global.json what version of the SDK is required and where it is supposed to be located, double clicks on a solution file, and.... nothing works. VS tells the contributor that it can't resolves types, find assemblies and build the solution. This is a very poor developer experience.
Building the command line is somewhat better - it'd tell that the required SDK is nowhere to be found, lists the available SDKs, but the end results is still poor - the contributor may not know how to install the required SDK.

A way to improve this experience could be the ability to optionally specify a custom message that'd be shown if the required SDK is absent. E.g., something like this:
```json
{
    "sdk": {
        "additionalPaths": [ ".dotnet" ],
        "installMessage": "Please run restore.cmd to install the required tooling.",
    }
}
```

akoeplinger:
> it'd be real nice to be able to just specify a .NET SDK version and have some global.json parameter to order NuGet to fetch it from an official source

This is essentially how most dotnet GitHub repos work, the build.sh/.ps1 scripts for commandline builds use a helper script called [tools.sh](https://github.com/dotnet/arcade/blob/main/eng/common/tools.sh)/[tools.ps1](https://github.com/dotnet/arcade/blob/main/eng/common/tools.ps1) which looks into global.json and installs the corresponding SDK into a local `.dotnet` folder if the system dotnet isn't sufficient. Having an official way to do that from the CLI would be nice, but I think this is orthogonal to this proposal.

RussKie:
> > > it'd be real nice to be able to just specify a .NET SDK version and have some global.json parameter to order NuGet to fetch it from an official source
> > 
> > 
> > This is essentially how most dotnet GitHub repos work, the build.sh/.ps1 scripts for commandline builds use a helper script called [tools.sh](https://github.com/dotnet/arcade/blob/main/eng/common/tools.sh?rgh-link-date=2023-12-04T19%3A16%3A36Z)/[tools.ps1](https://github.com/dotnet/arcade/blob/main/eng/common/tools.ps1?rgh-link-date=2023-12-04T19%3A16%3A36Z) which looks into global.json and installs the corresponding SDK into a local `.dotnet` folder if the system dotnet isn't sufficient. Having an official way to do that from the CLI would be nice, but I think this is orthogonal to this proposal.
> 
> I like the idea of having a standardized acquisition process, though it's not entirely straightforward because the default sources for the SDK are not necessarily sufficient to obtain the SDK (e..g you might need non-official locations).

I avoided mentioning this, as I have [already floated that idea](https://github.com/dotnet/sdk/issues/8254#issuecomment-860995649) and it was rejected by @richlander citing ["security concerns"](https://github.com/dotnet/sdk/issues/8254#issuecomment-1079407194). Although, that rejection was countered by @akoeplinger in [this comment](https://github.com/dotnet/sdk/issues/8254#issuecomment-1259884639).
I appreciate the security angle, but I'm with @akoeplinger on this, the security issue exists regardless but the rejection of this option makes the developer experience substandard.

proposed/local-sdk-global-json.md
ViktorHofer: > Not sure why anyone would want to do that, but I'm sure someone will come up with a use-case at some point.

That's basically what the Arcade script does that acquires the .NET SDK. It first checks if a direct match (based on the sdk version) is found in the %PATH% location and only if not then downloads the SDK to the local ".dotnet" directory.

That script could honor the selection here to make that behavior more configurable.

agocke: Nit: `$machine$` doesn't mean anything to me. I'd call this `$default_install$` or similar.

And yeah, if we have a way to refer to the default install I would change `additionalPaths` to `installPaths` and just search those paths precisely in order.

agocke: Actually, we still need to know whether to check `$PATH`.

sharwell:
Windows Update KB5033741 just broke developers working on Roslyn.sln, again. We are in ongoing desperate need of this fix.

agocke:
FYI this is not a fix. This would be a new feature. No matter what we implement, it will not be backported to .NET 8.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,112 @@
...
+```json
+{
+    "sdk": {
+        "additionalPaths": [ ".dotnet" ],
+    }
+}
+```
+
+In this configuration the host resolver would find a compatible SDK, if it exists in `.dotnet` or a machine wide location. The host resolver will consider the `additionalPaths` in the order they are defined and will stop at the first match. If none of the locations have a matching SDK then
+the host resolver will fall back to the existing SDK resolution strategy: `%PATH%` followed by machine wide installations.
```
elinor-fung, line 56: > existing SDK resolution strategy: `%PATH%` followed by machine wide installations.

As of .NET 7, the only place the host's resolution logic will look for the SDK is relative to the running `dotnet` executable. That is, it won't look in the machine-wide registered path per Windows registry or `/etc/dotnet/install_location` or the default machine-wide path (`%ProgramFiles%\dotnet`, `/usr/share/dotnet`, `/usr/local/share/dotnet`).

The host also does not use `%PATH%` (and I'd rather it stay that way). Since `dotnet` may be from `%PATH%`, the host may appear to be affected, but it is just based on the running `dotnet`.

proposed/local-sdk-global-json.md
elinor-fung: I'd also rather not add environment variable expansion until there's a more compelling scenario.

proposed/local-sdk-global-json.md
elinor-fung: I'd prefer not having a separate `additionalPathsOnly` property.

Given that the magic is much less in 7+ (we only look for the SDK relative to the running `dotnet` and I don't recall fallout from it), I'd expect `$machine$`/`$default_install$` shouldn't end up in most uses.

Since the fallback logic would be to look relative to the running `dotnet`, I'd still consider that only local deployment and I could see always looking next to running `dotnet` (after `additionalPaths`/`installPaths`) being reasonable.

agocke:
Thanks for all the feedback everyone!

Given that the muxer is the riskiest component in the SDK, I'd like to propose a narrower change that hopefully meets the same needs. I propose we add exactly one property of the following form:

``````json
{
    "sdk": {
        "paths": [ "$host$", ".dotnet" ]
    }
}
```

When the `paths` variable `null` or missing the SDK resolution algorithm proceeds as it does today.

When present, `paths` effectively overrides the search paths entirely. The paths will be searched in order to find an SDK with the requested version. Resolution will halt whenever one is found.

 There is exactly one special variable: `"$host$"`. It resolves to the path of the currently running muxer (`dotnet.exe`) executable. It is equivalent to the current SDK search directory. Any number of paths may be added into the list. If they are absolute paths, they will be resolved according to standard path resolution. If they are relative paths, they will be resolved relative to the `global.json` file.

And that's it. No other customization is provided. Would this address the most pressing scenarios?

vitek-karas:
I think it's a good solution for now, and we'll see if it works for everybody or it needs more tweaking.
I only question the support for absolute paths. I don't see it being very useful, plus it feels like potential security trouble (but I don't have a specific case in mind).

agocke:
> I only question the support for absolute paths. I don't see it being very useful, plus it feels like potential security trouble (but I don't have a specific case in mind).

Agreed on the questionable usefulness. I hope it doesn't pose any security risks, as I think you could emulate the behavior by using a relative path with a very large number of `..` directories.

RussKie:
What about the developer experience part? The current "the SDK not found" experience is really bad.

agocke:
No opinion on the proposed customizable error message. @elinor-fung , @vitek-karas any thoughts there?

vitek-karas:
I think it would make sense to be able to add some custom message to the error.
Ideally, when we fail to find the SDK, we should:
* Print out the fact that we could not find an SDK and specify which version we were looking for
* Print out the locations we looked into (basically the resolved list of paths from the `global.json` if it's used)
* Append the custom message if there is one

elinor-fung:
I think just the paths support already addresses the main pain, but a customizable error message would be reasonable. Currently, after listing the requested version, global.json file, and installed SDKs, we print a generic instruction like:
```
Install the [9.0.100-alpha.1.23615.4] .NET SDK or update [C:\helloworld\global.json] to match an installed SDK.
```
which could be replaced by any custom message.

proposed/local-sdk-global-json.md
jaredpar: I think I may be missing the distinction between the host and the resolver here. The key is for the core resolution to respect the new design such that products like VS and VS Code are effectively ignorant of this change. They just ask the resolver for the SDK as they've always done and the fact that we have a new strategy for finding it is an implementation detail. 

@rainersigwald, @baronfel as they pushed me in the current direction.

proposed/local-sdk-global-json.md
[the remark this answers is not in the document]
jaredpar: Updated to specify it should line up with our install locations.

proposed/local-sdk-global-json.md
jaredpar: Going to remove environment variables from the doc. Will add it into a _possible future expansions_ section.

agocke:
OK, I think we can accept the design from https://github.com/dotnet/designs/pull/303#issuecomment-1902268838, plus a customizable error message field. I think we can also schedule this work for .NET 9.

@jaredpar Do you want to update this proposal to that effect, or should I create a new proposal?

jaredpar:
> Do you want to update this proposal to that effect, or should I create a new proposal?

I can update this proposal (or if you want feel free to push the changes).

.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "[markdown]": {
+    "editor.rulers": [80],
+    "editor.wordWrap": "bounded",
+    "editor.wordWrapColumn": 80
```
jaredpar, line 5: These settings match with the linting rules in this repo. I'd prefer the linting rules be changed (why 80 characters hard wrap? this is 2023, not 1980 😉 ) but at the moment leaning into previous decisions.

agocke (commented):
Two minor corrections, otherwise LGTM.

Just want to confirm that our editors/dev tools use this entry point.

jaredpar:
> Just want to confirm that our editors/dev tools use this entry point.

@rainersigwald can you confirm this?

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+deployed .NET SDKs
+- [docs and scripts][example-scripts-build] to setup the environment and launch
+VS so it can find the deployed .NET SDKs.
+- [scripts][example-scripts-dotnet] that wrap `dotnet` to find the  _correct_
+`dotnet` to use during build.
+
+These scripts are not one offs, they are increasingly common items in repos in
+`github.com/dotnet` to attempt to fix the disconnect. Even so many of these
+solutions are incomplete because they themselves only consider local deployment.
+They dont't fully support the full set of ways the SDK can be deployed.
```
rolfbjarne, line 68:
```suggestion
They don't fully support the full set of ways the SDK can be deployed.
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+
+This problem also manifests in how customers naturally want to use our
+development tools like Visual Studio or VS Code. It's felt sharply on the .NET
+team, or any external customer who wants to contribute to .NET, due to how
+arcade infrastructure uses xcopy deployment into `.dotnet`. External teams
+like Unity also feel this pain in their development:
+
+- This [issue][cases-sdk-issue] from 2017 attempting
+to solve this problsem. It gets several hits a year from customers who are
```
rolfbjarne, line 77:
```suggestion
to solve this problem. It gets several hits a year from customers who are
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+}
+```
+
+In a scenario where the `.dotnet` directory had 7.0.200 SDK but there was a
+machine wide install of 7.0.300 SDK, the host resolver would pick 7.0.200 out
+of `.dotnet`. That location is considered first, it has a matching .NET SDK and
+hence discovery stops there.
+
+This design requires us to only change the host resolver. That means other
+tooling like Visual Studio, VS Code, MSBuild, etc ... would transparently
```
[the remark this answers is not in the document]
vitek-karas: I'll leave the detailed analysis to @elinor-fung , but I think it will work out that way. The changes will be in `hostfxr` and we've been advocating for all of the other tools to use `hostfxr` to locate the SDK. Specifically `MSBuildLocator` does use `hostfxr`. That said, I think it would be good to include testing of some of the obvious use cases as part of the feature work on this (we've been surprised in the past unfortunately).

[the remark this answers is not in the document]
vitek-karas: Additionally, the fact this lives in `hostfxr` will mean that it's versioning behavior will be somewhat "unexpected". As long as one has the .NET 9 version with this change installed, it will work regardless of the SDK version chosen in the `global.json` because we always use the latest `hostfxr` available in a given installation. This includes previews, so just installing a preview of .NET 9 will make this work basically everywhere on the machine.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+like Unity also feel this pain in their development:
+
+- This [issue][cases-sdk-issue] from 2017 attempting
+to solve this problsem. It gets several hits a year from customers who are
+similarly struggling with our toolings inability to handle local deployment.
+- This [internal discussion][cases-internal-discussion] from a C# team member.
+They wanted to use VS as the product is shipped to customers and got blocked
+when we shipped an SDK that didn't have a corresponding MSI and hence VS
+couldn't load Roslyn anymore.
+- [VS code][cases-vscode] having to adjust to consider local directories for SDK
```
RussKie, line 83:
```suggestion
- [VS Code][cases-vscode] having to adjust to consider local directories for SDK
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+
+These scripts are not one offs, they are increasingly common items in repos in
+`github.com/dotnet` to attempt to fix the disconnect. Even so many of these
+solutions are incomplete because they themselves only consider local deployment.
+They dont't fully support the full set of ways the SDK can be deployed.
+
+This problem also manifests in how customers naturally want to use our
+development tools like Visual Studio or VS Code. It's felt sharply on the .NET
+team, or any external customer who wants to contribute to .NET, due to how
+arcade infrastructure uses xcopy deployment into `.dotnet`. External teams
```
RussKie, line 73:
```suggestion
.NET Arcade infrastructure uses xcopy deployment into `.dotnet`. External teams
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+## Detailed Design
+
+The global.json file will support two new properties under the `sdk` object:
+
+- `"paths"`: this is a list of paths that the host resolver should
+consider when looking for compatible SDKs. In the case this property is `null`
+or not specified, the host resolver will behave as it does today.
+- `"errorMessage"`: when the host resolver cannot find a compatible .NET SDK it
+will display the contents of this property instead of the default error message.
+In the case this property is `null` or not specified, the current error message
```
RussKie, line 95:
```suggestion
In the case this property is `null` or not specified, the default error message
```

RussKie (commented):
🎉

proposed/local-sdk-global-json.md
jaredpar: Hmm ... is this my problsem or your problsem?

proposed/local-sdk-global-json.md
jaredpar: Dont't understand this feedback ... i'm so bad at spelling

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+will be displayed.
+
+The values in the `paths` property can be a relative path, absolute path or
+`$host$`.  When a relative path is used it will be resolved relative to the
+location of the containing global.json. The value `$host$` is a special value
+that represents the machine wide installation path of .NET SDK for the
```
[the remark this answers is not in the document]
elinor-fung: From .NET 7 and later, the host only looks next to the running `dotnet.exe`. I don't think we want anything representing the previous (.NET 6 and below, only different on Windows) behaviour.

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+considered _in addition_ to the current set of locations. Further if resolution
+failed the resolver would display the contents of `errorMessage` instead of
+the default error message.
+
+## Motivation
+
+There is currently a disconnect between the ways the .NET SDK is deployed in
+practice and what the host resolver can discover when searching for compatible
+SDKs. By default the host resolver is only going to search for SDKs in machine
+wide locations: `C:\Program Files\dotnet`, `%PATH%`, etc ...  The .NET SDK
```
elinor-fung, line 33:
```suggestion
SDKs. By default the host resolver is only going to search for SDKs next to
the running `dotnet`. This often means machine-wide locations, since users
and tools typically rely on `dotnet` already being on the user's path when
launching, instead of specifying a full path to the executable. The .NET SDK
```

proposed/local-sdk-global-json.md
```diff
@@ -0,0 +1,215 @@
...
+though is commonly deployed to local locations: `%LocalAppData%\Microsoft\dotnet`,
+`$HOME/.dotnet`. Many repos embrace this and restore the correct .NET for their
+builds into a local `.dotnet` directory.
+
+The behavior of the host resolver is incompatible with local based deployments.
+It will not find these deployments without additional environment variable
+configuration and instead search in machine wide locations. That means tools
```
elinor-fung, line 40:
```suggestion
configuration and only search next to the running `dotnet`. That means tools
```

proposed/local-sdk-global-json.md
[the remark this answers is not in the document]
agocke: I will take this to mean VS will partially work, VS Code will work fully. As a VS Code user, I am fine with this 😄

proposed/local-sdk-global-json.md
agocke: Not only will `dotnet exec app.dll` not be affected, `dotnet app.dll` will also not be affected. This change will affect _SDK_ resolution only, not runtime resolution. `global.json` is not used for runtime execution.

proposed/local-sdk-global-json.md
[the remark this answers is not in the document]
vitek-karas: VS will work the way it always was - even today if you set `PATH` and `DOTNET_ROOT` which is what the various scripts in repos do, before starting VS, VS will still run the .NET Framework version of MSBuild it ships with for design builds and the other one for build. But it should still use `hostfxr` to find the SDK to use for the given app. So this change should make it unnecessary to use the special scripts - which is probably 80% of the value of this feature :-)
That is the one scenario we definitely need to test when developing this feature.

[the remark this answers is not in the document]
RussKie: ...including the out-of-proc scenarios (e.g., the Windows Forms designers). /cc: @merriemcgaw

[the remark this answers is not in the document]
vitek-karas: > VS uses the .NET Framework version of MSBuild it ships with, period, for evaluation and for all builds including design time and F5 builds.

Thanks for the explanation!

proposed/local-sdk-global-json.md
elinor-fung: > This change will affect SDK resolution only, not runtime resolution.

I think this is they key part. But with potentially confusing aspects:
1. In order to run an SDK command, the host clearly needs to load same runtime. For this case, we should look for the runtime based on where the SDK was found. So in a sense it does affect runtime resolution, but only for running the SDK command, not for 'normal' app execution.
2. `dotnet exec` looks like an SDK command, but is not. It just runs the app and the SDK is not involved.
3. `dotnet run` is an SDK command that also runs the app. The SDK command launches the app, [setting the `DOTNET_ROOT` environment variable](https://github.com/dotnet/sdk/blob/9fd3b1eec66c9106433387749a3d490d8aec59f9/src/Cli/dotnet/commands/dotnet-run/RunCommand.cs#L261-L269) based on the running `dotnet`.
4. Other SDK commands [set `DOTNET_HOST_PATH`](https://github.com/dotnet/sdk/blob/9fd3b1eec66c9106433387749a3d490d8aec59f9/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs#L45) based on the running `dotnet`. Tools running as part of an SDK command look at its value.

For 3 and 4 in their current state, it would mean that the SDK command could be from some local deployment relative to global.json, but the command would then set variables such that child processes resolve the runtime based on the running `dotnet`, not where the SDK is running from. This may be odd? I can see the expectation that anything started due to one SDK command would be based on where that SDK was resolved. However, that would also mean `dotnet run` vs `dotnet app.dll` would be different in terms of where the runtime could come from.

vitek-karas: `dotnet run` and if I remember correctly also `dotnet test` actually have a somewhat more complex behavior. They will only propagate the SDK's runtime if the target is running on the same arch. They both need to support running apps/tests which target code incompatible with the SDK's runtime (for example running x86 target from x64 SDK). But for the same architecture they both try to make the target run on the same runtime as the SDK.

vitek-karas: The mechanics in `dotnet test` are different, but the overall goal was to achieve similar behavior to `dotnet run`.

MarcoRossignoli: Correct we set VSTEST_WINAPPHOST_* vars and we use it in case of apphost usage(testhost.exe) https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.TestHostProvider/Hosting/DotnetTestHostManager.cs#L488

Before the `--arch` support we used the default naming `DOTNET_ROOT_*` but architecture switch uses these and so we had to "pass" the sdk root information in a different one.

Architecture switch uses the ``DOTNET_ROOT_*` ` to look for the correct muxer that can be set by users and can be different than the sdk one https://github.com/microsoft/vstest/blob/main/src/Microsoft.TestPlatform.CoreUtilities/Helpers/DotnetHostHelper.cs#L153

@nohwnd  something to add?

jaredpar: > I'd like to see the "doesn't affect dotnet exec app.dll" part called out explicitly. Naively I would have expected it to, though I think it's ok that it doesn't.

I agree that tripped me up to so I will call it out. It materially impacts the scenarios here cause it is common in builds to effectively build a .NET core based tool then launch it with `dotnet exec`. That is done many times in the Roslyn build. 

At the same time it seems like we can continue this with a small tweak. Just need to flip to using `dotnet run` instead of `dotnet exec` in our scripts. That fixes other problems like having to hard code in the location of the built binary.

jaredpar: > Other SDK commands [set DOTNET_HOST_PATH](https://github.com/dotnet/sdk/blob/9fd3b1eec66c9106433387749a3d490d8aec59f9/src/Cli/Microsoft.DotNet.Cli.Utils/MSBuildForwardingAppWithoutLogging.cs#L45) based on the running dotnet. Tools running as part of an SDK command look at its value.

Want to make sure that when we run `dotnet build` that `DOTNET_HOST_PATH` gets set to the _resolved_ runtime / SDK. Example it would get set to the app local .NET SDK. Pretty sure based on reading that is the case but wanted to double check.

agocke (approved):
LGTM

proposed/local-sdk-global-json.md
```diff
@@ -182,6 +182,24 @@ for configuration around best versus first match is much stronger. There would
...
 that case.
 
+### dotnet exec
+
+This proposal only impacts how .NET SDK commands do runtime discovery. The
+command `dotnet exec` is not an .NET SDK command but instead a way to invoke
+the app directly using the runtime installed with `dotnet`.
+
+It is reasonable for complex builds to build and use small tools. For example
+building a tool for linting the build, running complex validation, etc ... To
```
RussKie, line 192:
```suggestion
building a tool for linting the build, running complex validation, etc. To
```

proposed/local-sdk-global-json.md
jaredpar: Want to elaborate on `DOTNET_HOST_PATH` a bit. Let's assume I have the following: 

- `dotnet.exe` is installed at `C:\Program Files\dotnet\dotnet.exe` and is on `%PATH%`. This location includes a 7.0.400 .NET SDK install
- The current directory is `c:\source` and has the following: 
  - `.dotnet` which is a directory that has 8.0.100 .NET SDK installed
  - `global.json` that requires `8.0.100` .NET SDK and has `path: ["$host$", ".dotnet" ]`. 

Now in the current directory I run `dotnet build` inside `c:\source`. That will eventually decide to use the 8.0.100 SDK from `c:\source\.dotnet` because the machine wide installation does not work.

What will the value of `%DOTNET_HOST_PATH%` be inside of msbuild here? That needs to be `c:\source\.dotnet` otherwise this proposal falls apart. The msbuild environment depends on this to `dotnet exec` tools inside of it so it must be a `dotnet.exe` that will have access to the 8.0.100 runtime. 

@elinor-fung, @agocke

simonferquel:
Just asking for confirmation:
if I have a custom MSBuild SDK (that extends the .net SDK) will I be able to use those paths instead of relying on nuget to resolve it ? - the spec only mentions the .net SDK, not custom ones.

agocke:
@MartyIX  you can do this today by installing .NET 9, but using global.json to specify a .NET 8 SDK version. That will cause your build to always use .NET 8. When you want to test on .NET 9 you can temporarily bump that version to .NET 9, then revert the change when you’re done.

xoofx:
@agocke do you think this work could be prioritized for .NET 10? We are currently blocked by this and it is causing internally lots of confusion and pain when building with a custom SDK that is not same used by the IDE.

PR #304, opened 2023-11-09
Sign CLI:  signature provider plugins

dtivel:
Currently, [Sign CLI](https://github.com/dotnet/sign) only supports signing with Azure Key Vault.  Sign CLI should support pluggable signature providers to enable other private key storage options.  This is necessary to keep pace with recent CAB Forum changes to private key storage requirements for publicly trusted code signing certificates.

CC @clairernovotny, @richlander

### DIFF
--- a/accepted/2023/sign-signature-provider-plugins.md
+++ b/accepted/2023/sign-signature-provider-plugins.md
@@ -1,0 +1,470 @@
+# Sign CLI Signature Provider Plugins
+
+**Owners** [Damon Tivel](https://github.com/dtivel) | [Claire Novotny](https://github.com/clairernovotny)
+
+Recent CAB Forum updates to baseline requirements<sup>[1](#r1)</sup> strengthened private key storage requirements for publicly trusted code signing certificates.  While older, less secure storage options (e.g.:  [PKCS #12 & PFX](https://wikipedia.org/wiki/PKCS_12)) became obsolete, more secure options (e.g.:  [HSM](https://wikipedia.org/wiki/Hardware_security_module)) became standard.  Many existing code signing tools do not support the new standard.
+
+[Sign CLI](https://github.com/dotnet/sign) already supports the new standard with [Azure Key Vault](https://learn.microsoft.com/azure/key-vault/general/overview#securely-store-secrets-and-keys) digest signing.  However, Sign CLI does not support other cloud providers, code signing services, or HSM tokens.  For that, Sign CLI needs a formal abstraction for signature providers and a signature provider plugin model that enables users to install the appropriate plugin for their situation.
+
+Note that a signature provider plugin is agnostic of signature format (e.g.:  Authenticode, VSIX, NuGet, etc.).  A plugin accepts a digest and provides a raw signature which Sign CLI embeds in the appropriate signature format.
+
+There is nothing in this proposed plugin model that precludes creation of a plugin that enables signing with a PFX file, and such a plugin might be welcome to a subset of users.  However, given the relative lack of support in existing signing tools for more secure private key storage options, the primary driver for this proposal is the need to support CAB Forum's more secure key storage requirements.
+
+## Scenarios and User Experience
+
+It is assumed that Sign CLI has already been installed (e.g.:  [`dotnet tool install --global sign --version 0.9.1-beta.23530.1`](https://www.nuget.org/packages/sign/0.9.1-beta.23530.1)).
+
+_All plugin names and examples below are fictitious and for illustration purposes only.  Also for illustration purposes, assume that Sign CLI's existing Azure Key Vault support has already been moved out of Sign CLI itself into a separate plugin._
+
+### Sign artifacts using Azure Key Vault
+
+First, the Azure Key Vault plugin must be installed.  The following command would download and install the latest version of the plugin.
+
+```text
+sign plugin install Microsoft.Azure.KeyVault.Sign
+```
+
+Executing `sign code -?` will show the new available command:
+
+```text
+...
+Commands:
+  azure-key-vault <file(s)>  Use Azure Key Vault.
+```
+
+Similarly, executing `sign code azure-key-vault -?` will show help for the new command and its options.
+
+```text
+Description:
+  Use Azure Key Vault.
+
+Usage:
+  sign code azure-key-vault <file(s)> [options]
+
+Arguments:
+  <file(s)>  File(s) to sign.
+
+Options:
+  -kvc, --azure-key-vault-certificate                    Name of the certificate in Azure Key Vault.
+  <azure-key-vault-certificate> (REQUIRED)
+  -kvi, --azure-key-vault-client-id                      Client ID to authenticate to Azure Key Vault.
+  <azure-key-vault-client-id>
+  -kvs, --azure-key-vault-client-secret                  Client secret to authenticate to Azure Key Vault.
+  <azure-key-vault-client-secret>
+  -kvm, --azure-key-vault-managed-identity               Managed identity to authenticate to Azure Key Vault.
+  -kvt, --azure-key-vault-tenant-id                      Tenant ID to authenticate to Azure Key Vault.
+  <azure-key-vault-tenant-id>
+  -kvu, --azure-key-vault-url <azure-key-vault-url>      URL to an Azure Key Vault.
+  -an, --application-name <application-name>             Application name (ClickOnce).
+  -d, --description <description> (REQUIRED)             Description of the signing certificate.
+  -u, --description-url <description-url> (REQUIRED)     Description URL of the signing certificate.
+  -b, --base-directory <base-directory>                  Base directory for files. Overrides the current working
+                                                         directory. [default: F:\git\sign]
+  -o, --output <output>                                  Output file or directory. If omitted, input files will be
+                                                         overwritten.
+  -pn, --publisher-name <publisher-name>                 Publisher name (ClickOnce).
+  -fl, --file-list <file-list>                           Path to file containing paths of files to sign within an
+                                                         archive.
+  -fd, --file-digest <file-digest>                       Digest algorithm to hash files with. Allowed values are
+                                                         'sha256', 'sha384', and 'sha512'. [default: SHA256]
+  -t, --timestamp-url <timestamp-url>                    RFC 3161 timestamp server URL. [default:
+                                                         http://timestamp.acs.microsoft.com/]
+  -td, --timestamp-digest <timestamp-digest>             Digest algorithm for the RFC 3161 timestamp server. Allowed
+                                                         values are sha256, sha384, and sha512. [default: SHA256]
+  -m, --max-concurrency <max-concurrency>                Maximum concurrency. [default: 4]
+  -v, --verbosity                                        Sets the verbosity level. Allowed values are 'none',
+  <Critical|Debug|Error|Information|None|Trace|Warning>  'critical', 'error', 'warning', 'information', 'debug', and
+                                                         'trace'. [default: Warning]
+  -?, -h, --help                                         Show help and usage information
+```
+
+The new command can then be used to sign artifacts.
+
+```text
+sign code azure-key-vault -kvu https://my.vault.azure.test/ -kvc MyCertificate -kvm -d Description -u http://description.test -b C:\ClassLibrary1\ClassLibrary1\bin\Debug\net8.0 ClassLibrary1.dll
+```
+
+### Sign artifacts using Windows certificate store
+
+First, the plugin must be installed.  The following command would download and install the latest version of the plugin.
+
+```text
+sign plugin install Microsoft.Windows.CertificateStore.Sign
+```
+
+The new command can then be used to sign artifacts.
+
+```text
+sign code certificate-store --store-location CurrentUser --store-name My --sha1fingerprint da39a3ee5e6b4b0d3255bfef95601890afd80709 -d Description -u http://description.test -b C:\ClassLibrary1\ClassLibrary1\bin\Debug\net8.0 ClassLibrary1.dll
+```
+
+## Requirements
+
+### Goals
+
+* Create a plugin model that enables pluggable signature providers.  A signature provider plugin will offer an alternate implementation of [`System.Security.Cryptography.RSA`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.rsa) and the [`System.Security.Cryptography.X509Certificates.X509Certificate2`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.x509certificate2) corresponding to the private key.
+* Make Sign CLI plugin-neutral.  While Sign CLI may install some core plugins (TBD), most plugins should be installed separately from Sign CLI itself, and Sign CLI's only interactions with any plugin should be through this plugin model.
+* Enable Sign CLI and plugins to version and release independently.
+
+### Non-Goals
+
+* Create a distribution channel for plugins.  Sign CLI is a .NET tool and is [available on https://nuget.org](https://www.nuget.org/packages/sign/).  Plugin packages can be published to any NuGet feed, including <https://nuget.org>.
+* Create a dynamic discovery mechanism for plugins.  Initially, we'll probably have a web page that lists common plugins and where to get them.
+* Manage (list, update, uninstall) installed plugins.
+* Add support for other signature algorithms (e.g.:  ECDSA).
+* Enable localizability of plugin command and option descriptions.
+
+## Design
+
+### High-level approach
+
+1. Create and publish a new _interfaces-only_ NuGet package that defines plugin-specific interfaces to be implemented by plugins.
+1. Implement the [dependency inversion](https://wikipedia.org/wiki/Dependency_inversion_principle) pattern by having Sign CLI and plugins reference the interfaces package.
+1. Move Azure Key Vault-specific implementations currently in Sign CLI into an Azure Key Vault-specific plugin.
+1. Augment Sign CLI commands at runtime with contributions from installed plugins (like [these options](https://github.com/dotnet/sign/blob/ef0e6b3ef8281dff1d62cea34445bd88fc3e6714/src/Sign.Cli/AzureKeyVaultCommand.cs#L25-L31) for Azure Key Vault).
+1. Enable Sign CLI to install new plugins and discover locally installed plugins.
+
+This design roughly follows [.NET's existing plugin model](https://learn.microsoft.com/dotnet/core/tutorials/creating-app-with-plugin-support).
+
+### Interfaces package
+
+We will create a new .NET assembly that contains only public interfaces to be implemented by plugins.  Sign CLI will implement a new command for a plugin that loads and interacts with the plugin implementation entirely by interfaces defined in the interfaces assembly.  This approach will enable Sign CLI and plugins to rev their implementations without either having any extraneous compile-time or runtime dependencies.
+
+Proposed interface:
+
+```C#
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+namespace Sign.Plugins.SignatureProvider.Interfaces
+{
+    public interface ISignatureProviderPlugin
+    {
+        Task InitializeAsync(IReadOnlyDictionary<string, string?> arguments, IServiceProvider serviceProvider, CancellationToken cancellationToken);
+        Task<X509Certificate2> GetCertificateAsync(CancellationToken cancellationToken);
+        Task<RSA> GetRsaAsync(CancellationToken cancellationToken);
+    }
+}
+```
+
+Sign CLI will pass all plugin arguments as a read-only dictionary of name-value pairs.  The plugin is responsible for all argument parsing and validation.
+
+The new interfaces assembly will be packaged and published to <https://nuget.org>, similar to [NuGetRecommender's contracts-only package](https://www.nuget.org/packages/Microsoft.DataAI.NuGetRecommender.Contracts).  The Sign CLI team will manage the source code repository for this package and publish the package to <https://nuget.org>.
+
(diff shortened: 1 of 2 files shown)


accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+
+The new interfaces assembly will be packaged and published to <https://nuget.org>, similar to [NuGetRecommender's contracts-only package](https://www.nuget.org/packages/Microsoft.DataAI.NuGetRecommender.Contracts).  The Sign CLI team will manage the source code repository for this package and publish the package to <https://nuget.org>.
+
+The interfaces package itself can have package dependencies; however, because Sign CLI and all plugins would inherit new interfaces package dependencies, we should exercise restraint and caution before adding new dependencies.  An example of a package dependency worth having is [`Microsoft.Extensions.Logging.Abstractions`](https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions).  Sign CLI uses [`ILogger`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.ilogger) ubiquitously, it makes sense for plugins to write to a shared logger.  From the provided `System.IServiceProvider` argument, plugins can request an `ILogger` instance for logging.
```
agr, line 196: > From the provided `System.IServiceProvider` argument

Example above uses concrete `ServiceProvider`.  The text should be aligned with the code sample.

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+* Make Sign CLI plugin-neutral.  While Sign CLI may install some core plugins (TBD), most plugins should be installed separately from Sign CLI itself, and Sign CLI's only interactions with any plugin should be through this plugin model.
+* Enable Sign CLI and plugins to version and release independently.
+
+### Non-Goals
+
+* Enable signature formats that require a certificate to sign without a certificate.
+* Enable signature formats to support asymmetric algorithms they do not already support.
+* Create a distribution channel for plugins.  Sign CLI is a .NET tool and is [available on https://nuget.org](https://www.nuget.org/packages/sign/).  Plugin packages can be published to any NuGet feed, including <https://nuget.org>.
```
agr, line 113: If interaction with arbitrary NuGet feeds is expected to be supported shouldn't the spec cover how user would tell sign CLI which feed to use? What about authentication to feeds that require one? V2/V3 support?

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+
+As part of this process, Sign CLI will use [`System.Runtime.Loader.AssemblyLoadContext`](https://learn.microsoft.com/dotnet/api/system.runtime.loader.assemblyloadcontext) and [`System.Runtime.Loader.AssemblyDependencyResolver`](https://learn.microsoft.com/dotnet/api/system.runtime.loader.assemblydependencyresolver) to load a plugin assembly and its dependencies strictly from the directory cone of the plugin's entry point assembly.  For example, if the plugin's entry point is `lib/net6.0/Microsoft.Azure.KeyVault.Sign.dll`, then Sign CLI will attempt to resolve assemblies under `lib/net6.0`.
+
+Sign CLI should log:
+
+* high-level plugin loading information at the [`Information`](https://learn.microsoft.com/dotnet/api/microsoft.extensions.logging.loglevel?view=dotnet-plat-ext-7.0#fields) log level.
+* assembly loading details at `Debug` / `Trace` log levels
+* errors at the `Error` log level
+
+### Sign CLI commands
```
agr, line 444: Should "list plugins" and "remove plugin" commands also exist?

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+
+New commands will be added to Sign CLI to manage plugins.
+
+#### `sign plugin <install>`
+
+The `sign plugin` command will expose subcommands for managing plugins.
+
+##### `sign plugin install <PluginPackageId> [--version <Version>]`
+
+This command will install the latest release version of the plugin package identified by `PluginPackageId` using existing package sources and NuGet's default NuGet.config lookup order.  If the latest version is already installed, the command will no-op.
```
agr, line 454: What if a previous version of a plugin is installed? What should happen?

accepted/2023/sign-signature-provider-plugins.md
richlander: I'm also wondering if NuGet-based plugins should be developed as a separate feature. Seems useful.

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+
+> Note:  If a plugin author wanted to remove support for older versions of a plugin, we could achieve that by specifying a minimum version of the plugin package in Sign CLI.  Older versions would be ignored, and plugin interfaces for those older versions could safely be removed from the interfaces package, provided that they are no longer needed.  Then, it would be possible for a plugin author to remove support for older versions and modify existing interfaces (vs. add new interfaces) in one step.  This remains true to the spirit of earlier guidance, that the interfaces package should preserve interfaces for all _supported_ plugin versions.  Enabling plugin authors to drop support for older versions of a plugin is out of scope for this specification.
+
+### Plugins
+
+A Sign CLI signature provider plugin:
+
+* extends Sign CLI functionality
+* contains a [`plugin.json`](#plugin-json-file) file in its root directory
```
richlander, line 217: I'm worried that this might lead to conflicts if people look at this work as a reference implementation. We might end up with a model where lots of .NET apps require a plugin.json for their plugins. I can imagine a case where some plugin can be a plugin for multiple apps.

richlander:
Looks good.

I'm seeing a lot over similarity / overlap with .NET tools. It would be good to talk with the tools folks to see if there can be sharing here. For example, tools are a specific package type.

On roll-forward, `LatestMajor` is probably the wrong setting. Context here: https://github.com/dotnet/sdk/pull/37231

It seems like this project is 90%+ of the way to defining a generic plug-in model system for .NET. That's really interesting. We should talk to the app model folks about that.

@vitek-karas @baronfel

accepted/2023/sign-signature-provider-plugins.md
vitek-karas: I would agree - would it be better to base the plugin discovery on patterns?
* Name of the file (`Microsoft.Azure.KeyVault.SignPlugin.dll` -> `*.SignPlugin.dll`) ?
* Use `System.Reflection.Metadata` to load the assembly (but not for running) and inspect it - from there you could read the necessary metadata (name, description, ...) - one option would be assembly level attributes for example (just like version information is stored today).

accepted/2023/sign-signature-provider-plugins.md
vitek-karas: One other consideration - the current description seems to hint at the nuget package being just downloaded as a zip file and extracted onto disk - no real NuGet logic involved. This will mean that the host will have to implement some of the logic NuGet client already implements (like TFM resolution). Would it make more sense to resolve the package via nuget client instead? That would provide all of the resolution logic as well as authentication and so on.

agocke:
Sorry, I haven't had time to read the full proposal, but I see that it relies on an assembly load from disk as an extensibility point. This security model would devolve to arbitrary code execution, mediated by the security of the file system and the file loading code.

That seems like a very large security surface area. I'm not ruling out that assembly load may be necessary for this level of extensibility, but does pose a substantial risk. I think we should at minimum do a full security review of the plugin loading design. Alternatively, if we can build an extensibility model that doesn't require assembly load, that might be easier to secure.

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+```
+
+## Requirements
+
+### Goals
+
+* Create a plugin model that enables pluggable signature providers.  A signature provider plugin will offer an alternate implementation of [`System.Security.Cryptography.AsymmetricAlgorithm`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.asymmetricalgorithm) and the [`System.Security.Cryptography.X509Certificates.X509Certificate2`](https://learn.microsoft.com/dotnet/api/system.security.cryptography.x509certificates.x509certificate2), if available, corresponding to the private key.  Note that while there may be valid scenarios for signing with a raw asymmetric key pair, some signing operations require a certificate and will fail if a certificate is not available.
+* Make Sign CLI plugin-neutral.  While Sign CLI may install some core plugins (TBD), most plugins should be installed separately from Sign CLI itself, and Sign CLI's only interactions with any plugin should be through this plugin model.
+* Enable Sign CLI and plugins to version and release independently.
```
vitek-karas, line 107: What is the reason for this requirement?
Or rather, why do we think it's necessary to allow pretty much anybody to build a plugin? I understand that there are several providers which would want to add a plugin here, but is it really necessary to fully decouple them?
For example, why not build all of these in-repo and ask the providers to simply send a pull request?
At runtime the plugins could still be loaded dynamically, but you would avoid the complexity of acquisition, install and security.

agr: What's the reason to restrict the ability for anyone to develop and distribute those on their own? Delivery mechanisms exist, it's not like whole new infrastructure is built just for distributing plugins.
What if someone doesn't want their code to be open source or disagrees with Sign CLI license? What if someone doesn't want a third party to control the release of their code?

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+`description` | yes | string | N/A | N/A
+`aliases` | yes | array of strings | N/A | N/A
+`dataType` | no | string | `Text` | `Text`, `Boolean`
+`defaultValue` | no | string | N/A | N/A
+`isRequired` | no | `true`/`false` | `false` | `true`, `false`
+
+Example:
+
+```JSON
+{
```
agr, line 337: Do we need some kind of indicator that it is a plugin for Sign CLI specifically?

agr: Alternatively, NuGet package type mechanism could be used for that.

accepted/2023/sign-signature-provider-plugins.md
jkotas: > What if someone doesn't want their code to be open source or disagrees with Sign CLI license? What if someone doesn't want a third party to control the release of their code?

They can fork the project and build their own distribution of the tool with their own modifications.

agr: And then that fork would have to be maintained and we'll end up with several Sign CLI distributions with different signing capabilities. Why require extra gestures from third parties? The more friction is added to the process, the less desire there would be to participate.
Unless there are security concerns with plugin delivery and loading, efforts for adding plugins should be reduced.

vitek-karas: Sorry if I wasn't clear - I'm not saying we should not do it. My point was about cost/complexity versus benefit. The added complexity and cost to support separate distribution of plugins looks substantial. I was mostly looking for a discussion about the needs and if there are other ways to solve the problem.

There's a somewhat similar concept in the .NET SDK - workloads. The requirements there are obviously different, but it was (and still is) a substantial effort to develop and maintain this (and I don't think we even allow 3rd part workloads).

Also, the entire "dotnet tool" functionality is similar in some ways as well.

richlander: >  the entire "dotnet tool" functionality is similar in some ways as well.

I had the same view. We shouldn't build this feature as spec'd in isolation of dotnet tools. They are (broad brush stroke) effectively identical. We should either do what Vitek is proposing (which is probably best) or build a first class add-in model that this tool can use.

The more I think about it, dotnet tools with source forking/sharing give you pretty much what the spec is proposing, and I think, via simpler mechanisms that are easier to understand and definitely cheaper to build.

agocke: Removing the plugin load requirement would also allow this tool to be compiled with Native AOT.

That would be very useful for systems which either don't have/want a pre-installed .NET runtime, or don't allow JITing. I know there are some secure systems which don't allow JITing today, so I would not rule out that requirement in the future.

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+    /// non-support.
+    ///
+    /// Some signature formatters work with raw asymmetric key pairs without a certificate.  To support only these
+    /// signature formatters, a plugin need not implement this interface.
+    ///
+    /// Some signature formatters (like Authenticode) do require a certificate and will fail without a certificate.
+    /// To support these signature formatters, the plugin should implement this interface.  Signature formatters may
+    /// fail if they require a certificate but a signature provider plugin does not implement this interface.
+    /// </remarks>
+    public interface ICertificateProvider
```
JeromySt, line 185: I'd imagine that it would be desirable for `ICertificateProvider` to be a special type of `ISignatureProvider` as well. The definition could be updated to `public interface ICertificateProvider : ISignatureProvider` such that those interfaces that only support `ISignatureProvider` can still use a certificate acquired from a `ICertificateProvider`?

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+    {
+        Task<X509Certificate2> GetCertificateAsync(CancellationToken cancellationToken);
```
JeromySt, line 187: I can foresee a use case where this interface would also need to provide access to the certificate chain of the certificate (this is needed in the COSE Sign1 world).  Would it be prudent to add that API surface to this interface?

accepted/2023/sign-signature-provider-plugins.md
clairernovotny: Hi,

Trying to answer a few things here around the intents:

- The binary separation of the plugins was in part to enable any service capable of providing a signature the ability to do so via this tool. 
- We wanted one tool, not forks, to keep it as simple as possible for end users. Forks are hard to maintain and there's also the possibility that multiple providers could be needed for more complex scenarios down the road.
- We wanted to isolate a plugin's dependencies from the main tool. That could be parts of the Azure SDK for Key Vault, or could be the AWS SDK should AWS choose to write a plugin. Could be other things entirely. 
- We didn't want to be gatekeepers of who gets to play. That includes having to review PR's for additional providers. That doesn't really scale well.

richlander: > We wanted to isolate a plugin's dependencies from the main tool. 

The proposed fork model does that even more naturally.

clairernovotny: There is a significant downside to forking -- maintenance. It is anticipated that we'll have at least 15-20 signature providers. For each of those to fork, which puts a significant burden on them to stay up to date with any changes and create new releases each time. That is likely more effort than many/most of those organizations providing the signatures would like to sign up for--it's a much easier lift for them to create a plugin once and update it if the interface changes than to continually release updated versions of an app. It also creates support issues for each of the providers as now they're front-line support for the app and not just their plugin.

If an app gets out of date, then that creates significant confusion for end users as to what features or bugs exist in which version of the app. This has already happened with private forks of the codebase to support other providers, where their version contains bugs already fixed in the main code. The support cost has also deterred public release of that fork.

richlander: How often do you think that one of these signature providers would need to issue an update in the global tool model?

accepted/2023/sign-signature-provider-plugins.md
dtivel: @agr, @vitek-karas, good questions.  We should use NuGet API's.  This would enable TFM resolution, support alternate package feeds, authentication, etc.  The spec should detail this.

@richlander, I agree a general model does seem like it might be useful.  @clairernovotny and I were hard-pressed to find close precedents here, so what we're doing seemed novel.  We're open to working with others on a generalized model if it would be useful.

accepted/2023/sign-signature-provider-plugins.md
dtivel: The intention was this spec would represent v1 of the feature, which would not have those commands.  Those commands could be added later.

Alternatively, we could spec them now and prioritize implementation after everything else.  (The same could be said about other potential features, but we were trying to expedite v1 by not speccing too much.)

accepted/2023/sign-signature-provider-plugins.md
dtivel: @agr, the intention was that it would install alongside the previous version without removing the previous version.

accepted/2023/sign-signature-provider-plugins.md
dtivel: @richlander, @vitek-karas, I understand your concerns.  We drew inspiration from dotnet/templating's template.json file as a reference on how to declare assets and contributions.  Here, plugin.json provides 2 things:  a mechanism for locating a plugin's main assembly and CLI option metadata.

I'm not sure how to entirely eliminate your concerns without tackling a generic .NET tool plugin model at the same time, such that dotnet/sign follows the same pattern.  A general .NET tool plugin model wasn't originally in scope, but we're open to working with others to find out how we can deliver this in alignment with a generic .NET tool plugin model.  There is pressure for dotnet/sign to deliver something in the very near future.

We could use an assembly file name pattern instead and remove that section from plugin.json.  plugin.json still specifies CLI option metadata, and we wanted to leave the door open for localization, even though it's out of scope now.   @clairernovotny and I talked about using attributes in the assembly to provide CLI option metadata, potentially generating plugin.json at build time (similar to dotnet/templating), but that all felt nice-to-have not MVP.

Are you suggesting that we eliminate plugin.json entirely or that a plugin.json file should conform to a standard .NET tool plugin schema?

accepted/2023/sign-signature-provider-plugins.md
dtivel: @agr, if we move to a more generic .NET tool plugin pattern, then probably yes.

accepted/2023/sign-signature-provider-plugins.md
```diff
@@ -0,0 +1,524 @@
...
+    1. If a plugin contains net6.0, net7.0, and net8.0 assemblies and Sign CLI's net7.0 assemblies are loaded by the .NET 8 runtime, Sign CLI should load the plugin's net7.0 assemblies.
+    1. If a plugin only had net6.0 and net8.0 assemblies, Sign CLI should load the net6.0 assemblies or report that plugin isn't available for the current runtime.
+1. Load the assembly at the entry point's `filePath` location.
+1. Create an instance of the type `implementationTypeName`.
+1. Cast the instance to an interface defined in the interfaces package with type name `interfaceTypeName`.
+
+As part of this process, Sign CLI will use [`System.Runtime.Loader.AssemblyLoadContext`](https://learn.microsoft.com/dotnet/api/system.runtime.loader.assemblyloadcontext) and [`System.Runtime.Loader.AssemblyDependencyResolver`](https://learn.microsoft.com/dotnet/api/system.runtime.loader.assemblydependencyresolver) to load a plugin assembly and its dependencies strictly from the directory cone of the plugin's entry point assembly.  For example, if the plugin's entry point is `lib/net6.0/Microsoft.Azure.KeyVault.Sign.dll`, then Sign CLI will attempt to resolve assemblies under `lib/net6.0`.
```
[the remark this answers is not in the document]
dtivel: @baronfel, we're open to an out-of-process design.  We thought an in-process design was something we could deliver quickly.  An out-of-process design does have benefits over an in-process design, as mentioned elsewhere in this review.

baronfel:
I had a chat with @clairernovotny a bit ago and we covered a few points of interest to this design. There are a few major points that we covered around

* Discoverability (both end-user knowledge of what plugins are available to be installed, and tool-plugin knowledge of what plugs are _currently_ installed)
* UX of the CLI interface to the resulting `sign` command: `dotnet sign providerA` may have drastically different arguments than `dotnet sign providerB`, and so each provider needs to be able to communicate its own set of arguments to the 'core' sign tool for areas like Help, Tab Completions, etc.
* Hosting/Invocation of each loaded plugin - aka the protocol that each plugin would need to adhere to to actually sign things

To me the areas in which this tool differs from the extensibility model of systems like [Git credential helpers](https://git-scm.com/docs/gitcredentials) or [Docker credential helpers](https://github.com/docker/docker-credential-helpers) are these latter two areas. These credential-helper extensibility models have a very limited and defined protocol, with little in the way of CLI-based customization. They are also fairly stateless - each invocation is distinct. The proposed model for plugins implies more persistence - based on my conversation with @clairernovotny each plugin is called repeatedly in a single signing session in a way that would require a whole rework of the protocol to support with an out-of-proc model in order to not tank performance. That could be done (@dtivel has signaled above that such designs might be useful) but it would definitely cause more delay here.

One potential comparison point here would be the `git` CLI itself - it has convention-based lookups for commands of the form `git-MY_COMMAND_HERE`, but it does so at the cost of potentially losing consistency of help/tab completions/etc - is that a cost we'd be willing to pay for that tool?

I think these differences are unique enough to not stop work here while we figure out some unified plugin plan.

anangaur:
@dtivel @baronfel  @clairernovotny please take a look at prior art here: https://youtu.be/FRrslz_AHzE?list=PLdo4fOcmZ0oWiK8r9OkJM3MUUL7_bOT9z&t=460 This is 1p tool [Upgrade Assistant](https://learn.microsoft.com/en-us/dotnet/core/porting/upgrade-assistant-overview?WT.mc_id=dotnet-35129-website#upgrade-with-the-cli-tool)

jkotas:
> @dtivel @baronfel @clairernovotny please take a look at prior art here: https://youtu.be/FRrslz_AHzE?list=PLdo4fOcmZ0oWiK8r9OkJM3MUUL7_bOT9z&t=460 This is Microsoft owned dotnet tool [Upgrade Assistant](https://learn.microsoft.com/en-us/dotnet/core/porting/upgrade-assistant-overview?WT.mc_id=dotnet-35129-website#upgrade-with-the-cli-tool)

I do not see the extensions feature in the current release of the `upgrade-assistant` tool. Was it removed? (It would not surprise me given the versioning issues - like the ones discussed here - that comes with it.)

```
C:\>dotnet tool install -g upgrade-assistant
Skipping NuGet package signature verification.

C:\>upgrade-assistant extensions --help

Error: Unknown command 'extensions'.

       extensions --help
       ^^^^^^^^^^ No such command


C:\>upgrade-assistant --help
USAGE:
    upgrade-assistant [OPTIONS] <COMMAND>

EXAMPLES:
    upgrade-assistant upgrade <PROJECT_PATH>
    upgrade-assistant upgrade <PROJECT_PATH> --operation Inplace --targetFramework net6.0
    upgrade-assistant upgrade <PROJECT_PATH> --operation SideBySide --targetFramework LTS --destination
<NEW_PROJECT_NAME_OR_PATH>

OPTIONS:
    -h, --help       Prints help information
    -v, --version    Prints version information

COMMANDS:
    upgrade    Upgrade project or its features. This command lets you pick upgrade options and guides through upgrade
               process
```

PR #305, opened 2023-11-22
Add feature switch attribute design doc

sbomer:
Adds a design for an attribute-based model for feature switches, to be supported in the trim/AOT analyzers and in ILLink/ILCompiler. There's significant overlap with @terrajobst's capability API analyzer draft in https://github.com/dotnet/designs/pull/261, but this approaches the problem more specifically with trimming support in mind.

### DIFF
--- a/accepted/2023/feature-switch-attributes.md
+++ b/accepted/2023/feature-switch-attributes.md
@@ -25,10 +25,10 @@ The attribute model is heavily inspired by the capability-based analyzer [draft]
   - [Validating correctness of feature guards](#validating-correctness-of-feature-guards)
   - [Feature guards and constant propagation](#feature-guards-and-constant-propagation)
 - [Feature check attributes](#feature-check-attributes)
-- [Relationship between feature checks and feature guards](#relationship-between-feature-checks-and-feature-guards)
-  - [Feature checks are also guards](#feature-checks-are-also-guards)
-  - [Feature guards may also be feature checks](#feature-guards-may-also-be-feature-checks)
-  - [Referencing features from FeatureGuard and FeatureCheck](#referencing-features-from-featureguard-and-featurecheck)
+- [Modeling dependencies between features](#modeling-dependencies-between-features)
+  - [Relationship between feature checks and feature guards](#relationship-between-feature-checks-and-feature-guards)
+  - [Feature guards may also have feature switches](#feature-guards-may-also-have-feature-switches)
+  - [Referencing features from `FeatureCheck` and `FeatureDependsOn`](#referencing-features-from-featurecheck-and-featuredependson)
 - [Unified view of features](#unified-view-of-features)
   - [Unified attribute model for feature checks and guards](#unified-attribute-model-for-feature-checks-and-guards)
 - [Comparison with other analyzers](#comparison-with-other-analyzers)
@@ -38,16 +38,15 @@ The attribute model is heavily inspired by the capability-based analyzer [draft]
 - [Namespace and visibility of feature switches](#namespace-and-visibility-of-feature-switches)
 - [Possible future extensions](#possible-future-extensions)
   - [Feature checks with inverted polarity (false means supported/available)](#feature-checks-with-inverted-polarity-false-means-supportedavailable)
-  - [Feature guards with inverted polarity](#feature-guards-with-inverted-polarity)
+  - [Feature dependencies with inverted polarity](#feature-dependencies-with-inverted-polarity)
   - [Feature attributes with inverted polarity](#feature-attributes-with-inverted-polarity)
   - [Validation or generation of feature check implementation](#validation-or-generation-of-feature-check-implementation)
-  - [Versioning support for feature attributes/checks/guards](#versioning-support-for-feature-attributeschecksguards)
+  - [Versioning support for feature attributes/checks](#versioning-support-for-feature-attributeschecks)
   - [Feature attribute schemas](#feature-attribute-schemas)
 - [Alternate API shapes](#alternate-api-shapes)
   - [Separate types for feature and Requires attribute](#separate-types-for-feature-and-requires-attribute)
-  - [Feature switches without feature attributes](#feature-switches-without-feature-attributes)
+  - [Feature switches without feature types](#feature-switches-without-feature-types)
   - [Generic attributes with interface constraint](#generic-attributes-with-interface-constraint)
-- [Implementation notes](#implementation-notes)
 
 ## Background
 
@@ -323,16 +322,48 @@ The attribute needs to reference the feature somehow, whether as:
 
   However, the Roslyn analyzer would have enough information from this attribute alone.
 
-## Relationship between feature checks and feature guards
+## Modeling dependencies between features
 
-### Feature checks are also guards
-A feature check property is also a feature guard for the feature it is defined by. We could encode this in one of two ways:
-- `FeatureCheck` only, with a unified representation that includes the mapping to both the attribute and feature name, or
-- Require both `FeatureCheck` and `FeatureGuard`, where both attributes together provide a mapping to the attribute and the feature name
+`FeatureGuard` expresses that a property depends on another feature. Because this information is on the property, not the type, it supports using the property as a guard, but would not support using a feature type (or feature attribute) as a guard. For example, if we had `RequiresDynamicCodeCompilationAttribute`, we would like this to act as a guard for `RequiresDynamicCode`:
 
-### Feature guards _may_ also be feature checks
+```csharp
+if (RuntimeFeature.IsDynamicCodeCompiled)
+    APIWhichRequiresDynamicCode(); // OK due to `FeatureGuard` on `IsDynamicCodeCompiled`
 
-A feature guard might also independently be a feature check property. We should be careful to avoid violating the assumptions of the feature guard by controlling the property via a feature switch.
+[RequiresDynamicCodeCompilation]
+static void UseDynamicCodeCompilation() {
+    APIWhichRequiresDynamicCode(); // should not warn
+}
+```
+
+To make this work, we could instead place `FeatureDependsOn` at the type level, which along with `FeatureCheck` would convey the same dependency that was previously expressed via `FeatureGuard`:
+
+```csharp
+[RequiresDynamicCodeCompilation]
+static void UseDynamicCodeCompilation() {
+    APIWhichRequiresDynamicCode(); // OK due to `FeatureCheck` and `FeatureDependsOn`
+}
+
+class RuntimeFeature {
+    [FeatureCheck(typeof(RequiresDynamicCodeCompilationAttribute))]
+    public static bool IsDynamicCodeCompiled => IsDynamicCodeSupported;
+}
+
+[FeatureDependsOn(typeof(RequiresDynamicCodeAttribute))]
+class RequiresDynamicCodeCompilationAttribute : Attribute {
+    // ...
+}
+```
+
+### Relationship between feature checks and feature guards
+
+A feature check property is by definition also a feature guard for the feature it is defined by. It may also be a guard for other features that it depends on
+
+`FeatureGuard` would allow the definition of a guard property without a separate feature check. However, we propose modeling feature dependencies at the type level via `FeatureDependsOn`, so that a feature guard property is defined by introducing a feature check for a feature that depends on another feature. In other words, every feature guard is also a feature check, namely one that has dependencies on other features.
+
+### Feature checks _may_ also have feature switches
+
+A feature check might also have a feature switch. We should be careful to avoid violating the assumptions of feature dependencies by controlling the property via a feature switch.
 
 For example, if we had a separate feature switch for `IsDynamicCodeCompiled`, this would allow setting `IsDynamicCodeCompiled` to `true` even when `IsDynamicCodeSupported` is `false`.
 
@@ -342,7 +373,6 @@ This is essentially what we did for features like `StartupHookSupport`, which ca
 class StartupHookProvider
 {
     [FeatureCheck(typeof(RequiresStartupHookSupport))]
-    [FeatureGuard(typeof(RequiresUnreferencedCodeSupport))]
     private static bool IsSupported => AppContext.TryGetSwitch("System.StartupHookProvider.IsSupported", out bool isSupported) ? isSupported : true;
 
     private static void ProcessStartupHooks()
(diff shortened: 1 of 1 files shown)


accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,719 @@
...
+
+  Both models allow multiple "feature switch" properties to be defined by the same feature, but the latter can prevent the same property from being marked as the "feature check" for multiple different features that could potentially conflict, if we set `AllowMultiple = false` on `FeatureSwitchAttribute`.
+
+- API shape: target of `CapabilityGuard` vs `FeatureGuard`
+
+  These are the same idea, and are represented very similarly in both models, targeting the guard property, with `AllowMultiple = true` to support a property that guards multiple features. The capability API draft lists the attribute targets as `AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Field`. We might want to start with just `AttributeTargets.Property`.
+
+## Comparison with platform compatibility analyzer
+
+The platform compatibility analyzer is semantically very similar to the behavior described here, except that it doesn't come with ILLink/ILCompiler support for removing removing branches that are unreachable when publishing for a given platform.
```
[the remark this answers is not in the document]
akoeplinger: > except that it doesn't come with ILLink/ILCompiler support for removing removing branches that are unreachable when publishing for a given platform

This is kind of implicitly working because `OperatingSystem.Is<OS>()` APIs are hardcoded to return false for the non-matching platforms so ILLink/ILCompiler constant propagation takes care of that.

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,719 @@
...
+
+- Teach the ILLink Roslyn Analyzer about the substitution XML
+
+  We don't want to teach the analyzer to read the substitution XML. The analyzer is the first interaction that users typically have with trimming and AOT warnings. This should not be burdened by the XML format. Even if we did teach the analyzer about the XML, it would not solve the problem because the analyzer must not globally assume that `IsDynamicCodeSupported` is false as ILCompiler does.
+
+- Define a model with the full richness of the supported OS platform attributes
+
+  We will focus initially on a model where feature switches are booleans that return `true` if a feature is enabled. We aren't considering supporting version checks, or feature switches of the opposite polarity (where `true` means a feature is disabled/unsupported). We will consider what this might look like just enough to gain confidence that our model could be extended to support these cases in the future, but won't design this fully in the first iteration.
```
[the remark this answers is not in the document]
akoeplinger: ah right, I got confused :)

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,718 @@
...
+        Name = name;
+    }
+}
+```
+
+The analyzer would validate that `FeatureGuardAttribute` and `FeatureSwitchAttribute` attribute arguments are derived classes of `FeatureAttribute`, and that `FeatureNameAttribute` is only placed on derived classes of `FeatureAttribute`.
+
+We could use this model even for feature switches similar to `StartupHookSupport`, which don't currently have a `StartupHookSupportAttribute`. There's no need to actually annotate related APIs with the attribute if that is overkill for a small feature. In this case the attribute definition would just serve as metadata that allows us to define a feature switch in a uniform way.
+
+Note that this makes it possible to define custom `Requires` attributes for arbitrary features. While the immediate goal of this proposal is not to enable analyzer support for analysis warnings based on such custom attributes, the model intentionally allows for this so that we could easily open up the analyzer to work for custom features. However, initially support for analysis warnings would be limited to `RequiresUnreferencedCodeAttribute`, `RequiresDynamicCodeAttribute`, and `RequiresAssemblyFilesAttribute`.
```
vitek-karas, line 413: There are actually technical challenges with support custom attributes for custom features - roslyn analyzer has to declare all produced diagnostics, so this would mean that all of the custom attributes would produce the same warning code - which is not ideal. Maybe we'll find a solution to this in the future.

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,718 @@
...
+  ```
+
+  ```csharp
+  if (Feature.IsSupported)
+    CallSomeAndroidAPI(); // no warning for guarded call
+  ```
+
+The platform compatibility analyzer also has some additional functionality, such as annotating _unsupported_ APIs, and including version numbers.
+
+## Possible future extensions
```
vitek-karas, line 505: One other future thing to look at would be the getter body for a property with a `FeatureSwitch` attribute. Currently these should follow a set of rules for the system to work correctly. Having either a way to validate it, or to generate these would help with consistency and robustness. These new attributes should provide enough information for this to be possible.

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,718 @@
...
+      [FeatureName]
+      public string PlatformName { get; }
+  }
+  ```
+
+  In this example, `FeatureName` indicates that the instances of `SupportedOSPlatformAttribute` should be differentiated based on the value of this parameter.
+
+## Alternate API shapes
+
+### Feature switches without feature attributes
```
vitek-karas, line 616: Personally I like another design: Have a separate type which represents the feature alone - and acts just as metadata.

```csharp
[FeatureName("System.Runtime.CompilerServices.RuntimeFeature.IsDynamicCodeSupported")]
[FeatureRequirement(typeof(RequiresDynamicCode))]
static class DynamicCodeSupportedFeature {
}

class RequiresDynamicCode : Attribute { }

class RuntimeFeature {
    [FeatureSwitch(typeof(DynamicCodeSupportedFeature))]
    public static bool IsDynamicCodeSupported => // ...

    [FeatureGuard(typeof(DynamicCodeSupportedFeture))]
    public static bool IsDynamicCodeCompiled => // ..
}
```

The current proposal basically collapses the `RequiresDynamicCodeAttribute` with the `DynamicCodeSupportedFeature`. The disadvantage of that is:
* Features which don't have a requirement attribute would still need to define one - which is confusing

Another consideration is visibility. Whichever type we use to "define the feature" (either its own type, or the requires attribute) - if that type is not public, it's not possible to define custom feature guard properties for the feature. This might be useful, but could also become a problem.

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,718 @@
...
+This will silence warnings from the analyzer without any indication that `RequiresDynamicCode`-annotated code might be reached at runtime. We should be extremely clear in our documentation that this is not the intended use.
+
+### Validating correctness of feature guards
+
+Better would be for the analyzer to validate that the implementation of a feature check includes a check for `IsDynamicCodeSupported`. This could be done using the analyzer infrastructure we have in place without much cost for simple cases:
+
+```csharp
+class Feature {
+    [FeatureGuard(typeof(RequiresDynamicCodeAttribute))]
+    public static bool IsSupported => RuntimeFeature.IsDynamicCodeSupported && SomeOtherCondition(); // OK
```
AaronRobinsonMSFT, line 172: The above is perhaps okay, but the following wouldn't be.

```csharp
public static bool IsSupported => SomeOtherCondition() && RuntimeFeature.IsDynamicCodeSupported;
```

Consider the case where `SomeOtherCondition()` performs a check with side-effects. Do we have an opinion on that? Also what about with `||` instead of `&&`?

```csharp
public static bool IsSupported => RuntimeFeature.IsDynamicCodeSupported || SomeOtherCondition();
```

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,718 @@
...
+}
+```
+
+```csharp
+class Feature {
+    [FeatureGuard(typeof(RequiresDynamicCodeAttribute))]
+    public static bool IsSupported => SomeOtherCondition(); // warning
```
AaronRobinsonMSFT, line 179: I assume this is only a warning if `SomeOtherCondition()` isn't also annotated with a matching `FeatureGuard`.

accepted/2023/feature-switch-attributes.md
vitek-karas: Note that marking something as a FeatureGuard may not necessarily say anything about the trimmer behavior - I could see us implementing it such that the trimmer still has to call the property getter in that situation, unless it can prove that it's constant (through constant propagation).
This behavior would make a lot of sense even for NativeAOT - that doesn't do constant propagation, and wouldn't need to in the future. The scanner would be happy with just the branch removal, but leaving the getter call in place. And then it's up to JIT to inline/remove the getter call if it can prove it's constant.

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,766 @@
+# Attribute-based model for feature switches
+
+.NET has [feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switch.md) which can be set to turn on/off areas of functionality in our libraries, with optional support for removing unused features when trimming or native AOT compiling. What we describe overall as "feature switches" have many pieces which fit together to enable this:
+
+- MSBuild property
+- RuntimeHostConfigurationOption MSBuild ItemGroup
```
terrajobst, line 6:
```suggestion
- `RuntimeHostConfigurationOption` MSBuild item group
```

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,766 @@
...
+- `runtimeconfig.json` setting
+- AppContext feature setting
```
terrajobst, line 8:
```suggestion
- `AppContext` feature setting
```

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,766 @@
...
+- **ILLink.Substitutions.xml**
+- **static boolean property**
```
terrajobst, line 10:
```suggestion
- **static Boolean property**
```

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,766 @@
...
+- **Requires attributes**
+
+The bold pieces are the focus of this document. [Feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switch.md) describes how settings flow from the MSBuild property through the AppContext (for runtime feature checks) or `ILLink.Substitutions.xml` (for feature settings baked-in when trimming). This document aims to describe an attribute-based model to replace some of the functionality currently implemented via ILLink.Substitutions.xml, used for branch elimination in ILLink and ILCompiler to remove branches that call into `Requires`-annotated code when trimming.
```
terrajobst, line 13:
```suggestion
The bold pieces are the focus of this document. [Feature switches](https://github.com/dotnet/designs/blob/main/accepted/2020/feature-switch.md) describes how settings flow from the MSBuild property through the `AppContext` (for runtime feature checks) or `ILLink.Substitutions.xml` (for feature settings baked-in when trimming). This document aims to describe an attribute-based model to replace some of the functionality currently implemented via ILLink.Substitutions.xml, used for branch elimination in ILLink and ILCompiler to remove branches that call into `Requires`-annotated code when trimming.
```

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,766 @@
+# Attribute-based model for feature switches
```
terrajobst, line 1: The introduction (all the text until the Background header) is very hard to read because it mixes defining terms and listing existing functionality with problem statement.

I think you'll want to describe (in simple non-jargon terms) what UX currently exists, what problem it has, and what (at a high level) you're going to propose to address this.

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,766 @@
...
+
+- Teach the ILLink Roslyn Analyzer about the substitution XML
+
+  We don't want to teach the analyzer to read the substitution XML. The analyzer is the first interaction that users typically have with trimming and AOT warnings. This should not be burdened by the XML format. Even if we did teach the analyzer about the XML, it would not solve the problem; for example, the analyzer must not globally assume that `IsDynamicCodeSupported` is false as ILCompiler does.
+
+- Define a model with the full richness of the supported OS platform attributes
+
+  We will focus initially on a model where feature switches are booleans that return `true` if a feature is enabled. We aren't considering supporting version checks, or feature switches of the opposite polarity (where `true` means a feature is disabled/unsupported). We will consider what this might look like just enough to gain confidence that our model could be extended to support these cases in the future, but won't design this fully in the first iteration.
+
+## Feature guard attribute
```
terrajobst, line 86: It would help if you could list all the features you'll want to guard today. The only examples that currently exist seem to be around dynamic code. That doesn't seem to be too compelling, especially because those already have attributes.

Also, do you imagine that users can extend the set of features? Or are the features being guarded a closed set that is defined by the .NET platform?

accepted/2023/feature-switch-attributes.md
```diff
@@ -0,0 +1,766 @@
...
+}
+```
+
+The attribute instance needs to reference the feature somehow, whether as:
+
+- a reference to the feature attribute:
+
+  ```csharp
+  class Feature {
+      [FeatureGuard(typeof(RequiresDynamicCodeAttribute))]
```
terrajobst, line 109: Using a type seems unfortunate because that would require on attribute per feature and would disallow more complex shapes (eg an attribute with flags).

jtschuster:
Personally, I think it makes the most sense to have a single identifier to associate guards, switches, and requirements to a feature. To me, using the feature name everywhere rather than an attribute makes the most sense.

I think this generalizes well to other feature switches, and fits well with `AppContext.TryGetSwitch`. Any guarded block (or method body with a `RequiresFeatureAttribute`) is assumed to have the guarded features "enabled" within it. Calls to APIs with the `RequiresFeatureAttribute` would compare the required feature name to the list of enabled feature names for that block.

The trimmer and Native AOT could hard code the (new) "UnreferencedCode" and "DynamicCode" feature names as `false` and could accept other feature names and values at the command line to do branch elimination for arbitrary feature switches.

I also like the proposed `NamedFeatureAttribute` as the base `RequiresFeatureAttribute` so we don't need to require `FeatureNameAttribute` to only be on `RequiresFeatureAttribute`-derived classes, and can apply the attribute ad-hoc without defining a new derived attribute.

```C#
class RequiresFeatureAttribute : Attribute {
    public string FeatureName { get; }

    public NamedFeatureAttribute(string name) => FeatureName = name;
}
```

```C#
// [FeatureName("MyLibrary.Features.MyFeature")]
// class RequiresMyFeature : RequiresFeatureAttribute { }

class RequiresMyFeature : RequiresFeatureAttribute("MyLibrary.Features.MyFeature") { }

class Features
{
    [FeatureSwitch("MyLibrary.Features.MyFeature")]
    public static bool MyFeature => AppContext.TryGetSwitch("MyLibrary.Features.MyFeature" out bool val) && val;

    [FeatureSwitch("MyLibrary.Features.MyOtherFeature")]
    public static bool MyOtherFeature => AppContext.TryGetSwitch("MyLibrary.Features.MyOtherFeature" out bool val) && val;

    [FeatureGuard("MyLibrary.Features.MyOtherFeature")]
    [FeatureGuard("MyLibrary.Features.MyFeature")]
    public static bool BothFeatures => MyFeature && MyOtherFeature;
}

public class Program
{
    [RequiresMyFeature]
    public static void MyFeatureMethod() { }

    [Requires("MyLibrary.Features.MyOtherFeature")]
    public static void MyOtherFeatureMethod() { }
}
```

PR #308, opened 2023-12-07
Add a design for SdkAnalysisLevel

baronfel:
This is the first pass at SdkAnalysisLevel, a unified flag to help users manage their SDK warning levels in situations where they may not be able to pin SDKs via global.json or other means.  The property is intended to communicate to tooling like the compilers an explicit intent - "treat me as if I were SDK X.Y.Z" without requiring the user to install or manage other SDKs. The first well-defined use of this is the Roslyn field member breaking change, where members of the Roslyn team hope to use this flag in conjunction with other user signals to determine if a user should get warnings about the field breaking change.

I've spoken with most of the folks mentioned in the stakeholders/reviewers section about this before - now it's time to get broader input and push the design to done.

### DIFF
--- a/proposed/sdk-analysis-level.md
+++ b/proposed/sdk-analysis-level.md
@@ -163,10 +163,12 @@ The more interesting question is where such a value might be set. Users typicall
 * as a MSBuild local property in a Project file (or MSBuild logic Imported by a project file)
 * as a MSBuild local property in a Directory.Build.props file (this is broadly the same as option 2 but happens implicitly)
 
-If we would like users to be able to set this new property in any of these, the SDK cannot set a default value until after the Project file has been evaluated.
+If we would like users to be able to set this new property in any of these, the SDK cannot determine the 'final' value of `SdkAnalysisLevel` until after the Project file has been evaluated.
 If we want all of the build logic in `.targets` files to be able to consume or derive from the value then we must set the value as early as possible.
 These two constraints point to setting the property as early as possible during the `.targets` evaluation of the SDK - for this reason
 we should calculate the new property [at the beginning of the base SDK targets](https://github.com/dotnet/sdk/blob/558ea28cd054702d01aac87e547d51be4656d3e5/src/Tasks/Microsoft.NET.Build.Tasks/targets/Microsoft.NET.Sdk.targets#L11).
+
+While an earlier location could be chosen (e.g. Microsoft.NET.Sdk.props) this would open up a layering/timing issue where the value of `SdkAnalysisLevel` could be set to a default, then computed by some other MSBuild props before the users' input had been evaluated. This would be confusing and error-prone for users, and would require a lot of extra work to ensure that the value was always correct. For this reason we choose to set the value in the base SDK targets as described above.
 
 ### Relation to existing WarningLevel and AnalysisLevel properties
 
--- a/proposed/sdk-analysis-level.md
+++ b/proposed/sdk-analysis-level.md
@@ -1,0 +1,167 @@
+# SDK Analysis Level Property and Usage
+
+
+**Owner** (PM) [Chet Husk](https://github.com/baronfel) | (Engineering) [Daniel Plaisted](https://github.com/dsplaisted)
+
+Today, users of the .NET SDK have a large degree of control over the way that the .NET SDK and the
+tools bundled in it emit diagnostics (warnings and errors). This control is provided in part by a
+series of MSBuild properties that control the severity levels of specific warnings, if certain messages
+should be treated as diagnostics, even a coarse-grained way to set the entire analysis baseline for a project.
+
+The default values for these properties are often driven by the Target Framework(s) chosen for a project, and
+as a result users have developed an expectation of behaviors around changing the Target Framework of a project.
+It's generally understood that when a TFM is changed, new kinds of analysis may be enabled, and as a result what
+was a successful build may now have errors.
+
+This cadence is predictable and repeatable for users - new TFMs are generally only introduced once a year - but
+for tooling developers this cadence means that important changes can generally only be tied to a new TFM release.
+New diagnostics can be introduced mid cycle, but they can only be enabled by default in a new TFM. Failure to adhere
+to this pattern results in pain for our users, as they are often not in control of the versions of the SDK used
+to build their code, often because their environment is managed by an external force like an IT department or
+a build environment.
+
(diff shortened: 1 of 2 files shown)


baronfel:
CC the stakeholders/reviewers from the doc:

| Team | Representatives |
| ---- | --------------- |
| SDK | @marcpopmsft @dsplaisted |
| MSBuild | @rainersigwald @ladipro |
| NuGet | @aortiz-msft @nkolev92 |
| Roslyn | @jaredpar @CyrusNajmabadi |
| F# | @vzarytovskii @KevinRansom |

nkolev92 (approved):
Looks great.

I'm guessing we're not including the warning strengthening policy here?

proposed/sdk-analysis-level.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Where to define `SdkAnalysisLevel`
+
+The more interesting question is where such a value might be set. Users typically set values for properties in one of three ways:
+
+* as a MSBuild global property via an environment variable or `-p` option of MSBuild/.NET CLI
+* as a MSBuild local property in a Project file (or MSBuild logic Imported by a project file)
+* as a MSBuild local property in a Directory.Build.props file (this is broadly the same as option 2 but happens implicitly)
+
+If we would like users to be able to set this new property in any of these, the SDK cannot set a default value until after the Project file has been evaluated.
```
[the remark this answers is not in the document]
baronfel: Great point - I'll clear up the language here in the next revision

baronfel:
Note from @timheuer - cover why use this vs global.json? Changing the SDK chosen vs Changing the behavior of the chosen SDK.

baronfel:
> Looks great.
> 
> I'm guessing we're not including the warning strengthening policy here?

Yes exactly - we can define the policy/support matrix for teams after this is implemented. IIRC the main question was "how long is long enough to establish a lower bound for tools to support older SDKs", and some answers were "previous 2 LTSs", "previous LTS", and "forever".

proposed/sdk-analysis-level.md
```diff
@@ -0,0 +1,187 @@
...
+| PM | @KathleenDollard @MadsTorgersen |
+
+## Design
+
+### Valid values of `SdkAnalysisLevel`
+
+The implementation of `SdkAnalysisLevel` is itself quite straightforward - the default value of `SdkAnalysisLevel` is always
+the stable 'SDK Feature Band' of the SDK that is currently being run, and the only valid values for the property are other SDK Feature Bands.
+An SDK Feature Band is a Semantic Version of the form `<Major>.<Minor>.<Patch>`, where the `<Patch>` version is a multiple of 100. Some
+valid feature bands for this discussion might be 6.0.100, 7.0.300, or 8.0.200.
```
ladipro, line 156: Is the scenario of preview / RC SDK updates excluded on purpose? It could be useful to say "keep me on SdkAnalysisLevel 9.0.100-preview.5" if Preview 6 introduced a change that broke my build. I understand that it would mean more internal bookkeeping but if we included the `SdkAnalysisLevel` overrides in anonymous telemetry, it could help guide product decisions within a single major.minor.patch release.

baronfel: Yes - preview-level changes aren't in scope for this document. Users opting into previews are inherently accepting changes, this proposal is mostly around how users of stable versions can mitigate behavioral changes temporarily.

Evangelink:
@baronfel Feel free to ping us (test team) on such changes :) This is interesting as we are also feeling this "limitation".

baronfel:
I've created https://github.com/dotnet/sdk/pull/37778 with a potential implementation of the new field to help illustrate this doc.

tboby:
Both the scenarios described in this design are ones I have experienced in the past 6 months, and the changes proposed seem like the right way to start to reduce their impact. 
 
I note this design doesn't actually detail what should be done with this new level, so it's hard to say if it would solve the issue, but I hope giving a real world example might be useful. 

I think another important scenario is that of mixed target frameworks within a single build environment.

My context is a highly regulated industry where changes cannot be made lightly. For example, a single repository may have:

- a maintenance branch targeting 6 LTS
- main containing an rc targeting 7
- a feature branch targeting 8

We might simultaneously need pipelines to support:
- safely building the main branch if we need to release, which must be stable
- building and running new integration tests or security analysis tools on the maintenance branch - without changing the product
 - upgrading the feature branch to target the latest runtime/SDK to evaluate whether we can upgrade and get the longest support for the next release 

In order to simplify managing builds and enforce consistency our pipeline definitions are shared across repositories. 

We could theoretically migrate repositories to the next major SDK versions individually but:
- that involves making many small changes to the pipelines where errors are hard to detect
- branch specific pipelines are fragile at best or not supported at all

In addition windows cicd runners are complicated, and it's not practical to have multiple runners with  different sdks. It may not be acceptable nor practical to add global.json files to many frozen projects. 

> This version introduced a new warning which is treated as an error

This doesn't have to be a new warning; a bug fix to an existing warning is sufficient. This happened to us with the release of 8. A bug fix to CA2017 suddenly started failing builds due to a relatively minor logging issue.

This time round the number of repos, and the timing in relation to our releases meant we could just fix code. However, in two years I imagine we'll be too big for that to be practical, and a similar change in a minor SDK release would be equally disruptive. 

How this would help:
I would set sdk analysis level in every solution alongside the target framework so that builds do not break on SDK upgrade. I would potentially set up a parallel build pipeline which unset the level but did not fail builds to help assess impact. 

Questions:
1. Would any changes to analysers be considered a change in level?

If yes, how practical is it for dotnet to manage backwards compatibility? 

If no, how is this explained, and does it it undermine the value if setting the analysis level only sometimes works. 

2. What previous levels would be supported? Presumably all in support SDK versions at the time of release?

3. When would support be dropped? This might not be as complicated as I'm thinking, but would a patch SDK release of net 6 today maintain the analyser level for the previous LTS (which would be years out of support)

4. Re the non-goal "Users keep using SdkAnalysisLevel indefinitely": I can't see a practical way of using the property without making it a permanent feature in my solution build props alongside tfm. If it isn't already set then the SDK upgrade will break previously stable builds. It may be appropriate to keep using it until that runtime is out of support if the product lifetime is similar.

nkolev92:
@baronfel 

As far as I can tell, this design is implemented in https://github.com/dotnet/installer/pull/18687/files, should this PR be merged?

baronfel:
It definitely could - was mainly leaving it open as a way to collect more feedback, but we can 'snap' at this point-in-time.

PR #310, opened 2024-01-05
Handle part of Swift frozen struct ABI in the projection layers

jkoritzinsky:
Due to high cost of implementation in RyuJIT, update the Swift frozen type handling design to require the first lowering pass to be done at the projection layer so the signatures the JIT sees are already broken down to primitive type sequences if necessary.


jakobbotsch (approved):
Thanks! This proposal will greatly simplify things on our side.

proposed/swift-interop.md
```diff
@@ -118,6 +118,8 @@ When calling a function that returns an opaque struct, the Swift ABI always requ
...
 
+For frozen structs and enums, Swift has a complicated lowering process where the struct or enum type's layout are recursively flattened to a sequence of primitives. If this sequence is length 4 or less, the values of this type are split into the elements of this sequence for parameter passing instead of passing the struct as a whole. Structs and enums that cannot be broken down in this way are passed by-reference to their specified frozen layout. Due to high implementation cost in the RyuJIT, in particular in the `UnmanagedCallersOnly` scenario, we should implement this first pass of lowering in the projection layer; the function pointer signature and `UnmanagedCallersOnly` signatures should only have primitives, our special Swift register types, and pointer types. For reference, this lowering pass is done in the Swift compiler when lowering from Swift IL to LLVM IR. This design decision reinforces our direction of having the Runtime layer of Swift interop support similar features as the LLVM IR representation of Swift.
```
jkotas, line 121:
```suggestion
For frozen structs and enums, Swift has a complicated lowering process where the struct or enum type's layout are recursively flattened to a sequence of primitives. If this sequence is length 4 or less, the values of this type are split into the elements of this sequence for parameter passing instead of passing the struct as a whole. Structs and enums that cannot be broken down in this way are passed by-reference to their specified frozen layout. Due to high implementation cost in the RyuJIT, in particular in the `UnmanagedCallersOnly` scenario, we should implement this first pass of lowering in the projection layer; the only types allowed for `CallConvSwift` calling convention in method or function pointer signatures are primitives, our special Swift register types, and pointer types. For reference, this lowering pass is done in the Swift compiler when lowering from Swift IL to LLVM IR. This design decision reinforces our direction of having the Runtime layer of Swift interop support similar features as the LLVM IR representation of Swift.
```
I assume that we will require this higher lowering to be done in all cases, even for DllImport.

jkoritzinsky: Yes, this higher level lowering will be required in all cases.

It may be worth implementing it such that this portion of lowering is separate from the rest of the projection tooling for ease of implementation, but I don't think we need to write that all out here now.

JulieLeeMSFT:
Thanks for the design that simplifies the JIT work!

kotlarmilos (approved):

PR #313, opened 2024-01-19
Add sections talking about Swift SIMD types and interop with them

jkoritzinsky:
Swift's SIMD types are named/categorized along a different axis (number of elements) than the .NET types (vector width). Additionally, .NET historically has a few vector types that were defined based on number of elements for single-precision floating point vectors that have generally had poor support in interop scenarios.

These sections discuss both of these problems and how I believe we should solve them for our Swift interop experience.


proposed/swift-interop.md
```diff
@@ -118,7 +118,13 @@ When calling a function that returns an opaque struct, the Swift ABI always requ
...
...ype's layout are recursively flattened to a sequence of primitives. If this sequence is length 4 or less, the values of this type are split into the elements of this sequence for parameter passing instead of passing the struct as a whole. Structs and enums that cannot be broken down in this way are passed by-reference to their specified frozen layout. Due to high implementation cost in the RyuJIT, in particular in the `UnmanagedCallersOnly` scenario, we should implement this first pass of lowering in the projection layer; the only types allowed for `CallConvSwift` calling convention in method or function pointer signatures are primitives, our special Swift register types, and pointer types. We will allow struct types as the return type in a signature to support scenarios where a return value is passed in multiple registers instead of passed indirectly; this occurs in some scenarios such as some larger-than-register SIMD types. For reference, this lowering pass is done in the Swift compiler when lowering from Swift IL to LLVM IR. This design decision reinforces our direction of having the Runtime layer of Swift interop support similar features as the LLVM IR representation of Swift.
```
jkotas, line 121: > We will allow struct types as the return type in a signature to support scenarios where a return value is passed in multiple registers instead of passed indirectly

Just return types and no arguments? 

Is this behavior architecture specific?

jkoritzinsky: I've looked into this a little more closely. I'm updating the text to better describe what happens.

proposed/swift-interop.md
```diff
@@ -118,7 +118,13 @@ When calling a function that returns an opaque struct, the Swift ABI always requ
...
...s. If this sequence is length 4 or less, the values of this type are split into the elements of this sequence for parameter passing instead of passing the struct as a whole. Structs and enums that cannot be broken down in this way are passed by-reference to their specified frozen layout. When a frozen struct or enum with a valid primitive sequence of 4 elements or less is returned from a function, it is returned if it were a structure of the elements of the primitive sequence. Due to high implementation cost in the RyuJIT, in particular in the `UnmanagedCallersOnly` scenario, we should implement this first pass of lowering in the projection layer. The only types allowed for `CallConvSwift` calling convention in method or function pointer parameters are primitives, our special Swift register types, and pointer types. In return types, we will also allow structure types to support returning the primitive type sequences correctly. For reference, this lowering pass is done in the Swift compiler when lowering from Swift IL to LLVM IR. This design decision reinforces our direction of having the Runtime layer of Swift interop support similar features as the LLVM IR representation of Swift.
```
jkotas, line 121: Does this match how LLVM deals with it? Arguments are handled in Swift IL lowering, but return values are left to codegen.

jkoritzinsky: Both parameters and return values are inspected in Swift IL lowering and lowered to primitive type sequences of 4 or less primitives if possible. Parameters that are lowered to this sequence are passed as separate parameters, one for each element of the sequence. If there is a valid primitive sequence for the return type, the actual return type in LLVM IR is a struct of the elements of the type sequence, not the original struct type. Processing this struct into which registers to return it through or return it by a return buffer is then handled by LLVM.

I've validated this by looking at the IR emitted by the Swift compiler on Compiler Explorer: https://godbolt.org/z/o1h6Y5de8

jkotas, line 121: > a valid primitive sequence

What is a valid primitive sequence? What would be an invalid primitive sequence?

jkotas: Is there a good reason for the different handling of the return values vs. arguments in Swift/LLVM toolchain? Does this difference show up in the Swift public surface or is it just an internal implementation detail of the Swift toolchain that can change in future without breaking the Swift ABI?

It looks weird to standardize the different handling of the return values vs. arguments in public surface. On the other hand, we should be able to add the struct handling for arguments in future if needed, without breaking anything. So I guess it is ok to start with it.

jkotas (approved):

proposed/swift-interop.md
jakobbotsch, line 121: > When a frozen struct or enum with a valid primitive sequence of 4 elements or less is returned from a function, it is returned if it were a structure of the elements of the primitive sequence.

... it is returned **as** if it were a structure?

proposed/swift-interop.md
```diff
@@ -118,7 +118,13 @@ When calling a function that returns an opaque struct, the Swift ABI always requ
...
+
+##### SIMD Types
+
+We will pass the `System.Runtime.Intrinsics.VectorX<T>` types in SIMD registers as we do with the managed calling convention. We will treat the `Vector2/3/4` types as non-SIMD types (and block their usage directly as parameters in the `CallConvSwift` signature as is the case with other structs).
```
jakobbotsch, line 125:
```suggestion
We will pass the `System.Runtime.Intrinsics.VectorX<T>` types in SIMD registers. We will treat the `Vector2/3/4` types as non-SIMD types (and block their usage directly as parameters in the `CallConvSwift` signature as is the case with other structs).
```

I don't think the managed convention does this on all platforms (today).

Do we allow interop with these types in other interop scenarios? It sounds like it is going to add https://github.com/dotnet/runtime/issues/8300 + https://github.com/dotnet/runtime/issues/9578 as part of the work.

proposed/swift-interop.md
jkoritzinsky: I think this is the only way that Swift could implement the "lower to a type sequence of primitives" consistently for return values and parameters while still allowing enregistering return values.

proposed/swift-interop.md
jkoritzinsky: I believe we have support on ARM64 due to HFA/HVA support. If I'm wrong, then yes this would add in those two issues as part of this work.

jakobbotsch: Yes, on ARM64 we support it, but not on x64.

Is the SIMD interop important enough to warrant implementing it for x64? Those two issues on their own are large work items.

proposed/swift-interop.md
jkotas: Why would it be the only way? They could have done all lowering in the LLVM codegen part as part of Swift calling convention handling. Is there anything fundamental preventing that?

proposed/swift-interop.md
```diff
@@ -118,7 +118,13 @@ When calling a function that returns an opaque struct, the Swift ABI always requ
...
...ives. If this sequence is length 4 or less, the values of this type are split into the elements of this sequence for parameter passing instead of passing the struct as a whole. Structs and enums that cannot be broken down in this way are passed by-reference to their specified frozen layout. When a frozen struct or enum with a primitive sequence of 4 elements or less is returned from a function, it is returned as if it were a structure of the elements of the primitive sequence. Due to high implementation cost in the RyuJIT, in particular in the `UnmanagedCallersOnly` scenario, we should implement this first pass of lowering in the projection layer. The only types allowed for `CallConvSwift` calling convention in method or function pointer parameters are primitives, our special Swift register types, and pointer types. In return types, we will also allow structure types to support returning the primitive type sequences correctly. For reference, this lowering pass is done in the Swift compiler when lowering from Swift IL to LLVM IR. This design decision reinforces our direction of having the Runtime layer of Swift interop support similar features as the LLVM IR representation of Swift.
```
tannergooding, line 121: This sounds overall similar to the struct splitting and HFA support done for Unix System V.

proposed/swift-interop.md
jkoritzinsky: It looks like macOS x64 is still going to be widely supported when .NET 9 releases, so it depends on if the libraries we want to support are high enough priority. For example, the [Accelerate framework](https://developer.apple.com/documentation/accelerate) has many APIs that take the SIMD types.

@kotlarmilos what are the Apple libraries that we're targeting for .NET 9?

tannergooding: Why would `Vector2/3/4` be blocked? They have always been supported for interop and have been treated the equivalent of user-defined structs containing 2, 3, or 4 `float` fields (which is exactly how they are defined).

`Vector64/128/256/512<T>` and `Vector<T>` are all blocked from interop. The former set because Windows doesn't correctly handle SIMD returns today (this isn't vectorcall, but rather missing handling for the default x64 calling convention) and the latter because it doesn't make sense from an interop perspective today.

-----

> Yes, on ARM64 we support it, but not on x64.

This should exist for Unix already as well and only be missing for Windows x64, since that doesn't pass vectors differently (only returns them differently). `__vectorcall` would be required for Windows x64 HFA/HVA support and is still desirable long term so that we better optimize such perf critical functions; it just hasn't bubbled up in priority yet.

jkoritzinsky: `Vector2/3/4` would be blocked because they're already not handled as SIMD for interop and are allowed in P/Invokes (we have many tests validating that we don't break this behavior).

Swift on Windows is not something we're looking at supporting in the .NET 9 time frame, so that might help reduce cost here.

tannergooding: I don't understand why this means they'd be blocked? Presumably they should work and be treated as any other user defined struct with 2-4 float fields, no?

jkoritzinsky: Yes, and we block any structs as parameters because we are requiring the "primitive type sequence" lowering to be done at the projection level. As a result, treating Vector2/3/4 as a regular struct means that they're blocked.

kotlarmilos: > what are the Apple libraries that we're targeting for .NET 9?

We should start from https://github.com/dotnet/runtime/issues/95636. I don't expect any SIMD intrinsics in these libraries.

jakobbotsch: > > Yes, on ARM64 we support it, but not on x64.
> 
> This should exist for Unix already as well and only be missing for Windows x64, since that doesn't pass vectors differently (only returns them differently). `__vectorcall` would be required for Windows x64 HFA/HVA support and is still desirable long term so that we better optimize such perf critical functions; it just hasn't bubbled up in priority yet.

I don't see us passing `VectorX` in SIMD registers on either SysV or win-x64. But you're right that we do support passing structs classified as such in SIMD registers if the SysV classifier says so, so at least there is some support in the backend for this. We wouldn't be able to use the SysV classifier here, however.

It helps that we don't have to consider Windows, and given was @kotlarmilos said it sounds like we can consider the full x64 support to be a stretch goal for now.

proposed/swift-interop.md
jakobbotsch: I don't think this is similar to SysV. In the SysV classifier, there is just one struct layout and the classifier splits it out over a set of registers or passes the entire thing on stack. It's not hard to represent as a first-class ABI constraint in the JIT.

I think this is more like ARM32. In ARM32, you can have single arguments that are passed partly in registers and partly on stack. The JIT has special support for it (`FEATURE_ARG_SPLIT`) and a special node `PUTARG_SPLIT` that is only used on ARM32 (and win-arm64 which has this peculiar case in an edge case around varargs too).

Swift takes this to the next level. In Swift, you can have a single argument that interacts arbitrarily with the underlying calling convention. For example, a single struct argument can simultaneously have parts of it that are passed as implicit byrefs (copied to caller stack and a pointer passed), other parts of it that get passed in registers, other parts that are on stack, other parts that are HFAs.... There really is no limit to how many ABI constraints result from a single argument. That's what makes this harder to represent in a first-class way than any other ABI (especially in the UCO case).

proposed/swift-interop.md
jakobbotsch: (FWIW, I would still suggest incorporating my suggestion, since we don't pass VectorX types in SIMD registers across the board.)

proposed/swift-interop.md
jkoritzinsky: I mean this is the only way in LLVM IR to represent multiple return values.

If they took a different approach and did all of the lowering in LLVM instead of in the Swift compiler, they wouldn't have needed a representation for the midway point. However, I assume there was an issue with costing or scope.

@jakobbotsch requested that we split the lowering similarly in the Swift interop due to cost reasons with RyuJIT's architecture, so I wouldn't be surprised if the Swift team had the same problems.

I'd prefer to have all of this lowering handled in the callconv handling in the VM/JIT and not expose the ABI differences to the projection layer, but (as per the discussion that led to #310) the cost is too high for the JIT.

proposed/swift-interop.md
jkoritzinsky: Implementing the primitive-type-sequence lowering in the JIT was deemed too expensive (see the comments on #310 for some comments from the JIT team).

I agree that I'm not a fan of pushing these details up to the projection layer. However, since our consumers are expected to use the Swift interop projection tooling to interact with Swift and not use the callconv handling directly, it was deemed acceptable.

We expect all Swift interop users to use Swift as the source-of-truth. Manually written "blittable P/Invokes" that are directly used like with our C interop story with a nice API shape as a C# source-of-truth are out-of-scope.

jkotas: I think we should re-evaluate the decision now that we understand the problem space in more detail. The split that is currently proposed does not feel right.

The JIT lowering does not need to produce best code quality. Does it make the problem easier?

jkoritzinsky: From @jakobbotsch below:

> I don't think this is similar to SysV. In the SysV classifier, there is just one struct layout and the classifier splits it out over a set of registers or passes the entire thing on stack. It's not hard to represent as a first-class ABI constraint in the JIT.
>
> I think this is more like ARM32. In ARM32, you can have single arguments that are passed partly in registers and partly on stack. The JIT has special support for it (FEATURE_ARG_SPLIT) and a special node PUTARG_SPLIT that is only used on ARM32 (and win-arm64 which has this peculiar case in an edge case around varargs too).
>
> Swift takes this to the next level. In Swift, you can have a single argument that interacts arbitrarily with the underlying calling convention. For example, a single struct argument can simultaneously have parts of it that are passed as implicit byrefs (copied to caller stack and a pointer passed), other parts of it that get passed in registers, other parts that are on stack, other parts that are HFAs.... There really is no limit to how many ABI constraints result from a single argument. That's what makes this harder to represent in a first-class way than any other ABI (especially in the UCO case).

Based on what we've learned (in particular that the type-sequence lowering goes down to primitives), I don't think the situation is as dire. I think the following are the basic pieces of how a frozen struct could be passed:

If the primitive type sequence is longer than 4 primitives:

- The whole struct is passed as implicit byref.

If the primitive type sequence is less than 4 primitives, each element of the sequence can be passed one of the following ways:

- In a general purpose register
- In a SIMD register
- On the stack (once registers are exhausted)

Because Swift is not supported on x86 Windows, we don't need to worry about the implicit byref case as no primitive is passed as an implicit byref on 64-bit platforms (to my knowledge).

We don't need to worry about the HFA case as the struct will already be broken down to its constituent float/double primitives.

@jakobbotsch does that simplify the scenario enough to make this implementable in the JIT?

jakobbotsch: I share the concerns raised by others around exposing the detail to users. Just to be clear, the original discussion was around whether we could move some of this lowering from the JIT into the VM. My suggestion was that we would have the VM handle the lowering by creating IL stubs, in the same way that is done today if marshalling is needed. It would be transparent to the user and the user could have signatures with (frozen) structs and tuples.

The proposed design was already handling tuples in the projection layer, so it did not seem like a reach to handle frozen structs in the same way.

As I understand it the hard case with doing this in the VM is again the UCO case. How hard is it to create wrapper stubs in the VM that can handle the case even for UCO? Even if it is hard to do today, it seems like work that is similar to work we are going to end up needing for async2 efforts anyway. So if we can do that work today and avoid the complication in the JIT then I think this is way better. 

> The JIT lowering does not need to produce best code quality. Does it make the problem easier?

For calls I think we can handle the transformation in the JIT. We can split arguments as needed and represent them as the simpler ones with their normal ABI constraints (implicit byrefs, HFAs, registers, stack...). It probably requires some work around how we represent ABI constraints (today all arguments have associated metadata types, which they wouldn't have here), but this seems like "just work".

For UCO the only reasonable way to handle it is by spilling these to stack immediately, which as you say will result in terrible CQ. That's the case I'm worried about, since it requires making the backend work with a single parameter that has arbitrarily complicated ABI constraints associated with it. It's not impossible -- but I think doing this lowering in another place, such as the VM, will both be simpler and result in better codegen.

jkotas: > How hard is it to create wrapper stubs in the VM that can handle the case even for UCO? 

Very complicated. Impacts type system, diagnostics, reflection, ... .

> For UCO the only reasonable way to handle it is by spilling these to stack immediately, which as you say will result in terrible CQ.

If we went with wrapper stubs generated by VM, I expect that the CQ would be even worse.

jakobbotsch: > From @jakobbotsch below:
> 
> > I don't think this is similar to SysV. In the SysV classifier, there is just one struct layout and the classifier splits it out over a set of registers or passes the entire thing on stack. It's not hard to represent as a first-class ABI constraint in the JIT.
> > I think this is more like ARM32. In ARM32, you can have single arguments that are passed partly in registers and partly on stack. The JIT has special support for it (FEATURE_ARG_SPLIT) and a special node PUTARG_SPLIT that is only used on ARM32 (and win-arm64 which has this peculiar case in an edge case around varargs too).
> > Swift takes this to the next level. In Swift, you can have a single argument that interacts arbitrarily with the underlying calling convention. For example, a single struct argument can simultaneously have parts of it that are passed as implicit byrefs (copied to caller stack and a pointer passed), other parts of it that get passed in registers, other parts that are on stack, other parts that are HFAs.... There really is no limit to how many ABI constraints result from a single argument. That's what makes this harder to represent in a first-class way than any other ABI (especially in the UCO case).
> 
> Based on what we've learned (in particular that the type-sequence lowering goes down to primitives), I don't think the situation is as dire. I think the following are the basic pieces of how a frozen struct could be passed:
> 
> If the primitive type sequence is longer than 4 primitives:
> 
> * The whole struct is passed as implicit byref.
> 
> If the primitive type sequence is less than 4 primitives, each element of the sequence can be passed one of the following ways:
> 
> * In a general purpose register
> * In a SIMD register
> * On the stack (once registers are exhausted)
> 
> Because Swift is not supported on x86 Windows, we don't need to worry about the implicit byref case as no primitive is passed as an implicit byref on 64-bit platforms (to my knowledge).
> 
> We don't need to worry about the HFA case as the struct will already be broken down to its constituent float/double primitives.
> 
> @jakobbotsch does that simplify the scenario enough to make this implementable in the JIT?

It seems workable, but it sounds like it is based on the assumption that we do not need to handle tuples in the JIT. Are you suggesting we would allow frozen structs but not tuples in signatures?

jakobbotsch: > > How hard is it to create wrapper stubs in the VM that can handle the case even for UCO?
> 
> Very complicated. Impacts type system, diagnostics, reflection, ... .

Do you think the work will be avoided in the long run, or will it just be pushed into the future? Again I'm thinking about async2.

> > For UCO the only reasonable way to handle it is by spilling these to stack immediately, which as you say will result in terrible CQ.
> 
> If we went with wrapper stubs generated by VM, I expect that the CQ would be even worse on average.

I would not expect that. The JIT will be able to handle things efficiently through natural inlining and promotion of the constituent fields. We will have proper liveness that will allow the JIT to avoid the unnecessary spilling.

jkotas: Note that UCOs always need to move arguments to callee saved registers or spill them to stack as the very first thing to do `JIT_ReversePInvokeEnter` call.

jkotas: > Again I'm thinking about async2.

I think that the design for async2 is up in the air. It is more likely than not that the final async2 design is going to be quite different from what we have in the experiment prototype. I would not make any near term decisions based on async2.

> The JIT will be able to handle things efficiently through natural inlining and promotion of the constituent fields.

I do not think that the JIT will be able to eliminate the call overhead (fully inline the actual method) in average case. You can do a lot of spilling for the cost of a single call.

jkotas: We used to have the wrappers for UCOs on x86 and worked hard to get rid them.

jakobbotsch: > I do not think that the JIT will be able to eliminate the call overhead (fully inline the actual method) in average case. You can do a lot of spilling for the cost of a single call.

It does not seem hard to make these particular calls aggressively inlined if we think that's beneficial.

Regardless, I'll trust your intuition about it. Let's try to see what will be needed on the JIT side to make the spilling approach work once we get to that point.

@jkoritzinsky I'm still curious about the above. Can we assume that the JIT does not need to handle tuples?

tannergooding: My biggest concern here is that ABI is an extremely complex space and interop is one of those spaces where users want both simplicity and reduced overhead, especially when generating larger binding libraries.

Apple has also notably broken ABI in the past or deviated conventions from the norm on new platforms and so it is entirely possible some new platform comes on and now every single bit of `ObjC`/`Swift` interop code is DoA.

I think it is ultimately much better (even if its not what is done for the initial release due to timing constraints or w/e) that we have this support in the runtime as a detail of the `CallConv` support and that users are ultimately able to write a `delegate* unmanaged[CallConvSwift]<T, U, V>` that mirrors the underlying ObjC/Swift signature that would be exposed to C/C++ using the official Swift tooling.

jkotas: > It does not seem hard to make these particular calls aggressively inlined if we think that's beneficial.

These calls are typically going to have try/catch block in them to convert the .NET exception into switft error. You would have to implement inlining of methods with exception handling to make this work...

tannergooding: That is to say, a user should be able to export Swift bindings using official Apple tooling and then use another existing tool, such as ClangSharp, CppAst, etc; which can generate blittable P/Invoke bindings from a C header and expect it to work.

If we can't achieve that, I expect we will have a lot of downstream pain/headaches from the community, especially as it gets into more complex bindings and libraries.

jakobbotsch: > These calls are typically going to have try/catch block in them to convert the .NET exception into switft error. You would have to implement inlining of methods with exception handling to make this work...

Don't tempt me :-) (Note that this is actually part of our .NET 9 plan, and I also think it would be much more likely we end up with this support than appetite for improving the UCO Swift case in the future.)

jkoritzinsky: Re handling tuples: I think we can still handle tuples at the projection layer since the splitting of a tuple into separate arguments is done at the SIL layer and is very straightforward (it doesn't have nearly the same complexity as the "primitive type sequence" lowering) especially if JIT implementation cost for tuples would be too much.

Additionally, the primitive type sequence lowering happens after the tuple lowering (so each tuple element can be lowered to a sequence of up to 4 primitives), so handling tuples in the projection layer doesn't interfere with the primitive sequence handling.

proposed/swift-interop.md
jakobbotsch: > Why would it be the only way? They could have done all lowering in the LLVM codegen part as part of Swift calling convention handling. Is there anything fundamental preventing that?

FWIW, I was doing a little research and stumbled upon https://forums.swift.org/t/getting-a-better-understanding-of-how-the-swift-calling-convention-impacts-the-avr-abi/69397/2, and the essential thing seems to be that, as one would expect, LLVM's type system is not able to represent all Swift types.
I was wondering if we're going to run into similar problems around this. For example, how are we going to represent discriminated unions? Example: https://godbolt.org/z/6nss9hrvh -- the discriminated union seems to be lowered by the Swift frontend into three parameters `i64, i64, i8`.

jkoritzinsky: It will be up to the projection tooling to (based on the `.swiftinterface` file) determine the layout of the discriminated union. The projection tooling would determine that the layout is a struct of `{i64 (c1.0 and c2.0), i64 (c2.1), i8 (descriminator)}`. Then the generated signature with `CallConvSwift` would pass that struct as a parameter, which then would be lowered to equivalent behavior as the set of three parameters, `i64, i64, i8`.

jakobbotsch: As I understand it, the algorithm you are describing here is part of the ABI: https://github.com/apple/swift/blob/d1d9fd1a2e478189e6eec7c48a0b952d9063859b/docs/ABI/CallingConvention.rst#L926-L993
It sounds like we are going to end up with ABI specific handling within the projection tooling regardless of whether we handle structs within the runtime or not. If that is the case, should we go with the tried-and-tested LLVM approach until we have a good grasp around the exact details of this handling and are confident that it all maps reasonably well to structs that can be described in IL? Are we already confident enough about this?

jakobbotsch: My understanding now is that we are going to be running the first part of the ABI twice. We need to run it in projection tooling regardless, because it is necessary for types that cannot be represented in IL. It is going to result in a struct of primitives. We are then also running the algorithm in the runtime, once more, and hoping that the results of running the first part of the algorithm twice is the same as what Swift+LLVM end up implementing.

What is the benefit of this compared to doing what Swift+LLVM does and avoiding wrapping the primitive sequence in a struct unless necessary (returns)?
I can see one benefit, which is that Swift types that are directly representable in C# can be defined in C# and used directly in interop. However, because there are Swift types that cannot be represented in C# the general guidance is always going to be to use the projection tooling.

Am I understanding this correctly? Does the diversion from Swift+LLVM make sense in this light?

jkoritzinsky: Here's another example that I'm going to reference below: https://godbolt.org/z/Y8Yxvdc3W

Here's how I view it:

The projection will handle the struct/enum layout. So it will lower `X` in the example above to something like:

```csharp
[LayoutKind(Explicit)]
 struct X
{
    [FieldOffset(0)]
    private Foo f;
    
    private struct B
    {
         private Bar b;
         private int i;
         private int descriminator;
    }

    [FieldOffset(0)]
    private B b;
}
```

The projection layer does not need to lower X to a primitive type sequence, it just needs to determine the layout that Swift uses to represent each case and the descriminator.

Then the JIT/VM would handle lowering the `X` struct to a primitive type sequence.

Basically, the CallConvSwift signatures will always be able to use named types like Swift, and the JIT will handle all of the primitive type sequence logic and the register allocation logic in a combined pass that better fits RyuJITs architecture.

jakobbotsch: > Then the JIT/VM would handle lowering the X struct to a primitive type sequence.

How exactly would it do this? Are you confident that the intermediate results during the ABI handling of enums can always be described with structs in this way, and that running the "primitive type sequencing" algorithm on these structs will result in the right thing? Or are we expecting that we are going to reconstruct the Swift source of truth on the runtime side and trying to give all Swift types an IL representation that the runtime knows how to parse?

jkoritzinsky: I am very confident that with the struct layout mechanisms that exist in .NET, we can construct a C# struct type with a matching layout for any frozen enum type from a Swift source of truth, especially since 32-bit targets don't need to support Swift interop, and that we can build the tooling in a way that the VM/JIT's primitive type sequencing algorithm will end up with the same results.

proposed/swift-interop.md
```diff
@@ -118,7 +118,15 @@ When calling a function that returns an opaque struct, the Swift ABI always requ
...
+
+The projection tooling will provide blittable representations of any projected frozen struct or enum types, at minimum for any of these types that are passed by value to any Swift APIs, to support the above ABI handling. These blittable types are not required to be lowered to primitive types; they may have fields of other blittable representations of other Swift types or other blittable structs.  These blittable representations will be the "struct or enum type layouts" mentioned in the paragraph above.
+
+##### SIMD Types
+
+We will pass the `System.Runtime.Intrinsics.VectorX<T>` types in SIMD registers as we do with the managed calling convention. We will treat the `Vector2/3/4` types as non-SIMD types as we will not project any Swift SIMD types to these types (see the SIMD Types section in the projection design).
```
jkotas, line 127:
```suggestion
We will pass the `System.Runtime.Intrinsics.VectorX<T>` types in SIMD registers as we do with the default unmanaged calling convention. We will treat the `Vector2/3/4` types as non-SIMD types as we will not project any Swift SIMD types to these types (see the SIMD Types section in the projection design).
```
Managed calling convention is our implementation detail, it is not strictly set in stone.

jkoritzinsky: I want to rephrase it slightly as we currently don't pass these types correctly with the default unmanaged calling convention (which is why we block them).

```suggestion
We will pass the `System.Runtime.Intrinsics.VectorX<T>` types in SIMD registers in the same way the default unmanaged calling convention specifies. We will treat the `Vector2/3/4` types as non-SIMD types as we will not project any Swift SIMD types to these types (see the SIMD Types section in the projection design).
```

PR #315, opened 2024-02-21
Adding in folder for Tensor<T> info

michaelgsharp:
Adding in a folder under proposed for all the Tensor<T> information. This folder for now contains the type/memory comparisons for tensors in different frameworks. More documents will be added to this folder in the future.

### DIFF
--- a/proposed/tensor/type-and-memory-comparison.md
+++ b/proposed/tensor/type-and-memory-comparison.md
@@ -1,0 +1,52 @@
+
+| | libTorch | ONNX | Numpy | TensorFlow | Julia |
+| --- | --- | --- | --- | --- | --- |
+| **Types Supported** | | | | | |
+|  |  |  |  |  |
+| byte | X | X | | X | X | X |
+| sbyte | X | X | X | X | X |
+| ushort | X | X | X | X | X |
+| short | X | X | X | X | X |
+| int | X | X | X | X | X |
+| uint | X | X | X | X | X |
+| long | X | X | X | X | X |
+| ulong | X | X | X | X | X |
+| bool (byte) | X | X | X | X | X |
+| string (bytes) |  | X | X | X
+| string (unicode) |  |  | X | 
+| string (c++ std::string) |  | X |  | 
+| float | X | X | X | X | X |
+| float16 | X | X | X | X | X |
+| bfloat16 | X | X |  | X |
+| double | X | X | X | X | X |
+| complex64 | X |  | X | X | X |
+| complex128 | X |  | X | X
+| long double (either 96 or 128 bits) |  |  | X | X |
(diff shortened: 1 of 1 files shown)


stephentoub (approved):

NiklasGustafsson (approved):

tannergooding (changes requested):
The proposed folder isn't really used anymore, instead things are meant to go directly into the `accepted` folder (the PR being merged is what makes it "accepted") and is meant to follow the general folder structure.

Thus, this would go under `accepted/2024/tensors` instead.

CC. @terrajobst

tannergooding:
Correspondingly, you should also follow the general instructions under https://github.com/dotnet/designs/tree/main?tab=readme-ov-file#creating-a-proposal, to ensure it doesn't conflict with the design doc I'm working on and that it is correctly indexed

proposed/tensor/type-and-memory-comparison.md
```diff
@@ -0,0 +1,52 @@
...
+| byte | X | X | | X | X | X |
+| sbyte | X | X | X | X | X |
+| ushort | X | X | X | X | X |
+| short | X | X | X | X | X |
+| int | X | X | X | X | X |
+| uint | X | X | X | X | X |
+| long | X | X | X | X | X |
+| ulong | X | X | X | X | X |
+| bool (byte) | X | X | X | X | X |
+| string (bytes) |  | X | X | X
```
eiriktsarpalis, line 15: @NiklasGustafsson @tannergooding is using tensors with string elements common? It seems like it might be difficult to reconcile with a tensor type that has a generic number constraint.

proposed/tensor/type-and-memory-comparison.md
```diff
@@ -0,0 +1,52 @@
...
+| Handle to mutable resource |  |  |  | X |  |
+|  |  |  |  |  |
+| **Other** |  |  |  |  |  |
+|  |  |  |  |  |
+| Supports Contiguous Memory | X | X | X | X |  |
+| Supports Non-Contiguous Memory |  |  | X | X |  |
+| Supports Sparse Tensors | X |  | X | X |  |
+
+
+ * Both Numpy and TensorFlow support Non-Contiguous memory, but its often less performant
```
eiriktsarpalis, line 49:
```suggestion
 * Both Numpy and TensorFlow support Non-Contiguous memory, but it's often less performant
```

proposed/tensor/type-and-memory-comparison.md
eiriktsarpalis: Same concern applies to a few other types in this list.

michaelgsharp:
Closing as the info is being used directly in Tanner's PR.

PR #317, opened 2024-05-15
Add design document for issuing a warning when targeting netstandard1.x

terrajobst:
We'd like to issue a warning when building a project targeting .NET Standard 1.x and recommend the user to upgrade to .NET Standard 2.0.

The reason being that some developers still target .NET Standard 1.x even though it doesn't buy them any additional reach as all in supported flavors of .NET all support .NET Standard 2.0 or later.

The goal of this warning is allow people to set their projects more up for success; that is, have good reach while also being able to have a sizeable feature set (both in terms of framework APIs as well as NuGet packages they can reference).

Please note that this will not issue a warning for .NET Standard 2.0/2.1, nor are we changing anything around versions of .NET Standard can be built or referenced.

### DIFF
--- a/accepted/2024/net-standard-recommendation.md
+++ b/accepted/2024/net-standard-recommendation.md
@@ -96,6 +96,51 @@ The .NET SDK will check the `TargetFramework` property. If it uses
 
 ## Q & A
 
+## What is the downside of targeting .NET Standard 1.x?
+
+There is a major difference in how tooling works for .NET Standard 1.x vs 2.x:
+
+* .NET Standard 1.x was based on packages, that is, if you target .NET Standard
+  1.x you end up referencing about ~70-100 packages (depending on whether you
+  also target a specific runtime identifier). Those are also persisted in the
+  resulting NuGet package which causes consumers having to restore them as well.
+
+* .NET Standard 2.x doesn't use packages; the resulting NuGet package doesn't
+  depend on any packages, just like targeting specific frameworks generally
+  doesn't.
+
+In principle we could change the way tooling works for .NET Standard 1.x but we
+don't really see the value for two reasons:
+
+1. The additional platforms that .NET Standard 1.x reaches are mostly out of    
+   support.
+
+2. [.NET Standard itself is no longer updated][net-standard-future]. We believe
+   the new OS-specific frameworks starting with .NET 5 are a much easier and
+   better model therefore supersede .NET Standard.
+
+If you target .NET Standard 1.x you're limiting yourself to a subset of .NET
+Framework 4.5, which shipped about 12 years ago. A lot of innovation has
+happened since then that you're missing out on. If you need to support older
+frameworks that Microsoft doesn't support, this might be an intentional choice
+in which case suppressing the warning makes total sense.
+
+However, we believe the vast majority of people don't have such a requirement
+and would be better off targeting .NET Standard 2.0 or .NET 5+ instead.
+
+## Should I drop .NET Standard altogether?
+
+If you need to produce a NuGet package that works on .NET Framework as well as
+modern .NET flavors then using .NET Standard 2.0 is still the best solution for
+you -- and your consumers.
+
+If you drop .NET Standard and instead just multi-target for .NET Framework and
+.NET 5+, you force your consumers to do the same, which is generally not
+necessary and thus very much undesirable.
+
+However, if you don't have to support .NET Framework then you're likely better
+off just targeting .NET 5+.
+
 ### Didn't you promise that .NET Standard is supported forever?
 
 We promised that future .NET implementations will continue to support
--- a/accepted/2024/net-standard-recommendation.md
+++ b/accepted/2024/net-standard-recommendation.md
@@ -1,0 +1,128 @@
+# .NET Standard Targeting Recommendations
+
+**Owner** [Immo Landwerth](https://github.com/terrajobst) | [Viktor Hofer](https://github.com/ViktorHofer)
+
+We have already said that there is [no new version of .NET
+Standard][net-standard-future] and that the replacement moving forward is to
+just target the platform-neutral `netX.Y` [base framework][tfm] (with `x.y >=
+5.0`) of .NET Core.
+
+For customers that need to author code that still needs to be consumed by .NET
+Framework, the recommendation is to continue using `netstandard2.0`. There is
+very little reason to target .NET Standard 2.1 because you lose .NET Framework
+while only gaining very little few additional APIs. And if you don't care about
+.NET Framework support, then just target .NET Core directly; at least then you
+get access to a lot more functionality, in exchange for the reduced reach.
+
+There is basically no reason to target .NET Standard 1.x any more as all .NET
+implementations that are still supported, can reference .NET Standard 2.0
+libraries.
+
+However, we have found that many customers still target .NET Standard 1.x. We
+found out the reasons boil down to one of two things:
+
+1. Targeting the lowest possible versions is considered better
+2. It was like and there is no reason to change
+
+The first argument has merit in a world where .NET Standard continues to produce
+new versions. Since that's no longer the case you might as well update to a
+later version that allows you to use more functionality. We found the biggest
+jump in productivity is with .NET Standard 2.0 because it also supports
+[referencing .NET Framework libraries][netfx-compate-mode], which is useful in
+cases where you modernize existing applications and have some mixture until you
+can upgrade all the projects.
+
+The second reason is always valid; however, in cases when you're actively
+maintaining a code base it makes sense to upgrade to avoid missing out on
+features that could have made your life easier and you didn't realize exist
+because code completion simply didn't advertise them.
+
+## Scenarios and User Experience
+
+### Building a project targeting .NET Standard
+
+Jackie is assigned a work item to add a feature that requires writing code in
+the shared business logic project. That project happens to target .NET Standard
+1.4. She doesn't know why because the project was built before her time on the
+team.  The fist time she compiles in Visual Studio she sees a warning:
+
+> warning NETSDK1234: Contoso.BusinessLogic.csproj: Targeting .NET Standard
+> prior to 2.0 is no longer recommended. See \<URL\> for more details.
+
+Following the link, she finds a page that inform her that all supported .NET
+implementations allow using .NET Standard 2.0 or higher. She asks one of her
+colleagues and she remembers that they used .NET Standard 1.4 because at some
+point her team also maintained a Windows Phone app (which, to everybody's
+disappointment no longer exists).
+
+Jack decides to upgrade the project to .NET Standard 2.0. She later realizes
+that this was a good idea because it enables her to use many of the
+`Microsoft.Extensions` libraries that make her life a lot easier, specifically
+configuration and DI.
+
+## Requirements
+
+### Goals
+
+* Issue a warning when *building* a project targeting `netstandard1.x` with the
+  .NET 9 SDK
+
+### Non-Goals
+
+* Issue a warning when building a project targeting `netstandard2.x`
+* Issue a warning when consuming a library that was built for .NET Standard 1.x.
+* Issue a warning building with a .NET SDK prior to 9.0
+
+## Stakeholders and Reviewers
+
+* Libraries team
+* SDK team
+* Build team
+* NuGet team
+
+## Design
+
+The .NET SDK will check the `TargetFramework` property. If it uses
+`netstandardX.Y` with `X.Y < 2.0` it will produce the following warning:
+
+> Targeting .NET Standard prior to 2.0 is no longer recommended. See \<URL\> for
+> more details.
+
+* The warning should have a diagnostic ID (prefixed `NETSDK` followed by a four
+  digit number).
+* The warning should be suppressable via the `NoWarn` property.
+* The URL should be point to the documented ID [here][sdk-errors] (not sure why
+  it says `sdk-errors` -- I don't believe we have a section for warnings).
+
+## Q & A
+
+### Didn't you promise that .NET Standard is supported forever?
+
+We promised that future .NET implementations will continue to support
+*consuming* libraries built for .NET Standard, all the way back to 1.0. That is
+different from promising that we'll support *producing* .NET Standard binaries
+forever; that would only make sense if .NET Standard is still the best way to
+build reusable libraries. And [since .NET 5 and the platform-specific
+flavors][tfm] of .NET we think there is now a better way to do so.
+
+The original promise around consumption still stands: existing libraries built
+for .NET Standard 1.0 can still be consumed and we have no plans on changing
+that. So any existing investment in building a library that can be consumed in
+as many places as possible still carries forward.
+
+### Can I continue to build libraries targeting .NET Standard 1.x?
+
+Yes. There *may* come a time where the .NET SDK stops supporting building .NET
+Standard libraries, but we currently have no plans on doing so as the targeting
+packs for .NET Standard 1.x aren't bundled with the SDK and are only downloaded
+on demand already; thus, the benefit to removing the support for building .NET
+Standard 1.x seems marginal.
+
+We believe our ecosystem is better off targeting at least .NET Standard 2.0
+which is why we want to issue a warning, but the motivation isn't in us being
+able to remove the support for building.
+
+[net-standard-future]: https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/
+[tfm]: https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md
+[netfx-compate-mode]: https://learn.microsoft.com/en-us/dotnet/core/porting/#net-framework-compatibility-mode
+[sdk-errors]: https://learn.microsoft.com/en-us/dotnet/core/tools/sdk-errors/
--- a/accepted/2024/net-standard-recommendation.md
+++ b/accepted/2024/net-standard-recommendation.md
@@ -55,7 +55,7 @@ colleagues and she remembers that they used .NET Standard 1.4 because at some
 point her team also maintained a Windows Phone app (which, to everybody's
 disappointment no longer exists).
 
-Jack decides to upgrade the project to .NET Standard 2.0. She later realizes
+Jackie decides to upgrade the project to .NET Standard 2.0. She later realizes
 that this was a good idea because it enables her to use many of the
 `Microsoft.Extensions` libraries that make her life a lot easier, specifically
 configuration and DI.
--- a/accepted/2024/net-standard-recommendation.md
+++ b/accepted/2024/net-standard-recommendation.md
@@ -22,7 +22,7 @@ However, we have found that many customers still target .NET Standard 1.x. We
 found out the reasons boil down to one of two things:
 
 1. Targeting the lowest possible versions is considered better
-2. It was like and there is no reason to change
+2. It was already like that and there is no reason to change
 
 The first argument has merit in a world where .NET Standard continues to produce
 new versions. Since that's no longer the case you might as well update to a
@@ -44,7 +44,7 @@ because code completion simply didn't advertise them.
 Jackie is assigned a work item to add a feature that requires writing code in
 the shared business logic project. That project happens to target .NET Standard
 1.4. She doesn't know why because the project was built before her time on the
-team.  The fist time she compiles in Visual Studio she sees a warning:
+team.  The first time she compiles in Visual Studio she sees a warning:
 
 > warning NETSDK1234: Contoso.BusinessLogic.csproj: Targeting .NET Standard
 > prior to 2.0 is no longer recommended. See \<URL\> for more details.
--- a/INDEX.md
+++ b/INDEX.md
@@ -85,6 +85,7 @@ Use update-index to regenerate it:
 | 2023 | [Experimental APIs](accepted/2023/preview-apis/preview-apis.md) | [Immo Landwerth](https://github.com/terrjobst) |
 | 2023 | [Multi-threading on a browser](accepted/2023/wasm-browser-threads.md) | [Pavel Savara](https://github.com/pavelsavara) |
 | 2023 | [net8.0-browser TFM for applications running in the browser](accepted/2023/net8.0-browser-tfm.md) | [Javier Calvarro](https://github.com/javiercn) |
+| 2024 | [.NET Standard Targeting Recommendations](accepted/2024/net-standard-recommendation.md) | [Immo Landwerth](https://github.com/terrajobst), [Viktor Hofer](https://github.com/ViktorHofer) |
 
 ## Drafts
 


accepted/2024/net-standard-recommendation.md
```diff
@@ -0,0 +1,128 @@
+# .NET Standard Targeting Recommendations
+
+**Owner** [Immo Landwerth](https://github.com/terrajobst) | [Viktor Hofer](https://github.com/ViktorHofer)
+
+We have already said that there is [no new version of .NET
+Standard][net-standard-future] and that the replacement moving forward is to
+just target the platform-neutral `netX.Y` [base framework][tfm] (with `x.y >=
+5.0`) of .NET Core.
```
WeihanLi, line 8: Maybe better using `.NET` since we starting use `.NET` since .NET 5?
```suggestion
5.0`) of .NET.
```

accepted/2024/net-standard-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+because code completion simply didn't advertise them.
+
+## Scenarios and User Experience
+
+### Building a project targeting .NET Standard
+
+Jackie is assigned a work item to add a feature that requires writing code in
+the shared business logic project. That project happens to target .NET Standard
+1.4. She doesn't know why because the project was built before her time on the
+team.  The fist time she compiles in Visual Studio she sees a warning:
```
WeihanLi, line 47:
```suggestion
team.  The first time she compiles in Visual Studio she sees a warning:
```

terrajobst:
> Can also issue a warning when the target framework is 2.0/2.1? Because even 2.0/2.1 has many additional apis that are not available, a multi target framework is recommended.
> 
> Just as SqlClient has [removed support for netstandard 2.0/2.1,](https://github.com/dotnet/SqlClient/pull/2386) I believe there will be more library removal support in the future.

Do you mean .NET Standard 2.0/2.1? Then the answer would be no, we don't want to do this yet as .NET Framework is still supported and .NET Standard 2.0 is the only way to build a single binary that can be consumed by both .NET Framework and .NET Core.

If you mean .NET Core 2.x, we already issue a warning when using frameworks that are out of support, like so:

> warning NETSDK1138: The target framework 'netcoreapp2.0' is out of support and will not receive security updates in the future. Please refer to https://aka.ms/dotnet-core-support for more information about the support policy.

WeihanLi:
@terrajobst speaking of the dotnet framework support, I have a question, to support the dotnet framework for a new package, do we recommend using the `netstandard2.0` target or a framework target like `net462`?

terrajobst:
> @terrajobst speaking of the dotnet framework support, I have a question, to support the dotnet framework for a new package, do we recommend using the `netstandard2.0` target or a framework target like `net462`?

I generally recommend .NET Standard 2.0 for code that you want to work everywhere. If you don't use `<TargetFramework>netstandard2.0</TargetFramework>` and instead use multi-targeting (say `<TargetFrameworks>net462;net6.0</TargetFrameworks>`) then all your consumers also need to multi-target, which is extremely annoying and prone to errors.

dotMorten:
I don’t really see the point. If I don’t need anything netstd2 offers me there’s no harm in targeting an older version. It’ll still work with newer versions. A warning seems to indicate I’m doing something wrong/bad but I can’t think of a single bad thing here (apart from of course the smaller api surface which is expected from using an older version).

> The goal of this warning is allow people to set their projects more up for success; that is, have good reach

I don’t get how increasing my minimum will create better reach? It’ll only create less (many people are still stuck on older no longer supported targets that I might want to support even if Microsoft don’t). This doesn’t set me up for greater success - only more warnings and potentially just doing it even though I don’t need the extra APIs and not realizing I cut a user off who relied on older versions

accepted/2024/net-standard-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+> warning NETSDK1234: Contoso.BusinessLogic.csproj: Targeting .NET Standard
+> prior to 2.0 is no longer recommended. See \<URL\> for more details.
+
+Following the link, she finds a page that inform her that all supported .NET
+implementations allow using .NET Standard 2.0 or higher. She asks one of her
+colleagues and she remembers that they used .NET Standard 1.4 because at some
+point her team also maintained a Windows Phone app (which, to everybody's
+disappointment no longer exists).
+
+Jack decides to upgrade the project to .NET Standard 2.0. She later realizes
```
yaakov-h, line 58:
```suggestion
Jackie decides to upgrade the project to .NET Standard 2.0. She later realizes
```

JamesNK:
The problem with netstandard1.x if you need to bring in a bunch of nuget packages. e.g. https://github.com/JamesNK/Newtonsoft.Json/blob/2eaa475f8853f2b01ac5591421dcabd7a44f79ce/Src/Newtonsoft.Json/Newtonsoft.Json.csproj#L73-L90

It's fiddly, and annoying. Meanwhile netstandard2.0 just works. Yes, you get more reach with 1.x, but it's a pain to create a package that has 1.x targets, and netstandard2.0 gets you the targets you want anyway: modern net4.x and modern net5+.

> I don’t get how increasing my minimum will create better reach? It’ll only create less (many people are still stuck on older no longer supported targets that I might want to support even if Microsoft don’t).

Who are you trying to help by having netstandard1.x? I've said many times in the past that I wish NuGet gave states about which target in a NuGet package was being used. Unfortunately we don't have that. What I can say anecdotally is that when I dropped portable class library targets (remember those?) from Newtonsoft.Json, not a single person complained.

dotMorten:
> The problem with netstandard1.x if you need to bring in a bunch of nuget packages. 

Only if you need those APIs. That's one reason why you'd might want to move up to netstandard2.0, but my argument was: If I don't need the extended API surface, it seems more prudent to only target as low a version as I need. Giving me a warning because I don't need the full netstandard2.0 API surface seem silly.

> it's a pain to create a package that has 1.x targets

I don't get this. There's no practical difference in creating packages targeting 1.x vs 2.0.

JamesNK:
You need to figure out the right packages you need for the APIs you want to reference. It's not extremely hard, just annoying. IMO the juice isn't worth the squeeze. Another negative is netstandard1.x creates a larger NuGet dependency graph. They're slower to build (someone correct me if I'm wrong here)

Remember that this would just be a warning saying that you - a developer creating a library NuGet package - shouldn't need to do this because 1.x only platforms are out of support. If you want to keep making netstandard1.x packages, then you can ignore the warning.

I see this as a positive step to move the ecosystem forward by removing no longer necessary history.

dotMorten:
Are you officially deprecating netstandard 1.x ?
Since it's just a minimum list of APIs required for someone to implement to be vX.Y compliant, it just feels odd to be deprecating that list, because the currently known implementers of the various Microsoft .NET runtimes have all moved on.
How does this affect other .NET implementations, like Mono, Nano Framework, Micro Framework, Unity3D etc?

JamesNK:
I'm not doing anything other than providing my 2 cents to Immo's proposal.

Immo can say for sure, but it sounds like this is just adding a compiler warning and can probably be suppressed.

MichalStrehovsky:
> If I don’t need anything netstd2 offers me there’s no harm in targeting an older version

Just to give a concrete example of real harm when targeting older version, my favorite "project is targeting NetStandard < 2" bug:

There are [no guarantees around referential equality](https://github.com/dotnet/runtime/issues/17142) of `MemberInfo`s (except for runtime provided `Type`). NetStandard < 2 doesn't have `operator==` on `MemberInfo`, so any comparison is referential comparison. Because of this, comparisons involving `MemberInfo` will often end up [not equal on UWP](https://github.com/octokit/octokit.graphql.net/issues/262) (or Native AOT too). It's a real mystery bug unless you know to expect it. Recompiling for netstandard2.0 fixes those issues.

There are probably other examples where better overloads of various methods became available in netstandard2.0. Compiling for netstandard2.0 actually produces a different assembly, not just a different version of assembly references.

Perksey:
Ultimately .NET Standard isn’t a framework unto itself, it’s an API contract. I agree with @dotMorten’s point. Consider cases like OpenGL, which is also an API contract with many implementations, generally you only target the minimum version for the functionality you actually need. It is true that most people target NS 2.0 by default because generally this is the path of least resistance for the users looking to build for all .NET implementations, but this needn’t detract from the contract vs framework argument.

MichalStrehovsky:
> Consider cases like OpenGL, which is also an API contract with many implementations, generally you only target the minimum version for the functionality you actually need

OpenGL is a C API and C doesn't have function overloading. Merely changing the version of the API contract will not change what method is called, it just makes a different set of methods available. In .NET, changing the version of the API contract can and will change what method is called due to overloading. Usually the compiler will select a better/more specialized version of the method. Targeting a lower version of the contract when it's not actually necessary to go that low produces a worse assembly not just because there are fewer APIs to call consciously, but also fewer APIs for the compiler to bind to. It can even [lead to bugs](https://github.com/dotnet/designs/pull/317#issuecomment-2113989817) that got fixed later.

huoyaoyuan:
> According to the [documentation](https://github.com/dotnet/standard/blob/release/3.0/docs/versions.md),. Net standard version 1.4 to 2.0, the corresponding. Net framework version is 4.6.1, and. Net framework 4.6.1 is no longer supported.

@Varorbc A correction for this: the .NET Standard versions do not *correspond* to .NET Framework 4.6.1, they are supported by .NET Framework 4.6.1 **and above**. The must-to-support is the *highest .NET Standard version* that is supported by *lowest supported .NET Framework version*. In this case, the highest supported .NET Standard version for .NET Framework 4.6.2-4.8.1 is all .NET Standard 2.0.

LuisApplivity:
Both sides of the argument make valid points: As a STANDARD, it makes no sense to deprecate and/or issue a warning against a subset of an otherwise-encouraged standard (.NET Standard 2.0). So props to @dotMorten and @Perksey on that.

On the other hand, and this is 100% Microsoft's fault for deliberately obfuscating/removing references and explanations in older documentation over time, there are the facts that both @JamesNK and @MichalStrehovsky alluded to, which is that .NET Standard 1.x is NOT purely a subset of .NET Standard 2.0 (and 2.1). The inner workings are different. This is also part of why .NET Standard 2.0 was NOT called .NET Standard 1.7, and the final 1.x version remained at 1.6. There are various, UN-intuitive under-the-hood changes in how they function, in what exactly they support, and so on.

In my personal view, I wholeheartedly agree .NET Standard 2.0 should be favored, for the developer's sake, BUT there is absolutely NO reason working .NET Standard 1.x projects should be nagged with any form of warning, so I ultimately disagree on @terrajobst's proposal. **Because all supported versions of .NET implement .NET Standard 1.x, no version of .NET Standard should ever be deprecated**, nor have warnings issued against. Calling it a "standard" would become a lie if that happens.

There is one more reason for my disagreement, which is that **both .NET Standard 2.0 and 2.1 LACK a few features that are exclusive to .NET Standard 1.5 and 1.6**, but missing in 2.0 and 2.1, meaning those working 1.5/1.6 projects have a **risk** of **breaking** when moving "up" to 2.0 or 2.1, for absolutely no benefit. Admittedly, these features are very few, and rarely ever used, but are ultimately there. The caveats mentioned by @JamesNK and @MichalStrehovsky will not matter for robust, mature, working projects that need not be touched. That being said, I do agree that, when/if possible, 1.5/1.6 projects should be moved either down to 1.3 (better reach), or up to 2.0 (better almost-everything), to avoid these types of confusion and special caveats.

As a side bonus, .NET Standard 1.3 and lower are implemented by .NET Framework 4.6, which is the latest .NET to run on Windows Vista, even if they are no longer supported. Still worth mentioning. @JamesNK is right, though, in that usually people using such will also not care if a newer Package version no longer supports their peculiar use-case etc. (I miss PCL by the way! You can count me as your first "complaint"! 😉 What if I want to target the Xbox 360 with a reusable class library? It still plays games in 2024!).

huoyaoyuan:
> which is that .NET Standard 1.x is NOT purely a subset of .NET Standard 2.0 (and 2.1). The inner workings are different. This is also part of why .NET Standard 2.0 was NOT called .NET Standard 1.7, and the final 1.x version remained at 1.6.

This is not true. @MichalStrehovsky 's example is more about how *adding* stuff can make a change. It is a strict superset. The major change of .NET Standard 2.0 is *how things are assembled together*.

> As a side bonus, .NET Standard 1.3 and lower are implemented by .NET Framework 4.6, which is the latest .NET to run on Windows Vista, even if they are no longer supported.

Worth considering that the SDK doesn't issue a warning when targeting unsupported `net46`, unlike lower versions of .NET Core.

LuisApplivity:
> > which is that .NET Standard 1.x is NOT purely a subset of .NET Standard 2.0 (and 2.1). The inner workings are different. This is also part of why .NET Standard 2.0 was NOT called .NET Standard 1.7, and the final 1.x version remained at 1.6.
> 
> This is not true. @MichalStrehovsky 's example is more about how _adding_ stuff can make a change. It is a strict superset. The major change of .NET Standard 2.0 is _how things are assembled together_.

Wrong, it is **not** a **strict** superset, a very tiny number of APIs are exclusive to 1.5 and 1.6. There was a "rollback" with .NET Standard 1.5, because it originally supported .NET Framework 4.6.2, but NOT 4.6.1, because some of its APIs were not available to 4.6.1. This was later deemed as a mistake by Microsoft, since 4.6.1 was very active at the time, and it was desired for 4.6.1 to consume .NET Standard 1.5+ packages.

This is also why it was often recommended to jump from .NET Standard 1.4 straight to 2.0 at the time, and skip 1.5 and 1.6 entirely.

Nowadays that info is seriously buried/deleted, but here are some links I still happen to have containing some artifacts of this:

https://web.archive.org/web/20170817143725/https://docs.microsoft.com/en-us/dotnet/standard/net-standard

https://web.archive.org/web/20160416214114/https://github.com/dotnet/corefx/blob/master/Documentation/architecture/net-platform-standard.md

Finally, I brought up what @MichalStrehovsky said in the context that there are little known, unintuitive differences between .NET Standard 1.x and 2.x.

MichalStrehovsky:
> Finally, I brought up what @MichalStrehovsky said in the context that there are little known, unintuitive differences between .NET Standard 1.x and 2.x.

To be clear, the differences I wrote about are pretty much:

```csharp
class C
{
    // Exists in NetStandard X
    public static void Do(object x);

    // Added in NetStandard X + 1
    public static void Do(int x);
}
```

The code will compile in both NetStandard X and X + 1, but if compiled against NetStandard X + 1, `Do(42)` will call the more specialized and efficient overload. This will always be the case. More efficient overloads get added to the framework all the time and just bumping the target framework lets one pick up the benefit for free, without any edits to the C# source file. Sticking to an old one is compatible with old runtimes, but comes at a cost.

LuisApplivity:
> [...] There was a "rollback" with .NET Standard 1.5, [...]

Minor correction, the "rollback" was with .NET Standard 2.0, in that it is a 100% API superset of .NET Standard 1.4, but left behind a very small number of .NET Standard 1.5 and 1.6 APIs behind.

I wish someone would find the exact Microsoft (GitHub?) link/discussion that addressed this. There was even mention that this decision was made, because they believed that the few 1.5/1.6-exclusive features would impact only a very small amount of existing applications.

slang25:
A reason I try to avoid netstandard < 2.0 is that I get an explosion of packages. Am I remembering that right? For me this is reason enough to target ns2.0 even though I'm only using a subset.

accepted/2024/net-standard-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+* SDK team
+* Build team
+* NuGet team
+
+## Design
+
+The .NET SDK will check the `TargetFramework` property. If it uses
+`netstandardX.Y` with `X.Y < 2.0` it will produce the following warning:
+
+> Targeting .NET Standard prior to 2.0 is no longer recommended. See \<URL\> for
```
JonDouglas, line 88: Does this need a URL? What about something like:

> .NET Standard 1.X is no longer recommended. Please upgrade to .NET Standard 2.0 or higher. For example: <TargetFramework>netstandard2.0</TargetFramework>

JonDouglas (approved):
🚢🐿️

accepted/2024/net-standard-recommendation.md
terrajobst: I think we'll want a URL to explain what this warning means and when it's OK to suppress; we've done this or all other `NETSDKXXXX` warnings/errors.

jzabroski:
Agree.

Can't believe anyone would waste finger movements typing reasons to still target ns1.x but

JamesNK (approved):

jzabroski:
It's easier to write poly fills when targeting netstandard2.0 than

```xml
<TargetFrameworks>net48;net8.0</TargetFrameworks>
```

I only use net48 for "startup projects" like console programs, desktop applications, web services, websites.

LuisApplivity:
@jzabroski Please read the discussion more carefully: These were **not** reasons to target ns1.x, they were reasons _not to issue a warning_. Completely different meaning.

While there are and will be _multiple good reasons_ to target it, as demonstrated above, we all seem to unanimously agree .ns2.0 is simply best in practice. However, as a subset, issuing a **warning** against ns1.x is simply senseless, even though that's exactly what's underway, as we can see.

jzabroski:
I did. In terms of SDK obsolescence etiquette, It is necessary but not sufficient to emit a warning discouraging target of pre-netstandard2.0.

By not supporting the warning, project developers need to pull in many dependencies, often with nuget semver package versions different than assembly versions, and debug assembly binding errors that mist developers are untrained and unskilled in how to debug. I have supported two very popular open source projects where at least fifty issues have been logged related to developer confusion over  targeting older versions of .NET Framework than 4.8. The reality is, Immo fixed this mess by halting iteration on netstandardX.Y altogether, and now he is closing this chapter completely for the best. I could go into depth discussing this and why wildcard transitive dependency resolution was a very bad idea in the first place, but it's a waste of time as nobody except Azure Functions users has submitted such issues to my projects since Immo cleaned all this up as part of .NET 5. Azure Functions is still a mess, but they can have their own little playground to mess around in.  People who genuinely care can use chiseled containers on Azure Functions now and bypass a lot of the messy stuff.

LuisApplivity:
> There were growing pains and we really only got the quirks ironed out with .NET Standard 2.0. So targeting 1.x has a lot of rough edges and is something that users should only do with vary careful consideration [...]

100% agreed.

> We have a multitude of links to documentation on this topic and some of the quirks for the various 1.x targets listed under https://learn.microsoft.com/en-us/dotnet/standard/net-standard

Unfortunately, that link doesn't contain anything to help others understand what is so different about targeting 1.x as opposed to 2.0 anywhere. All it contains is text recommending 2.0 and, at most, a footnote to point out .NET Framework 4.6.1 support is flakey, without specifying the actual reason (real reason was deleted long ago, as I showed earlier), which has to do with both 1.x and 2.0.

> but even then you're more likely better off simply multitargeting the out of support .NET Framework versions at that point.

Like .NET Framework **4.6** etc.? Agreed for most cases. However, keep in mind .NET Standard 1.x is still useful for **supported** versions as well, such as .NET Framework 4.8 and .NET 8, because it's often useful to engineer your code and patterns within a subset, such as .NET Standard 1.x, and have it enforced by targeting it, as a tunnel that helps avoid undue complexity. Similar arguments also apply to setting <langVersion>7.3</langVersion> for C#, and other such things.

In other words, people interested in .NET Standard 1.x are also those who **are** targeting **supported** runtimes, as well, not only out-of-support ones, and a warning is simply an unnecessary nag and hindrance. Let's not forget, this warning practice has **only** been done in the past for **unsupported runtimes such as earlier .NET Core**, but all supported runtimes such as .NET 8 and .NET Fr. 4.8 implement and will keep implementing .NET Standard 1.x (and thus also 2.0).

> You can always ignore the warning, you can always continue supporting any such customers on an older version of your package that they are known to be on, etc.

Until they remove it from a later version of Visual Studio as they did with some of the PCL (Portable Class Library) targets, and state the reason of removal as "we already were issuing a warning against it anyway". Then you need older VS versions, and other older tooling, but customer won't have or allow it, and so on. See the headache snowball? It starts with the size of a tiny pebble like this.

This also damages people's confidence in .NET Standard 2.0, and can/will be used as a precedent to warn/attack/remove it eventually, too, possibly some years later, despite being so much more useful than multi-targeting in most cases. It's the usual "cooking the frog" pattern, since ultimately Microsoft doesn't want multiple targets, but just their own tightly-controlled .NET (Core) platform, and only "accept" implementations that are forking from it from intimate partners, such as Samsung's Tizen.

It's good for Microsoft, but bad for people, the industry as a whole, and for developers in general. No benefit will come out of the issuing of warning. It's a bad practice. Relatively few people target .NET Standard as a whole, most don't even understand what it's about even to this day, let alone an older standard like 1.x, so the ones affected by the warning will be those people that **do** understand it and **will** have their reasons for using it.

jzabroski:
> it's often useful to engineer your code and patterns within a subset, such as .NET Standard 1.x, and have it enforced by targeting it, as a tunnel that helps avoid undue complexity

There are two kinds of complexity - internal and external. Newer versions of .NET generally simplify external complexity with respect to performance, while adding external complexity with respect to overloads (e.g., .NET 9 Span and variadic arguments improvements).

> This also damages people's confidence in .NET Standard 2.0, and can/will be used as a precedent to warn/attack/remove it eventually, too, possibly some years later, despite being so much more useful than multi-targeting in most cases.

Microsoft and its competitors are commercial entities and will do what is necessary to stay in business while making the majority of its customers happy. The idea of having a "minimum version" to support was something the Android SDK team originally tried.  As of 2021, Google set the global minimum version for new uploads to the Play Store to be _30_.

<details><summary>Details on Android Levels</summary>
<p>

Important to note, Android Levels are not compile-time type-checked like .NET Standard is.  The main problem with these Levels is they still require global compile-at-head behavior. Plus, after awhile, the regression suite required to backtest all the devices going back to the original "minimum version" is enormous, and very few developers in practice can actually reason about what the min-max API surface area even _looks like_, much less _predict_ what will actually happen. The problem is _so hard_ that Google built the world's most sophisticated buildchain tool, https://bazel.build/ , with explicit support for debugging and reporting transitive package references.  And people _still_ ship Play Store apps that rate 2 stars or lower.  But the main point here is that Google, with its uber fancy build tool that explicitly manages crazy dependencies, still wants developers to use _30_ as its minimum version number.  That is _years_ of API levels that are obsolete.  (I believe the Android decision coincided with ARM processor architecture adoption, allowing Mac developers to use Android Studio on an M1 computer.)

The only API that has not seen much deprecation that is widely used commercially is JavaScript in the browser.  There, there is no standard, and developers detect features and use progressive enhancement or graceful degradation or informal minimum version as enforced by either choice of stylesheeting technology or web development framework.  However, the behavior is relatively standardized now due to webkit engine being open source and all major browsers implementing it as the engine.

</p>
</details>

teo-tsirpanis:
> it's often useful to engineer your code and patterns within a subset, such as .NET Standard 1.x, and have it enforced by targeting it, as a tunnel that helps avoid undue complexity

I don't understand this argument. Complexity in a codebase originates from the overall structure of the code, not the specific external APIs it uses, or even is able to use. There are better ways to restrict yourself from using certain APIs if you want to, like the [banned API analyzer](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BannedApiAnalyzers/), team code review, or just self-discipline.

Besides the transitive dependency explosion, targeting .NET Standard 1.x limits you so much in both the inbox APIs and the third-party packages you can use, and combined with the fact that all frameworks supported by NS1.x and not NS2.0 are unsupported, I don't see any valid technical reason to use NS1.x in a maintained codebase.

LuisApplivity:
> I don't understand this argument. Complexity in a codebase originates from the overall structure of the code, not the specific external APIs it uses, or even is able to use. There are better ways to restrict yourself from using certain APIs if you want to, like the [banned API analyzer](https://www.nuget.org/packages/Microsoft.CodeAnalysis.BannedApiAnalyzers/), team code review, or just self-discipline.

You are definitely correct in that self-discipline, team code review and "banned API analyzer" all aid to this goal, but all those 3 have a cost, and can nonetheless be used **together** with NS1.x, rather than in its place.

For example, some team members won't have that much self-discipline, and/or can be relieved from some of it via the restriction. Setting up "banned API analyzer" costs some time to do, and also adds some complexity, it's not as straightforward and simple as just merely targeting NS1.x. Team code reviews are indispensable, but not free, as they eat up production time, in some workforces way more than others, however by keeping the API surface thinner, you simplify the code reviews, and remove spent time overhead by nature, and the longer the project, the greater the benefit of doing so. Some people even get fed up and leave certain companies due to excess regularity or time spent in code reviews, so the more streamlined it is, the more we also avoid this problem, so we get all kinds of real-world benefits by following a "less is more" principle.

That being said, thank you for pointing out "banned API analyzer", that's an excellent resource I hadn't yet come across that I will definitely make use of here on out.

> Besides the transitive dependency explosion, targeting .NET Standard 1.x limits you so much in both the inbox APIs and the third-party packages you can use, and combined with the fact that all frameworks supported by NS1.x and not NS2.0 are unsupported, I don't see any valid technical reason to use NS1.x in a maintained codebase.

Personally, despite everything, I like to target NS2.0 myself by default, not 1.x. That being said, that is **just me and you** (and others who do the same as us). By that, I mean to say other people choosing 1.x **will** have their reasons, we cannot anticipate all of them, the real-world market is incredibly varied with all kinds of use cases and individuals, and it's those individuals that should be making the calls, without any discouragement or warnings, because they understand their business and their environment as insiders, rather than us.

tannergooding:
> however by keeping the API surface thinner, you simplify the code reviews

This is very subjective and quite often not the case.

Minimizing the standard library surface area can theoretically make code reviews smaller, but in practice don't because it really comes down to a very simple steps of:
* Do I need to do 'x'?
  * If yes, does the BCL provide an API in box?
    * If yes, great, nothing further to do
    * If no, well now you have to go write your own implementation
  * If no, then an API existing or not existing doesn't matter

So in practice, there is very little that targeting a more minimal surface area gives. You might see slightly fewer items in IntelliSense, but that's really not going to be distinguishable for typical users since the bulk of the API surface is in namespaces like `System`, `System.Collections.Generic`, `System.IO`, `System.Linq`, `System.Net.Http`, `System.Threading`, and `System.Threading.Tasks` anyways (which are also the default using statements new files get access to by default).

Instead, what ends up happening is that you have less functionality in box so PRs end up being larger and needing more scrutiny due to having to polyfill all the bits that are missing. Then, when it actually comes to platform support you aren't really winning anything anyways.

For reference, the number of APIs available and supported target platforms is:
* NS1.0 - 7949/37118 APIs, .NET Core 1.0+, .NET Framework 4.5+, Mono 4.6+, Xamarin.iOS 10.0+, Xamarin.Mac 3.0+, Xamarin.Android 7.0+, UWP 8.0+, Unity 2018.1+
* NS1.1 - 10239/37118 APIs
* NS1.2 - 10285/37118 APIs, .NET Framework 4.5.1+, UWP 8.1+
* NS1.3 - 13122/37118 APIs, .NET Framework 4.6+, UWP 10.0+
* NS1.4 - 13140/37118 APIs, .NET Framework 4.6.1+
* NS1.5 - 13355/37118 APIs, UWP 10.0.16299+
* NS1.6 - 13501/37118 APIs
* NS2.0 - 32638/37118 APIs, .NET Core 2.0+, Mono 5.4+, Xamarin.iOS 10.14+, Xamarin.Mac 3.8+, Xamarin.Android 8.0+
* NS2.1 - 37118/37118 APIs, .NET Core 3.0+, Drops .NET Framework, Mono 6.4+, Xamarin.iOS 12.16+, Xamarin.Mac 5.16+, Xamarin.Android 10.+, Drops UWP, Unity 2021.2+

Where .NET 5+ expands this significantly and each subsequent version of .NET (6, 7, 8, and even 9 when it ships this November) is effectively a new standard baseline that has expanded upon this further. https://apisof.net/diff can be used to see what APIs were added or changed between different versions of .NET, which can be quite useful to find out what has been added or changed.

So given that, by targeting NS2.0 you're potentially missing out on `.NET Framework 4.5-4.6` support, legacy Mono/Xamarin, and `UWP 8.0-10.0` support (all of which are unsupported, no longer getting bug fixes, etc) and significantly restricting the surface area available thus requiring you to do more overall work. All while also giving a worse consumption experience to downstream consumers of your package, causing their deployment size to increase drastically due to the number of facade assemblies required, and reducing performance/startup/compilation time due to the number of additional references that exist.


> By that, I mean to say other people choosing 1.x will have their reasons

Indeed, and such users who believe they are in such a specialized scenario can suppress the warning. Ecosystems have a need to move forward and part of that includes surfacing diagnostics to users around things that are obsolete/deprecated and which may be problematic. Users have the freedom to ignore that warning, but they are then opting into the risks and downsides that come about from that.

jzabroski:
@LuisApplivity The key thing I caught you say was:

> Personally, despite everything, I like to target NS2.0 myself by default, not 1.x. That being said, that is just me and you (and others who do the same as us). By that, I mean to say other people choosing 1.x will have their reasons, we cannot anticipate all of them

Everyone who is an expert on .NET pretty much can _and does_ know the list of problems people would run into targeting 1.x and the lift addressing it - and it's pretty huge - think 10,000 person-hours of labor:

There would be a lot of actual work to make me _want_ to use .NET Standard 1.6 or lower.  Off the top of my head, assuming we ignore the polyfill challenges with Async Interfaces, and other libraries:

1. Microsoft Docs.  Figuring out what the lowest "standard" version a given API supports is time consuming and clicky-tappy.  It's doable, but if you select ".NET Standard 1.6" for an API that does not support .NET Standard 1.6, Microsoft Docs forwards you to the latest major version of (modern) .NET.
2. Libraries.  Joel Verhagan has written a nuget.org package analyzer (https://github.com/nuget/insights) and a [kusto query language](https://learn.microsoft.com/en-us/azure/data-explorer/external-azure-storage-tables-query) front-end to write custom queries - the reality is the incoming velocity to .NET Standard <=1.6 is _extremely low_.
3. In-box vs. out-of-box. With .NET Standard 2.0, there was a push to move more Microsoft-sponsored packages "in-box" (batteries included), so that you just needed the right TargetFramework and SDK and you would get access to a lot of packages without needing to declare dependency versions. With the introduction of tree-shaking (dead code elimination) for slim libraries, you get a whole bunch of libraries and then the compiler shakes out what you don't need.
4. Hosting and Host Builders. With .NET Core App 3.0, .NET 5, and .NET 6, there were progressive iterations to building application hosts to run executables, and while these are end-point concerns, normally a .NET Standard 1.6 library would write helpers for host builders as part of Microsoft DI support - and so targeting .NET Standard 1.6 and lower is a pain because you'd need to ship two sets of binaries for DI support.  It can be done, but it's double the binaries, double the documentation, and more than double the support requests/bugs.
5. Open Telemetry Support - Any modern big development budget app is going to want to have the benefit of distributed tracing and monitoring, and you don't get that with .NET Standard 1.6 or lower.  There would have to be backports not just from Microsoft, but key partners like Datadog.
6. Loss of performance optimizations - Insurmountable?  Some things in modern .NET just require modern technology to work the way users expect at scale - and unethical in an era of climate change to push users to inferior performance.  API Analyzers allow you to workaround this insurmountable obstacle by curtailing whatever APIs you don't want to use.  There are big companies like LL Bean that have used tools like NDepend to analyze APIs and build modular applications, even before API Analyzers existed.

For all these reasons and the ones already mentioned, nobody actually wants to target .NET Standard 1.6 or lower.  Really.  The ergonomics are terrible.

JonDouglas:
For what it is worth, .NETStandard 1.x is < 2%(all versions combined) of assemblies on NuGet.org. .NETStandard 2.0 is ~23% and .NETStandard 2.1 is ~2%. Just a quick query, but hopefully helps in this conversation.

In addition, people can use [NuGet.org's newish Search by TFM feature ](https://devblogs.microsoft.com/nuget/introducing-search-by-target-framework-on-nuget-org/) to help find specific packages w/ those assets.

.NET Standard 1.X:
image

.NET Standard 2.0:
image

.NET Standard 2.1:
image

LuisApplivity:
I believe we all made as many meaningful contributions to the main discussion as we could by this point, and hopefully it will assist future readers into making better-informed decisions.

On a mostly-unrelated note, and I emphasize such an offtopic matter is not being brought up by myself, but rather I'm simply **answering** it:

> [...] and unethical in an era of climate change to push users to inferior performance. [...]

Oh, please, spare us your hypocrisy: With each release of .NET, there are more and more platforms dropped and made incompatible. Windows 11 itself dropped support for a **gigantic** amount of hardware, encouraging it all to go to the landfill, and is not alone in this deed (look at Apple, Samsung etc.), so the more .NET goes "forward", the less eco-friendly it gets. You want eco-friendly .NET? Use Mono and MonoDevelop. Not to mention the ever-increasing bloat and telemetry, both in the newer OSes and newer .NET, which not only waste CPU cycles, but also is frequently used in a privacy-infringing matter. I'm not sure how you could have made your "climate change" remark any more ironic.

You cannot accurately put ".NET" and "ethical" in the same sentence.

You would also never want to waste CPU cycles on non-natively-compiled platforms. Unless all you do is AHT (ahead-of-time) compilation, try ANSI C89 / ISO C90 for actual true cross-platform, universal OS, compiler **and** hardware support. You cannot go more "green" than that. **Remember: 1. Reuse, 2. Repair, 3. Recycle, in this exact order.**

Furthermore, since we are already having this ridiculous discussion: as previously pointed out, an older OS like Windows Vista SP2 supports .NET Standard 1.3 maximum, and it, like Windows versions up to (non-updated) Windows 7 SP1, is many times more eco-friendly and ethical, because it doesn't have, out of Microsoft's own self-admittance, most of the keyloggers and other spyware/malware built into later versions of Windows itself, all of which make sure every single one of those CPU cycles are not only wasted, but used unethically and to our individual **and** collective detriment.

So by targeting .NET Standard 1.3, hey, guess what, you at least can use your Class Library project in one more OS that isn't as blatantly unethical and polluting as the only OSes modern .NET can target. For Windows XP support, even .NET Standard won't do, but PCL will, and so based on your line of argumentation, we should reinstate full PCL support in the next Visual Studio release. Because "climate change".

accepted/2024/net-standard-recommendation.md
```diff
@@ -0,0 +1,128 @@
...
+5.0`) of .NET Core.
+
+For customers that need to author code that still needs to be consumed by .NET
+Framework, the recommendation is to continue using `netstandard2.0`. There is
+very little reason to target .NET Standard 2.1 because you lose .NET Framework
+while only gaining very little few additional APIs. And if you don't care about
+.NET Framework support, then just target .NET Core directly; at least then you
+get access to a lot more functionality, in exchange for the reduced reach.
+
+There is basically no reason to target .NET Standard 1.x any more as all .NET
```
dsplaisted, line 17: I think it would be worth mentioning some of the drawbacks of targeting .NET Standard 1.x.  You get a LOT of packages downloaded/in your graph, 73 by my count without a RuntimeIdentifier specified, and 103 if you do specify a RuntimeIdentifier.

dsplaisted (approved):

accepted/2024/net-standard-recommendation.md
terrajobst: I believe I have addressed this in 83157fdf99fb4e472cb9ef024273837ee73054e1.

terrajobst:
Thanks everyone for your feedback!

I've added a few more explanations in 83157fdf99fb4e472cb9ef024273837ee73054e1 which addresses two major points:

1. What's the downside of targeting .NET Standard 1.x
2. Should you drop .NET Standard altogether.

I understand that some of you might not like the warning because you have requirements that make you keep targeting .NET Standard 1.x. That's why the warning is suppressible and just like calling OS-specific APIs, intentional suppression is very much part of the design. In end, our goal is tool the guidance that makes sense for the vast majority of our customers. Given complexities around most features, no guidance is going to be true in every case; I believe not provide tooling in those cases is a disservice to our community because it basically means a minority of advanced users holds the simpler experience hostage, which is generally not aligned with the design philosophy of .NET.

Hence, I'm going to merge this now.

PR #322, opened 2024-10-02
Add initial design document for parsing `net10`

terrajobst:


jaredpar (approved):

WeihanLi:
> currently, the SDK supports net8, would the SDK drop the support? Since net10 may conflict between net10.0 and netfx 1.0

This relates to my question before 

https://github.com/dotnet/sdk/issues/42920#issuecomment-2351296651

Since we're going to use `net10.0` in the previous design proposal, why not just treat the `net10` as the dotnet framework 1.0, and `net10.0` as dotnet 10, think it would also keep it simpler

https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#what-about-net-10

It would also introduce the complexity and inconsistency across SDKs versions

accepted/2024/net10.md
```diff
@@ -0,0 +1,62 @@
...
+Considering that .NET Framework 1.0 is out of support for a while and Visual
+Studio doesn't even support building for anything older than .NET Framework 2.0,
+it seems very doable to make `net10` mean .NET 10. The question is what happens
+with all the other frameworks, such as `net20`, `net452` etc. Quite a few of
+those are still supported and actively used.
+
+Here is the proposal:
+
+* Make `net10` mean .NET 10 and `net11` mean .NET 11 but keep all higher
+  versions as-is (`net20` still means .NET Framework 2.0).
```
jaredpar, line 49: Should we consider a warning for `net20` and ask that customers make it explicit by changing it to `net2.0`? Basically start getting ahead of this ambiguity.

terrajobst:
To be clear, `net10.0` already means .NET 10 and is the preferred syntax (because it's unambiguous).

This proposal is about how we should address the problem that some people will use `net10`, thinking it will mean .NET 10 when in fact it already means .NET Framework 1.0.

GrabYourPitchforks:
If even first-party tools msbuild and nuget don't handle this properly, shouldn't this sound the alarm that third-party ecosystem tools also won't handle it properly? It seems like us special-casing `net10` to try to make it work might end up causing harm by falsely lulling developers into believing that something works, when in reality it's potentially masking a failure elsewhere. Seems like this - not `net12` - might be the right time to make it an error. With an improved message, of course.

terrajobst:
> If even first-party tools msbuild and nuget don't handle this properly, shouldn't this sound the alarm that third-party ecosystem tools also won't handle it properly?

What do you mean by this? `net10` has been supported forever because that was the old syntax.

accepted/2024/net10.md
```diff
@@ -0,0 +1,62 @@
+# Parsing `net10`
+
+**Owner** [Immo Landwerth](https://github.com/terrjobst)
```
KexyBiscuit, line 3:
```suggestion
**Owner** [Immo Landwerth](https://github.com/terrajobst)
```

INDEX.md
```diff
@@ -86,6 +86,7 @@ Use update-index to regenerate it:
 | 2023 | [Multi-threading on a browser](accepted/2023/wasm-browser-threads.md) | [Pavel Savara](https://github.com/pavelsavara) |
 | 2023 | [net8.0-browser TFM for applications running in the browser](accepted/2023/net8.0-browser-tfm.md) | [Javier Calvarro](https://github.com/javiercn) |
 | 2024 | [.NET Standard Targeting Recommendations](accepted/2024/net-standard-recommendation.md) | [Immo Landwerth](https://github.com/terrajobst), [Viktor Hofer](https://github.com/ViktorHofer) |
+| 2024 | [Parsing `net10`](accepted/2024/net10.md) | [Immo Landwerth](https://github.com/terrjobst) |
```
KexyBiscuit, line 89:
```suggestion
| 2024 | [Parsing `net10`](accepted/2024/net10.md) | [Immo Landwerth](https://github.com/terrajobst) |
```

accepted/2024/net10.md
terrajobst: LOL

INDEX.md
terrajobst: LOLv2

huoyaoyuan:
A retrocompute question - is it ever possible to build for .NET Framework 1.0 with current .NET SDK?

I'd suggest to keep `net10` meaning .NET Framework 1.0 with more dedicated warning/error message.

terrajobst:
> A retrocompute question - is it ever possible to build for .NET Framework 1.0 with current .NET SDK?

I don't believe so because .NET Framework 1.x doesn't even have a targeting pack; those came to be with .NET Framework 2.0.

GrabYourPitchforks:
> What do you mean by this? net10 has been supported forever because that was the old syntax.

Depends on what you mean. ".NET 10" has never been supported by virtue of the .NET 10 product and SDK never existing. Unless I misread your doc, ".NET Framework 1.0" has not been supported by modern toolchains for years. This seems like a case where `net10` is ambiguous, and in the case of ambiguity, the typical preference is for APIs / tools to throw rather than guess.

> So what about net472? Should this be an error too?

I only suggested that the target framework `net10` be considered an error. It's a modification to your current proposal, where you contemplate that `net10` be a warning and `net12` be an error. Per your proposal, `net20` and beyond (which presumably includes `net472`) retain their current meaning. I think this part of your proposal is sound and you should keep it as-is.

terrajobst:
The project file/SDK didn't invent the framework syntax, we simply made it the NuGet syntax.

As a result, you can put `<TargetFramework>net10</TargetFramework>` in the project file today (or `net10` as a folder in a NuGet package) and it means .NET Framework 1.0. See https://nugettools.azurewebsites.net/6.11.1/parse-framework?framework=net10.

> I only suggested that the target framework net10 be considered an error. It's a modification to your current proposal, where you contemplate that net10 be a warning and net12 be an error

Gotcha. I just updated the PR description. Does the "alternatively" section represent your proposal then? If so, I'd be OK with that too.

GrabYourPitchforks:
> Does the last sentence represent your proposal then? If so, I'd be OK with that too.

Yup! Thanks. :)

Again - I'm not terribly passionate one way or another. It's just a proposal for others' consideration. Your doc is very well thought though.

GrabYourPitchforks (approved):

rolfbjarne:
Would it possible / make sense to make nuget.org reject packages that use the old format without periods? In particular for `net1X` that might be doable, since it would only be a breaking change for anyone trying to submit a .NET Framework v1.0 package.

Also this page should be updated to use periods: https://learn.microsoft.com/en-us/dotnet/standard/frameworks (and probably other doc pages too).

accepted/2024/net10.md
```diff
@@ -0,0 +1,75 @@
...
+Considering that .NET Framework 1.0 is out of support for a while and Visual
+Studio doesn't even support building for anything older than .NET Framework 2.0,
+it seems very doable to make `net10` mean .NET 10. The question is what happens
+with all the other frameworks, such as `net20`, `net452` etc. Quite a few of
+those are still supported and actively used.
+
+Here is the proposal:
+
+* Make `net10` mean .NET 10 and `net11` mean .NET 11 but keep all higher
+  versions as-is (`net20` still means .NET Framework 2.0).
```
akoeplinger, line 49: this seems backwards to me. if the goal is to eventually guide people to always use periods then I don't think we should complicate our lives by allowing `net10` and potentially cause more issues in downstream tools that parse the version.

I'd change the rule to:
1. Warn about missing dot for (currently) non-ambigious versions like net20, net472
2. Error for ambigious versions like net10, net11.

And change to error everywhere in net12 like you suggested.

sundhaug92:
Could it be an idea to make it a warning by default but an opt-in error?

accepted/2024/net10.md
RussKie: > 1. Warn about missing dot for (currently) non-ambigious versions like net20, net472

We don't have any alternatives to `net20` or `net472` to fix the warning, do we? If not, we shouldn't warn for those.

akoeplinger: you can use `net2.0` and `net4.7.2` instead.

RussKie: TIL

goswinr:
How about `dotnet10` ?

terrajobst:
@rolfbjarne 

> Would it possible / make sense to make nuget.org reject packages that use the old format without periods?

That's a good point. It should be easy because NuGet tooling could already normalize the syntax.

terrajobst:
@WeihanLi 

> If we want to produce a warning before producing an error directly, would it be applicable to produce a warning for net9 target and an error for net10 target usage since net9.0 has not GA

I don't think it's practical to make such a change before .NET 9 GA. We could, however, likely make that change to the SDK post GA (because the SDK versions gets shipped more frequently). It's probably considered a source breaking change, but I'd consider that acceptable.

akoeplinger:
I'm not sure we need to do anything for nuget.org or nuget packages, `net10` can continue to mean net1.0 there since nobody is creating these by hand. I thought the main motivation for this is the TFM property in .csproj.

terrajobst:
@akoeplinger 

That's a fair point actually.

accepted/2024/net10.md
terrajobst: Agreed. I've changed the proposal.

accepted/2024/net10.md
```diff
@@ -0,0 +1,202 @@
...
+That is the preferred syntax and will continue to work. The point of this work
+is not to make `net10` work because it's nicer -- it's to promote "framework
+names should be specified with a period", but we don't want to make this an
+error (because many, many people use `net472` today) but a warning. However,
+when using `net10` will most likely fail because the SDK doesn't support
+targeting it, which will drown out the warning telling you to use period.
+Failing it only for `net10` would be possible but feel odd because we'd prefer
+periods for all framework names, not just for `net10.0`.
+
+# Why don't we make `net10` just work work?
```
danfma, line 190:
```suggestion
# Why don't we make `net10` just work properly?
```

accepted/2024/net10.md
```diff
@@ -0,0 +1,202 @@
...
+<TargetFramework>netX.Y.Z</TargetFramework>
+```
+
+Examples include `net9.0`, `net10.0`, `net4.5.1`. This syntax makes it
+unambiguous.
+
+Proposal:
+
+* Fail the build if the TFM is `net10` or `net11` with an error message
+  > This framework syntax is unsupported because it's unambiguous. If you mean
```
RussKie, line 52:
```suggestion
  > This framework syntax is unsupported because it's ambiguous. If you mean
```

accepted/2024/net10.md
```diff
@@ -0,0 +1,202 @@
...
+* Fail the build if the TFM is `net10` or `net11` with an error message
+  > This framework syntax is unsupported because it's ambiguous. If you mean
+  > .NET Framework 1.0, use `net1.0`. If you mean .NET 10, use `net10.0`.
+* If the TFM is neither `net10` nor `net11`, produce a warning if it doesn't
+  contain a period:
+  > This framework syntax is obsolete. You should use periods to separate
+  > version components.
+
+The reason to fail the build is because in practice someone trying to target
+.NET 10 will most likely fail anyway because the either don't have the .NET
```
akoeplinger, line 60:
```suggestion
.NET 1.0 will most likely fail anyway because they either don't have the .NET
```

accepted/2024/net10.md
```diff
@@ -0,0 +1,202 @@
...
+Framework 3.5 targeting pack -- or worse -- they do, but the code doesn't
+compile because it's meant for modern .NET Core, not a 25 year old .NET
+Framework 1.0. By failing it early it turns an unintelligible error message into
```
akoeplinger, line 63:
```suggestion
Framework 1.0. By failing early it turns an unintelligible error message into
```

akoeplinger (approved):
LGTM but the issue description needs to be updated to reflect the current proposal.

accepted/2024/net10.md
```diff
@@ -0,0 +1,202 @@
...
+Failing it only for `net10` would be possible but feel odd because we'd prefer
+periods for all framework names, not just for `net10.0`.
+
+### Why don't we make `net10` just work?
+
+Considering that .NET Framework 1.0 is out of support for a while and Visual
+Studio doesn't even support building for anything older than .NET Framework 2.0,
+it seems very doable to make `net10` mean .NET 10.
+
+There are two downsides to this:
```
rolfbjarne, line 196:
```suggestion
There are three downsides to this:
```

rolfbjarne (approved):

terrajobst:
I no longer work at Microsoft so I’m gonna close this. If someone on the team wants to move this forward, they should open their own PR.

PR #323, opened 2024-10-16
Remove mentions about watchOS.

rolfbjarne:
watchOS isn't supported in .NET, so remove any mentions of it.

### DIFF
--- a/accepted/2020/platform-checks/platform-checks.md
+++ b/accepted/2020/platform-checks/platform-checks.md
@@ -405,9 +405,6 @@ namespace System
         public static bool IsTvOS();
         public static bool IsTvOSVersionAtLeast(int major, int minor = 0, int build = 0, int revision = 0);
 
-        public static bool IsWatchOS();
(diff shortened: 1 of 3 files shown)


accepted/2020/platform-checks/platform-checks.md
```diff
@@ -405,9 +405,6 @@ namespace System
         public static bool IsTvOS();
         public static bool IsTvOSVersionAtLeast(int major, int minor = 0, int build = 0, int revision = 0);
 
-        public static bool IsWatchOS();
-        public static bool IsWatchOSVersionAtLeast(int major, int minor = 0, int build = 0, int revision = 0);
```
akoeplinger, line 409: we did add these APIs though

akoeplinger (commented):
I don't think this is worth doing, these are just design docs.
The actual supported OSes are documented in https://github.com/dotnet/core/blob/main/release-notes/8.0/supported-os.md and we don't mention watchOS there

akoeplinger:
Closing this for now.

PR #324, opened 2024-10-16
Update how the default OS version is calculated for Apple platforms.

rolfbjarne:
Also change the corresponding example to use Android instead of iOS, because
it's not correct for iOS.


rolfbjarne:
I'm also updating the docs: https://github.com/dotnet/docs/pull/43089

jonathanpeppers (approved):
The notes about Android look correct. For .NET 9, `net9.0-android` is `net9.0-android35`. If 36 came out in the middle of the year you would have to put `net9.0-android36` in your project to use it.

PR #327, opened 2025-01-17
Note implemented security mitigations

jkotas:


jkotas:
Following up on feedback that was shared some time ago.

richlander (approved):

richlander:
Thanks for doing that. I think this new context is quite useful.

PR #328, opened 2025-01-29
Introduce a repo-centric .NET SDK config file and system

baronfel:
We have about 15 (/s) different configuration files and formats that rev independently by different teams. Users would be well-served by a single source of truth that all tools can read from, that has unified documentation and comes accompanied by helper libraries to ensure consistency of behavior.  The first step along that journey is having a shared understanding of the config system and its intended uses, so that we can build tooling on top of that shared understanding.

### DIFF
--- a/proposed/dotnet-sdk-unified-config.md
+++ b/proposed/dotnet-sdk-unified-config.md
@@ -1,0 +1,96 @@
+# A unified configuration file and experience for the .NET SDK
+
+**Owner** [Chet Husk](https://github.com/baronfel) | [Daniel Plaisted](https://github.com/dsplaisted)
+
+The .NET SDK is a product that consists of a number of different tools, all owned by different teams, all operating at slightly different levels of interaction. Some tools operate on the entire repository, some tools on a single project, and yet other tools on a single file.
+
+Because of the disparate nature of these tools, the configuration for each tool is different. This makes it difficult for users to understand how to configure the tools, and difficult for the tools to share configuration information. This document proposes a single, centralized document for repository-wide configuration that tools belonging to the SDK (as well as SDK-adjacent 1st- and 3rd-party tools) can use to share configuration information in a way that has
+
+* a consistent user experience
+* a well supported IDE experience
+* is shareable across team members and trackable in source control
+
+## What exists today
+
+Today there are several places where teams have put configuration for repo- or project-level configuration for tooling:
+
+* global.json files
+* NuGet.config
+* .config/dotnet-tools.json
+* .runsettings (dotnet test)
+
+In addition, several tools are only configurable via environment variables, which are not easily shareable or trackable in source control (due to the lack) of support for .env or other such conventions in the `dotnet` CLI.
+
+All of these files and systems arose independently, have different formats, and the amount of tooling support for them varies significantly. We have 'shipped the org structure' in terms of our configuration.
+
+## What is the need?
+
+There are [several dotnet/sdk items](https://github.com/dotnet/sdk/issues?q=label%3Aconfig-system-candidate) where the concept of a unified configuration system would be beneficial. These include:
+
+* expressing user behavioral preferences around build/pack/publish Configurations
+* if binlogs should be made when performing builds
+* the use or non-use of advanced terminal behaviors for build
+* default verbosities for build/pack/publish
+* custom aliases for CLI commands
+* unified opt-out of telemetry
+* storing tooling decisions like which mode `dotnet test` should execute in
+
+All kinds of repository-wide configuration need a unified place to be expressed that we lack today.
+
+## Proposal
+
+I propose that we create a new, repo-level configuration file that will be consumed by `dotnet` and `dotnet`-related tooling to provide a unified configuration experience. This file will be called `poohbear.config`[^1] and will be located at the root of the repository.
+
+### Composition and heirarchy
+
+This file will be hierarchical in a very limited fashion. The final realized configuration will be the result of merging the `poohbear.config` file in the repository root with an optional `poohbear.config` file in the user's `$HOME` directory (specifically, the SpecialFolders.UserProfile location in the .NET BCL). This will allow users to set defaults for their own use, and then override those defaults on a per-repository basis for sharing with coworkers.
+
+### Nested keys
+
+We should encourage tools to stake a claim on a top-level 'namespace' for keys in this configuration file. Keys should either be namespaced with the tool's name or be namespaced with a common prefix that is unique to the tool. This will prevent key collisions between tools and make it easier for users to understand what each key does. This allows for the format to be open to extensibility, as new tools can add their own keys without fear of collision.
+
+### File format
+
+>[!NOTE]
+> This is going to be a bike-shedding point. Please try to focus first on the use cases and tool-level use, and then we can discuss the file format.
+
+The `poohbear.config` file should be parseable by tooling and pleasant to read and edit by users.
+
+The choice of format is in some ways limited by engineering considerations. The easiest thing would be to choose a format that is already supported by the BCL, which would lead us to JSON or XML. Both of these formats have downsides, though: JSON doesn't support comments and has an anemic data model, and XML is historically verbose and difficult to parse in a consistent way.
+
+The recent wave of programming languages and programming-related tools have popularized the use of more human-readable formats like YAML, TOML, and INI. These formats are more human-readable and have more expressive power than JSON or XML. None of them have parsers in the BCL, though, which would mean that we would have to write our own parser or use a third-party library. Writing a parser is a potentially error-prone task, and using a third-party library would require onboarding the parser to source build as well as doing a thorough security audit. YAML specifically has such a lax parser that it might be a security risk to use it.
+
+My personal suggestion is that we should use INI or TOML and advocate for parsers for these in the BCL. TOML is widely used, being used for the huge python ecosystem and the up and coming Rust ecosystem, and strikes a good balance between human-readability and expressiveness.
+
+### Tooling support
+
+CLI commands (dotnet config) a la `git config` will be provided to allow users to set and get configuration values. This will allow users to set configuration values without having to edit the file directly. We should allow reading
+
+We should provide a library to parse the file and integrate its data into `Microsoft.Extensions.Configuration` so that tools can easily consume the configuration data.
+
+We must ensure that both the CLI and VS support the configuration file and the configuration system.
+
+### Secondary goals
+
+Because the configuration file will be read into a Microsoft.Extensions.Configuration context this opens up the possibility for environment variables to be implicitly defined for all configuration keys. This would allow users to set configuration values via environment variables, which is a common pattern and an improvement from the mismatched support for environment variables that we have today.
+
+### Future work
+
+With adoption, we could soft-deprecate some of the sources of standalone configuration mentioned above. Each configuration file could migrate to one or more namespaced keys in the `poohbear.config` file. For example, .NET SDK tools could live in a `local-tools` namespace that listed the tools and versions.
+
+### Non-goals
+
+We do not aim to replace solutions with this config file.
+We do not aim to provide 'build configurations' - descriptions of projects to build and properties to apply to that build.
+
+## Suggested configuration keys/namespaces for v1
+
+> TBD, perhaps as part of discussion
+
+* dotnet
+  * settings related to cross-cutting dotnet cli behaviors: telemetry, CLI aliases, etc
+* test
+  * settings related to the behavior of `dotnet test`
+
+
+[^1]: Name is a placeholder, and will be replaced with a more appropriate name before implementation. A final name might change the extension to represent the file type chosen.
--- a/proposed/dotnet-sdk-unified-config.md
+++ b/proposed/dotnet-sdk-unified-config.md
@@ -56,11 +56,11 @@ We should encourage tools to stake a claim on a top-level 'namespace' for keys i
 
 The `poohbear.config` file should be parseable by tooling and pleasant to read and edit by users.
 
-The choice of format is in some ways limited by engineering considerations. The easiest thing would be to choose a format that is already supported by the BCL, which would lead us to JSON or XML. Both of these formats have downsides, though: JSON doesn't support comments and has an anemic data model, and XML is historically verbose and difficult to parse in a consistent way.
+The choice of format is in some ways limited by engineering considerations. The easiest thing would be to choose a format that is already supported by the BCL, which would lead us to JSON, XML, or INI. JSON and XML formats have downsides, though: JSON doesn't support comments and has an anemic data model, and XML is historically verbose and difficult to parse in a consistent way.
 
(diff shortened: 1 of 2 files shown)


proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,96 @@
...
+
+The recent wave of programming languages and programming-related tools have popularized the use of more human-readable formats like YAML, TOML, and INI. These formats are more human-readable and have more expressive power than JSON or XML. None of them have parsers in the BCL, though, which would mean that we would have to write our own parser or use a third-party library. Writing a parser is a potentially error-prone task, and using a third-party library would require onboarding the parser to source build as well as doing a thorough security audit. YAML specifically has such a lax parser that it might be a security risk to use it.
+
+My personal suggestion is that we should use INI or TOML and advocate for parsers for these in the BCL. TOML is widely used, being used for the huge python ecosystem and the up and coming Rust ecosystem, and strikes a good balance between human-readability and expressiveness.
```
akoeplinger, line 63: there's already support for INI in Microsoft.Extensions.Configuration via https://www.nuget.org/packages/Microsoft.Extensions.Configuration.Ini#readme-body-tab

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+
+Because of the disparate nature of these tools, the configuration for each tool is different. This makes it difficult for users to understand how to configure the tools, and difficult for the tools to share configuration information. This document proposes a single, centralized document for repository-wide configuration that tools belonging to the SDK (as well as SDK-adjacent 1st- and 3rd-party tools) can use to share configuration information in a way that has
+
+* a consistent user experience
+* a well supported IDE experience
+* is shareable across team members and trackable in source control
+
+## What exists today
+
+Today there are several places where teams have put configuration for repo- or project-level configuration for tooling:
```
MattKotsenas, line 15: May or may not be relevant, but I'd also consider `Directory.Build.rsp` as another of these types of configuration sources.

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+
+### Composition and heirarchy
+
+This file will be hierarchical in a very limited fashion. The final realized configuration will be the result of merging the `poohbear.config` file in the repository root with an optional `poohbear.config` file in the user's `$HOME` directory (specifically, the SpecialFolders.UserProfile location in the .NET BCL). This will allow users to set defaults for their own use, and then override those defaults on a per-repository basis for sharing with coworkers.
+
+### Nested keys
+
+We should encourage tools to stake a claim on a top-level 'namespace' for keys in this configuration file. Keys should either be namespaced with the tool's name or be namespaced with a common prefix that is unique to the tool. This will prevent key collisions between tools and make it easier for users to understand what each key does. This allows for the format to be open to extensibility, as new tools can add their own keys without fear of collision.
```
MattKotsenas, line 50: Consider from the outset some type of namespace of "vendor" prefix, even if it's just guidance without code enforcement. I get the point of not wanting to appear to privilege our own tools over 3rd parties, however I'm also sensitive to the risk of picking suboptimal names to avoid backwards incompatibility or collisions at the top-level.

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+#### Programmatic access
+
+We (BCL) should provide a library to parse the file and integrate its data into `Microsoft.Extensions.Configuration` so that tools can easily consume the configuration data.
+
+#### IDE onboarding
+
+We must ensure that both the CLI and VS support the configuration file and the configuration system.
+
+#### Usage within the dotnet CLI and supporting tools
+Inside the `dotnet` CLI (and tools that ship within the CLI), we will initialize a `ConfigurationBuilder` with the `poohbear.config` file and merge it with the user's `poohbear.config` file. We will also seed that builder with environment variables. Tools that need the configuration within the `dotnet` CLI will read the appropriate values and apply them to their execution.
```
MattKotsenas, line 95: If the tool / format starts to support env vars or other sources, consider adding a `dotnet config dump` or `dotnet config get --all` command or similar so that folks can debug setup issues.

proposed/dotnet-sdk-unified-config.md
MattKotsenas: Similarly, we'll need guidance for when to put something here versus `Directory.Build.props`. For instance, where should I put `TreatWarningsAsErrors`? I think that has a surprisingly big impact for MSBuild, given that one's a default property and one's a global property.

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+and the following command would get the user-global value only
+```shell
+> dotnet config get dotnet.telemetry --global
+false
+```
+
+The [git config][git-config] command is a good source of inspiration.
+
+
+#### Programmatic access
```
MattKotsenas, line 86: A different type of programmatic access, but should this file be available to Roslyn as well? For instance, should it be a goal, non-goal, etc. to allow something like stylecop.json or BannedSymbols.txt to be here instead?

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+
+All kinds of repository-wide configuration need a unified place to be expressed that we lack today.
+
+## Proposal
+
+I propose that we create a new, repo-level configuration file that will be consumed by `dotnet` and `dotnet`-related tooling to provide a unified configuration experience. This file will be called `poohbear.config`[^1] and will be located at the root of the repository.
+
+### Composition and heirarchy
+
+This file will be hierarchical in a very limited fashion. The final realized configuration will be the result of merging the `poohbear.config` file in the repository root with an optional `poohbear.config` file in the user's `$HOME` directory (specifically, the SpecialFolders.UserProfile location in the .NET BCL). This will allow users to set defaults for their own use, and then override those defaults on a per-repository basis for sharing with coworkers.
```
MattKotsenas, line 46: Somewhat silly question for me, but how do you know what the repository root is? Are you searching up for the first you find? What if they nest?

Possible options:

- Allow nesting similar to .editorconfig, optionally with an `is_root` style property
- Find the first; if so, what if the first found isn't in `$HOME` but is "above" the .git folder?
- Something else?

`git config` doesn't have these problems because you never actually interact with the file; here it's visible to the user (which is good) but that means you can and will end up with every combination possible :)

baronfel:
Feedback from @KathleenDollard: have a story in here about how to version defaults for this config

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+
+* expressing user behavioral preferences around build/pack/publish Configurations
+* if binlogs should be made when performing builds
+* the use or non-use of advanced terminal behaviors for build
+* default verbosities for build/pack/publish
+* custom aliases for CLI commands
+* unified opt-out of telemetry
+* storing tooling decisions like which mode `dotnet test` should execute in
+
+All kinds of repository-wide configuration need a unified place to be expressed that we lack today.
```
jkotas, line 38: Would it be better to extend global.json instead of introducing yet another file?

baronfel: Worth having the conversation, but global.json first and foremost belongs to the runtime today. I'd like to have something that tooling can work with that has guaranteed zero impact on runtime scenarios.

There is a bit of https://xkcd.com/927/ in this proposal, though.

jkotas: > global.json first and foremost belongs to the runtime today

From my point of view, global.json belongs to the SDK. There is a tiny bit of code in in dotnet/runtime that reads it to find the version of SDK to call. This code only ever runs when `dotnet` is used to invoke SDK.

Also, different parts of SDK or SDK extensions have started to use global.json to store random configuration. Check what's in global.json in the dotnet/sdk repo: https://github.com/dotnet/sdk/blob/main/global.json

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+
+The `poohbear.config` file should be parseable by tooling and pleasant to read and edit by users.
+
+The choice of format is in some ways limited by engineering considerations. The easiest thing would be to choose a format that is already supported by the BCL, which would lead us to JSON, XML, or INI. JSON and XML formats have downsides, though: JSON doesn't support comments and has an anemic data model, and XML is historically verbose and difficult to parse in a consistent way.
+
+The recent wave of programming languages and programming-related tools have popularized the use of more human-readable formats like YAML, TOML, and INI. These formats are more human-readable and have more expressive power than JSON or XML. None of them have parsers in the BCL, though, which would mean that we would have to write our own parser or use a third-party library. Writing a parser is a potentially error-prone task, and using a third-party library would require onboarding the parser to source build as well as doing a thorough security audit. YAML specifically has such a lax parser that it might be a security risk to use it.
```
baronfel, line 61: Add [KDL](https://kdl.dev/)

baronfel:
Add risks for parsers for non-BCL-included formats.

proposed/dotnet-sdk-unified-config.md
slang25: I was going to suggest KDL, it's so good

zkat: Wow uhhh yeah that would be awesome.

fwiw, KDL might be standardizing officially with Ecma going forward. It would also be VERY easy to map msbuild concepts to it (see [this translated csproj example](https://github.com/kdl-org/kdl/blob/main/examples/nuget.kdl)).

It’s worth noting that KDL was created specifically because I ended up not wanting to use TOML as a configuration file. Some reasons being practical (like the difficulty with nesting and how strict the formatting needs are), and others being ideological (I just do not want to touch things so closely associated with that author, a sentiment many in my circles share).

Going from there, KDL was very carefully designed to bring in a lot of the practicality that people want from YAML with much more safety. In all but a few rare cases with specific Unicode shenanigans (for example, certain invisible spaces that Unicode specifically requires not be treated as spaces), a KDL document will be exactly what it seems to be to a human  reader with any given editor.

Anyway I’m more than happy to answer questions about it, its tooling and adoption status, impressions from existing tools that use it for configuration, and even do some porting of example configs to show off potential benefits of KDL-specific features. You can also reach me internally, although I’m OOF until mid-March right now.

proposed/dotnet-sdk-unified-config.md
normj: As ubiquitous as Git is these days surely there are still customers that don't use Git. Why wouldn't the hierarchy follow the similar pattern of `global.json` and look for a file up the parent directory till you find a file or reach the root. And of course take into account the $HOME directory.

nohwnd:
For the tooling, I would love if there was a supported and standard way of preserving this configuration into the finished artifacts, and merging the configurations from multiple artifacts into one. This way the configuration is not just for the build-time, but also for execution.

 Lot of this is probably already covered by the config subsystem, but here are the the usecases I have in mind:

- User defines their default test run preferences using poohbear.config. They build / publish their solution. Then they move their bin folder to 3 different agents to run tests on windows, macos, linux. These agents don't have dotnet sdk. 

    On the target system they run the app via dotnet exec. Their configuration should be the same as if they used dotnet test on the original system, and not require additional copy of poohbear.config from the root and all applicable poohbear.config files.

- User does the above, but they have 10 and not 1 dll to run. A test run orchestrator (that does not require dotnet sdk), should be able to collect configuration from all sources, and merge them to get the correct "solution" configuration, and correct per-project configuration. ( definition of which setting applies for whole "solution" and which one applies per project is up to us, but preferably there should be an easy way to read configurations, and merge them).

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+## What exists today
+
+Today there are several places where teams have put configuration for repo- or project-level configuration for tooling:
+
+* global.json files
+* NuGet.config
+* .config/dotnet-tools.json
+* .runsettings (dotnet test)
+
+In addition, several tools are only configurable via environment variables, which are not easily shareable or trackable in source control (due to the lack) of support for .env or other such conventions in the `dotnet` CLI.
```
am11, line 22: Glad to see [`.env`](https://hexdocs.pm/dotenvy/dotenv-file-format.html) mentioned here. It may as well be a candidate if we are looking for something other than JSON or XML because of its simplicity and its ability to expand values dynamically from environment variables. Python, Docker Compose, Node.js (`node --env-file` new in v20) etc. ecosystems are using it with great success.

am11: FWIW, corerun (the internal host for coreclr testing) also uses `.env` since https://github.com/dotnet/runtime/commit/f3c705ef291ff89b53220a31d8321355471d1937, so using it right inside corehost won't be a problem either (if it comes to that).

proposed/dotnet-sdk-unified-config.md
martincostello: 🚲🏠: As already referenced in another comment, my only concern here is that within .NET I've never had to use TOML before, so by using it we're adding a new format/standard for users to learn here if they haven't previously had exposure to it.

Whatever the eventual choice, the addition of the git/npm-like ability to get and set the configuration is a great mitigation for this as users can leverage the CLI itself to query-set values in any automation they already have without having to add additional tooling/knowledge to be able to interact with it (e.g. using `jq` to read JSON).

proposed/dotnet-sdk-unified-config.md
```diff
@@ -0,0 +1,150 @@
...
+### Suggested configuration keys/namespaces for v1
+
+> TBD, perhaps as part of discussion
+
+* dotnet
+  * settings related to cross-cutting dotnet cli behaviors: telemetry, CLI aliases, etc
+* test
+  * settings related to the behavior of `dotnet test`
+
+### Prior art
```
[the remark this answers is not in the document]
JonDouglas: while the unified SDK config proposal is focused on tool configuration and not project metadata, I’m curious if there’s future thinking around combining both, similar to the approach taken in Python’s `pyproject.toml`, Rust’s `Cargo.toml`, or other modern ecosystems. 

Is there interest in evolving toward a single file that serves as both project and tool configuration for .NET?

proposed/dotnet-sdk-unified-config.md
rolfbjarne: Not everything is a (git) repository either:

```shell
dotnet new myapp
echo "answer = 42" > poohbear.config
# poohbear.config doesn't work???
```

IMHO it would make more sense to just walk up the directory hierarchy.

jjonescz:
> for example `dotnet run-api`

coincidentally, `dotnet run-api` is already a built-in dotnet command (used by file-based apps), so your example probably wouldn't work without some more explicit subcommand like `dotnet task run-api`

WeihanLi:
> > for example `dotnet run-api`
> 
> coincidentally, `dotnet run-api` is already a built-in dotnet command (used by file-based apps), so your example probably wouldn't work without some more explicit subcommand like `dotnet task run-api`

Oh, thanks for pointing it out, I forgot that, or it may be a fallback of dotnet run file when file not found, or `dotnet task run-api` etc

Nirmal4G:
For the file name, I propose `dotnet.config` or `dotnet.toml` or `dotnet.json` whichever data format we lean into. But yes, I would appreciate a Json 5 or TOML-like language parser in BCL.

We could have `core`, `languages`, `frameworks`, `platforms` or `runtimes`, `sdks` and `tools` as initial top level keys. The `tools` can contain configuration for both internal tools such as `msbuild`, `nuget`, `paket`, `csc`, `vbc` and `fsc` and external tools such as `try`, `format`, `migrate`, etc...

PR #330, opened 2025-03-07
First draft: caller unsafe

agocke:


proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Goals
+
+The existing unsafe system does not clearly identify which methods need hand-verification to use safely, and it doesn't indicate which methods claim to provide that verification (and produce a safe API).
+
+We want to achieve all of the following goals:
+
+* Clearly annotate which methods require extra verification to use safely
+* Clearly identify in source code which methods are responsible for performing extra verification
+* Provide an easy way to identify in source code if a project doesn't use unsafe code
```
richlander, line 18: How about for a binary? Should we set `UnverifiableCodeAttribute` or similar?

agocke: Good question. In this design it’s allowed to define a CallerUnsafe method without an unsafe block if you never actually call the method. If you did, there would be an unsafe block and the assembly would have unverifiable code. I’m not sure if defining such a method, but never using it except through other callers unsafe methods, should be unverifiable. 

My inclination is yes: anything that contains unsafe code should be marked, even if it’s not called from safe code.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+
+```C#
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor)]
+public sealed class CallerUnsafe : System.Attribute
+{
+    public WarnByDefault { get; init; } = true;
+}
+```
+
+This would be the dual of the existing `unsafe` context. The `unsafe` context eliminates unsafe errors, indicating that the code outside of the unsafe context should be considered safe. The `CallerUnsafe` attribute does the opposite: it indicates that the code inside is unsafe unless specific obligations are met. These obligations cannot be represented in C#, so they must be listed in documentation and manually verified by the user of the unsafe code. Only when all obligations are discharged can the code be wrapped in an `unsafe` context to eliminate any warnings.
```
richlander, line 34: It would be great to see an example of this, a method that is unsafe with the `CallerUnsafe` attribute. I'd like to see examples of obligations that can only be represented in a comment and how they are discharged in a caller. I'm certain that readers would find such an example very informative.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+* Memory safety
+* No access to uninitialized memory
+
+In this document **memory safety** is strictly defined as: safe code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime.
+
+No access to uninitialized memory means that all managed memory is either never read before it has been initialized by C# code, or it has been initialized to a zero value.
+
+Some examples of APIs or features that are unsafe due to exposing uninitialized memory include:
+
+* ArrayPool.Rent
```
richlander, line 55: `ArrayPool.Rent` has two sides to uninitialized memory. As the new "renter", I cannot see signs of the old renter and the older renter cannot see signs of what I'm doing.

Staying with renters ... we want to ensure that old and new renters cannot post to social media about the bizarre paint choices of the other ... assuming that the landlord paints the walls to an unappealing beige between tenants, as a matter of policy (and the door keys are never exchanged until the paint has dried).

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+
+While existing `unsafe` is useful, it is limited by only applying to pointers and ref-like lifetimes. Many methods may be considered unsafe, but the unsafety may not be related to pointers or ref-like lifetimes. For example, almost all methods in the System.RuntimeServices.CompilerServices.Unsafe class has the same safety issues as pointers, but do not require an `unsafe` block. The same is true of the System.RuntimeServices.CompilerServices.Marshal class.
+
+## Goals
```
richlander, line 10: It might be worth some time thinking forward to policies that organizations with strict security policies might have given these new mechanisms. 

For example, I could imagine `audit` being expanded to block on restore of packages with unsafe code, either absolutely or allowing unsafe code from only certain publishers. Similarly, I could imagine CI breaks on new unsafe blocks.

Unsafe code is a reality, in all general purpose programming languages. It's a question of what the default posture is and how that gets implemented in everyday workflows.

Some of that is out of scope of this doc. It's entirely possible that "easy way to identify" covers all the cases. It's worth a bit more thought. The workflows around safety don't end at the compiler.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Proposal
+
+We need to be able to annotate code as unsafe, even if it doesn't use pointers.
+
+Mechanically, this would be done with a new attribute:
+
+```C#
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor)]
+public sealed class RequiresUnsafe : System.Attribute
```
jkotas, line 28: I agree we need an attribute like this to encode the fact that the callers must be called unsafe.

However, I think we should strongly consider changing the rules for `unsafe` keyword so that it works more sensible and more similar to Rust. If one sees `void unsafe M() { ... }`, it is natural to think that the method signature is unsafe and that the callers must be unsafe. I had to explain many times that it is not the case today. I am proposing that we change this to mean that the method is annotated with `[RequiresUnsafe] void unsafe M() { ... }`.

If we do not do this, we will have to be explaining why `unsafe` in C# is non-sensical and different from Rust. When we have to be explaining, we are failing.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+{
+    public WarnByDefault { get; init; } = true;
+}
+```
+
+This would be the dual of the existing `unsafe` context. The `unsafe` context eliminates unsafe errors, indicating that the code outside of the unsafe context should be considered safe. The `RequiresUnsafe` attribute does the opposite: it indicates that the code inside is unsafe unless specific obligations are met. These obligations cannot be represented in C#, so they must be listed in documentation and manually verified by the user of the unsafe code. Only when all obligations are discharged can the code be wrapped in an `unsafe` context to eliminate any warnings.
+
+The `WarnByDefault` flag is needed for backwards-compatibility. If an existing API is unsafe, adding warnings would be a breaking change. If `WarnByDefault` is set to `false`, warnings are not produced unless a project-level property, `ShowAllRequiresUnsafeWarnings`, is set to true.
```
jkotas, line 36: Is the idea going to be that `WarnByDefault` is effectively going to mean: API introduced before .NET 10 (or whatever release we are going to skip this in)? I do not think we want to introduce a wart like this that will have to live in the API surface forever. Also, it is probably going to problematic for projects that multi-target.

I think that it would be better to deal with the compatibility concern at project level, similar how it was done for nullability. My proposal would be to extend existing `AllowUnsafeBlocks` property (or introduce a new msbuild property) with a new value that opts-in the project into a new `unsafe` policy where `RequiresUnsafe` annotations are respected, `unsafe` in the method signature implies `RequiresUnsafe` annotation, etc.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+
+In this document **memory safety** is strictly defined as: safe code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime.
+
+No access to uninitialized memory means that all managed memory is either never read before it has been initialized by C# code, or it has been initialized to a zero value.
+
+Some examples of APIs or features that are unsafe due to exposing uninitialized memory include:
+
+* ArrayPool.Rent
+* The `stackalloc` C# feature used with `SkipLocalsInit` and no initializer
+
```
rolfbjarne, line 57: All P/Invokes would also be unsafe, even ones that don’t need the unsafe keyword.

Even a P/Invoke with no return value or parameters is unsafe:

```cs
[DllImport ("libc")]
static extern void strlen();
```

This will crash when called.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+Since only warnings are an output of the above design, the feature could be implemented as an analyzer for C#.
+
+### Definition of Unsafe
+
+`RequiresUnsafe` is only useful if there is an accepted definition of what is considered unsafe. For .NET there are two properties that we already consider safe code to preserve:
+
+* Memory safety
+* No access to uninitialized memory
+
+In this document **memory safety** is strictly defined as: safe code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime.
```
jkotas, line 49: > In this document **memory safety** is strictly defined as: safe code can never acquire a reference to memory that is not managed by the application.

This does not look like a sufficient definition. Consider the following example:
```
M(Random.Shared.Next());

void M(int x)
{
    unsafe
    {
        *(int*)x = 42;
    }
}
```
The safe code never acquired a reference to memory that is not managed by the application in this example, but this code has major memory safety issue.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
+
+# Annotating unsafe code with RequiresUnsafe
```
jkotas, line 2: @jaredpar started writing a draft for this as well some time ago: https://github.com/jaredpar/csharplang/tree/unsafe . 

(It proposes to reuse the `unsafe` keyword that I think is much cleaner design.)

proposed/caller-unsafe.md
richlander: I don't think of operations that reliably crash as unsafe. I would assert that stack overflow is safe in .NET because `StackOverflowException` is uncatchable and reliably causes a crash/fail fast.

Crashing is the same as a fuse that trips (and cuts the power), which prevents home wiring from transitioning from warm to burny temps. I love analogies since they enable us to see our problems from a distance in another domain. Do you agree with this one?

That said, in previous conversations I believe we consider all P/Invokes (as you suggest) as unsafe. It is hard to consider them otherwise. There is no way to "discharge obligations" at a `static extern` method signature.

proposed/caller-unsafe.md
agocke: I agree that if we were starting from scratch, the retrofit is simpler. Starting from where we are now, I’m not so convinced. 

Let’s say we add a new project flag that makes all uses of `unsafe` on members or types mean RequiresUnsafe. That implies that both the producer and consumer need to opt-in to see any changes. Producer is not hard — the core libraries will be updated to set the flag and the next .NET version will produce appropriate metadata.

The consumer side is not simple. Because the scope is all-or-nothing, consumers will have to opt-in to the breaking change. Because the breaking change severity is so high, we will never be able to enable the feature by default. Because the feature is all-or-nothing we will also have no way to force errors for new code, even if there’s no breaking change risk. 

Conversely, if we go with the attribute approach we gain two degrees of freedom. We can both opt-in individual methods, and we can force new errors without consumer involvement. That would allow us to make new unsafe APIs always error. 
We could also potentially consider breaking changes for specific API surface. 

Even six years later, adoption of nullable isn’t pervasive, and I’m not sure it ever will be at this point. This feature is far less prominent that nullable and I expect even fewer users will go out of their way to enable it. 

Lastly, from an operational perspective, if we need the unsafe modifier to cause an attribute to be emitted, we’re now forced to implement this feature in the compiler and we’ll have to do language design review for the language changes. The analyzer approach is lighter weight and could be implemented by any team with spare bandwidth.

proposed/caller-unsafe.md
agocke: You’re right, the definition needs to be more precise. It’s along the lines of: the above definitions are for _valid_ code. All code _must_ meet those definitions to be considered valid. Unsafe code is merely code that doesn’t have validity checked and enforced by the compiler/runtime. I’ll need to wordsmith this. (And yes, I’m aware this is retreading our definitions of “verifiable”, but I’m not sure we want to import that legacy wholesale).

proposed/caller-unsafe.md
jkotas: Yes, I agree that the adoption will take a while. There is a good chance that opt-in into this feature will become mandatory for Microsoft 1P code at some point that would increase the adoption significantly.

I do not think enabling errors for new APIs and keeping existing APIs as warnings is interesting. I do not expect that we are going to grow unsafe API surface a lot, so the number of APIs on the error plan would be minimal in foreseeable future. If we wanted to create groups of APIs (I am not a fan - it makes the story more complicated), it should probably be on the same plan as Obsolete.

> we’ll have to do language design review for the language changes

I think it is a good thing. `unsafe` has been a part of the language since forever, and any extensions of it should be reviewed starting from that context. We really to build fully aligned and coherent story here. The lack of alignment between language and libraries is why we are looking into this.

proposed/caller-unsafe.md
jaredpar: The sentence says _project_. That should be very simple to achieve: does the project have `AllowUnsafeBlocks` or not? That is the barrier today for allowing `unsafe` usage and that should not change. If the desire is to detect in the binary of we had this attribute or not then yes we'd need to add more metadata.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+
+We want to achieve all of the following goals:
+
+* Clearly annotate which methods require extra verification to use safely
+* Clearly identify in source code which methods are responsible for performing extra verification
+* Provide an easy way to identify in source code if a project doesn't use unsafe code
+
+## Proposal
+
+We need to be able to annotate code as unsafe, even if it doesn't use pointers.
```
jaredpar, line 22: This is the critical part.

proposed/caller-unsafe.md
jaredpar: > That implies that both the producer and consumer need to opt-in to see any changes

I think this is the only rational way to approach the problem. The number of APIs that require `unsafe` to call is going to increase quite significantly compared to today. Don't see any way to do this without the ability for the consumer to opt in separately from the producer. And I think that is totally fine. As long as it's easy for us to check in build systems and productions if a build / binary has opted into the new version of `unsafe` then we can use tooling / policy to push parties to move to the new model. For 3P customers who are uninterested in the new `unsafe` they can remain as they are today without changes.

proposed/caller-unsafe.md
rolfbjarne: Assuming the rules are going to be the same for everyone (BCL / SDKs / everybody else), wouldn't we need the opposite as well?

> We need to be able to annotate code as safe, _even if it uses pointers._

Otherwise it seems to me that this will become viral (similar to the trimming attributes), where a lot of APIs are considered unsafe, because they eventually do something unsafe in some code path.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+`RequiresUnsafe` is only useful if there is an accepted definition of what is considered unsafe. For .NET there are two properties that we already consider safe code to preserve:
+
+* Memory safety
+* No access to uninitialized memory
+
+In this document **memory safety** is strictly defined as: safe code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime.
+
+No access to uninitialized memory means that all managed memory is either never read before it has been initialized by C# code, or it has been initialized to a zero value.
+
+Some examples of APIs or features that are unsafe due to exposing uninitialized memory include:
```
rolfbjarne, line 53: I think `StructLayout.Explicit` can be unsafe as well.

proposed/caller-unsafe.md
agocke: That’s what unsafe does today. It lets you call unsafe things, but has no requirements on the callers.

olivier-spinelli: I'm a bit lost here.
Does this mean that to produce "safe" packages I'll have to get rid of all the code the uses Unsafe.As, CollectionsMarshal and others?
How can I "guaranty" to others that my implementation IS safe (even if I use a `ImmutableCollectionsLarshal.AsImmutableArray` in a single place in my code base)?
Will all my packages be categorized as "unsafe" because in one low-level assembly I do use a `Unsafe.As`?
Also, I fully agree with @rolfbjarne: FieldOffset are unsafe... 
So the question is: what code base remains on the "safe" side?

rolfbjarne: > That’s what unsafe does today. It lets you call unsafe things, but has no requirements on the callers.

Ah, I (think I) see, this new attribute would need to be added manually, it won't be added automatically whenever someone uses unsafe code.

EgorBo: > I'm a bit lost here. Does this mean that to produce "safe" packages I'll have to get rid of all the code the uses Unsafe.As, CollectionsMarshal and others? How can I "guaranty" to others that my implementation IS safe (even if I use a `ImmutableCollectionsLarshal.AsImmutableArray` in a single place in my code base)? Will all my packages be categorized as "unsafe" because in one low-level assembly I do use a `Unsafe.As`? Also, I fully agree with @rolfbjarne: FieldOffset are unsafe... So the question is: what code base remains on the "safe" side?

Does this doc mention marking packages as unsafe? I don't see it

proposed/caller-unsafe.md
EgorBo: What makes Explicit structs memory unsafe? Presumably, the runtime already guarantees that you can't mess up with the GC references (wrong alignment or aliased with some other type)

proposed/caller-unsafe.md
agocke: OK, sounds like we've reached consensus. @jaredpar do you want me to update this PR with the language changes? Or should I move stuff over to a different PR?

jaredpar: I'd say sketch out the general language changes here, if we go forward wit the design then we can write a more formal spec for csharplang.

hamarb123, line 28:
```suggestion
public sealed class RequiresUnsafeAttribute : System.Attribute
```

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,60 @@
...
+
+* ArrayPool.Rent
+* The `stackalloc` C# feature used with `SkipLocalsInit` and no initializer
```
hamarb123, line 56:
```suggestion
* The `stackalloc` C# feature with no initializer
```
`SkipLocalsInit` is irrelevant as `stackalloc`s are not guaranteed to be initialized.

EgorBo: that's not true. we consider any non-initialized stackalloc as a bug if no SkipLocalsInit is set. We might want to make an adjustment to the ECMA with that.

proposed/caller-unsafe.md
hamarb123: Do we count accessing improperly aligned memory as unsafe? It is reasonable for APIs to expect their `ref T` to be correctly aligned if they need that to work & not check it. If so (which I think would be reasonable), then we should treat it as `unsafe`, and also any use of any such field that isn't guaranteed to be correctly aligned. Explicit layout also allows accessing padding values as non-padding, which could read uninitialized memory.

proposed/caller-unsafe.md
hamarb123: This is what the c# standard specifies, it doesn't guarantee you get `localloc` which has the required guarantees.

proposed/caller-unsafe.md
richlander: @olivier-spinelli -- Fair question. As @EgorBo  says, that's not what is being communicated. In terms of helping people understand what we're thinking, this was probably a poor doc to go first. Expect some higher-level docs that better explain a plan.

proposed/caller-unsafe.md
EgorBo: > Explicit layout also allows accessing padding values as non-padding, which could read uninitialized memory.

not sure I follow, what uninitialized memory with explicit layout you're referring to?

hamarb123: > not sure I follow, what uninitialized memory with explicit layout you're referring to?

Hmm, I think you're right about this actually. I wasn't thinking clearly enough about definite assignment rules with this (I think I thought that you could just write 1 field in a union-like one & then immediately read out another, but you obviously can't do that). But the other concerns I mentioned still remain valid I think.

proposed/caller-unsafe.md
hamarb123: I personally think the `unsafecallers` should just be some seperate keyword, e.g., `public unsafecallers void X() => ...;` that way it's seperate from the existing `unsafe` keyword - I think it's also more descriptive "this code is unsafe" vs "it is unsafe to call this code".

Regardless, I think it would be good if we could have an `unsafe` expresssion, because 1. so that it can be used instead of how `unsafe` on methods is used today, 2. it allows limiting the scope of `unsafe` (thus having better safety checking) without forcing a bunch of lines (to declare the temp, do the unsafe operation + the 3 lines for unsafe, { and }), e.g., with the syntax suggested above: `public unsafe /*using the unsafe callers only meaning*/ void MyUnsafeFunctionThatUsesUnsafe() => /*unsafe expression*/ unsafe(...);`, and also `MyType x = myVariable.DoSomething(5, myObject, unsafe(*theOnlyUnsafeThing.field), () => ..., ...)` vs putting the whole function as unsafe so you don't have to add a bunch of lines for the small, contained unsafe operations or having to write this to get the same level of safety
```csharp
//this has the samed level of safety, but is much more verbose, hence people usually just make the whole method unsafe
SomeType tmp;
unsafe
{
    tmp = *theOnlyUnsafeThing.field;
}
MyType x = myVariable.DoSomething(5, myObject, tmp, () => ..., ...);
```
It would also be nice if you could declare a local unsafe, e.g.,
```csharp
unsafe int* myLocal = ...;
... //safe code that I want checked fully
...
...
...
...
//now I need to use myLocal again
int value = unsafe(*myLocal += 6);
```
These features would certainly vastly reduce the amount of code I have that is in an unsafe context.

proposed/caller-unsafe.md
rolfbjarne: > What makes Explicit structs memory unsafe? Presumably, the runtime already guarantees that you can't mess up with the GC references (wrong alignment or aliased with some other type)

You're right, aliasing objects and value types doesn't seem to work anymore:

>  System.TypeLoadException: Could not load type 'S' from assembly 'console, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' because it contains an object field at offset 0 that is incorrectly aligned or overlapped by a non-object field.

proposed/caller-unsafe.md
agocke: I added a line about P/Invoke. I agree P/Invoke is unsafe because an incorrectly written signature can produce arbitrary memory safety problems.

proposed/caller-unsafe.md
agocke: I updated the doc to use the `unsafe` keyword.

Unsafe expressions seems useful, but separate.

proposed/caller-unsafe.md
agocke: I agree that this is out of scope -- but an interesting challenge for a different document.

proposed/caller-unsafe.md
```diff
@@ -21,40 +21,83 @@ We want to achieve all of the following goals:
...
+### Globally-valid properties
 
-`RequiresUnsafe` is only useful if there is an accepted definition of what is considered unsafe. For .NET there are two properties that we already consider safe code to preserve:
+The overall goal is to ensure .NET code is "valid," in the sense that certain properties are always true. Generating a complete list of such properties is out of scope of this document. However, at least the following properties are required:
 
 * Memory safety
 * No access to uninitialized memory
 
-In this document **memory safety** is strictly defined as: safe code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime.
+In this document **memory safety** is strictly defined as: code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime.
```
richlander:
```suggestion
In this document **memory safety** is strictly defined as: code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime.
```

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+
+* Clearly annotate which methods require extra verification to use safely
+* Clearly identify in source code which methods are responsible for performing extra verification
+* Provide an easy way to identify in source code if a project doesn't use unsafe code
+
+## Proposal
+
+We need to be able to annotate code as unsafe, even if it doesn't use pointers.
+
+Mechanically, this would be done with a modification to the C# language and a new property to the compilation. When the compilation property "EnableRequiresUnsafe" is set to true, the `unsafe` keyword on C# _members_ would require that the call appear in an unsafe context. An `unsafe` block would be unchanged -- the statements in the block would be in an unsafe context, while the code outside would have no requirements.
```
jkotas, line 24:
```suggestion
Mechanically, this would be done with a modification to the C# language and a new property to the compilation. When the compilation property "EnableRequiresUnsafe" is set to true, the `unsafe` keyword on C# _members_ would require that their use appear in an unsafe context. An `unsafe` block would be unchanged -- the statements in the block would be in an unsafe context, while the code outside would have no requirements.
```
This should use a more general term to cover things like fields or delegate construction.

For example:
```
class MyType
{
    unsafe int _x;

    int M() => _x; // error, _x requires unsafe context
}
```

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+unsafe void M() { }
+```
+
+In the case of `Caller1`, the call to `M()` doesn't produce an error because it is inside an unsafe context. However, calls to `Caller1` will now produce an error for the same reason as `M()`.
+
+`Caller2` will also not produce an error because `M()` is in an unsafe context. However, this code creates a responsibility for the programmer: by presenting a safe API around an unsafe call, they are asserting that all safety concerns of `M()` have been addressed.
+
+Notably, unsafe did not change the requirement that the code in the block must be correct. It merely offset the responsibility from the language and the runtime to the user in verification.
+
+For more precise details on the error semantics of unsafe blocks and unsafe members, the rules will mirror the rules defined for "Requires" attributes defined in [Feature attribute semantics](https://github.com/dotnet/runtime/blob/main/docs/design/tools/illink/feature-attribute-semantics.md#requiresfeatureattribute). The only addition is the presence of the `unsafe` block, which effectively provides a local `Requires` context.
```
jkotas, line 60: Behavior of "Requires" attributes differs between classes and structs. Is it desirable to match that in this C# language feature?

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+```
+
+### Globally-valid properties
+
+The overall goal is to ensure .NET code is "valid," in the sense that certain properties are always true. Generating a complete list of such properties is out of scope of this document. However, at least the following properties are required:
+
+* Memory safety
+* No access to uninitialized memory
+
+In this document **memory safety** is strictly defined as: code can never acquire a reference to memory that is not managed by the application. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime, or memory that is acquired for legal use by the runtime through any other means.
```
jkotas, line 80:
```suggestion
In this document **memory safety** is strictly defined as: code can never acquire a reference to memory that is not managed by the runtime. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime, or memory that is acquired for legal use by the runtime through any other means.
```

proposed/caller-unsafe.md
agocke: Do we anticipate any future IL analysis here? That’s one of the things that makes structs tricky.

jkotas: I see this as a C# language feature. I do not think it makes sense to think about IL analysis in the context of C# language feature.

rolfbjarne (approved):

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+unsafe void M() { }
+```
+
+In the case of `Caller1`, the call to `M()` doesn't produce an error because it is inside an unsafe context. However, calls to `Caller1` will now produce an error for the same reason as `M()`.
+
+`Caller2` will also not produce an error because `M()` is in an unsafe context. However, this code creates a responsibility for the programmer: by presenting a safe API around an unsafe call, they are asserting that all safety concerns of `M()` have been addressed.
+
+Notably, unsafe did not change the requirement that the code in the block must be correct. It merely offset the responsibility from the language and the runtime to the user in verification.
+
+For more precise details on the error semantics of unsafe blocks and unsafe members, the rules will mirror the rules defined for "Requires" attributes defined in [Feature attribute semantics](https://github.com/dotnet/runtime/blob/main/docs/design/tools/illink/feature-attribute-semantics.md#requiresfeatureattribute). The only addition is the presence of the `unsafe` block, which effectively provides a local `Requires` context.
```
jjonescz, line 60: It looks like RequiresFeature applied to a class has an effect only on static members of the class. That's not what I expected for unsafe nor how it behaves today in C#. Is that intentional?

hamarb123 (commented):
I've left some comments primarily of questions about this precise approach. I'm not certain what I think the correct answer to them is, but I personally think they are important questions that need some consideration, so I've left them with some of my thoughts about them for everyone to think about to ensure we come up with the best version of the feature possible.

Thanks if you read the whole of my comments 😅

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Implementation
+
+In addition to compiler enforcement, the following attribute will be added for annotating unsafe members. It is an error to use this attribute directly in C#. Instead, the `unsafe` keyword should be used.
+
+```C#
+[AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor)]
```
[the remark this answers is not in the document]
jjonescz: > a type itself cannot be unsafe

FWIW, pointer types themselves are an example of "unsafe types". For example, it might be considered inconsistent if you couldn't create your custom `struct MyPointer` with same unsafe semantics as `int*`.

Anyway, how would it work if RequiresUnsafe wouldn't work for types? You can already declare types as unsafe in C# today and that propagates to its members (except iterators) automatically. If you excluded unsafe types from RequiresUnsafe semantics, would you require users to mark the members separately as unsafe? Or just error on unsafe types?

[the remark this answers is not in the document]
hamarb123: > > a type itself cannot be unsafe

> FWIW, pointer types themselves are an example of "unsafe types".

This was explaining how it works in rust, not how it works today in c#:

> my understanding is that rust doesn't allow this for the following reason: a type itself cannot be unsafe, only operations can be

In rust, the following function is not unsafe-callers-only: `fn a(x: *const i32)` ([link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c72823ae8db7ed8fc62010d80f974126)), however, attempting to perform an operation on the pointer does require `unsafe` ([link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c653cc3713bb0056892e8f7e3d53c33d)).

I think this is the ideal setup, but I'm not certain it's possible to actually do in c#, since we have `default` on `struct`s (and `new()`), which might represent an `unsafe` state on structs (rust avoids this problem, since `default` is a trait & you are not required to implement it). It is probably possible to make this work w/o requiring unsafe-callers-only types if we can mark for `default` for a `struct` being unsafe & if we can keep track of where `default` is, or may (e.g., in generics) be, used on it & complain somewhere about doing that (for generics, this would probably be at passing as generic parameter time, unless it's marked with some `allows nondefault` anti-constraint).

> Anyway, how would it work if RequiresUnsafe wouldn't work for types? Users can already declare types as unsafe in C# today and that propagates to its members (except iterators) automatically.

It also doesn't propogate to `async` members. I personally think marking something unsafe-callers-only should be a seperate toggle to marking it unsafe-impl anyway (which is what the `unsafe` keyword does in all positions today). Since we are considering massively changing the meaning of the `unsafe` keyword means anyway, I think we should change it to whatever makes most sense. I know for a fact that all the types I've added `unsafe` to were just for convenience, since they had a lot of members that worked with pointers, but it didn't make the *types* themselves *necessarily* actually `unsafe` in *all* possible contexts - I suspect this is also the case for most (or all, excepting the `default`/`new()` thing on structs I mentioned earlier) other uses.

Additionally, the current proposal doesn't even make marking a type with `unsafe` keep the same members having an unsafe context - it explicitly excludes most instance members on class types, explicitly excludes nested types, and includes any generator & async functions not otherwise excluded - these are all differences to what gets an unsafe-impl-context today from the `unsafe` keyword on a type.

It also seems to differ here
> Note that this does not create a feature requirement for nested types or for members of base classes or interfaces implemented by the type.

Since when you mark a type unsafe-impl today these members do get an unsafe-impl context, but if we change the meaning to getting an unsafe-callers-only context, they shouldn't get that (which makes sense for base/interface members that aren't unsafe-callers-only, but is also another change with what putting `unsafe` on a type means today).

Also:

> [AttributeUsage(AttributeTargets.Class | AttributeTargets.Struct | AttributeTargets.Method | AttributeTargets.Property | AttributeTargets.Constructor)]

These targets also seem wrong imo, I didn't notice it earlier as I must have skimmed over that line, but e.g., if I have a pointer field (or any other type we mark unsafe-callers-only), does that require me to mark the whole type unsafe-callers-only, even if it has safe operations on it & just happens to use pointers or other unsafe functionality? Do we need to strictly disallow unsafe-callers-only types as instance fields in classes, since we cannot mark them as unsafe-callers-only, and the current spec specifies that it doesn't get passed down to instance members in classes (other than constructors), thus making it impossible to have them as instance fields?

> Or just error on unsafe types when EnableRequiresUnsafe is enabled for project?

If we change the meaning of `unsafe` on a type to what the current spec specifies (of which I pointed out a number of differences above), then imo we should at least warn at every pre-existing usage to say "this now means something completely different to what it did before, fix it" anyway.

Lastly, https://github.com/dotnet/designs/pull/330#discussion_r1986013375

> If we do not do this, we will have to be explaining why `unsafe` in C# is non-sensical and different from Rust. When we have to be explaining, we are failing.

I think it is inevitable that we will end up with some differences to rust (due to the reasons I'm about to explain), but we could make it the same if we really wanted to, but this would require a massive set of changes to `unsafe` - we would need to make pointers safe but operations on them unsafe, we would need to make `unsafe` mean `unsafe-callers-only`, and we would need to disallow `unsafe` on the class/struct level. This is probably the best solution, but I haven't mentioned it explicitly until now as I see it as too breaking, but if we're already breaking whether `unsafe` means unsafe-impl or unsafe-callers-only, then I don't see any good reason to stop halfway at fixing what `unsafe` means & how it works. The reason I proposed making `unsafecallersonly` a seperate keyword initially, was purely because I wanted to avoid making a massive break in what the `unsafe` keyword means, but if we're going to change its meaning anyway as much as is currently proposed, then we should do it right imo (& I think rust's solution is more or less the ideal one - we probably need to make a few adjustments for c# (like allowing on properties, events, fields, constructors, & default in addition to methods - as these are all achieved in different ways in rust), but I think it's pretty much right).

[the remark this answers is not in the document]
agocke: > Tbh, I don't think it should be allowed on a type level - my understanding is that rust doesn't allow this for the following reason: a type itself cannot be unsafe, only operations can be, and I personally think that is pretty sensible.

In Rust impls can be unsafe, which serves a similar purpose here. C# does not have a meaningful difference between type declaration and implementation.

We will need some way to handle the following problem:

```C#
class UnsafeEnumerable : IEnumerable<T>
{
   IEnumerator<T> GetEnumerator() => ...; // use unsafe code
}
```

We have to guard virtual invocations (interface methods), otherwise you can trivially bypass unsafe. But we can't mark `IEnumerable<T>` itself unsafe. So we need some way to handle `UnsafeEnumerable`. The way we do it with trimming is that you mark the type itself, which in turn marks all the constructors, but hides errors for the instance members. Because if you managed to construct the type, you already used `unsafe`.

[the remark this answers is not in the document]
hamarb123: > C# does not have a meaningful difference between type declaration and implementation.

Right, so the purpose of marking a type unsafe-callers-only is to indicate that 1 or more of the interfaces it implements is implemented unsafely? That is really not clear from anything imo (not clear syntactically, also not clear from this document), but I suppose it makes sense if the intent is that it's the only reason you'd use it.

proposed/caller-unsafe.md
huoyaoyuan: There is also a widely-seen formalism to disallow `unsafe` syntax totally, instead using the `Marshal` family which is inefficient and typically more unsafe. We should provide a path for moving people away from that.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+public sealed class RequiresUnsafeAttribute : System.Attribute
+{
+}
+```
+
+### Globally-valid properties
+
+The overall goal is to ensure .NET code is "valid," in the sense that certain properties are always true. Generating a complete list of such properties is out of scope of this document. However, at least the following properties are required:
+
+* Memory safety
```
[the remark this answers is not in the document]
jkotas: > Type safety and GC safety should get highest severity because of the ability to break runtime

Nearly all memory safety bugs can crash the runtime. For example, incorrect use of `NativeMemory.Alloc/Free` can cause very hard to diagnose crash in the runtime code as well.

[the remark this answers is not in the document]
agocke: The safety property I suggested, memory safety, is derived from an analysis of severe security breaches across the industry. Things in this category are known to be very dangerous and award attackers with significant power. I don’t think there’s further subdivision of this category that is interesting. Unsafe.As and Unsafe.AsPointer are both capable of producing memory safety problems and are therefore both unsafe.  

The only other property I’m suggesting is viewing uninitialized memory. This is less severe. We could drop it if necessary. At the moment I’m proposing no other properties that would be covered by the unsafe feature and therefore further categorization is irrelevant.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+    {
+        M();
+    }
+}
+unsafe void M() { }
+```
+
+In the case of `Caller1`, the call to `M()` doesn't produce an error because it is inside an unsafe context. However, calls to `Caller1` will now produce an error for the same reason as `M()`.
+
+`Caller2` will also not produce an error because `M()` is in an unsafe context. However, this code creates a responsibility for the programmer: by presenting a safe API around an unsafe call, they are asserting that all safety concerns of `M()` have been addressed.
```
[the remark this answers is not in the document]
agocke: This document doesn’t just present an annotation, it presents a formalism. By definition, safe code cannot cause memory safety issues because the runtime guarantees the safety. This means that applying unsafe is not a judgment call — if code can cause a memory safety issue, it needs to be unsafe. If not, it does not.

If too much code is unsafe, we would have to narrow the property we’re attempting to prove. Otherwise the formalism is unsound.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+
+* Clearly annotate which methods require extra verification to use safely
+* Clearly identify in source code which methods are responsible for performing extra verification
+* Provide an easy way to identify in source code if a project doesn't use unsafe code
+
+## Proposal
+
+We need to be able to annotate code as unsafe, even if it doesn't use pointers.
+
+Mechanically, this would be done with a modification to the C# language and a new property to the compilation. When the compilation property "EnableRequiresUnsafe" is set to true, the `unsafe` keyword on C# _members_ would require that their uses appear in an unsafe context. An `unsafe` block would be unchanged -- the statements in the block would be in an unsafe context, while the code outside would have no requirements.
```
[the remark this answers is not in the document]
agocke: > Would EnableRequiresUnsafe be something that is enabled by default at some point in the future?

This is a breaking change in the language, so that depends on the observed severity of the break.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+### Globally-valid properties
+
+The overall goal is to ensure .NET code is "valid," in the sense that certain properties are always true. Generating a complete list of such properties is out of scope of this document. However, at least the following properties are required:
+
+* Memory safety
+* No access to uninitialized memory
+
+In this document **memory safety** is strictly defined as: code can never acquire a reference to memory that is not managed by the runtime. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime, or memory that is acquired for legal use by the runtime through any other means.
+
+No access to uninitialized memory means that all memory is either never read before it has been initialized, or it has been initialized to a zero value.
```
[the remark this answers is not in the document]
agocke: I currently don’t see a reason why that method needs to be unsafe, but we could choose to apply the marker anyway.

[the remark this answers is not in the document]
jkotas: `GetUninitializedObject` is private reflection. It should get the same treatment as the rest of private reflection once we figure out what to do about it.

proposed/caller-unsafe.md
agocke: It does have an effect, look at the section above:

> When a class or struct has a feature requirement, the following members are in a feature available scope:

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+These properties are guaranteed by "safe" code through a combination of compiler and .NET runtime enforcement. For `unsafe` code, these properties must be guaranteed by the programmer.
+
+`unsafe` members are used to identify the places that cannot be automatically checked by the compiler and runtime for validity. Inside unsafe blocks, the programmer is responsible for ensuring that all requirements of the unsafe code are met, and that all code outside the block will have validity properly enforced by the system.
+
+
+### Examples and APIs
+
+**ArrayPool.Rent**
+
+This method is unsafe because it returns an array with unintialized memory. Code must not read the contents of the returned array without initialization.
```
stephentoub, line 95: How is uninitialized defined here? Is this because the initial array is allocated with GC.AllocateUninitializedArray, or is this because someone can rent an array, write data into it, return it, and then someone ride renting gets that data?

If it's the former, I'd rather just change the Rent implementation to use new[] instead of AllocateUninitializedArray. If it's the latter, that won't help.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
+
+# Annotating members as `unsafe`
+
+## Background
+
+C# has had the `unsafe` feature since 1.0. There are two different syntaxes for the feature: a block syntax that can be used inside methods and a modifier that can appear on members and types. The original semantics only concern pointers. An error is produced if a variable of pointer type appears outside an unsafe context. For the block syntax, this is anywhere inside the block; for members this is inside the member; for types this is anywhere inside the type. Pointer operations are not fully validated by the type system, so this feature is useful at identifying areas of code needing more validation. Unsafe has subsequently been augmented to also turn off lifetime checking for ref-like variables, but the fundamental semantics are unchanged -- the `unsafe` context serves only to avoid an error that would otherwise occur.
```
stephentoub, line 6: > The original semantics only concern pointers.

It also included things adjacent, like sizeof(T). As part of this effort, I'd like us to revisit those choices. Based on the direction in this doc, seems sizeof(T) should not be unsafe.

proposed/caller-unsafe.md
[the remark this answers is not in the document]
agocke: Good feedback, I think this doc needs some more examples and explanation

richlander:
Can you add a section on adoption? I imagine that we'll start marking existing APIs as unsafe, for example in .NET 11. Some users may want to move to .NET 11 but not want to deal with any of these new markings, at least to avoid blocking their .NET 11 adoption. This is somewhat similar to nullable, although that was opt-in.

Users should be able to correct their code progressively or (if they choose) not at all.

proposed/caller-unsafe.md
agocke: The former -- uninitialized means uninitialized by the "application" vs. the OS. If the pool used regularly .NET array allocation it would be zero-init and considered initialized.

agocke:
I do have a mention of the mechanics of adoption at the start:

> Mechanically, this would be done with a modification to the C# language and a new property to the compilation. When the compilation property "EnableRequiresUnsafe" is set to true, the `unsafe` keyword on C# _members_ would require that their uses appear in an unsafe context. An `unsafe` block would be unchanged -- the statements in the block would be in an unsafe context, while the code outside would have no requirements.

I don't have much of a plan beyond this. I'm also not sure this document is the right place to put that info, as tooling may change over time even if the change in language semantics remains the same.

proposed/caller-unsafe.md
jkotas: Uninitialized memory is not the only reason why `ArrayPool.Rent/Return` should be marked as unsafe since (the lifetime of) the memory returned by `ArrayPool.Rent/Return` is not managed by the runtime. Is this your interpretation of "managed by the runtime"?

richlander:
I see. I slightly misinterpreted what that was saying (my issue, not the text).

Nullable offers fine-grained control. We can likely wait for feedback on how important that is for this project.

KathleenDollard (commented):
First, it is fantastic that for all the complex conversations you have had on this, the document is rather short and straightforward. I read this as an informed, but "not us" kind of reader.

I made a couple of comments where a word or two might clarify things for readers that aren't us.

I do think the document could use with one more section, possibly just a paragraph. I'm not sure it's clear how this impacts users. Working from the goals:

* Annotations are covered well
* _Users receiving error when unsafe code is called outside an unsafe context_ is not a goal, but might be considered
* It's not obvious to me how this "clearly identifies ...methods ...responsible...for extra validation". The method responsible is the one with the unsafe block, and it seems there would be no marker on the method
* It's not obvious how this helps a user identify if a project does not use unsafe code. If you just means searching for `unsafe` is more reliable, that's probably worth noting. 

Is there an effort starting to have NuGet share whether `RequiresUnsafe` is set on packages? It seems that the confidence one could have in their lack of unsafe calls is only as good as the confidence they have that their libraries don't use unsafe, or at least use it with an awareness beyond pointers.

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+
+* Clearly annotate which methods require extra verification to use safely
+* Clearly identify in source code which methods are responsible for performing extra verification
+* Provide an easy way to identify in source code if a project doesn't use unsafe code
+
+## Proposal
+
+We need to be able to annotate code as unsafe, even if it doesn't use pointers.
+
+Mechanically, this would be done with a modification to the C# language and a new property to the compilation. When the compilation property "EnableRequiresUnsafe" is set to true, the `unsafe` keyword on C# _members_ would require that their uses appear in an unsafe context. An `unsafe` block would be unchanged -- the statements in the block would be in an unsafe context, while the code outside would have no requirements.
```
jkotas, line 24: > compilation property "EnableRequiresUnsafe"

Alternatively, this can be exposed as a new value for the existing `AllowUnsafeBlocks` property. For example, `<AllowUnsafeBlocks>strict</AllowUnsafeBlocks>`.

jaredpar: I much prefer this over adding another msbuild property into the equation.

agocke: I have no strong opinion on the name. However we may also want a property for “use caller-unsafe behavior, but don’t allow unsafe blocks.” That would catch the use of unsafe methods in referenced code, but would still have the effect of disallowing unsafe in the project.

richlander: `verystrict`

proposed/caller-unsafe.md
```diff
@@ -0,0 +1,103 @@
...
+
+* Memory safety
+* No access to uninitialized memory
+
+In this document **memory safety** is strictly defined as: code never accesses memory that is not managed by the runtime. "Managed" here does not refer to solely to heap-allocated, garbage collected memory, but also includes stack-allocated variables that are considered allocated by the runtime, or memory that is acquired for legal use by the runtime through any other means.
+
+No access to uninitialized memory means that all memory is either never read before it has been initialized, or it has been initialized to a zero value. This term is defined relative to the application context, not the system context. Therefore, allocating memory using a mechanism where .NET guarantees that the memory has been zeroed would be considered initialized. Reading memory from a function which does not guarantee zero initialization would not. Reusing memory that has previously been initialized but now may hold an unknown value due to previous application execution is considered safe. This may be incorrect logic from the application's perspective, but is valid code according to .NET.
```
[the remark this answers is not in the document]
jkotas: "This term is defined relative to the application context, not the system context." needs better explanation. 

Is it trying to make a distinction between unknown values produced locally vs. globally? How about this:

Reusing memory that has previously been initialized but now may hold an unknown value due to previous application execution is considered safe in a local scope. This may be incorrect logic from the application's perspective, but is valid code according to .NET.

If the unknown values were produced by code in your package or component, reusing such memory is considered safe. The consequence is that local object pools are considered safe.

If the unknown values were produced by code that may be unrelated to your package or component, reusing such memory is considered unsafe. The consequence is that global allocators or global object pools that return uninitialized memory such as `GC.AllocateUninitializedArray` or `ArrayPool<byte>.Shared` are considered unsafe.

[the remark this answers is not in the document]
EgorBo: > but the potential severity of misuse with ArrayPool<byte> is more limited

It's probably quite the opposite in fact if we talk about ArrayPool specifically. It's used in many places (a lot of uses inside the base class library and many in various 3P) and a single mis-use in one place is enough for any of those places to corrupt the global state in some non-obvious way. Recently, someone filed a bug against `DefaultInterpolatedStringHandler` in the runtime and it turned out it was a mis-use of ArrayPool in user's code. https://github.com/dotnet/runtime/issues/118292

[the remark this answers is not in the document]
mriehm: To be clear, though, the risks with `ArrayPool` aren't only related to the lack of zero initialization that this document is using to form the definition `unsafe`. Suppose that some library creates a public static `ZeroInitArrayPool` that zero-initializes each array before renting it out to a consumer. It sounds like that does not fall within any definition of `unsafe` being discussed so far. Yet, in the case that a consumer accidentally writes to an array after calling `ArrayPool.Return()`, or a consumer accidentally calls `Return()` twice, then we still have the problematic corruption of the buffer for each consumer.

It seems that what we really want to protect against is widely shared mutable state. For example, I haven't used Rust, but I believe their compiler tracks object ownership transfer and prevents object use after ownership has been transferred away (e.g., after `ArrayPool.Return()`). I would be excited if a similar ownership concept were ever added to C#, though I won't pretend to understand language design enough to know if it could integrate well with the rest of C#.

BTW, thanks for the time you took to respond to my comments. I definitely agree that `ArrayPool` is unsafe in the colloquial sense, and it would be great to find ways to mitigate that. I'm only skeptical that `unsafe` scopes are an appropriate or effective tool for that mitigation.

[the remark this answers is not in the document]
agocke: One alternative might be careful phrasing about defined values. That is, uninitialized memory from the heap or stack have undefined values. Memory handled by .NET should have defined values as long as you stay outside of undefined behavior, which is mostly unsafe code anyway. The tricky bit is race conditions (again). Technically the value should be defined, but it may be multiple possible defined values.

proposed/caller-unsafe.md
agocke: As currently designed, we don’t have a way to properly categorize safe lifetime management. So, for the purposes of this doc, all of the lifetime issues are considered safe. Unfortunately that does leave a large category of safe constructs that are known to have security impact, but I’ll leave that in the same category as directory traversal right now, I.e. out of support of the current proposal.

proposed/caller-unsafe.md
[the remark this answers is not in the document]
agocke: > If the unknown values were produced by code in your package or component, reusing such memory is considered safe. The consequence is that local object pools are considered safe.

This is an interesting path. It relies on a somewhat fuzzy definition of local vs global, but is still pretty well understood in practice. One unfortunate property is that it builds an unsafe construct out of purely safe components. I don’t particularly like this, but it may be better than the alternatives.

agocke:
I think this PR has had good feedback, but is starting to pull in a lot of stuff. I'm going to merge this PR now, and we should address further changes through further PRs.

PR #332, opened 2025-04-15
Project simplification proposal

KathleenDollard:


proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
+# Project file simplification and improved change management
+
+// Contents of Type should match template name as much as possible
+
+## Summary
+[summary]: #summary
+
+This proposal is to simplify the project file while giving the user better control over breaking changes and new warnings.
+
+The approach is versioned default properties specified by a `Base` attribute in the project file, a `Directory.Build.props`, or an `Import` file. This attribute specifies a version of the file containing the default properties. The project file is further simplified by `Type` attribute on the `Project` element. The `Type` is not versioned, although versioned default properties can be added as part of the `Base` evaluation. The `Type` attribute replaces the current `SDK` and the `Type` file contains both the `SDK` name and properties that apply to the application `Type`.
```
jaredpar, line 10: A worry I have about `Type` is that it's likely going to create _dialects_ of SDK projects. Essentially, are you using old school `Sdk` or new `Type`. I'm not sure the value of this attribute outweighs the cost of having competing dialects in the ecosystem. 

I don't have the same worry about `Base`. I see that as say `SdkVersion` which is something that large repos will care about and will be in centralized places, much like say CPM is today. It will be a _big repo_ property but likely won't impact most .NET development. An _expert_ feature.

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+```
+
+If there is a `Directory.Build.props` or `Import` file, it can contain the `Base` attribute, although it cannot contain the `Type` attribute. This is because the `Type` attribute specifies the SDK instructions for build evaluation.
+
+There would be no change to existing projects. They will continue to work as today.
+
+## Motivation
+[motivation]: #motivation
+
+The fundamental benefit of this effort is to speed up and remove friction for users that wish to adopt tooling improvements quickly, while also allowing users to delay incorporating change until it is not disruptive. To do this, we need to allow developers a friction free experience on upgrade, which means that with no more work than previously they have control over the upgrade process.
```
jaredpar, line 27: What does friction free mean here though? There are always behavior differences when updating between major versions of the .NET SDK. Are we saying that all behaviors are tied to `Base`, just diagnostics, etc ... Need to be clear where the line is with this.

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+All breaking changes and new warning are introduced because we believe the positive impact outweighs the pain of the change. If a change is not breaking and does not add new warnings, we just add the feature and they are available when the user upgrades. If the change breaks everyone, we are unlikely to make the change. This proposal affects the occasional features that may break a few projects some of the time.
```
jaredpar, line 29: If we hide every breaking change behind this flag what does installing a new major version of the SDK actually do when `Base` points to the previous version? Is it expected to be 100% a no-op? If so then why are customers upgrading? What is the experience that they are expecting?

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+The break may be a new warning. We treat new warnings as breaking changes because they can be disruptive to projects that rely on `TreatWarningsAsErrors` to guard against regressions in their own code. We want to make it easy for users to incorporate changes and warnings as quickly and smoothly as possible to achieve these benefits. An example of a new feature that created new warnings and clearly benefits almost all projects is the `NuGetAudit` flag which informs you if you are using a vulnerable package directly or transitively.
```
jaredpar, line 31: Agree there is value in having new warnings across our tooling stack tied to a single property. But we have that today with `SdkVersion` so have to ask why is that not working but we think that `Base` will?

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+We realize that your capacity to accommodate change and possible changes to your code varies widely by team, time of year, business considerations, and whether your already working on the code. These changes also affect all of the projects you are working on or building. Thus, breaks and warning that occur on tooling upgrades are undesirable.
+
+Your approach to upgrading the `TargetFramework` of your project varies from wanting the newest features and performance improvements as quickly as possible to postponing as long as possible to ease the upgrade effort.
+
+### Motivation for `Base` attribute
```
jaredpar, line 37: Why does `Base` need to be an attribute? Think we need to be specific about that here. Not having it as a property seems like it will be a problem:

1. Cannot test changes to a repo by using `-p:Base=11.0.0` on the command line
2. Will not be usable in non-SDK projects who do not have these attributes

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+```XML
+<Project Base="10.0.100" 
+         Type="Console">
+    <PropertyGroup>
+        <TargetFramework>net11</TargetFramework>
+    </PropertyGroup>
+</Project> 
+```
+
+For users that are not ready to upgrade their `TargetFramework` but want any new warnings that may improve their application, they can use a higher version of tooling defaults:
```
jaredpar, line 85: The `Base` version is the same in all three examples here. Did you mean for it to change in this example (based on this text)?

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+<Project> 
+    <PropertyGroup>
+        <TargetFramework>net10</TargetFramework>
+    </PropertyGroup>
+</Project> 
+```
+
+In this case, `TargetFramework` should be set to `net11` because it is closer to te user.
+
+We are particularly concerned about precedence being predictable because the results may be subtle or _spooky action at a distance_ because the properties set may affect a very different part of the build. Also, debugging MSBuild via binlogs and the structured log viewer is a skill many of our users do not have because they rarely or never need to debug MSBuild logic.
```
jaredpar, line 144: Think you need to illustrate why this would not work if `Base` were a property. also this is likely to be _expert only_ territory so why not say that `Base` must be a property and it can only be set in a Dir.Build.props file? There can be an error if it's modified anywhere else.

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+We will add two new attributes to project files. This will be an optional/alternative way to declare the SDK and the property values that were previously in the project file when created from templates. The existing approach to project files will remain unchanged.
+
+In addition, we will provide a mechanism for declaring a default property value. When this file is read a check will be made - if the property has already been set, the existing value will be retained and no action will be performed.
+
+#### `Type` attribute
+
+_The name `Type` is not final._
+
+The `SDK` attribute on the `Project` element can be replaced by a `Type` attribute. It is not legal to have both a `Type` and an `SDK` attribute. If neither exists, the project will behave in the same manner as projects without a `SDK` attribute behave today.
```
jaredpar, line 266: For Type to replace Sdk think we need some very strong motivating examples cause it will create dialects of the language.

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+The key element of the `Base` design is that it is versioned and released with each SDK. This allows you to initially experiment with, and then adopt all the breaking changes and new warnings with a single value. The proposed structure of this version is the corresponding SDK version number, such as `10.0.100`, with a possible simplification to just `10`. The `Base` defaults include the `TargetFramework` so the user manages a single version number, in most cases.
+
+However, key to this design is that `Base` is not required to match either your SDK or your `TargetFramework`. For example, if you use an LTS strategy and want to ensure you get the latest warnings to benefit your code and prepare for future upgrade, you might have:
+
+```XML
+<Project Base="11.0.100" 
+         Type="Console"> 
+   <PropertyGroup>
+        <TargetFramework>net10<TargetFramework>
```
akoeplinger, line 51:
```suggestion
        <TargetFramework>net10.0<TargetFramework>
```

this applies throughout the doc

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+When a there is a breaking change or a new warning, there is a new MSBuild property (as today). For breaking changes or new warnings, the feature will be turned off if it `false` or the equivalent `disable`. The defaults specified by `Base` will turn on the features we strongly recommend. If you disagree with this for your project, you can reset the value to `false`.
+
+The key element of the `Base` design is that it is versioned and released with each SDK. This allows you to initially experiment with, and then adopt all the breaking changes and new warnings with a single value. The proposed structure of this version is the corresponding SDK version number, such as `10.0.100`, with a possible simplification to just `10`. The `Base` defaults include the `TargetFramework` so the user manages a single version number, in most cases.
```
webreidi, line 43: If I simplify to just "10", does that take the earliest or latest 10.x.x?

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+It is an error for a `Type` attribute to appear anywhere but a project file `Project` element. For example, it cannot appear in `Directory.Build.props`, The reason for this restriction is that the logic run when the `Type` attribute appears determines the SDK (the tasks and targets) that are used. `Directory.Build.props` is located as part of an SDK task.
+
+There is a file delivered as part of the SDK that provides the `Type` information. The effect of this file will be the same as if it was a props file with an `SDK` attribute on the `Project` element and a series of properties. It is an error for a `Type` file to be missing the SDK, to contain any MSBuild element except `PropertyGroup`. Since we will create and deliver these files, we might not do the work to enforce this restriction. The purpose of this restriction is that SDKs already allow adding information to the build, and anything more complex is probably better done via an SDK.
```
martincostello, line 272:
```suggestion
There is a file delivered as part of the SDK that provides the `Type` information. The effect of this file will be the same as if it was a `.props` file with an `SDK` attribute on the `Project` element and a series of properties. It is an error for a `Type` file to be missing the SDK or to contain any MSBuild element except `PropertyGroup`. Since we will create and deliver these files, we might not do the work to enforce this restriction. The purpose of this restriction is that SDKs already allow adding information to the build, and anything more complex is probably better done via an SDK.
```

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+<Project Base="11.0.100" 
+         Type="Console">
+    <PropertyGroup Label="These properties added on upgrade from 10.0.100 to 11.0.100"
+       <UseFeatureA>disable</UseFeatureA>
+       <UseFeatureB>disable</UseFeatureB>
+    </PropertyGroup>
+</Project> 
+```
+
+This approach is simple for a project file. It might be extended to complex repos by placing the `PropertyGroup` containing the new properties at the start of whichever file (project file, `Directory.Build.props` or `Import` file) the `Base` attribute appeared in. However, the command would probably be run in the same directory as the file containing the `Base` attribute and specify the file if there were multiple props files.
```
martincostello, line 322:
```suggestion
This approach is simple for a project file. It might be extended to complex repos by placing the `PropertyGroup` containing the new properties at the start of whichever file (project file, `Directory.Build.props` or `Import` file) the `Base` attribute appeared in. However, the command would probably be run in the same directory as the file containing the `Base` attribute and specify the file if there were multiple `.props` files.
```

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+|Command|Result|
+|-|-|
+|dotnet project upgrade 11.0.100 --whatif|List the properties that would be added, along with any inline comments|
+|dotnet project upgrade 11.0.100 --preserve|Actually run the upgrade, resulting in new properties to match previous behavior|
+|dotnet project upgrade 11.0.100 --no-preserve|Actually run the upgrade, without new properties|
+
+## Drawbacks
+[drawbacks]: #drawbacks
+
+This introduces a new concept to .NET users, and some additional complexity in the overall MSBuild evaluation. Users would see this in areas such as understanding binlogs. We think the trade off is worthwhile because users get small clean project files the ability to control new warnings/breaking changes independent of both SDK version and TargetFramework.
```
martincostello, line 335:
```suggestion
This introduces a new concept to .NET users, and some additional complexity in the overall MSBuild evaluation. Users would see this in areas such as understanding binlogs. We think the trade off is worthwhile because users get small clean project files with the ability to control new warnings/breaking changes independent of both SDK version and `TargetFramework`.
```

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Alternatives
+[alternatives]: #alternatives
+
+We could continue to tie new features to TFM. THe drawback is that users generally upgrade their TFM to keep their projects in support. While this avoids breaking changes/warnings occurring on tooling upgrade, it may either delay adopting desirable new features, or result in breaking changes/warnings happening when the user is just trying to stay in support.
```
martincostello, line 340:
```suggestion
We could continue to tie new features to TFM. The drawback is that users generally upgrade their TFM to keep their projects in support. While this avoids breaking changes/warnings occurring on tooling upgrade, it may either delay adopting desirable new features, or result in breaking changes/warnings happening when the user is just trying to stay in support.
```

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+## Open questions
+[open]: #open-questions
+
+### What capacity to pin and float should we offer for `Base` versions?
+
+### What about multi-targeting?
```
martincostello, line 347: This was the main question I had going in 😃. For example, in November the _current_ plan is to ship Swashbuckle.AspNetCore multi-targeting `net8.0`, `net9.0` and `net10.0` (plus `netstandard2.0` where relevant, but I'm tempted to drop this...). It's quite a large codebase (at least compared to my own), so it would be interesting to see what the challenges (or not) of having .NET 10 "opinions" applied to the code is, particularly if there are any features enabled that don't work down-level. As an example of this, today we use `#if NET` (or similar) defines in various places to make analysers happy about using newer/better APIs that aren't available to `netstandard2.0`.

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+### Should we restrict `Type` to the project file?
+
+Stated differently, should we allow `Type` to be in  `Directory.Build.props`. Since `Type` determines the SDK and that contains the targets that run most build operations, this is tricky. Also, it may be best for users to see this in the project file for clarity.
```
martincostello, line 351:
```suggestion
Stated differently, should we allow `Type` to be in `Directory.Build.props`? Since `Type` determines the SDK and that contains the targets that run most build operations, this is tricky. Also, it may be best for users to see this in the project file for clarity.
```

ClxS:
I feel like one thing which needs to be outlined here is how does adding this actually simplify the typical project structure? From my POV this just adds an additional way to achieve what system is already capable of, while adding additional complexity in that there is yet another way to achieve the same goals.

jcotton42:
Something that isn't clear to me is how does `Type` work with SDKs pulled down from nuget?

From https://learn.microsoft.com/en-us/dotnet/core/project-sdk/overview#project-files

> To specify an SDK that comes from NuGet, include the version at the end of the name, or specify the name and version in the global.json file.
```xml
<Project Sdk="MSBuild.Sdk.Extras/2.0.54">
  ...
</Project>
```

proposed/project-file-simplification.md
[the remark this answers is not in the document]
jkoritzinsky: Technically, the `Sdk="Microsoft.NET.Sdk"` is a shorthand for inserting a `<Sdk Name="Microsoft.NET.Sdk" />` element immediately under the `<Project>` element (before the Aspire SDK), which is its own shorthand for another thing (which isn't important for this conversation). That's why you see two in the Aspire project.

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+An example of a new project file is:
+
+```XML
+<Project Base="10.0.100" 
+         Type="Console"> 
+</Project> 
+```
+
+If there is a `Directory.Build.props` or `Import` file, it can contain the `Base` attribute, although it cannot contain the `Type` attribute. This is because the `Type` attribute specifies the SDK instructions for build evaluation.
```
jkoritzinsky, line 20: Would a `Base` attribute specified in an `Import`ed file retroactively affect previously imported files (including in the Microsoft.NET.Sdk)? That would be very unintuitive to how MSBuild works today.

proposed/project-file-simplification.md
jkoritzinsky: I think the only way this feature could work with how MSBuild functions would be if specifying the `Base` attribute and `Type` would lower down to something like a property set before the SDK Imports.

For example, given:

```xml
<Project Base="10.0.100" Type="Console">
    <PropertyGroup>
         <TargetFramework>net10.0</TargetFramework>
    </PropertyGroup>
</Project>
```

It would be equivalent to:

```xml
<Project>
   <PropertyGroup>
       <ProjectSdkBase>10.0.100</ProjectSdkBase>
       <ProjectSdkType>Console</ProjectSdkType>
   </PropertyGroup>
   <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />

    <PropertyGroup>
         <TargetFramework>net10.0</TargetFramework>
    </PropertyGroup>

   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
</Project>
```

With this sort of design contract, you would be able to set `ProjectSdkBase` with `-p:ProjectSdkBase=11.0.0` and it would override the `Base` attribute.

proposed/project-file-simplification.md
jkoritzinsky: I would agree here.

Instead of allowing this and creating a fully separate evaluation model in MSBuild for just this feature, I'd prefer using something like the new MSBuild BuildChecks feature that would make specifying a `Base` value "too late" (ie not in a project file or a Directory.Build.props or not at the start of a project file) an error.

Otherwise, you end up further complicating the already confusing MSBuild evaluation mechanisms even more. Although this feature would be meant to simplify how MSBuild works, it would just make it even more difficult to understand what's going on when it doesn't meet the user's expectations (primarily for users who can read a binlog to solve very simple problems through experts in using MSBuild for enterprise-level build scripts but don't follow .NET release notes)

proposed/project-file-simplification.md
andrewlock: Based on comments later on, I understand this to mean earliest, i.e. `10.0.100`, which is the _opposite_ of what I expected intuitively 🤔

proposed/project-file-simplification.md
andrewlock: I believe the text is referring to the fact that `10.0.100` is a higher tooling version than the tfm of `net482`?

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+[drawbacks]: #drawbacks
+
+This introduces a new concept to .NET users, and some additional complexity in the overall MSBuild evaluation. Users would see this in areas such as understanding binlogs. We think the trade off is worthwhile because users get small clean project files the ability to control new warnings/breaking changes independent of both SDK version and TargetFramework.
+
+## Alternatives
+[alternatives]: #alternatives
+
+We could continue to tie new features to TFM. THe drawback is that users generally upgrade their TFM to keep their projects in support. While this avoids breaking changes/warnings occurring on tooling upgrade, it may either delay adopting desirable new features, or result in breaking changes/warnings happening when the user is just trying to stay in support.
+
+## Open questions
```
rolfbjarne, line 342: How would this all work for other SDKs, say the iOS SDK?

For instance, given this project file:

```xml
  <PropertyGroup>
    <TargetFramework>net10.0-ios</TargetFramework>
  </PropertyGroup>
</Project>
```

How would this be in the new world? I'm guessing something like this:

```xml
<Project Base="10.0.100" 
         Type="iOS"> 
</Project> 
```

* Where is the fact that `Type=iOS` is a valid type declared? How does this translate to letting the user know they have to install the iOS SDK if it's not already installed?
* How can the iOS SDK provide their own set of defaults?
* How can these iOS defaults be versioned? The release schedule / versioning of the iOS SDK does not match the .NET SDK at all.

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+```XML
+<Project Base="10.0.100" 
+         Type="Console"> 
+</Project> 
+```
+
+```XML
+<Project Base="11.0.100" 
+         Type="Console">
+    <PropertyGroup Label="These properties added on upgrade from 10.0.100 to 11.0.100"
```
JoeRobich, line 315:
```suggestion
    <PropertyGroup Label="These properties added on upgrade from 10.0.100 to 11.0.100">
```

proposed/project-file-simplification.md
```diff
@@ -0,0 +1,351 @@
...
+
+This would be a breaking change, because a user may be successful evaluating a project that has a torn state - even if it is fragile. As such we would provide a property that turns on the test for torn state, and it would be off by default. In the `Base` for .NET 10, we would enable the torn state test.
+
+#### User customization
+
+We have received several requests for customization for `Base` files. Our initial response is to clarify our intent in `Base` files and how they differ from `Import` files. `Import` files are a long-standing feature of MSBuild and allow project's to incorporate MSBuild elements from another file. This feature is widely used in large repos and we have no plans for any changes to it.
+
+`Base` files are intended as a mechanism to give user's more control of the changes Microsoft to rolls out.Fundamentally, any user entered properties will overwrite any Microsoft supplied properties - you're in control of your build. With this context, we think that `Import` is a better solution for customizing at a repo level and do not plan to support custom `Base` files at this time. We are interested in hearing about any compelling scenarios.
```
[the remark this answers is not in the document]
KathleenDollard: Thank you. That is not a scenario that I had in mind. I have added it as an open question.

proposed/project-file-simplification.md
KathleenDollard: Do you think this is any more than other features?

proposed/project-file-simplification.md
KathleenDollard: Many similar scenarios will require multi-targeting, so it's a tad more complex. Especially because I think it's an error to declare both a `TargetFramework` and `TargetFrameworks` I added to the open question on multi-targeting.

KathleenDollard: The Type can supply its own defaults.

For versioned defaults, we discussed looking for a file that next to the Base that defines Type specific defaults. Like:

- Base10.0.100.props
- iOS10.0.100.props

But, that didn't make it into the proposal. 

I need to review how workloads install as, at least for V1, we planned to restrict the location for Type and Base files to the .NET install repo. Hmmm.

proposed/project-file-simplification.md
DamianEdwards: I'm not sure floating makes sense unless we intend to make the base definitions acquirable via NuGet? What is the intended behavior if the version specified is newer than the base included in the SDK currently being used?

proposed/project-file-simplification.md
RussKie: > Users would be upgrading to get the new tools. 

Isn't what we already have - "leave the TFS as is, install the new SDK" though?

richlander:
I think we can consider this proposal abandoned.

PR #333, opened 2025-04-21
Add proposal for RID-specific .NET Tool packages

dsplaisted:


accepted/2025/rid-specific-tool-packages.md
```diff
@@ -0,0 +1,91 @@
+# RID-Specific .NET Tool Packages
+
+**Owner** [Daniel Plaisted](https://github.com/dsplaisted)
+
+## Summary
+
+This proposal adds the ability for a .NET Tool to have separate packages for each supported OS and architecture.
+
+In .NET, the combination of OS and architecture is represented by a Runtime Identifier (RID).  Today, .NET Tools support native assets, but the native assets for all of the supported Runtime Identifiers need to be included in the same package.  For tools with large native dependencies, or if the entire tool is native (for example via NativeAOT), this multiplies the package download size by the number of supported RIDs.
```
jkotas, line 9: The exact same problem exists for regular nuget packages too. Any chance we can generalize the solution to solve the problem for all types of nuget packages, and not just tools? 

Related question: Is the downloader for the proposed dependent packages going to be implemented in nuget or in the SDK?

accepted/2025/rid-specific-tool-packages.md
```diff
@@ -0,0 +1,91 @@
...
+
+To support NativeAOT .NET apps (or non .NET apps), we should support an additional `executable` runner for executables that can be launched natively by the operating system.  On Unix-like operating systems, we can create a symlink in the global tools folder to the tool entry point executable.  On Windows, we may create a batch file that launches the entry point executable.  If this has drawbacks we might create a new type of shim/launcher.
+
+An alternative would be to copy the whole entry point executable to the global tools folder.  However, there might be other files in the same folder that the tool depends on, so we don't currently plan to do this.
```
rolfbjarne, line 68: > However, there might be other files in the same folder that the tool depends on, so we don't currently plan to do this.

1. A tool using [ClangSharp](https://github.com/dotnet/ClangSharp) would be an excellent test case for a rid-specific tool package, because the native clang dependency is _big_.
2. A tool using ClangSharp would have a dependency on a native dynamic library (dll/dylib), and as such copying just the executable would not work.

accepted/2025/rid-specific-tool-packages.md
```diff
@@ -0,0 +1,91 @@
...
+
+The tool manifest for the primary package would look like this:
+
+```xml
+<DotNetCliTool Version="1">
+  <Commands>
+    <Command Name="dotnet-say" />
+  </Commands>
+  <RuntimeIdentifierPackages>
+    <RuntimeIdentifierPackage RuntimeIdentifier="win-x64" Id="dotnet-say.win-x64" Version="1.0.0" />
```
joelverhagen, line 48: This feels a lot like the dependencies section of .nuspec. Akin to @jkotas's [comment above](https://github.com/dotnet/designs/pull/333/files#r2051960057), could this concern be considered more generally and expressed in the .nuspec as a RID-specific dependency?

joelverhagen: There is the `runtime.json` file in some packages... but I forget how this works...

accepted/2025/rid-specific-tool-packages.md
```diff
@@ -0,0 +1,91 @@
...
+```xml
+<ItemGroup>
+	<ToolPackageRuntimeIdentifier Include="win-x64"/>
+	<ToolPackageRuntimeIdentifier Include="linux-x64"/>
+</ItemGroup>
+```
+
+If `ToolPackageRuntimeIdentifier` is non-empty, then packing the project without a RuntimeIdentifier will create the primary package, and packing the project with a RuntimeIdentifier will create the corresponding RID-specific package.  If feasible, we could do an automatic inner pack of the ToolPackageRuntimeIdentifiers before creating the primary tool package.
+
+The package name for the RID-specific packages will be `<ToolPackageName>.<RuntimeIdentifier>`.  If not specified, the version number of the RID-specific package will be the same as that of the primary package.  `Version` metadata on the `ToolPackageRuntimeIdentifier` item can be used to set the RID-specific package version.
```
joelverhagen, line 87: Why have optional `Version`? Won't it be known at pack time? I am unsure why we should allow a missing version at all.

joelverhagen, line 87: Is `Version` a version or version range? Seems like version. What's the error experience when the RID-specific package cannot be found?

accepted/2025/rid-specific-tool-packages.md
```diff
@@ -0,0 +1,91 @@
...
+This proposal adds the ability for a .NET Tool to have separate packages for each supported OS and architecture.
+
+In .NET, the combination of OS and architecture is represented by a Runtime Identifier (RID).  Today, .NET Tools support native assets, but the native assets for all of the supported Runtime Identifiers need to be included in the same package.  For tools with large native dependencies, or if the entire tool is native (for example via NativeAOT), this multiplies the package download size by the number of supported RIDs.
+
+We would like to add support for RID-specific .NET Tool packages.  This would enable a tool to have a separate package for each supported RID, and when the tool is installed only the package for the correct RID would need to be downloaded.
+
+### Goals
+
+- Tools can create RID-specific packages
+- Installing a tool with RID-specific packages is transparent but only downloads the package for the current RID
```
joelverhagen, line 16: It will download both the top level (generic) package and the RID specific package (2) right? Not 1?

accepted/2025/rid-specific-tool-packages.md
```diff
@@ -0,0 +1,91 @@
...
+<DotNetCliTool Version="1">
+  <Commands>
+    <Command Name="dotnet-say" EntryPoint="dotnet-say.dll" Runner="dotnet" />
+  </Commands>
+</DotNetCliTool>
+```
+
+## Design
+
+A tool with RID-specific packages will consist of a single primary package and RID-specific packages for each supported RID.  The primary package will include a tool manifest that lists the RIDs supported by the tool, and the package name and package version of the RID-specific tool package for each one.  The primary package will not include any tool implementation assets or shims.  The RID-specific packages will have the same layout and format as tool packages currently have.  The only difference will be that the NuGet package type will be set to a new `DotnetToolRidPackage` type, in order to prevent tool search results from being cluttered with RID-specific tool packages (the primary package is the one that should show up in the results).
```
joelverhagen, line 36: Package type has versioning. Currently `DotnetTool` is the default `0.0` version. If it is bumped to `DotnetTool, 2.0.0` then they should still be searchable by `DotnetTool` but also have a persistent marker that could be later used for filtering, without having to crack open the package again to parse the `DotnetToolSettings.xml`. Right now we only have package type name searching enabled but we could enable version filtering in the future. 

Conceptually, the package type is moving to version 2 as shown in your XML.

joelverhagen, line 36: Will the RID-specific package be validate to have `DotnetToolRidPackage`? Or will any existing package be accepted and treated as a tool package? 

The relationship is only 2 layers deep, right? `1 DotnetTool` to `N DotnetToolRidPackage`, and no `DotnetTool -> DotnetTool -> DotnetToolRidPackage`?

joelverhagen, line 36: If the `DotnetTool` has multiple applicable RIDs, such as `win` and `win-x64`, which is selected?

accepted/2025/rid-specific-tool-packages.md
joelverhagen, line 87: Just curious, what is the scenario for having a different version on the RID-specific packages?

accepted/2025/rid-specific-tool-packages.md
akoeplinger: https://github.com/NuGet/Home/issues/10571 is an issue about the problem for regular nuget packages

accepted/2025/rid-specific-tool-packages.md
dsplaisted: If you are building a NativeAOT app, then you mostly have to build for each Runtime Identifier on a separate machine.  So depending on how your build/pipeline system works, you might not have the same version for all of the different packages.

dsplaisted: It's an exact version.  If that version isn't found then it should error out, probably in a similar way to what would happen if the tool couldn't be found.

accepted/2025/rid-specific-tool-packages.md
dsplaisted: That's correct.  I didn't find a way to word this to be precisely correct without being confusing.

accepted/2025/rid-specific-tool-packages.md
dsplaisted: The RID-specific package won't have `DotnetToolRidPackage` in its manifest, that's only for the primary package to redirect to the RID-specific packages.  Theoretically a tool package could redirect to any other tool package.  There's currently only layers though, it wouldn't keep following redirects.

We might also decide we want to support tool redirects.  I think that might be useful as part of the [one-shot / dnx / dotnet tool exec work](https://github.com/dotnet/sdk/issues/31103) in order to hide the difference between a tool package name and its command name.  But that's not part of a proposal yet.

dsplaisted: We'll select the "best" (or most specific) one.  This is something we do in various places in the SDK.

accepted/2025/rid-specific-tool-packages.md
dsplaisted: See here: https://github.com/dotnet/designs/pull/333#discussion_r2055077258

accepted/2025/rid-specific-tool-packages.md
dsplaisted: Interesting.  I'm not sure we knew that you could version package types when we originally implemented tools, or maybe at that time it wasn't possible to version them.  Since old SDKs don't look at the package type we'll have to update the version in the tool manifest anyway so that old SDKs will error out when trying to install a newer tool.  So I'm not sure there's much extra value in also updating the package type version.

joelverhagen: It would allow NuGet.org to filter these by version, and analyze their adoption in NuGet Insights (separate from V1 tools). It could allow customized UI on NuGet.org to show the RID-specific dependencies (after analyzing `DotnetToolSettings.xml`). In other words, it's the "NuGet" way of versioning package types.

Package type versioning has been there since the beginning. I made it :)

It's there to use if you want it. I think it would be good to have these new packages marked differently, in a nuget-specific way if possible.

joelverhagen: Sounds good. Might be worth having a small note about this in the design. I'm a bit curious what the algorithm is for best RID. I am familiar with best TFM in NuGet and it seems similar.

joelverhagen, line 36: Will package source mapping, vulnerability auditing, signature validation and policies (etc) all apply to the RID-specific package? I think what I'm getting at, is: will security features that are added to restore happen here too automatically?

dsplaisted: I think it's actually a NuGet API that we call to help us process the RID graph and choose the best match.  As I understand it, it starts with the current RID and walks the parent chain up to the root (which I think is `any`).  Then from the available RIDs for an asset, the best match is the one that appears first in the chain.

joelverhagen: Okay, thanks for clarifying. It's unfortunate that some security features don't work for current .NET tools. That seems like a problem but I see how it's orthogonal to this change. I think the reason I asked is because this feature feels like a mini restore operation with a relatively shallow graph. Since there are now transitive nodes it may be harder for users to reason about the package that they use and depend on.

Maybe the only bit of feedback here is to consider in the design how users can discover and audit the RID-specific package with tooling, similar to how we have dependency package tooling.

accepted/2025/rid-specific-tool-packages.md
joelverhagen: If the version is not same for all of the different packages, how would the top level tool package come to know about the list of `(RID, version)` tuples, i.e. to produce the `DotnetToolSettings.xml` file?

I see the example .csproj XML:
```xml
<ItemGroup>
	<ToolPackageRuntimeIdentifier Include="win-x64"/>
	<ToolPackageRuntimeIdentifier Include="linux-x64"/>
</ItemGroup>
```

This seems to be the non-NativeAOT case, where the whole graph (3 packages in this case?) is produced on one machine.

If the RID-specific packages are being produced on another machine, there seems to be a _fan-out then collect_ type flow.

I'm trying to visualize how the differing version numbers will flow from one build step to the next. It feels like a bit of a mess if the version numbers are allowed to diverge. If we say "all version numbers must match" then there is no need for build steps to happen in a certain order or flow information (RID + selected version number for that RI) from one build step to the next. I see how it can be technically possible but I wonder what the .csproj will look like to handle the split-up, RID-specific build steps.

accepted/2025/rid-specific-tool-packages.md
joelverhagen: I think we can close this thread in favor of https://github.com/dotnet/designs/pull/333#discussion_r2051960057.

NinoFloris:
It seems like some concept of associated packages should instead be added to nuget packages proper. 

If the package manifest could hold a list of associated package entries (e.g. {packagetype, name, optional? version range}) nuget could provide common authoring tooling and insights, and it would not be entirely opaque to nuget.org either.

Entries with certain package types could have some builtin resolution behavior or otherwise be ignored during restore. A spec like this would then define the custom resolution rules for the their package type.

As libraries have the exact same issue with rid splitting it strongly suggests things should not be tied to a custom file and schema that nuget does not understand. Sharing this part of the investment would make it much more likely we'll actually be able to get the library rid splitting issue solved as well.

PR #334, opened 2025-05-23
Add a design proposal for dotnet tool exec and dnx

marcpopMSFT:
This is an attempt to summarize our discussions from the WIP for dotnet tool exec and offline discussions about having a shortcut dnx:
https://github.com/dotnet/sdk/issues/31103
https://github.com/dotnet/sdk/pull/48443

For dnx, we need to decide if there are some flags automatically set like --ignore-failed-sources or --yes.

### DIFF
--- a/accepted/2025/direct-tool-execution.md
+++ b/accepted/2025/direct-tool-execution.md
@@ -24,7 +24,7 @@ The `dotnet tool exec` command will support the following parameters:
 
 | Parameter                | Description                                                                                  |
 |--------------------------|----------------------------------------------------------------------------------------------|
-| `<package-id>`           | The NuGet package ID of the tool to execute.  Supports <package-id>@<version> format to install a specific version without using the --version flag  |
+| `<package-id>`           | The NuGet package ID of the tool to execute.  Supports `<package-id>@<version>` format to install a specific version without using the --version flag  |
 | `--version <version>`    | (Optional) The specific version of the tool to run. If not specified, the latest is used.    |
 | `--prerelease`           | (Optional) Allows installing and running prerelease versions.                                |
 | `--framework <tfm>`      | (Optional) Specifies the target framework to use for the tool.                               |
@@ -69,7 +69,7 @@ Note: The version option would work the same as other dotnet tool commands suppo
 
 There is the potential of supporting calling a tool by its command name rather than the package ID. This approach could improve usability, but it is potentially problematic because .NET tools do not have a standard naming convention for commands, and there are existing protections on nuget.org to guard against spelling errors and package squatting that apply only to package IDs. Further discussion and feedback are needed to determine the best approach for this scenario.
 
-We could support a standard naming convention for the package like "dotnet-<command>" as that's used in dotnet-ef and many other tools but that would open the question of what to do about existing tools.
+We could support a standard naming convention for the package like `dotnet-<command>` as that's used in dotnet-ef and many other tools but that would open the question of what to do about existing tools.
 
 ---
 


dsplaisted:
> The name `dnx` was [previously used](https://weblogs.asp.net/imranbaloch/k-kvm-kpm-klr-kre-in-asp-net-vnext) after the rename from the K-series CLI tools (`kvm` et al), before all the tool names were [consolidated under `dotnet`](https://www.hanselman.com/blog/exploring-the-new-net-dotnet-command-line-interface-cli). Re-using that name could lead to confusion from people finding older docs or LLMs mushing up the two.

My opinion is that after 10 years it's OK to reuse the name `dnx`.  Nowadays it's probably going to look more familiar as an analog to `npx`, rather than people getting confused about it because it's a name we used during previews of .NET Core.

dsplaisted:
> @dsplaisted My CLI can run both as a "normal" CLI tool and a MCP server - see the link above.

I think the normal CLI should also be runnable via `dotnet tool exec` / `dnx`.  So the command to run the MCP would be `dnx ErikEJ.DacFX.TSQLAnalyzer.Cli -mcp`, while `dnx ErikEJ.DacFX.TSQLAnalyzer.Cli` would run the normal CLI.

ErikEJ:
@dsplaisted Cool, got it, thanks

accepted/2025/direct-tool-execution.md
```diff
@@ -0,0 +1,99 @@
...
+- Scripting and automation where ephemeral tool usage is desired.
+
+## CLI Parameters
+
+The `dotnet tool exec` command will support the following parameters:
+
+| Parameter                | Description                                                                                  |
+|--------------------------|----------------------------------------------------------------------------------------------|
+| `<package-id>`           | The NuGet package ID of the tool to execute.                                                 |
+| `--version <version>`    | (Optional) The specific version of the tool to run. If not specified, the latest is used.    |
```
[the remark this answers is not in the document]
joelverhagen: Asked another way, is this similar semantics as a PackageReference where you often specify a version like `1.0.0` but this actually expands to `[1.0.0, )` (take lowest) but supports `1.*` (floating).

accepted/2025/direct-tool-execution.md
```diff
@@ -0,0 +1,99 @@
...
+| `--prerelease`           | (Optional) Allows installing and running prerelease versions.                                |
+| `--framework <tfm>`      | (Optional) Specifies the target framework to use for the tool.                               |
+| `-- [tool arguments]`    | Arguments to pass to the tool being executed.                                                |
```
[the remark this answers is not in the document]
marcpopMSFT: Heh, I was just rewatching our design meeting on this and I literally commented on this as I was going over this section. Good catch though.

accepted/2025/direct-tool-execution.md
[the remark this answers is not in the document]
marcpopMSFT: I would expect it to work the way --version works for other tool commands which is to say that we do support the `1.*` floating syntax but I don't believe we support full nuget `[1.0.0,)` syntax. I added a note to clarify that.

accepted/2025/direct-tool-execution.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Reducing typing and cognitive load for frequent tool users.
+- Providing a migration path for users familiar with the legacy `dnx` command.
+- Enabling easier scripting and automation.
+
+## Detailed Design
+
+- On Windows, `dnx.cmd` will be installed next to `dotnet.exe`.
+- On non-Windows platforms, an executable file named `dnx` (no extension) will be installed next to the `dotnet` host and marked as executable.
+- The script will invoke `dotnet tool exec` with all arguments passed through unchanged.
```
jkotas, line 92:
```suggestion
- The script will invoke `dotnet tool dnx` with all arguments passed through unchanged.
```
?

accepted/2025/direct-tool-execution.md
```diff
@@ -0,0 +1,108 @@
...
+
+We could support a standard naming convention for the package like "dotnet-<command>" as that's used in dotnet-ef and many other tools but that would open the question of what to do about existing tools.
+
+---
+
+## `dnx` Summary
+
+This proposal introduces a new launcher script, `dnx`, which acts as a thin wrapper for `dotnet tool exec`. The script will be installed as `dnx.cmd` on Windows and as an executable `dnx` (no extension) on non-Windows platforms, and will reside alongside the `dotnet` host executable.
+
+When invoked, `dnx` will forward all arguments to `dotnet tool dnx` which will handle parsing the arguments and passing them through to `dotnet tool exec`, enabling a familiar and concise entry point for running .NET tools directly. The script will not contain any additional logic, ensuring consistency and simplifying maintenance across platforms.
```
jkotas, line 76: What the rationale for the extra `dnx` layer?

dsplaisted: We want to keep the script super simple and hopefully it will never change.  Having the logic in the CLI (and I would put it under `dotnet dnx` instead of `dotnet tool dnx`) leaves the door open for `dnx` to support executing other things, like C# "script" files or projects.  That way the CLI can decide how to interpret `dnx` instead of having to implement behavior in the script.

jkotas: Keeping the script minimal makes sense.

I have been wondering about this:

> `dotnet tool dnx` which will handle parsing the arguments and passing them through to `dotnet tool exec

This suggests that `dotnet dnx` will parse the arguments, does something with them (what is that exactly?), creates a new command line to pass to `dotnet tool exec`, `dotnet tool exec` parses the arguments again and executes the tool. Can `dotnet dnx` just do the thing?

marcpopMSFT: When I was originally writing the doc, I had the idea of dnx detecting if it was a tool, an exe, a dll, or a .cs file and running different commands. I think the parsing text was left over from that as we decided to move that to a non-goal to scope and simplify the feature. I can't think of anything that dnx should do that tool exec couldn't.

I'll clarify that we won't actually create a new command line for dotnet tool exec but just be using the same command logic internally. I definitely don't want to spin up an extra dotnet process.

accepted/2025/direct-tool-execution.md
```diff
@@ -0,0 +1,114 @@
...
+- Scripting and automation where ephemeral tool usage is desired.
+
+## CLI Parameters
+
+The `dotnet tool exec` command will support the following parameters:
+
+| Parameter                | Description                                                                                  |
+|--------------------------|----------------------------------------------------------------------------------------------|
+| `<package-id>`           | The NuGet package ID of the tool to execute.                                                 |
+| `--version <version>`    | (Optional) The specific version of the tool to run. If not specified, the latest is used.    |
```
dsplaisted, line 28: While we should support `--version`, I think the preferred way to specify the version should be via the package@version syntax, for example `dotnet tool exec dotnetsay@1.2.3 -- Hello, World!`.

accepted/2025/direct-tool-execution.md
```diff
@@ -0,0 +1,114 @@
...
+
+## `dnx` Summary
+
+This proposal introduces a new launcher script, `dnx`, which acts as a thin wrapper for `dotnet tool exec`. The script will be installed as `dnx.cmd` on Windows and as an executable `dnx` (no extension) on non-Windows platforms, and will reside alongside the `dotnet` host executable.
+
+When invoked, the `dnx` script will forward all arguments to a .NET CLI command `dotnet dnx` which will initially just be an alias for `dotnet tool exec`, enabling a familiar and concise entry point for running .NET tools directly. The script will not contain any additional logic, ensuring consistency and simplifying maintenance across platforms.
+
+## Motivation
+
+The `dnx` command provides a short, memorable alias for running .NET tools, inspired by the historical DNX (DotNet Execution Environment) command. It simplifies tool invocation, especially in cross-platform scripts and documentation, and provides a consistent experience across operating systems.
```
dsplaisted, line 84: I think it's worth noting that using `dnx` for this is also inspired by `npx`, and the similarity is helpful since many devs are already familiar with `npx`.

dsplaisted (approved):

accepted/2025/direct-tool-execution.md
joelverhagen: I'm a bit surprised by this. Do we use the `@` syntax much in .NET? The `dotnet package add` syntax uses `--version` so it felt natural to me. I disagree per se, just have an initial question on whether it would be inconsistent.

dsplaisted: We've started adding support for `id@version` syntax, first I think in workloads and now in some tool and NuGet commands.  We think it's a nicer syntax and want to support it generally. /cc @baronfel

accepted/2025/direct-tool-execution.md
```diff
@@ -24,7 +24,7 @@ The `dotnet tool exec` command will support the following parameters:
 
 | Parameter                | Description                                                                                  |
 |--------------------------|----------------------------------------------------------------------------------------------|
-| `<package-id>`           | The NuGet package ID of the tool to execute.                                                 |
+| `<package-id>`           | The NuGet package ID of the tool to execute.  Supports <package-id>@<version> format to install a specific version without using the --version flag  |
```
jjonescz, line 27: `<tags>` are rendered as HTML in markdown so this needs code block

```suggestion
| `<package-id>`           | The NuGet package ID of the tool to execute.  Supports `<package-id>@<version>` format to install a specific version without using the --version flag  |
```

accepted/2025/direct-tool-execution.md
```diff
@@ -0,0 +1,114 @@
...
+
+- The tool is ephemeral and not discoverable via `dotnet tool list`.
+- Caching behavior follows standard NuGet cache policies.
+- There should be minimal impact to the state of the machine beyond the install to the NuGet cache.
+
+### Calling by Command vs. Package
+
+There is the potential of supporting calling a tool by its command name rather than the package ID. This approach could improve usability, but it is potentially problematic because .NET tools do not have a standard naming convention for commands, and there are existing protections on nuget.org to guard against spelling errors and package squatting that apply only to package IDs. Further discussion and feedback are needed to determine the best approach for this scenario.
+
+We could support a standard naming convention for the package like "dotnet-<command>" as that's used in dotnet-ef and many other tools but that would open the question of what to do about existing tools.
```
jjonescz, line 72:
```suggestion
We could support a standard naming convention for the package like `dotnet-<command>` as that's used in dotnet-ef and many other tools but that would open the question of what to do about existing tools.
```

PR #335, opened 2025-06-23
Add `dotnet install` capability to the SDK and as a Standalone Executable. Install CLI UX.

nagilson:
# DNVM / Dotnet Bootstrapper End-to-End Installation Mechanisms
View the rendered version here: https://github.com/dotnet/designs/blob/a4e76c1e409f96e45e3813669b0782449eda3de1/proposed/dotnet-bootstrapper-cli-experience.md

The officially supported `DNVM` and `DNX` project was replaced by the `.NET CLI` as `dotnet` in `2017`.

As of 2025, `DNVM` is an unofficial product, but internally developed and leveraged.

We are proposing an official `DNVM` product (name subject to change).
This is an offering similar to `nvm` or `rustup` but for `.NET`.
There have already been many internal documents/discussions. This is a 'strawman proposal' for us to dig at further. For all mentions of the term `dnvm`, this is subject to change to an applicable product name.

Also see [another document by dsplaisted](https://microsoft.sharepoint.com/teams/ManagedLanguagesIDE/Shared%20Documents/CLI/CLI%20OneNote/Planning.one#DNVM%20Scenarios&section-id=%7BA625BF1A-DBE8-4361-8CEC-0AB872CEFB49%7D&page-id=%7B7246EF16-EBAF-41DD-AC8C-7107E48035C1%7D&end ) for inspiration.
Please see the attached document. No opinions or statements are final.

The suggestions in this proposal incorporate feedback/efforts from many individuals, not just my own.


proposed/dotnet-bootstrapper-cli-experience.md
```diff
@@ -0,0 +1,256 @@
...
+| `dotnet update version` | Only update the specified SDK managed by `dnvm` |
+| `dotnet self update` | Updates the `dnvm` product |
+
+#### Notes
+`dotnet track` is not provided.
+`dotnet untrack` is not provided.
+
+## 🗑️ Uninstall Experience
+
+`dotnet uninstall <channel>` - Uninstalls the specified SDK. Uninstall is simply a file deletion. No updates to the `PATH` are made.
```
nagilson, line 248: Perhaps revisit this part if this is the last remaining SDK / consider self uninstall of the toolchain.

proposed/dotnet-bootstrapper-cli-experience.md
```diff
@@ -0,0 +1,256 @@
...
+
+The officially supported `DNVM` and `DNX` project was replaced by the `.NET CLI` as `dotnet` in `2017`.
+
+As of 2025, `DNVM` is an unofficial product, but internally developed and leveraged.
+
+We are proposing an official `DNVM` product (name subject to change).
+This is an offering similar to `nvm` or `rustup` but for `.NET`.
+There have already been many internal documents/discussions. This is a 'strawman proposal' for us to dig at further. For all mentions of the term `dnvm`, this is subject to change to an applicable product name.
+
+The `dnvm` product will be AOT bundled alongside the apphost to enable shipment via an external standalone executable, that is also callable from within the .NET SDK. The implementation details of this will be settled and discussed within another document, but it's necessary to remark on this here to explain the usage of 'dotnet' as a root command.
```
JoeRobich, line 13: > The `dnvm` product will be AOT bundled alongside the apphost to enable shipment via an external standalone executable

For `dnx` [it appears the plans](https://github.com/dotnet/designs/blob/main/accepted/2025/direct-tool-execution.md#dnx-summary) are to ship a script that then invokes dotnet CLI commands. Would it be a good idea to implement `dnvm` similarly?

proposed/dotnet-bootstrapper-cli-experience.md
```diff
@@ -0,0 +1,256 @@
...
+
```
jkotas, line 14: Should this mention somewhere that this is not going to be applicable to .NET installed via OS installers (distro package managers, etc.)?

proposed/dotnet-bootstrapper-cli-experience.md
```diff
@@ -0,0 +1,256 @@
...
+## 🗃️ Installation Experience
+
+```
+dotnet install
+dotnet install <channel>
+```
+
+### MVP:
+`dotnet install` → installs latest STS/LTS SDK, or the 'latest' acceptable SDK based on `global.json` if specified. The installation modifies the user-level `PATH`. (This is complicated and will get its own document.)
```
yaakov-h, line 23: Is the thinking here along the lines of [actions/setup-dotnet](https://github.com/actions/setup-dotnet), so I could `cd project && dotnet install` and it would install whatever matches `project/global.json`?

proposed/dotnet-bootstrapper-cli-experience.md
```diff
@@ -0,0 +1,256 @@
...
+|----|-----------|
+| **Windows** | `/dnvm` |
+| **Mac** | `/Library/Application\ Support/dnvm/` |
+| **Linux** | `/.local/dnvm` |
+
+### Additional Features
+| Command | Description |
+|---------|-------------|
+| `dotnet install 9` | Installs latest 9.X SDK |
+| `dotnet install 9.0.1xx` | Installs latest 9.0.100 SDK. Equivalent to `rollForward: latestFeature` in `global.json` |
```
yaakov-h, line 78: Is this a literal, i.e. `x` is a wildcard character?

proposed/dotnet-bootstrapper-cli-experience.md
```diff
@@ -0,0 +1,256 @@
...
+          "type": "globalJson",
+          "path": "C:\\projects\\myproject\\global.json"
+        }
+      ]
+    },
+    {
+      "version": "9.0.200-preview.1",
+      "mode": "sdk",
+      "architecture": "x64",
+      "path": "C:\\Users\\username\\dnvm\\installs\\9.0.200-preview.1",
```
yaakov-h, line 149: Where would installations for non-system architectures go?

For example, `dotnet install -a arm64 9.0.200-preview.1` assumedly would not install into the same directory as the x64 build?

proposed/dotnet-bootstrapper-cli-experience.md
nagilson: I think that's a great point, this is context I did not include here. The interaction with this tool, the PATH, and VS are all important here. We are deliberately 'putting' the admin problem, besides the PATH

proposed/dotnet-bootstrapper-cli-experience.md
```diff
@@ -0,0 +1,256 @@
...
+We are proposing an official `DNVM` product (name subject to change).
+This is an offering similar to `nvm` or `rustup` but for `.NET`.
+There have already been many internal documents/discussions. This is a 'strawman proposal' for us to dig at further. For all mentions of the term `dnvm`, this is subject to change to an applicable product name.
+
+The `dnvm` product will be AOT bundled alongside the apphost to enable shipment via an external standalone executable, that is also callable from within the .NET SDK. The implementation details of this will be settled and discussed within another document, but it's necessary to remark on this here to explain the usage of 'dotnet' as a root command.
+
+## 🗃️ Installation Experience
+
+```
+dotnet install
```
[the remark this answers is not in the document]
jkotas: BTW: If you are going to overload `dotnet` like this, you need to be super careful about escalation of privilege security bugs. 

It is very problematic when admin installed application probes non-admin installed locations for runtimes by default.

proposed/dotnet-bootstrapper-cli-experience.md
nagilson: Yes. @richlander had a great point that we should try to align some of the behaviors of the github action here.

proposed/dotnet-bootstrapper-cli-experience.md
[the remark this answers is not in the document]
nagilson: 💯 , we are going to consult @blowdart and @GrabYourPitchforks when we make that design doc. It should come pretty soon. I think it's very important we get this right. We are working with @agocke on this since he owns the host.

proposed/dotnet-bootstrapper-cli-experience.md
nagilson: That's a great addition, I will add this context.

nagilson:
> One general thing to bring up here is Microsoft Update.
> 
> If you take the unzip approach such things will not get automatically updated by an OS's update mechanism as you're not laying down the artifacts Microsoft Update (or apt-get) needs to deliver security updates.
> 
> That alone would require a very high-level exception to the SDL to ship.

My initial proposal was to spawn a daemon process that automatically updated these SDK zips, but this was decided to be lower priority. If that's the case, maybe this changes. We should talk.

jkotas:
> My initial proposal was to spawn a daemon process that automatically updated these SDK zips

Ugh. We owned background services that run all the time on the machine in .NET Framework days. Owning background service like that is super expensive.

joeloff:
> > My initial proposal was to spawn a daemon process that automatically updated these SDK zips
> 
> Ugh. We owned background services that run all the time on the machine in .NET Framework days. Owning background service like that is super expensive.

Why don't we defer this to users? On windows they can use schedule tasks and on Linux they can do cron jobs.

joeloff:
> One general thing to bring up here is Microsoft Update.
> 
> If you take the unzip approach such things will not get automatically updated by an OS's update mechanism as you're not laying down the artifacts Microsoft Update (or apt-get) needs to deliver security updates.
> 
> That alone would require a very high-level exception to the SDL to ship.

This would only be a problem if we installed into Program Files. And that should just be off limits altogether. We have ways to acquire and maintain global installs on Windows, plus it gets really messy when you bring Visual Studio into the mix.

The default solution for this would be to use CCM deployments, but detection poses a problem so we should not even venture down this route at all.

nagilson:
Yes, both `rustUp` and `nvm`, which are motivating examples, reference a different executable, and different executable name. That name distinction is one of the core differences between these approaches. `dotnet` is already a multi-plexing system, and the belief was that the best UX for this tool would be to interface as 'dotnet.' I certainly view this as higher risk than enabling a unique tool name and then possibly integrating it into `dotnet` later. However, a lot of folks like the idea of doing `dotnet install`.

We may be able to view the releases.json as an 'update manifest' if the content is signed.

proposed/dotnet-bootstrapper-cli-experience.md
yaakov-h: Keep in mind that while it's not common, a single OS installation can be multi-architecture-capable. I believe macOS at the moment is capable of a single installation being able to boot into both x86_64 and arm64 depending on what CPU is present - practically, probably when installed to an external disk or as a VM. I suspect Windows can't do the same but I wouldn't be surprised if Linux could.

This edge case may not be a practical concern, or it could be explicitly unsupported, but it is quite possible then that I could install x64 on an arm64 machine, then boot into x64 on the same disk, and it would look in an unprefixed directory and load arm64 .NET on an x64 host 😄.

proposed/dotnet-bootstrapper-muxer-path.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+### Create a C# Muxer
+
+The muxer would become a bootstrapper and the muxer itself. It would cease to become C++ and become ported C# code that is run as native AOT code. This is high-risk and also likely still lower performance than a C++ application.
+
+
+## PATH Issue
+
+Visual Studio and the MSI or 'global' .NET Installer set the System `PATH` on Windows and not the user level `PATH`. The system `PATH` always takes precedence over the user `PATH`. If `dnvm` sets the user path to enable a folder of local .NET SDKs to be discovered by the `dotnet.exe` application host at `DOTNET_ROOT`, via the `PATH`, those installs would become 'undetectable' to the runtime, debugger, etc when Visual Studio would update and modify the System `PATH`.
```
richlander, line 32: This is a very significant issue. Agreed.

proposed/dotnet-bootstrapper-muxer-path.md
```diff
@@ -0,0 +1,68 @@
...
+
+There have been several methods suggested to handle this. None of them are ideal.
+
+### Fail on Windows if System PATH set, if `paths` in `global.json` is not set
+
+In this approach, we prioritized defined and replicable behavior over the easier UX behavior. System administrator installs are created in predictable locations, especially those from Visual Studio and/or package manager feeds.
+
+What we could do is look for those feeds per install operation and decline to do anything when a user tries to run `dotnet install`. The error message would point the user towards leveraging the `sdk: paths []` option in their repository's `global.json` file, which would specify to the host or `dotnet.exe` where to look for the installs of the .NET SDK or .NET runtimes. The host is already aware of this as of .NET 10.
+
+This is subject to backlash, especially as only 2 to 6% of users are on a `global.json` according to internal metrics.
```
richlander, line 44:
```suggestion
This is subject to negative feedback, especially as only 2 to 6% of users are on a `global.json` according to internal metrics.
```

proposed/dotnet-bootstrapper-muxer-path.md
```diff
@@ -0,0 +1,68 @@
...
+
+However, there is precedence in the sense that `nvm` is a Linux and macOS only tool, from an official standpoint. An alternative would be to make `dotnet install` only applicable to Linux and macOS. In this case, I would be more strongly opposed to using `dotnet` as the root command, and instead leaning towards making this an 'opinionated' installer that resolves the Linux and macOS scenario. This creates a split between OS support.
+
+### Update the Installers / VS to Set the User PATH
+
+The Windows official guidance is to set the system level PATH, not to set the global `PATH`. However, there are some possible issues with setting the system `PATH`. For one, we believe that the Windows API for setting the user path does a `COM` broadcast to every single open window on the OS and waits for a response. `chrome.exe`, for example, hangs for 1 minute when receiving this broadcast, which means setting the API for setting the system PATH can take a minute, mostly dead waiting.
```
elinor-fung, line 50: I'm a bit confused by the terminology here. I think of 'system' and 'global' as the same. Where we currently set 'system', but are considering setting 'user'.

elinor-fung, line 50: Do we know that the broadcast is only for the user environment? I'd guess it'd be something like a `WM_SETTINGCHANGE` message and it seems surprising that setting the user environment does a broad notification, but setting the (broader in scope) system environment does not.

proposed/dotnet-bootstrapper-muxer-path.md
```diff
@@ -0,0 +1,68 @@
...
+
+This is also applicable to `source-built` SDKs and the `pkg` installer.
+
+There have been several methods suggested to handle this. None of them are ideal.
+
+### Fail on Windows if System PATH set, if `paths` in `global.json` is not set
+
+In this approach, we prioritized defined and replicable behavior over the easier UX behavior. System administrator installs are created in predictable locations, especially those from Visual Studio and/or package manager feeds.
+
+What we could do is look for those feeds per install operation and decline to do anything when a user tries to run `dotnet install`. The error message would point the user towards leveraging the `sdk: paths []` option in their repository's `global.json` file, which would specify to the host or `dotnet.exe` where to look for the installs of the .NET SDK or .NET runtimes. The host is already aware of this as of .NET 10.
```
elinor-fung, line 42: > where to look for the installs of the .NET SDK or .NET runtimes

The host only looks for SDKs in `sdks: paths []`, not runtimes.

nagilson:
I will convert this back to a draft for now based on the controversial / significant nature of the changes, and update based on further conversations once we are more aligned and have discussed the broader document from the PM/Mgmt chain.

proposed/dotnet-bootstrapper-muxer-path.md
nagilson: I did not know that. Thank you for clarifying!

proposed/dotnet-bootstrapper-muxer-path.md
nagilson: You're right, it appears to be any environment variable from the registry via `SetEnvironmentVariableFromRegistry`.

nagilson: Great point, this wording is confusing. I rewrote this portion and some other parts. Some of this was incorrect.

proposed/dotnet-bootstrapper-muxer-path.md
nagilson: I rewrote this to focus on the bad experience / work it creates for users, rather than the negative feedback. Thank you.

AustinWise:
It might be worthwhile to to mention the existing [`dotnet-install` scripts](https://learn.microsoft.com/en-us/dotnet/core/tools/dotnet-install-script) in the "related offerings" section. It seems like there is some overlap with those scripts and this proposal for initially acquiring the .NET SDK. What this proposal adds is the ability to easily update an existing install, which would be quite nice.

nagilson:
I think there are several security implications about this project proposal at large that we should review before others invest more time in this document. Thank you all for taking the time to provide your feedback. These documents will be used in our security discussion going forward. I will update the PR after our discussions.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+> 📓 Technical Details:
+> A registry key is preferred over an environment variable because it is an admin-protected context. We don't want user-local controllable behavior to influence/break the behavior of an administrative installer, as that is a security concern. A policy key may also be used, but a policy key is essentially equivalent to a registry key in this context.
+>
+> On Unix, we could set this via a protected sentinel file.
+
+3. Under step 2, the `PATH` is updated to a dotnet application host (dotnet.exe) that contains the local installs managed by `bootstrapper`. Visual Studio and the other installers will no longer set the System `PATH` until the registry key or sentinel change is reverted. The option to undo this would be available, by removing the key, perhaps via an `uninstall all` or `reset` CLI feature from `bootstrapper`. The exact semantics of such are not pertinent here.
+
+4. The existing `global.json` `paths` feature would still enable other scenarios like for Visual Studio to specify any SDK. In this sense, it is already possible to enable local .NET SDKs to work with VS.
```
dsplaisted, line 32: This seems a bit unclear to me.  The `paths` feature would still work, but it doesn't let Visual Studio specify an SDK.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+5. The user is prompted to see if they would like to install the local versions of the administrative installs currently managed by Visual Studio/MU/Package Manager, as well as the update option they prefer for those installs (ala global.json `rollFoward` values, as described in the cli-experience document.)
```
dsplaisted, line 34: I think it's clearer to use "channel" here.  If we have to change the term later then fine, but for now I think it makes it easier to understand.

```suggestion
5. The user is prompted to see if they would like to install the local versions of the administrative installs currently managed by Visual Studio/MU/Package Manager, as well as the update channel they prefer for those installs (ala global.json `rollFoward` values, as described in the cli-experience document.)
```

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+> 📓 Technical Details: Update
+> Upon the action of step 5, we would need to take on the responsibility of updating these SDKs. Managing a daemon is resource-intensive, so it may not be optimal. SDL requirements specify a required update policy; we cannot hook into MU and IT administrators rely on administrative install behaviors. One approach is to mimic the update behavior of Visual Studio upon every interaction with `bootstrapper` by suggesting updates and asking the user to specify an update cadence when the new local mode is applied. IT administrators should be able to enforce a policy that requires a specific update cadence, and the provided update cadence should be compliant with SDL requirements, such as checking for updates on a daily cadence when the tool is used. Visual Studio would need to trigger this check as well to maintain compliance.
```
dsplaisted, line 40: To start with, I believe it's enough to notify the user of available updates and allow them to install those updates (via an explicit action).  I don't think we need any sort of automatic update in order to comply with the SDK (or to have a good experience).

I do think that the update notifications should not be limited to interactions with the bootstrapper or installation related commands if we make them part of `dotnet`.  Typical `dotnet` development commands such as `dotnet build` should notify the user that an update is available.

I also don't think we need (at least to start with) any IT policy around this, beyond possibly disabling user installs entirely.  If IT wants to manage .NET installs, then they should probably be using admin installs.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+A disadvantage of this setup is that older .NET Installers that do not respect the setting would still interfere with `bootstrapper` by modifying the system `PATH`. (For example, the `pkg` installer uses `tee` which overwrites the `PATH` file rather than just appending to it.) While we can potentially service these installers, older versions would still cause this issue. Addressing this comprehensively would require modifying the application host or `dotnet.exe`, which is not achievable within the .NET 10 timeframe and would need to be targeted for .NET 11. The advantage is that the application host always runs on the latest version, which would enable it to handle older installers appropriately.
```
dsplaisted, line 48: This disadvantage doesn't apply as much as it may seem.  The installer component that sets the path is the one that lays out dotnet.exe and hostfxr, which updates in-place to the latest version.  So as long as any admin version of .NET 10 was installed, installing or updating an earlier version of .NET would not update this component and hence wouldn't modify the system PATH.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Multi-level-lookup
+
+`DOTNET_MULTILEVEL_LOOKUP` was designed to enable or disable the host/muxer/`dotnet.exe`'s ability to find SDKs or Runtimes in other locations. We could re-enable the muxer to look specifically at the directory of local installs that are placed by `bootstrapper`, as proposed to be `DOTNET_HOME` in the other specification, [`dotnet-bootstrapper-cli-experience.md`](./dotnet-bootstrapper-cli-experience.md).
+
+This would enable the lookup to work without modification to the `PATH`. We could add back this feature in a limited capacity and only with explicit user consent. However, `DOTNET_MULTILEVEL_LOOKUP` was deprecated and removed in .NET 7 due to user confusion (when it would find global SDKs in a user local context) and for performance reasons. The `host` team and others indicated that this feature presented significant challenges when it was available. When installers set conflicting paths, it creates coordination issues between different installation mechanisms, so it may be better to view this as an installer problem like above.
```
dsplaisted, line 58:
```suggestion
### Allow user and admin installs to work together

Instead of switching to a mode where only user installs are available, we could have a way that both user installs and admin installs would be accessible.  This would reduce the impact to Visual Studio, essentially installing a user install of .NET would be no different than installing the same version as an admin install.  In either case, installing a new version would simply make an additional SDK version available, without removing access to any versions already installed.

We had a feature called multi-level lookup that supported something similar to this.  It was removed in .NET 7 because it was confusing and unexpected behavior, among other reasons.  We believe we could add a new feature that would allow local and admin installs to be used together, but that would avoid some of the issues that multi-level lookup had.

However, it would still add complexity and given the negative experience around multi-level lookup, it might not completely mitigate all the issues.  So we propose to go with the simpler option of allowing the user to switch to a mode where user installs are used instead of admin installs.  If we get significant feedback that users would like to have both types available together, we can consider adding support for that.
```

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+What we could do is look for those installations and decline to do anything when a user tries to run `bootstrapper` if they are present on disk. The error message would point the user toward leveraging the `sdk: paths []` option in their project or repositories `global.json` file, which would specify to the host or `dotnet.exe` where to look for the installs of the .NET SDK. The host is already aware of this as of .NET 10.
+
+> 📓 Usage Statistic:
+> Only ~4% of users are on a `global.json` according to internal metrics, so this gesture will create additional work when using `dotnet` for a large number of customers.
+
+The downside of this approach is that it creates an additional step for a significant number of .NET users. Requiring extra configuration files to function properly presents a usability challenge and increases the barrier to entry for new users. However, the `paths` feature itself provides considerable value for repository-specific SDK management, which is why it has already been implemented and made available.
```
dsplaisted, line 69: Another downside is that as currently implemented, the `paths` feature only supports repo installs and not user installs (because it doesn't support variables/tokens to point to something like the home or LOCALAPPDATA fodler.  This means that different repos would not be able to share installations of the .NET SDK, wasting disk space for developers that use multiple repos.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+On the other hand, if `bootstrapper` set the system `PATH`, it would modify the behavior of Visual Studio, as Visual Studio would no longer find its installs. This could be considered a breaking change to Visual Studio. In this scenario, Visual Studio may begin to run on a different SDK than the one it shipped with, based on whichever SDKs the user has installed that can be used with the user project (based on the major.minor version of .NET and host resolution logic for the specific context.) This is already possible today by installing a different administrative SDK that is newer than the one Visual Studio has. However, it is not possible to install a downgrade of an administrative install in the specific version-set without uninstalling the SDK owned by Visual Studio first. This also means that updates of the SDK that Visual Studio includes would no longer take effect within Visual Studio, as the update mechanism would need to be handled by `bootstrapper`.
+
+An additional concern would be the `source-built` SDKs and the `pkg` installer.
+
+There have been several methods suggested to handle this, which are listed below.
```
dsplaisted, line 103: I think you can remove this line, I think it made sense when this section was at the beginning of the document, but not now.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Update the Installers / Visual Studio
+
+:star: A framing document and prototype made by the .NET SDK team relies on and suggests this proposal. We believe this to be the best proposal.
+
+One option is to set the machine into a 'local' SDK mode that users could opt into when users engage with `bootstrapper` to do their first local install. This 'mode' would tell the system-wide installers (both `pkg` and `msi`) to not set the system `PATH`.
+
+In this scenario:
+
+1. When the user interfaces with the `bootstrapper`, to install an SDK, they are prompted about the change to `local` installs. They are also informed about the impact this has on VS and can pick an update cadence.
```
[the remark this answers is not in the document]
nagilson: Yeah, I can see that being better UX. The guidance we have is to check for updates daily. In theory we could also utilize patch tuesday but we often ship off schedule nowadays. I think instead, we may want to mention that it would be helpful to have a story in place for VS calling our bootstrapper to run updates when it launches, since if the user doesn't interact with the CLI, they wouldn't get those daily updates, but the SDK would still be used.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
nagilson: Ah, I think I had meant to enable VS to use an 'admin' SDK rather than a local one.

nagilson: Which doesn't work very well due to the lack of env variables explained below

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
+# Integrating a 'dotnet' Bootstrapper with Visual Studio (PATH Issues)
+
+## Introduction
+
+How can we enable an `nvm` or `rustUp` like experience for `dotnet`? In this document, the command to do so will be aliased as `bootstrapper`, and we will only discuss PATH concerns. This may be `dotnet` going forward, or something else.
```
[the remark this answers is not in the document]
nagilson: Hopefully Ive updated it to be more clear (full commit push with everything coming soon)

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+
+One of the core differences in our ecosystem compared to other toolchains is the inclusion of .NET in VS. Visual Studio and the MSI or 'global' .NET Installer set the System `PATH` on Windows and not the user level `PATH`. The system `PATH` always takes precedence over the user `PATH`, which is how other toolchains manage user local install folders. Below are proposals to remedy this situation.
```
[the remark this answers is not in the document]
nagilson: Great feedback, I like this way of making the framing more concise. I tried to include more detail while keeping your improvements.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
```diff
@@ -0,0 +1,130 @@
...
+In this scenario:
+
+1. When the user interfaces with the `bootstrapper`, to install an SDK, they are prompted about the change to `local` installs. They are also informed about the impact this has on VS and can pick an update cadence.
+
+2. On Windows, a registry key is set in `HKLM` (the registry hive for local user contexts). `bootstrapper` must elevate to enable this registry edit when prompting the user.
+
+> 📓 Technical Details:
+> A registry key is preferred over an environment variable because it is an admin-protected context. We don't want user-local controllable behavior to influence/break the behavior of an administrative installer, as that is a security concern. A policy key may also be used, but a policy key is essentially equivalent to a registry key in this context.
+>
+> On Unix, we could set this via a protected sentinel file.
```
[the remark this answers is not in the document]
nagilson: The mac PATH is a root protected file, so I think we need to do something in this scenario. A sentinel file may not be the best approach, but I think something similar is just and we dont need to decide that now.

proposed/dotnet-bootstrapper/dotnet-bootstrapper-path.md
nagilson: I'm unsure of the technical aspects of this so I've reflected that in the doc, this is interesting

nagilson: Thanks for that context, I will update the document.

jkotas: > There is a separate installer for host FXR and the host. The host MSI installs dotnet.exe and the component responsible for the PATH.

Are these part of Runtime installer or SDK installer? For example, if I install ".NET Runtime" or ".NET Desktop Runtime" only, do I still get `dotnet` on PATH?

Can these PATH modifications affect non-SDK scenarios? If the PATH is modified as part Runtime installer, it is likely that there are applications out there that depend on dotnet.exe to be on the path, and switching the machine to local SDK mode may break them in various ways.

nagilson: >Can these PATH modifications affect non-SDK scenarios 

Yes, there are a lot of security boundaries we would have to consider w.r.t runtimes, such as another account relying on the system wide PATH for a runtime like described above.

PR #336, opened 2025-07-16
Add related links to binding design

richlander:
I am writing a doc and wanted to link to this doc and noticed that only half the related links I wanted were there.

I always forget where to find info about `COREHOST_TRACE` (because I always forget the name). Having one more link to it will help.


richlander:
@elinor-fung

PR #337, opened 2025-07-24
Add draft of combined tool + library packages

baronfel:
This adds a spec for an idea I've been discussing with @joelverhagen and @DamianEdwards now that we have a very lightweight tool-manifest package pattern with .NET 10's RID-specific tool packages.

The short form is that users should be able to author one package that can be used as a tool _and_ a PackageReference, without

a) impacting the behavior of either mode, 
b) incurring large download size increase costs, and
c) sacrificing simplicity of project files

### DIFF
--- a/proposed/combined-tool-library-packages.md
+++ b/proposed/combined-tool-library-packages.md
@@ -147,11 +147,11 @@ To make this work, changes will be needed on both the consumption and packaging 
 
 ### Consumption changes for tools
 
-* No changes required - as long as a package has the DotNetTool package type, `dnx` and all other tool-interaction commands will use the package as a tool.
+* No changes required - as long as a package has the DotnetTool package type, `dnx` and all other tool-interaction commands will use the package as a tool.
 
 ### Consumption changes for libraries
 
-The primary change that will be required is to loosen the checks made by the [CompatibilityChecker](https://github.com/NuGet/NuGet.Client/blob/4ce65d4a1c482eda1c8656fc032d1f5cf247763a/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs) to allow using packages with the explicit PackageType of `Dependency` as a library - regardless of their other attributes. Today, this checking takes many characteristics of the package into account, but it has a [hard deny](https://github.com/NuGet/NuGet.Client/blob/4ce65d4a1c482eda1c8656fc032d1f5cf247763a/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L339-L347) for `DotNetTool` packages. If this restriction is removed, the package can be used as a library without any other changes.
+The primary change that will be required is to loosen the checks made by the [CompatibilityChecker](https://github.com/NuGet/NuGet.Client/blob/4ce65d4a1c482eda1c8656fc032d1f5cf247763a/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs) to allow using packages with the explicit PackageType of `Dependency` as a library - regardless of their other attributes. Today, this checking takes many characteristics of the package into account, but it has a [hard deny](https://github.com/NuGet/NuGet.Client/blob/4ce65d4a1c482eda1c8656fc032d1f5cf247763a/src/NuGet.Core/NuGet.Commands/RestoreCommand/CompatibilityChecker.cs#L339-L347) for `DotnetTool` packages. If this restriction is removed, the package can be used as a library without any other changes. Packages with no package type will continue to be treated as Dependency packages.
 
 ### Packaging changes
 
--- a/proposed/combined-tool-library-packages.md
+++ b/proposed/combined-tool-library-packages.md
@@ -1,0 +1,266 @@
+# Combined Tool and Library Packages
+
+**Status**: Proposed
+
+**Owner** [Chet Husk](https://github.com/baronfel)
+
+## Summary
+
+This proposal introduces a new packaging model for .NET that allows developers to create packages that serve dual purposes: as command-line tools and as libraries. This approach enables a single package to contain both executable tool functionality and traditional library assets, allowing developers to invoke functionality directly via the command line while also being able to reference the same package as a dependency in their projects.
+
+## Motivation
+
+.NET tools provide a convenient way to distribute and consume command-line utilities, but they exist in isolation from the libraries they might be built upon. This proposal introduces "combined packages" that contain both executable tool functionality and traditional library assets (lib/, ref/, analyzer/, build/ items), enabling developers to both invoke functionality directly via `dnx` and reference the same package as a dependency in their projects. This mirrors patterns common in Node.js (`npm run <command>` / `npm bin <command>`) and Go (`go install` for CLI, regular import for library usage), bringing similar convenience to the .NET ecosystem.
+
+## Scenarios and User Experience
+
+### Scenario 1: Tool Author - Creating a Combined Package
+
+A developer building a JSON schema validator wants to provide both a CLI tool for CI/CD pipelines and a library for programmatic validation within .NET applications.
+
+```xml
+<!-- JsonSchemaValidator.csproj -->
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <PackAsTool>true</PackAsTool>
+    <ToolCommandName>json-validate</ToolCommandName>
+    <PackageId>JsonSchemaValidator</PackageId>
+    <Version>1.0.0</Version>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
+  </ItemGroup>
+</Project>
+```
+
+The project structure includes both a CLI entry point and library classes:
+
+```csharp
+// Program.cs - CLI entry point
+class Program
+{
+    static async Task<int> Main(string[] args)
+    {
+        if (args.Length == 0)
+        {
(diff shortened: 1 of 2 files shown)


proposed/combined-tool-library-packages.md
```diff
@@ -0,0 +1,267 @@
...
+To users this change is completely transparent - they interact with the package as a tool or as a library as in scenarios 2 and 3, but the package is now optimized for the platform it is running on.
+
+
+## Making it work
+
+To make this work, changes will be needed on both the consumption and packaging sides.
+
+### Consumption changes for tools
+
+* No changes required - as long as a package has the DotNetTool package type, `dnx` and all other tool-interaction commands will use the package as a tool.
```
joelverhagen, line 150: nit:
```suggestion
* No changes required - as long as a package has the DotnetTool package type, `dnx` and all other tool-interaction commands will use the package as a tool.
```

proposed/combined-tool-library-packages.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Scenario 1: Tool Author - Creating a Combined Package
+
+A developer building a JSON schema validator wants to provide both a CLI tool for CI/CD pipelines and a library for programmatic validation within .NET applications.
+
+```xml
+<!-- JsonSchemaValidator.csproj -->
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
```
joelverhagen, line 25: How does multi-targetting play in here? If I had `<TargetFrameworks>net10.0;net9.0</TargetFrameworks>` would this impact both the library and the tool assets?

proposed/combined-tool-library-packages.md
```diff
@@ -0,0 +1,267 @@
...
+#### Managing tool PackageReferences and Program.cs
+
+It's very possible for the Tool expression of a project to need different dependencies than the Library expression. For example, a tool might need a command-line parsing library. To handle this, we can use conditional references in the project file to detect the 'mode' we are in:
+
+```xml
+<PackageReference Include="CommandLineParser" Version="2.9.1" Condition="'$(OutputType)' == 'Exe'" />
+<PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.0" />
+```
+
+However, ideally none of the _entrypoint_ (or code only used by the entrypoint) of the application would be included in the library portion of the package. To achieve this, we can condition the removal of the Entrypoint-related code from the project:
```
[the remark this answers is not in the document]
baronfel: yeah, exactly - it's not _required_ per se, but I want to aim for the moon here in terms of not influencing the library code from the program if possible.

proposed/combined-tool-library-packages.md
```diff
@@ -0,0 +1,267 @@
...
+### Scenario 1: Tool Author - Creating a Combined Package
+
+A developer building a JSON schema validator wants to provide both a CLI tool for CI/CD pipelines and a library for programmatic validation within .NET applications.
+
+```xml
+<!-- JsonSchemaValidator.csproj -->
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <PackAsTool>true</PackAsTool>
```
joelverhagen, line 26: Seems like there should be a way to opt in to the new experience. This .csproj looks just like an existing tool project, but now it's magically adding Dependency assets? Or am I missing something?

proposed/combined-tool-library-packages.md
baronfel: good question! the answer depends on the _kind_ of tool. framework-dependent, platform-agnostic tools would bundle the N TFM variants of the tool in the package. framework-dependent, platform-specific tools could make the N platform-specific tools, but our existing model actually doesn't handle _TFM_ variants of platform-specific tools at this time. I don't actually know what would happen here. For self-contained tools, there is no need to have TFM-specific variants at all, and so we should strongly consider taking the 'newest' TFM and only making the platform-specific variants for that TFM.

proposed/combined-tool-library-packages.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+### Compatibility Considerations
+
+- Existing .NET tool packages continue to work unchanged
+- Existing library packages continue to work unchanged
+- Combined packages can be consumed as either tools or libraries by consumers unaware of the dual nature
+- NuGet clients that don't understand combined packages treat them as either tools or libraries based on which assets they recognize
+
+## Q & A
```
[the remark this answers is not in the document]
baronfel: publishes the project as a library, by default. This is because the project has no `OutputType` set, so the defaults it to a `Library` `OutputType`.

proposed/combined-tool-library-packages.md
```diff
@@ -0,0 +1,267 @@
...
+```
+JsonSchemaValidator.win-x64.1.0.0.nupkg
+├── tools/
+│   └── any/
+│       └── win-x64/
+│           ├── json-validate.exe
+└── [package metadata files]
+```
+
+### Compatibility Considerations
```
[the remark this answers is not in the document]
baronfel: that's good confirmation - we do today, and will continue to have in this proposal, separate binaries for the tool usage and the library usage. [This layout](https://github.com/dotnet/designs/pull/337/files#diff-9137432a9f8e2cf805c300061c5ea632ac0ca73cbd4124d868545a7113db1724R197-R210) is intended to show this separation. The idea would be that the entrypoint would _not_ be part of the library-use-case dlls in any way - only the tool-use-case dlls.

proposed/combined-tool-library-packages.md
```diff
@@ -0,0 +1,267 @@
...
+
+## Motivation
+
+.NET tools provide a convenient way to distribute and consume command-line utilities, but they exist in isolation from the libraries they might be built upon. This means that a developer must create multiple packages, push multiple package ids to feeds, consumers must search for and use seemingly-random package names, etc. Our current model forces a package author to choose which _modality_ of package consumption (tool or reference) gets the 'nice' package name, while the other unused modality gets whatever is left. Allowing for the nice/concise/expected package name to fill both needs helps with name recognition and makes a tool/library appear more cohesive. This mirrors patterns common in Node.js (`npm run <command>` / `npm bin <command>`) and Go (`go install` for CLI, regular import for library usage), bringing similar convenience to the .NET ecosystem.
+
+## Scenarios and User Experience
```
richlander, line 15: I'm looking at this space. I'd like to see the package format expand to allow tools and libraries in the same package. However, I think the idea of doing that with one project should be a separate proposal. Combining the two ideas is confusing.

javiercn:
We have a concrete, shipping use case for this in the Blazor gateway, and it maps almost exactly onto the two modalities described here.

The Blazor gateway (`Microsoft.AspNetCore.Components.Gateway` in dotnet/aspnetcore) is an ASP.NET Core executable that fronts *standalone* Blazor WebAssembly apps: it serves the static web assets, adds OTLP telemetry, and does service discovery / reverse proxying to backends. The gateway is multi-app by design (it can front several standalone WASM apps at once, as in an Aspire graph); the single-app case is just an optimization of that. We need to consume this one artifact both as a `PackageReference` and as a tool, without splitting it into separate package ids.

**Scenario 1: standalone Blazor WebAssembly dev inner loop (PackageReference).**
For development, we simply want a `PackageReference` to consume the gateway in the app. The package carries build logic (a `build/…targets`) that rewrites `RunCommand`/`RunArguments`, so `dotnet run` and F5 transparently launch the gateway to host the app locally (dev-time static assets, telemetry, backend routing). The developer adds one package and their run/launch stack changes to "run behind the gateway," with no project-file gymnastics and no second package to discover.

**Scenario 2: Aspire orchestration (tool, driven internally by the integration).**
In Aspire we want the AppHost to run the same gateway across the standalone WASM apps in the graph. Today we do this with a *file-based C# app* (`Gateway.cs`) that we pack into `Aspire.Hosting.Blazor` and copy into the AppHost output, then run. That is a workaround. What we actually want is to consume the real gateway through Aspire's existing dotnet-tool integration. The user does not write this: our Blazor integration calls `AddDotnetTool` internally and passes the gateway its arguments (static web assets manifests, endpoints manifests, path prefixes, logging filters, and so on):

```csharp
// inside Aspire's Blazor integration, not user code
builder.AddDotnetTool(name, "Microsoft.AspNetCore.Components.Gateway")
       .WithArgs(/* per-app manifests, path prefixes, ... */);
```

**Scenario 3: validating a published standalone app (tool).**
We also want an easy way to run the *published* output of a standalone WASM app to validate it (not just the dev build). Because the same package is a tool, a user (or CI) can acquire the gateway on demand and point it at the publish output to serve and smoke-test the real published assets, using the same tool-acquisition path rather than a bespoke server. No dev build, no extra package.

**Why the combined package matters to us.**
These experiences must stay in lockstep: same gateway behavior, same version, same configuration surface. If they diverge, a standalone project, an Aspire-orchestrated graph, and a published-output check behave differently, which is a support nightmare. A combined tool + library package as proposed here lets us ship a single package id that is simultaneously tool-acquirable and PackageReference-able with build targets, retire the file-based `Gateway.cs` in Aspire, and converge on the real gateway binary. Leaning on the RID-specific tool manifest model also keeps the reference modality small, which matters because the gateway is an executable.

**Desired end-state experience**

- *Standalone WASM dev:* add one PackageReference, `dotnet run`/F5 hosts the app behind the gateway.
- *Aspire:* the Blazor integration internally acquires the same gateway as a tool and runs it over the WASM apps in the graph, with no copied source files and no separate package.
- *Published-output validation:* acquire the same tool on demand and point it at the publish output to validate the real published app.

All three resolve the same package id at the same version.

PR #338, opened 2025-07-30
Add spec on cve.json schema

richlander:
We publish vulnerability fixes and matching CVE documentation nearly ever month. These are primarily [documented on GitHub in HTML](https://github.com/dotnet/announcements/issues?q=is%3Aissue%20state%3Aopen%20label%3ASecurity) for people to read. It is well understood that HTML is a presentation format and not appropriate for structured processing. We intend to offer CVE information in structured JSON format (and likely markdown, too).

This schema has been designed around a "core + extensions" model that reflects how many modern software platforms are structured - with core product components and an ecosystem of extensions. This pattern is prevalent across the industry (editors with extensions, runtimes with packages, apps with plugins), making this schema broadly applicable beyond its initial .NET use case. The design intentionally avoids product-specific terminology to support reuse by other offerings with similar security disclosure needs.

I also have a C# object model for the schema. I'll share that shortly. It's currently broken because I made a bunch of schema changes as a result of insight from writing the spec.

### DIFF
--- a/accepted/2025/cve-schema/cve.json
+++ b/accepted/2025/cve-schema/cve.json
@@ -1,0 +1,210 @@
+{
+  "date": "2024-07-09",
+  "cves": [
+    {
+      "id": "CVE-2024-30105",
+      "title": ".NET Denial of Service Vulnerability",
+      "severity": "Critical",
+      "cvss": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H/E:U/RL:O/RC:C",
+      "description": [
+        "A vulnerability exists in .NET when calling the JsonSerializer.DeserializeAsyncEnumerable method against an untrusted input using System.Text.Json may result in Denial of Service."
+      ],
+      "brand": ".NET",
+      "references": [
+        "https://github.com/dotnet/announcements/issues/315"
+      ]
+    },
+    {
+      "id": "CVE-2024-35264",
+      "title": ".NET Remote Code Execution Vulnerability",
+      "severity": "Critical",
+      "cvss": "CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:U/C:H/I:H/A:H/E:U/RL:O/RC:C",
+      "description": [
+        "A Vulnerability exists in ASP.NET Core 8 where Data Corruption in Kestrel HTTP/3 can result in remote code execution.",
+        "Note: HTTP/3 is experimental in .NET 6.0. If you are on .NET 6.0 and using HTTP/3, please upgrade to .NET 8.0.7"
+      ],
+      "brand": "ASP.NET Core",
+      "references": [
+        "https://github.com/dotnet/announcements/issues/314"
+      ]
+    },
+    {
+      "id": "CVE-2024-38081",
+      "title": ".NET Denial of Service Vulnerability",
+      "severity": "Critical",
+      "cvss": "CVSS:3.1/AV:L/AC:L/PR:L/UI:R/S:U/C:H/I:H/A:H/E:U/RL:O/RC:C",
+      "description": [
+        "A vulnerability exists in Visual Studio installer on Windowswhere an unprivileged user is allowed to manipulate the Visual Studio installation resulting in elevation of privilege."
+      ],
+      "brand": ".NET",
+      "platforms": [
+        "Windows"
+      ],
+      "references": [
+        "https://github.com/dotnet/announcements/issues/313"
+      ]
+    },
+    {
+      "id": "CVE-2024-38095",
+      "title": ".NET Denial of Service Vulnerability",
+      "severity": "Critical",
+      "cvss": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:L/E:U/RL:O/RC:C",
+      "description": [
+        "A vulnerability exists when System.Formats.Asn1 in .NET parses an X.509 certificate or collection of certificates, a malicious certificate can result in excessive CPU consumption on all platforms result in Denial of Service."
+      ],
+      "brand": ".NET",
+      "references": [
+        "https://github.com/dotnet/announcements/issues/312"
+      ]
+    }
+  ],
+  "core": [
+    {
+      "cve-id": "CVE-2024-38081",
+      "name": "Microsoft.NETCore.App.Runtime",
+      "min-vulnerable": "6.0.0",
+      "max-vulnerable": "6.0.31",
+      "fixed": "6.0.32",
+      "release": "6.0",
+      "commits": [
+        "0a0dd0e27560e692e11ee286ed9f45471b2131fa"
+      ]
+    },
+    {
+      "cve-id": "CVE-2024-38095",
+      "name": "Microsoft.NETCore.App.Runtime",
+      "min-vulnerable": "6.0.0",
+      "max-vulnerable": "6.0.31",
+      "fixed": "6.0.32",
+      "release": "6.0",
+      "commits": [
+        "979135d5ca4efaf6436ee13539cc3f1e039d570a"
+      ]
+    },
+    {
+      "cve-id": "CVE-2024-30105",
+      "name": "Microsoft.NETCore.App.Runtime",
+      "min-vulnerable": "8.0.0",
+      "max-vulnerable": "8.0.6",
+      "fixed": "8.0.7",
+      "release": "8.0",
+      "commits": [
+        "fa5b0d8f4a8b424732cc992158aa92842f8a2846"
+      ]
+    },
+    {
+      "cve-id": "CVE-2024-35264",
+      "name": "Microsoft.AspNetCore.App.Runtime",
+      "min-vulnerable": "8.0.0",
+      "max-vulnerable": "8.0.6",
+      "fixed": "8.0.7",
+      "release": "8.0",
+      "commits": [
+        "c5721fb7a65ddc13d1b445c2c08c27b72ab57cdc"
+      ]
+    },
+    {
+      "cve-id": "CVE-2024-38095",
+      "name": "Microsoft.NETCore.App.Runtime",
+      "min-vulnerable": "8.0.0",
+      "max-vulnerable": "8.0.6",
+      "fixed": "8.0.7",
+      "release": "8.0",
+      "commits": [
+        "4a8d5a007971d19f389ca17f7b8eb4f9bb199991"
+      ]
+    }
+  ],
+  "extensions": [
+    {
+      "cve-id": "CVE-2024-38081",
+      "name": "Microsoft.IO.Redist",
+      "min-vulnerable": "4.6.0-preview.18571.3",
+      "max-vulnerable": "6.0.0",
+      "fixed": "6.0.1",
+      "commits": [
+        "0a0dd0e27560e692e11ee286ed9f45471b2131fa"
+      ]
+    },
+    {
+      "cve-id": "CVE-2024-38095",
+      "name": "System.Formats.Asn1",
+      "min-vulnerable": "5.0.0-preview.7.20364.11",
+      "max-vulnerable": "6.0.0",
+      "fixed": "6.0.1",
+      "release": "6.0",
+      "commits": [
+        "979135d5ca4efaf6436ee13539cc3f1e039d570a"
+      ]
+    },
+    {
+      "cve-id": "CVE-2024-38095",
+      "name": "System.Formats.Asn1",
+      "min-vulnerable": "7.0.0-preview.1.22076.8",
+      "max-vulnerable": "8.0.0",
+      "fixed": "8.0.1",
+      "release": "8.0",
+      "commits": [
+        "4a8d5a007971d19f389ca17f7b8eb4f9bb199991"
+      ]
+    },
+    {
+      "cve-id": "CVE-2024-30105",
+      "name": "System.Text.Json",
+      "min-vulnerable": "8.0.0",
+      "max-vulnerable": "8.0.3",
+      "fixed": "8.0.4",
+      "release": "8.0",
+      "commits": [
+        "fa5b0d8f4a8b424732cc992158aa92842f8a2846"
+      ]
+    }
+  ],
+  "commits": {
+    "fa5b0d8f4a8b424732cc992158aa92842f8a2846": {
+      "repo": "runtime",
+      "branch": "release/8.0",
+      "hash": "fa5b0d8f4a8b424732cc992158aa92842f8a2846",
+      "org": "dotnet",
+      "url": "https://github.com/dotnet/runtime/commit/fa5b0d8f4a8b424732cc992158aa92842f8a2846"
+    },
+    "c5721fb7a65ddc13d1b445c2c08c27b72ab57cdc": {
+      "repo": "aspnetcore",
+      "branch": "release/8.0",
+      "hash": "c5721fb7a65ddc13d1b445c2c08c27b72ab57cdc",
+      "org": "dotnet",
+      "url": "https://github.com/dotnet/aspnetcore/commit/c5721fb7a65ddc13d1b445c2c08c27b72ab57cdc"
+    },
+    "0a0dd0e27560e692e11ee286ed9f45471b2131fa": {
+      "repo": "runtime",
+      "branch": "release/6.0",
+      "hash": "0a0dd0e27560e692e11ee286ed9f45471b2131fa",
+      "org": "dotnet",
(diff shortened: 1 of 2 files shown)


accepted/2025/cve-schema/spec.md
```diff
@@ -0,0 +1,483 @@
+# Publishing CVE information as structured data
+
+We publish vulnerability fixes and matching CVE documentation nearly ever month. These are primarily [documented on GitHub in HTML](https://github.com/dotnet/announcements/issues?q=is%3Aissue%20state%3Aopen%20label%3ASecurity) for people to read. It is well understood that HTML is a presentation format and not appropriate for structured processing. We intend to offer CVE information in structured JSON format (and likely markdown, too).
```
Frulfump, line 3: Typo "ever month"

accepted/2025/cve-schema/spec.md
```diff
@@ -0,0 +1,483 @@
...
+
+This schema has been designed around a "core + extensions" model that reflects how many modern software platforms are structured - with core product components and an ecosystem of extensions. This pattern is prevalent across the industry (editors with extensions, runtimes with packages, apps with plugins), making this schema broadly applicable beyond its initial .NET use case. The design intentionally avoids product-specific terminology to support reuse by other offerings with similar security disclosure needs.
+
+We've taken a significant deparature with this schema design, battle-testing it with `jq` and some other tools as part of the development process. If the `jq` queries look bad, the schema is bad. We've also noticed in this new age that `jq` query ease is a proxy for LLM consumption ease. Great `jq` use will likely lead to good or better LLM consumption while bad `jq` queryability will likely lead to bad or worse LLM consumption. We intend to design all schema this way going forward.
```
Frulfump, line 7: Nice insight! What's the departure?

accepted/2025/cve-schema/spec.md
```diff
@@ -0,0 +1,483 @@
...
+
+This new CVE format will be part of a larger queryable information graph that we'll document soon. For now, we'll focus on the CVE format.
+
+## Quick pitch
+
+Common queries answered by this schema:
+
+- "Which CVEs were reports this month?" → `.cves[].id`
```
Frulfump, line 15: Typo "were reports"

accepted/2025/cve-schema/spec.md
```diff
@@ -0,0 +1,483 @@
...
+### 1. Arrays vs Dictionaries
+
+We use different data structures based on access patterns:
+
+- **Arrays for data (discovery)**: `cves[]`, `core[]`, and `extensions[]` contain the actual CVE and affected component data as arrays because users discover this information rather than knowing it a priori. The naming directly reflects the "core + extensions" model
+- **Dictionaries for indices (lookup)**: `commits{}`, `cve-commits{}`, and `release-cves{}` provide fast lookups for known keys
+
+This clear separation - arrays for what happened, dictionaries for finding things - creates a consistent mental model and emphasizes the role of indices as the primary lookup mechanism.
+
+We started with `product` and `packages` since that's the "core + extensions" model for .NET. That didn't feel like a strong pairing, outside of great aliteration, so we fell back `core` and `extensions`.
```
Frulfump, line 32: fellback to?

accepted/2025/cve-schema/spec.md
richlander:
```suggestion
We publish vulnerability fixes and matching CVE documentation nearly every month. These are primarily [documented on GitHub in HTML](https://github.com/dotnet/announcements/issues?q=is%3Aissue%20state%3Aopen%20label%3ASecurity) for people to read. It is well understood that HTML is a presentation format and not appropriate for structured processing. We intend to offer CVE information in structured JSON format (and likely markdown, too).
```

accepted/2025/cve-schema/spec.md
richlander:
```suggestion
- "Which CVEs were reported this month?" → `.cves[].id`
```

accepted/2025/cve-schema/spec.md
richlander: Doing a better job? More seriously, I've been using LLMs a lot and have learned a lot about how they work.

One insight is that LLMs having trouble using existing products is almost always an indication that we prioritized one set of human users over another. It is rarely the case that there is a complete divide between LLM and human usage patterns.

accepted/2025/cve-schema/cve.json
```diff
@@ -0,0 +1,210 @@
+{
+  "date": "2024-07-09",
+  "cves": [
+    {
+      "id": "CVE-2024-30105",
+      "title": ".NET Denial of Service Vulnerability",
+      "severity": "Critical",
+      "cvss": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H/E:U/RL:O/RC:C",
```
omajid, line 8: Is the plan to use this field name even if the value later switches to CVSS 4.0 (or some other) format?

baronfel: This feels similar to me to how container tooling uses the `digest` field - a digest is a string of the form `<digest_type>:<digest_value>`, and the type tells tooling how to interpret the value. As long as the various CVE-description formats follow a similar philosophy, maybe the name of this field could be more generic ("characteristics"?) and the values in the field would be self-describing.

kaniini: My suggestion would be to do something like:

```json
"cvss": {"version": "3.1", "uri": "cvss:3.1:..."}
```

richlander:
One of my colleagues recommended I look at this spec for design insight: https://cve.mitre.org/cve/list_rules_and_guidance/cve_assignment_information_format.html. I'll post something about that shortly. It's pretty easy to see that it doesn't demand a re-design. I think two systems are already very close.

accepted/2025/cve-schema/cve.json
leecow: The current format (`CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:H/E:U/RL:O/RC:C`) is provided to us by the security team. The suggesting offered by @kaniini above looks good make the version explicit.

richlander: Resolved.

richlander:
I updated the spec.

richlander:
FYI: Almost done. I found the need for some more design changes.

richlander:
You will need to talk to @leecow about that. It's outside my scope.

richlander:
I just pushed a few more commits. I wanted to re-work the query section to make it more clear and have a better flow. I walso wanted to add some JSON -> MD examples.

accepted/2025/cve-schema/cve.json
```diff
@@ -0,0 +1,217 @@
...
+    },
+  "cve-releases": {
+      "CVE-2024-30105": ["8.0"],
+      "CVE-2024-35264": ["8.0"],
+      "CVE-2024-38081": ["6.0"],
+      "CVE-2024-38095": ["6.0", "8.0"]
+    },
+  "release-cves": {
+    "6.0": ["CVE-2024-38081", "CVE-2024-38095"],
+    "8.0": ["CVE-2024-30105", "CVE-2024-35264", "CVE-2024-38095"]
```
MichaelSimons, line 215: Are you applying a formatter or doing this by "hand"?  It seems inconsistent on when a single line vs multiple are used.

**Single**
```
      "commits": [
        "fa5b0d8f4a8b424732cc992158aa92842f8a2846"
      ]
```
```
      "references": [
        "https://github.com/dotnet/announcements/issues/312"
      ]
```

**Multiple**
`"CVE-2024-38095": ["6.0", "8.0"]`
`"CVE-2024-38095": ["979135d5ca4efaf6436ee13539cc3f1e039d570a", "4a8d5a007971d19f389ca17f7b8eb4f9bb199991"]`
`"cvss": {"version": "3.1", "uri": "CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:N/I:N/A:L/E:U/RL:O/RC:C"},`

richlander: Whitespace is not part of the spec. It's just what happens to be there. I didn't run it through a formatter. The original version was hand-written to me. It went through many schema changes. All of those were AI-assisted and I didn't pay any attention to the formatting.

richlander:
One thing I realized is that some/most/all LLMs cannot add `.diff` to the commit URLs to get LLM friendly content due to the way LLM service firewalls work. We will be adding `.diff` to all the URLs in the actual cve.json files we produce. Tools that want to provide links for humans with the web view can strip it.

I had this clunky experience with LLMs on this exact topic for a while. This specific realization didn't hit me until yesterday.

The basic point is that if an experience is intended to be tool/LLM friendly, it needs to deliver that all the way. Using `.diff` links for commits is an example of that.

I'll add this point to the spec but wanted to clarify it in a comment first.

MichaelSimons (approved):

richlander:
I'm done for now. I won't merge this until I get the C# OM updated and generate a schema. That won't be until next week. That leaves time for more feedback.

richlander:
BTW: This spec is massive overkill. I used it as a personal learning experience.

leecow:
> You will need to talk to @leecow about that. It's outside my scope.

Chatted with @dviererbe and others in our Wed meeting and followed up further. It doesn't look like sharing or linking to the information is possible for the cve.json effort.

accepted/2025/cve-schema/cve.json
```diff
@@ -0,0 +1,221 @@
...
+    "CVE-2024-35264": ["8.0"],
+    "CVE-2024-38081": ["6.0"],
+    "CVE-2024-38095": ["6.0", "8.0"]
+  },
+  "cve-commits": {
+    "CVE-2024-30105": ["fa5b0d8f4a8b424732cc992158aa92842f8a2846"],
+    "CVE-2024-35264": ["c5721fb7a65ddc13d1b445c2c08c27b72ab57cdc"],
+    "CVE-2024-38081": ["0a0dd0e27560e692e11ee286ed9f45471b2131fa"],
+    "CVE-2024-38095": ["979135d5ca4efaf6436ee13539cc3f1e039d570a", "4a8d5a007971d19f389ca17f7b8eb4f9bb199991"]
+  }
```
udlose, line 220: @richlander How are the commit SHA's correlated back to the actual fix? Is this a manual process that a dev is responsible for? Is there a tag/keywork used in the GH commit? etc.

I know this is just an example and possibly crafted by hand, but wanted to call this out in case these were run thru a process. There are commit SHA'a for CVE-2024-38081, and CVE-2024-38095 that I would have expected to have been included but are not listed here. Notably, this PR - [CG alert cleaning on VS17.10#10724](https://github.com/dotnet/msbuild/pull/10724/commits) specifically states it contains fixes for CVE-2024-38081 and CVE-2024-38095: 
1. However, none of the relevant commits in that PR are listed [here](https://github.com/dotnet/designs/pull/338/files#diff-1a1a5b4850827cf220bd56e59a538c795bf0c91e16540ae6e027347b733a0b8bR218) nor [here](https://github.com/dotnet/designs/pull/338/files#diff-1a1a5b4850827cf220bd56e59a538c795bf0c91e16540ae6e027347b733a0b8bR219).
2. WRT CVE-2024-38081, there is an associated commit that is missing in this example:
  https://github.com/dotnet/runtime/commit/64331033cae32fd6166b321301baaf8f004e2336

Image

accepted/2025/cve-schema/cve-schema.md
```diff
@@ -0,0 +1,1047 @@
+# Publishing CVE information as structured data
+
+We publish vulnerability fixes and matching CVE documentation nearly every month. These are primarily [documented on GitHub in HTML](https://github.com/dotnet/announcements/issues?q=is%3Aissue%20state%3Aopen%20label%3ASecurity) for people to read. It is well understood that HTML is a presentation format and not appropriate for structured processing. We intend to offer CVE information in structured JSON format (and likely markdown, too).
```
tboby, line 3: Forgive me if I'm missing something, but isn't the *primary* place CVEs are documented in the primary source that the Microsoft CNA provides to NVD?

E.g. https://nvd.nist.gov/vuln/detail/cve-2024-38081#range-16178023 provides https://msrc.microsoft.com/update-guide/vulnerability/CVE-2024-38081 as the primary source.

There is no link to GitHub, and I rarely come across vulnerability aggregators referencing the github issues: They're not linked to from the CVE record, so why would they be trusted?

Automated tools use NVD as the source of truth for CVEs. e.g. https://security.snyk.io/vuln/SNYK-ALPINE320-DOTNET6BUILD-7448438

accepted/2025/cve-schema/cve.json
richlander: It is currently manual. It isn't clear how to change that.

A big challenge is that the fix and the CVE process run in parallel. It's not a good model to require commits to have this information or at least if we want the information to be correct (and we're absolutely not getting into the history rewriting game). All the CVE details are subject to change after a fix is committed. In general, we know the fix is needed so we prioritize that. There can be a lot of discussion about the category and severity. Sometimes, a fix that was thought to be a significant security issue isn't actually and just becomes a regular fix.

accepted/2025/cve-schema/cve.json
```diff
@@ -198,20 +194,28 @@
...
+    "CVE-2024-38081": ["6.0"],
+    "CVE-2024-38095": ["6.0", "8.0"]
+  },
+  "cve-commits": {
+    "CVE-2024-30105": ["fa5b0d8f4a8b424732cc992158aa92842f8a2846"],
+    "CVE-2024-35264": ["c5721fb7a65ddc13d1b445c2c08c27b72ab57cdc"],
+    "CVE-2024-38081": ["0a0dd0e27560e692e11ee286ed9f45471b2131fa"],
+    "CVE-2024-38095": ["979135d5ca4efaf6436ee13539cc3f1e039d570a", "4a8d5a007971d19f389ca17f7b8eb4f9bb199991"]
   }
 }
```
Frulfump, line 221: Missing final new line
https://github.com/dotnet/designs/blob/ece1f853d2ba4c174151291056ff29d9101d0091/.editorconfig#L18-L21

accepted/2025/cve-schema/cve-schema.md
richlander:
```suggestion
We publish vulnerability fixes and matching CVE documentation nearly every month. These are primarily [documented in HTML](https://github.com/dotnet/announcements/issues?q=is%3Aissue%20state%3Aopen%20label%3ASecurity) for people to read. It is well understood that HTML is a presentation format and not appropriate for structured processing. We intend to offer CVE information in structured JSON format (and likely markdown, too).
```

richlander:
These are all good topics. Some of it is beyond the format but the information it contains.

I'll address the topics:

- data correctness: Yes, this is a key issue. Resolution: (A) clarify how "implementors" of the schema should document products, and (B) do that as an implementor. This will address ".NET Runtime" vs "SDK".
- CSV vs JSON: There's a straightforward transformation for that, while interop with HTML is way harder. Resolution: If enough people ask for JSON -> CSV, we can provide that. JSON -> markdown is our primary output target for now.
- "could easily result in immediate patching of all developer workstations": I think this is an existing problem made no worse by this schema. If you've got ideas on how to address that, we're listening.
- "packages which are versioned in line with .NET releases": This one I thought _was_ addressed by the schema. Tell me more.
- "component you provide (e.g. System.Text.Json) and a product that consumes that component": I think of this as another existing problem w/o an obvious solution, made no worse by this schema.

Great feedback. Let's just separate concern about the overal process from concerns about this proposed format.

mateusrodrigues (commented):
I noticed that not all CVE entries have `platforms` like this one. Does the omission of this property mean that *all* platforms are affected? If so, should that be marked explicitly instead of relying on the consumer to make this assumption?

My initial suggestion is something like this:
```
      ...
      "platforms": [
        "All"
      ],
      ...
```
I considered adding all known .NET platform targets to the list instead, but that carries the risk of leaving out any community-supported platforms we don't know of. In that case, an entry that says "All" seems to be the best approach.

richlander:
This is a good idea. I can make that change. I've learned that there can be a desire to elide properties with default values but that this generally isn't a good idea.

richlander:
I made some changes to address some general concerns raised. See "dotnet-guidance.md".

richlander:
I made some significant changes to the schema based on JSON-related insight acquired from another project. The primary change is that I moved from kebab-case-lower to snake_case_lower. There is a significant ergonomic different between these two styles, with snake_case_lower being far superior. The problem is that "-" means substraction in jq which breaks dot notation.

PR #339, opened 2025-08-04
Add proposal for `dnup` CLI-based acquisition tool

baronfel:
[Viewing Link](https://github.com/dotnet/designs/blob/cli-acquisition-design/accepted/2025/cli-acquisition-tool.md)

There is a [Discussion thread on this Repo](https://github.com/dotnet/designs/discussions/340) for more detailed, threaded discussions so we keep the PR itself more clear.

This is the content from https://hackmd.io/oyAjpxTNQKa60ipv-MmYSQ?view, polished up a bit for the format and general layout of this repo.

Note that the name `dnup` is not solid by any means - it's used in this document as a placeholder, and when we have a finalized name we should update this.


proposed/cli-acquisition-tool.md
```diff
@@ -0,0 +1,149 @@
...
+
+### Acquisition inside VS Code
+
+Hector installs the C# DevKit extension on an otherwise-clean Windows machine. The DevKit extension recognizes the lack of a .NET SDK install and prompts Hector for permission to install the latest one. Hector accepts, and the DevKit extension downloads and runs the latest version of the dnup-init bootstrapper, instructing it to install the latest .NET 10 SDK to `%LOCALAPPDATA%\dotnet`, as well as updating Hector's user-level PATH. DevKit then continues initialization and Hector is able to begin development.
+
+### Acquisition with an LLM
+
+Ricardo wants to check what .NET versions have security vulnerabilities and update them. He asks his local GitHub Copilot which installed .NET versions are out of date. The agent talks to an acquisition MCP server that lets him know he has two older .NET versions that have known vulnerabilities and asks him if he wants to update. Then it calls `dnup update` to update to the latest version for each.
```
nagilson, line 36: Smart call to include the LLM bit here, I like this proposal a lot - is there a broader issue we could link to here such as https://github.com/microsoft/vscode-dotnettools/issues/2090#issuecomment-3070467289

proposed/cli-acquisition-tool.md
```diff
@@ -0,0 +1,149 @@
...
+- there is a 10.0.100-preview.6 version available
+
+Since 8.0.404 is pinned by a specific repo, that version is not updated/removed, but 9.0.303 and 10.0.100-preview.5 are updated to 9.0.304 and 10.0.100-preview.6 respectively.
+
+## Requirements
+
+### Goals
+
+- **Unified SDK Management**: Provide a single tool for acquiring, updating, and managing .NET SDKs across all supported platforms
+- **User-local Installation**: Focus on user-writable locations that can be managed without elevation after initial setup
```
nagilson, line 63: Should we add that in the long term this may replace the install scripts/uninstall tool on mac but that we don't see it replacing the install scripts at least in the near-term? Similar with the vscode extension logic.

proposed/cli-acquisition-tool.md
```diff
@@ -0,0 +1,149 @@
...
+
+Ricardo wants to check what .NET versions have security vulnerabilities and update them. He asks his local GitHub Copilot which installed .NET versions are out of date. The agent talks to an acquisition MCP server that lets him know he has two older .NET versions that have known vulnerabilities and asks him if he wants to update. Then it calls `dnup update` to update to the latest version for each.
+
+### Onboarding to a repository using .NET
+
+Mei wants to contribute to a .NET library that she found on GitHub. She clones the repository, but the repo uses a global.json to enforce a certain tooling version for all contributors be installed to a repo-local directory. Since Mei doesn't have that version, she runs `dnup install` (or her editor does so on her behalf) and that version is downloaded and installed into her local repository copy. After this install is done, the next time she launches her editor it detects the install and she begins work on the feature she's selected.
+
+### Updating all installed toolchains
+
+Gavin has a few different SDKs managed via `dnup install` on their machine in the following configuration:
```
nagilson, line 44: It might be worth differentiating here between what happens with `dnup install` if you're in a `local` mode or not, and whether that command can defer back to the setup experience if the user didn't use `-y`, etc.

proposed/cli-acquisition-tool.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Q & A
+
+### How will I get `dnup`?
+
+We intend to distribute `dnup` via a direct download initially. The dotnet download web site and the Learn documentation for getting started with .NET are the primary candidates. However, it is also common to distribute these kinds of bootstrappers via system package managers, so future efforts could include WinGet packages, Debian and Red Hat packages, a Homebrew cask, etc. The tool itself will be a Native AOT tool that is completely standalone, so acquisition should be relatively simple to set up.
+
+### How will this handle existing global installations?
+
+`dnup` will detect existing global installations but focus on managing user-local installs. Users with existing package manager installations will continue to use those tools. `dnup` will provide guidance when detecting unsupported installation types, and may offer to seed the user-local hive with the same versions so that there are no functionality gaps when starting with `dnup`.
```
nagilson, line 129:
```suggestion
`dnup` will detect existing global installations but focus on managing user-local installs. Users with existing package manager installations will continue to use those tools. `dnup` will provide guidance when detecting unsupported installation types, and may offer to seed the user-local hive with the same versions so that there are no functionality gaps when starting with `dnup`.
```


I like that you've called this out and thought of a lot of scenarios in the Q&A session. Another important point to me, is how the product clarifies that `dnup uninstall <program files sdk>` would not work, and how we message that - what type of guidance would we provide? A link/doc?

proposed/cli-acquisition-tool.md
```diff
@@ -0,0 +1,149 @@
...
+* install specific versions of the SDK tooling (10.0.100)
+* install logically-named versions of the SDK tooling (10, LTS, etc)
+* list the installed SDKs
+* uninstall specific versions of the SDK tooling
+* install the versions of components required by a given repo (as specified by global.json)
+* provide update checks/out-of-date information to end users about the installs they have
+
+### Initial setup
+
+On first-run, `dnup-init` will configure the users environment to support user-local hives. DOTNET_ROOT will be set to point to the `dnup` user-local root, the user's PATH will be appended to, and on Windows `dnup-init` will request to elevate to set a Windows Registry key that will prevent global .NET installers from setting the global .NET Install location on the System PATH (which would overwrite the user PATH always).
```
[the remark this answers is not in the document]
nagilson: I think it's good to call out how we'd want the tool to work if we don't elevate, +1

proposed/cli-acquisition-tool.md
```diff
@@ -0,0 +1,149 @@
...
+
+**Owner** [Chet Husk](https://github.com/baronfel) | [Daniel Plaisted](https://github.com/dsplaisted)
+
+Users of the .NET SDK today face a wide array of choices when getting our tooling, with multiple distribution channels that release at different cadences. While these options serve specific needs, several important user cohorts are not well served and have asked for a more unified approach to getting started with our tooling. This proposal introduces a standalone install and SDK management tool (tentatively named `dnup`) that provides a unified interface for acquiring and managing .NET tooling across all supported platforms, focusing on user-local installations that can be managed without elevation.
+
+## Scenarios and User Experience
+
+### Brand-new learner
+
+Sasha wants to learn C# and develop applications on her Mac. She searches and is led to a 'getting started' style page that directs her to download our bootstrapper. She navigates to `get.dot.net` and clicks the prominent link for `dnup-init`. She then clicks the downloaded exe, which runs the bootstrapper executable where she is prompted through an interactive setup:
```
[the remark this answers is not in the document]
baronfel: @mairaw would be our primary point of contact for anything around hosting bits with a nice url - but in this case the get.dot.net example is for illustration only. We wouldn't be able to promote this tool so heavily for the initial release.

For our initial releases we should probably just lean on an aka.ms if possible.

proposed/cli-acquisition-tool.md
baronfel: can you expand on this a bit? I'm not sure what you mean by `local` mode, etc.

nagilson: Yes, by this I mean when the user has enabled or disabled using the 'user local hive' by following what's described in `Initial setup`. 

We wanted to have a command to turn back off the 'user folder hive' and switch back to the 'system hive'.

If the user has switched back to the system hive, do we make `dnup install` run part of the `initial setup` experience again to go back to the 'user folder hive' or do we fail and prompt the user to rerun the initial executable? If we chose to fail, then I think that constrains us to the rustup design where there is a separate component that can only do initial setup and vice versa.

My vote would be to have `dnup install` rerun the intial setup if the user local hive is disabled (and to forward -y properly to auto accept, though this may require elevation, so we may want a separate setting for that, to say whether we want to elevate or not.)

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,217 @@
...
+She then clicks the downloaded executable, which runs the bootstrap process where she is prompted through an interactive setup:
+
+```bash
+Where should we install .NET to? *(~/Library/Application Support/dnup)*
+> <enter>
+What version should we install? *(latest)*
+> <enter>
+Should we set up your environment/PATH automatically? *(Y/n)*
+> <enter>
+Installing .NET SDK 10.0.100 to ~/Library/Application Support/dnup
```
dsplaisted, line 34: I think the installation folder should be named `dotnet`.

```suggestion
Where should we install .NET to? *(~/Library/Application Support/dotnet)*
> <enter>
What version should we install? *(latest)*
> <enter>
Should we set up your environment/PATH automatically? *(Y/n)*
> <enter>
Installing .NET SDK 10.0.100 to ~/Library/Application Support/dotnet
```

dsplaisted (approved):

accepted/2025/cli-acquisition-tool.md
baronfel: Fair - I do think the 'state' directory for this should be in a location that's more clearly-associated to the `dnup` tool though.

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,223 @@
...
+To help get ahead of this, we have begun work with the VS team to expand the set of cross-SDK/cross-VS version testing that is done in VS today.
+We will aim to keep the SDK and VS maximally-compatible so that VS can support consuming the SDKs across a broader range than previously documented.
+
+It's worth noting that this need for compatibility is not a problem unique to `dnup` - it exists today when users use locally-managed SDKs installed by Arcade, dotnet-install scripts, or other means.
+_Where_ the SDK is resolved from is largely orthogonal to _how compatible_ that SDK is with the IDE.
+
+### What future work does this enable?
+
+- Unifying IDE-specific and disparate install components that exist today
+  - Specifically components like the `dotnet-install` scripts and all downstream consumers of it (`actions/setup-dotnet`, `UseDotNet`)
```
martincostello, line 202: Will the tool have flags to enable replacing the scripts to acquire nightly builds too, not just the official releases?

martincostello (commented):
+1 to `dnup` not being the final name 😄

accepted/2025/cli-acquisition-tool.md
baronfel: Nightlies are a much harder thing but definitely something I am interested in as the tool grows. We will need to work with the VMR to create some parallels for the data we get from the releases.json manifests, but fundamentally I'd like this tool to eventually support
* stable releases
* nightlies from the VMR
* PRs from VMR (or SDK?)

And depending on how the long-term goal of 'factoring out the SDK into pay-as-you-go chunks' goes, you could potentially see grabbing PRs/nightlies of `specific_component_X` in the far future.

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,224 @@
...
+* on Windows `dnup-init` will request to elevate to set a Windows Registry key that will prevent global .NET installers from setting the global .NET Install location on the System `PATH` (which would always overwrite the user `PATH`).
+  * this behavior _may_ be optional - we may choose to only run it by default if the user has a global .NET install already
+
+This configuration will be necessity be shell-specific, so the tool will need to detect the user's shell and modify the appropriate profile file (e.g. `.bashrc`, `.zshrc`, etc.) to set the `DOTNET_ROOT` and `PATH` variables.
+We have some prior art in the `dotnet` CLI codebase for detecting the shell that is in use, which can be reused here.
+The different shells (and to some extent different OS platforms) have different ways of distributing such changes, so the tool will need to know about those to be a bit opinionated about how it makes these changes.
+
+This shell-specific configuration will need to be independently run via a command, because the user may choose a different shell after the initial setup.
+
+### Install locations
```
vcsjones, line 144: I don't know if this is an explicit goal or not, but something myself and other folks have struggled with is juggling multiple architectures for "dotnet". Windows has a few different architecture flavors for x86, x64, and ARM64. Will, or how, will this tool make it easier to juggle the "I need .NET, no, the other one" problem?

Karuption:
This would be an awesome tool for my team and solve some problems we have with onboarding. The thing that would limit us is that the dev environment is sandboxed. Having a verification tracked through the tool makes sense, but we would need to be able to host the sdks and tool updates ourselves. Npm has a lot of installers as packages and would completely satisfy our needs if both could be NuGet packages. I know that we are the minority, but tools requiring internet access completely takes them off the table for us.

ssg:
I think `dnup` is fine (and in-line with `rustup`), but I can't help myself recommending `dotnetize` as an alternative. :)

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,224 @@
...
+`dnup` needs to be able to update itself as well, via a command similar to `dnup update --self`.
+This command will check documented, signed release manifests for `dnup` and download + replace the current `dnup` binary with the new one.
+Provenance and signature validation of the new binary will also happen to ensure the content is as expected.
+This update capability should be able to be triggered manually or on an automatic cadence (e.g. every execution, once a day, etc.)
+
+## Q & A
+
+### How will I get `dnup`?
+
+We intend to distribute `dnup` via a direct download initially.
```
baronfel, line 171: note: rustup has a shell script for non-windows here - we need to look into _why_: sh.rustup.rs

collabskus:
how will this affect maintainers who package dotnet for various linux distributions (I am thinking about fedora in particular but I am sure other distributions are also affected?) To be clear, I am not with red hat. I am just curious.

agocke:
@Karuption Do you have _intranet_ access? I ask because I thought about this a little bit. https://dnvm.net lets you set the feed URL, so as long as you mirror the overall structure, it should work OK. But that still depends on having a webserver.

Karuption:
@agocke yes, we have NuGet feeds, web servers, and everything you would expect in a dev environment. We just can't access the Internet. We will pull stuff from the Internet into the sandbox with lots of testing and security auditing. 

The portion about verification of the SDKs and the fact that it referenced to get what the .NET team publishes explicitly had me worried.

Getting static files from a webserver would be fine. I would just want to be able to authenticate with creds/token with whatever is decided on. This would also mean that the ssl cert verification for the connection should be configurable, or be able to be disabled. I definitely would want the tool to still do all the integrity checks on the actual file once it's grabbed from the private server.

Karuption:
> @Karuption Do you have _intranet_ access? I ask because I thought about this a little bit. https://dnvm.net lets you set the feed URL, so as long as you mirror the overall structure, it should work OK. But that still depends on having a webserver.

@agocke yes, we have NuGet feeds, web servers, and everything you would expect in a dev environment. We just can't access the Internet. We will pull stuff from the Internet into the sandbox with lots of testing and security auditing.

The portion about verification of the SDKs and the fact that it referenced to get what the .NET team publishes explicitly had me worried.

Getting static files from a webserver would be fine. I would just want to be able to authenticate with creds/token with whatever is decided on. This would also mean that the ssl cert verification for the connection should be configurable, or be able to be disabled. I definitely would want the tool to still do all the integrity checks on the actual file once it's grabbed from the private server.

agocke:
> . I definitely would want the tool to still do all the integrity checks on the actual file once it's grabbed from the private server.

Yeah right now we don't publish signatures for the SDK tarballs, but as soon as we do I plan to add a switch to dnvm to turn of https validation and rely only on the signatures

Antaris:
RE: Keep track of `global.json` version pinning - how would this work exactly? I guess there are some automatic points:

- When VS/VSCode creates a `global.json`, it _registers_ it with `dnup` for tracking?
- When using the `dotnet add globaljson` command it does the same?

We'd also need a manual way of registering a `global.json` with `dnup` to allow it to handle those loose files?

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,224 @@
...
+_Where_ the SDK is resolved from is largely orthogonal to _how compatible_ that SDK is with the IDE.
+
+### What future work does this enable?
+
+- Unifying IDE-specific and disparate install components that exist today
+  - Specifically components like the `dotnet-install` scripts and all downstream consumers of it (`actions/setup-dotnet`, `UseDotNet`)
+  - The dotnet-core-uninstall tool, which is currently a separate tool that is not well integrated with the rest of the .NET ecosystem
+  - The VS Code SDK Install extension, which does a lot of this same logic but is tied to VS Code only
+- Enabling higher-level `dotnet`-CLI-driven workflows
+    - auto-acquire missing Runtimes for a worksapce
```
jnm2, line 207:
```suggestion
    - auto-acquire missing Runtimes for a workspace
```

accepted/2025/cli-acquisition-tool.md
baronfel: We haven't put _too_ much thought into this, but the current best-guess would be since we're already in the business of managing 'hives', we'd track separate hives per-arch. Install/update commands would let you specify which hive (default would be the current arch, flags to update across all, etc). There would be a specific command to 'switch' the active hive, and a way to tell which hive you were currently in.  Similar in spirit to how `nvm use` lets you set up a shell to work on a specific version of node.  Thoughts on this general approach? Happy to push it into the doc.

accepted/2025/cli-acquisition-tool.md
baronfel: Yeah - turns out we'd need the same, if for no other reason than a user would need to chmod the newly-downloaded binary to make it executable.

accepted/2025/cli-acquisition-tool.md
jkotas: > f I need to do development against an IA32-only component on AMD64, what would the developer inner loop experience be like if there was no 32-bit SDK, only runtime?

Add `<RuntimeIdentifier>win-x86</RuntimeIdentifier>` to your .csproj file and your app will be compiled and run as 32-bit with .NET 64-bit SDK. It is very similar to how e.g. mobile development works. It is also how a people often target linux-arm32. Native .NET SDK is incredibly slow on arm32 devices.

If you want the project to be arch neutral and you just want to have a one-off terminal window where it builds and runs as 32-bit, you can set RuntimeIdentifier=win-x86 in that window.

For one-off run as 32-bit, you can use `-a` argument, e.g. `dotnet run -a x86`.

accepted/2025/cli-acquisition-tool.md
baronfel: Yep! Looking at it in more detail this all seems a) doable and b) necessary for a good Linux onboarding experience. The bulk of the 'hard' logic (thinking here things like signature validation, knowledge of side-by-side installations and which parts of the layout are versioned vs unversioned) would still be in the NAOT binary, but the detection of _which_ binary to initially download would need to happen in the script.

nagilson (approved):

MattParkerDev:
I think dotnetup is a better name than dnup

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,223 @@
...
+
+Ricardo wants to check what .NET versions have security vulnerabilities and update them.
+He asks his local GitHub Copilot which installed .NET versions are out of date.
+The agent talks to an acquisition MCP server that lets him know he has two older .NET versions that have known vulnerabilities and asks him if he wants to update.
+Then it calls `dnup update` to update to the latest version for each.
+
+### Onboarding to a repository using .NET
+
+Mei wants to contribute to a .NET library that she found on GitHub.
+She clones the repository, but the repo uses a global.json to enforce a certain tooling version for all contributors be installed to a repo-local directory.
```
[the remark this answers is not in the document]
AraHaan: I feel like ``repo local`` should not be the default and instead default the global.json to local user profile (running as non-admin & non-sudo) and system wide when run as admin/in sudo (that should easily be a few lines of code to implement it lol).

And then in the global.json a field in that json that determines if it should be installed repo locally instead though, but can be overridden by an extra arg of ``--force-local-user`` or ``--force-system-global``.

Benefits to the arg override:
- When building the .NET component repos like dotnet/sdk, dotnet/winforms, dotnet/runtime, and dotnet/aspnetcore one could use these args and if they all share the same version of the .NET SDK required to build in their ``global.json`` the steps for resolving that .NET SDK could be skipped to save a bunch of disk space.
  - The reason for this is if the person is building on an SSD where the more they *use it the faster the SSD goes* and it is well known that while a 1 TB Seagate Baracuda is ~70$ or more, a 2 TB WD Blue is ~120$ or more (for some people these SSD's are expensive as not everyone who builds .NET lives in the US (I know of someone who lives somewhere where their monthly paycheck would convert to ~10 USD).
- Can result in faster local builds which means changes can be made faster to all the components to .NET. Faster changes made locally could mean more PR's per release which would also mean that each new version could potentially have more time to implement features that are nice to have but labeled as ``future`` due to time constraints.
  - Basically it can help relax the time constraints a bit and allow the .NET Team to be able to do more with the time they got as it would help free up critical time to review issues (including feature requests), review PR's, and merge them into .NET.
- Can be used to update the .NET runner images to the version of .NET SDK version shared between all the .NET Component repos which can speed up CI time as ``dnup`` could then be made to see the version installed in those runner images and then set the environment for the runner image to that already installed version on the runner.
  - This in turn would make it possible to see if any mistakes were pushed or even test failures much faster for PR's as well and would allow the person submitting the PR to see and fix the problem faster so then the review can be made much easier for everyone.

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,224 @@
...
+Users of the .NET SDK today face a wide array of choices when getting our tooling, with multiple distribution channels that release at different cadences.
+While these options serve specific needs, several important user cohorts are not well served and have asked for a more unified approach to getting started with our tooling.
+
+
+This proposal introduces a standalone install and SDK management tool (tentatively named `dnup`) that provides a unified interface for acquiring and managing .NET tooling across all supported platforms, focusing on user-local installations that can be managed (largely) without elevation.
+
+This tool is especially critical for users that are
+
+* new users, especially those from the Node or Rust ecosystems
+* Linux users who typically use packages from their distro, as not all distro maintainers support feature bands beyond 1xx
```
MichaelSimons, line 14: source build does not support > 1xx prior to 10 so this isn't possible at all today.  Starting in 10, distro maintainers will have the ability to support which ever feature bands they see a business need for.

MichaelSimons, line 14: One thing I have heard time and time again is that for a subset of enterprise customers, their processes allow for use of packages from the native package feeds.  Anything beyond them requires special approval if at all.  This is to say acquiring components outside the native feeds is significantly harder if not prohibited.  dnup does not change this, unless dnup is written in such a way to acquire from the package feeds.

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,224 @@
...
+- **Self-maintenance**: Keep the bootstrapper tool itself up to date
+
+### Non-Goals
+
+- **Package Manager Integration/Global Installations**: Will not support installers beyond .zip/.tar installs initially
+- **Replace dotnet.exe**: The bootstrapper will not be dotnet.exe itself
+- **NuGet/Tools Support**: Tools are not in scope for `dnup` - instead they would belong to a NAOT `dnx` or NAOT'd `dotnet` CLI
+- **SDK Size Reduction**: Size reduction will be tackled separately, focused on deduplication and factoring out slices of functionality
+
+## Stakeholders and Reviewers
```
MichaelSimons, line 108: Do you consider the Aspire team to be a stakeholder?

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Initial setup
+
+On first-run, `dnup-init` will configure the users' environment to support user-local hives.
+* `DOTNET_ROOT` will be set to point to the `dnup` user-local root,
+* `PATH` will have the `dnup` user-local root appended to it,
+* on Windows `dnup-init` will request to elevate to set a Windows Registry key that will prevent global .NET installers from setting the global .NET Install location on the System `PATH` (which would always overwrite the user `PATH`).
+  * this behavior _may_ be optional - we may choose to only run it by default if the user has a global .NET install already
+
+This configuration will be necessity be shell-specific, so the tool will need to detect the user's shell and modify the appropriate profile file (e.g. `.bashrc`, `.zshrc`, etc.) to set the `DOTNET_ROOT` and `PATH` variables.
```
MichaelSimons, line 138: Confusing wording
```suggestion
This configuration will be necessity be shell-specific, so the tool will need to detect the user's shell and modify the appropriate profile file (e.g. `.bashrc`, `.zshrc`, etc.) to set the `DOTNET_ROOT` and `PATH` variables.
```
```suggestion
This configuration is shell-specific, so the tool will need to detect the user's shell and modify the appropriate profile file (e.g. `.bashrc`, `.zshrc`, etc.) to set the `DOTNET_ROOT` and `PATH` variables.
```

accepted/2025/cli-acquisition-tool.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Install locations
+
+These installs will be located at a default _user-local hive_ instead of a globally-reachable system hive.
+This will allow users to manage new tooling versions without needing to elevate.
+The tool will support being configured to install to other locations via configuration file, so that simple use of the tool remains consistent.
+
+### Install provenance
+
+The data about installs will be sourced from the release manifests that are published on a regular cadence by the .NET Releases team.
```
MichaelSimons, line 152: Is there a need for this to be configurable so that the tool can point to a mirror/private copy of the release manifest for some enterprise scenarions?

accepted/2025/cli-acquisition-tool.md
baronfel: You're right - `dnup` doesn't change this situation. `dnup` is explicitly not targeting package feeds at all, users that need fleetwide or admin-level management of systems would still rely on those mechanisms.

accepted/2025/cli-acquisition-tool.md
baronfel: At the time of this writing they weren't on the radar, but they definitely are now.

baronfel:
Merging this to have a stake in the ground for the feature - will open a subsequent PR with a round of additional content based on some continued reviews/discussions.

PR #341, opened 2025-09-15
Memory safety plan

richlander:
[Rendered markdown](https://github.com/dotnet/designs/blob/memory-safety/accepted/2025/memory-safety/memory-safety.md)

We plan to make memory safety a focus for the next couple releases. This document describes the overall plan and motivation. It references https://github.com/dotnet/designs/pull/330, which is an important part of the plan. Memory safety has always been a critical design tenet of C# and .NET. Now is a good time to revisit memory safety design assumptions and product behaviors. We believe we have an opportunity to significantly improve security posture for all .NET developers with this project.

Our plan is significant, however, it is not a breaking change. We plan to make the proposed behaviors opt-in for existing code. Like nullable, new templates may opt users into the new plan, however, those decisions are still months away and not important at this stage.

This plan is the result of a collaboration across language, libraries, runtime, and security design teams. We will be publishing additional design docs as the project progresses. We plan to make  the dotnet/runtime repo the key focus for the project. Improvements to the core libraries lift all boats and also enable higher-level libraries and apps to be written with more accurate information on safety, per the design in the  caller unsafe spec. We also believe that few other projects have more expansive or stringent requirements as dotnet/runtime.

### DIFF
--- a/accepted/2025/memory-safety/memory-safety.md
+++ b/accepted/2025/memory-safety/memory-safety.md
@@ -19,6 +19,7 @@ We expect that code safety requirements will become more strict over time and th
 
 Minimum requirements:
 
+- Document [Unsafe code best practices](https://learn.microsoft.com/dotnet/standard/unsafe-code/best-practices)
 - Introduce new [C# unsafe annotation and compiler feedback](https://github.com/dotnet/designs/pull/330).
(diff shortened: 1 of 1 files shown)


EgorBo (approved):

richlander:
FYI: https://github.com/dotnet/csharplang/issues/9704

PR #342, opened 2025-09-16
Unsafe Code Guidelines

EgorBo:
A companion doc for https://github.com/dotnet/designs/pull/341 - a set of common unreliable patterns with recommendations.

cc @richlander @jkotas


jkotas:
This should be published as documentation at learning.microsoft.com. Similar to e.g. https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices .

EgorBo:
> This should be published as documentation at learning.microsoft.com. Similar to e.g. https://learn.microsoft.com/en-us/dotnet/standard/native-interop/best-practices .

@jkotas do you have any pointers where to

PR #353, opened 2025-12-02
Clarify cli acquisition milestones and provide ecosystem context

baronfel:
We want to make it clear that the public preview phase is still treating the tool as an experiment - we want to collect feedback and data to turn into a "should we make this a fully-supported product" decision through the public preview period.

Then, based on that go/no-go decision with management, we have SDL and fit-and-finish requirements to go implement.

from marcpopMSFT:
Added some ecosystem context that clarifying how other tools work today and what customers are gravitating towards.


accepted/2025/cli-acquisition-tool.md
```diff
@@ -184,6 +184,36 @@ This command will check documented, signed release manifests for `dnup` and down
...
 This update capability should be able to be triggered manually or on an automatic cadence (e.g. every execution, once a day, etc.)
 
+## Ecosystem Context
+
+### Key Observations from Other Ecosystems
+*	**npm, Rust, Python, Maven/Gradle:** Strong preference for per-user installs; admin installs are rare.
+*	**Go and Java:** Language runtimes often installed centrally, but developers increasingly use version managers or user-scoped installs to avoid admin rights.
+*	**PATH-based discovery** is universal across ecosystems.
+*	**Isolation mechanisms** (virtual environments, local installs) are highly valued for avoiding dependency conflicts.
+
```
jaredpar: Feel like we need to pick one or two of these and dive deeper. Basically what happens with nvm when there is a global install, how do they recommend it get handled and what is the user experience around it? Think knowing that for a few others helps put into context some of the challenges that we have here.

PR #363, opened 2026-03-05
Add proposal for async validation

halter73:
I'm just opening this temporarily for people to leave feedback. I plan to turn this into API Proposals for the runtime and aspnetcore repos. After that, I plan to close this without merging. Sorry if that's a misuse of this repo.

### DIFF
--- a/proposed/async-validation-plan.md
+++ b/proposed/async-validation-plan.md
@@ -185,7 +185,7 @@ namespace System.ComponentModel.DataAnnotations;
 public interface IAsyncValidatableObject : IValidatableObject
 {
     IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext validationContext) =>
-        throw new NotSupportedException(
+        throw new NotSupportedException("This object supports only asynchronous validation. Use the async validation APIs.");
 
     IAsyncEnumerable<ValidationResult> ValidateAsync(
         ValidationContext validationContext,
--- a/proposed/async-validation-plan.md
+++ b/proposed/async-validation-plan.md
@@ -1,0 +1,397 @@
+# Async Validation Support for System.ComponentModel.DataAnnotations
+
+## Problem Statement
+
+The entire `ValidationAttribute` → `GetValidationResult` → `IsValid` chain is synchronous. The ASP.NET Core validation framework (`IValidatableInfo.ValidateAsync`) is async at the orchestration level (recursive traversal of object graphs), but calls `attribute.GetValidationResult()` synchronously at each step. There's no way to write a `ValidationAttribute` that performs async work (e.g., database checks, external service calls).
+
+## Current Architecture
+
+### System.ComponentModel.Annotations (runtime)
+- **`ValidationAttribute`** — abstract base:
+  - `virtual bool IsValid(object?)` — deprecated, backwards compat
+  - `protected virtual ValidationResult? IsValid(object?, ValidationContext)` — recommended override point
+  - `GetValidationResult(object?, ValidationContext)` — public entry point (calls `IsValid`)
+  - `Validate(object?, ValidationContext)` — throws on failure
+  - Uses `_hasBaseIsValid` volatile bool for mutual-recursion override detection
+- **`Validator`** — static class, all sync:
+  - `TryValidateObject`, `TryValidateProperty`, `TryValidateValue` → return `bool`
+  - `ValidateObject`, `ValidateProperty`, `ValidateValue` → throw on failure
+  - All call `attribute.GetValidationResult()` internally
+- **`IValidatableObject`** — sync interface: `IEnumerable<ValidationResult> Validate(ValidationContext)`
+
+### ASP.NET Core Validation — Minimal APIs (`src/aspnetcore/src/Validation/`)
+- **`IValidatableInfo`** — `Task ValidateAsync(object?, ValidateContext, CancellationToken)`
+- **`ValidatablePropertyInfo.ValidateAsync`** — calls `attribute.GetValidationResult()` synchronously in a local `ValidateValue()` helper (lines 153-173)
+- **`ValidatableTypeInfo.ValidateAsync`** — calls `attribute.GetValidationResult()` synchronously in `ValidateTypeAttributes()` (lines 125-150), also calls `IValidatableObject.Validate()` synchronously
+- **`ValidatableParameterInfo.ValidateAsync`** — calls `attribute.GetValidationResult()` synchronously (lines 86-103)
+- **Source Generator** — emits `GeneratedValidatablePropertyInfo`/`GeneratedValidatableTypeInfo` subclasses; delegates `GetValidationAttributes()` to a reflection-based `ValidationAttributeCache`
+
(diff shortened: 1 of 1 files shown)


proposed/async-validation-plan.md
```diff
@@ -0,0 +1,397 @@
...
+
+// Public async entry point (counterpart to GetValidationResult)
+public ValueTask<ValidationResult?> GetValidationResultAsync(
+    object? value,
+    ValidationContext validationContext,
+    CancellationToken cancellationToken)
+{
+    // Calls IsValidAsync, ensures error message is populated (mirrors GetValidationResult logic)
+}
+```
```
halter73:
```suggestion
```

I thought I had already completely removed this, but I missed some.

proposed/async-validation-plan.md
```diff
@@ -0,0 +1,383 @@
...
+Sync methods remain unchanged — they will throw if an `AsyncValidationAttribute` is encountered (via the `IsValid` override).
+
+#### 1.4 `IAsyncValidatableObject.cs` — New Interface
+```csharp
+namespace System.ComponentModel.DataAnnotations;
+
+public interface IAsyncValidatableObject : IValidatableObject
+{
+    IEnumerable<ValidationResult> IValidatableObject.Validate(ValidationContext validationContext) =>
+        throw new NotSupportedException(
```
halter73:
```suggestion
        throw new NotSupportedException("This object supports only asynchronous validation. Use the async validation APIs.");
```

proposed/async-validation-plan.md
```diff
@@ -0,0 +1,383 @@
...
+- Aligns with modern BCL async patterns
+
+### Sync Path Behavior
+1. **Traditional `ValidationAttribute`**: sync path works normally (unchanged)
+2. **`AsyncValidationAttribute`**: sync path throws `NotSupportedException` with clear message pointing to async API (via `IsValid` override)
+3. **Never** silently skip async validation
+4. **Never** do sync-over-async
+
+### `IAsyncValidatableObject` Returns `IAsyncEnumerable`
+Allows streaming results and is aligned with the pattern where validation may involve multiple async checks. If deemed too complex for V1, `Task<IEnumerable<ValidationResult>>` is an acceptable simplification.
```
javiercn, line 332: It's not directly related to this, but I think the way async validation will execute deserves more discussion. This has a big impact on the UX for UI frameworks, and can have very unpredictable outcomes in API contexts too.

Is the expectation that ValidateAsync will walk through the validations as it does today and `await` each async validation before triggering the next one?

proposed/async-validation-plan.md
```diff
@@ -0,0 +1,383 @@
...
+10. `ValidatableObjectAdapter.cs:42` — MVC IValidatableObject validation
+11. `EditContextDataAnnotationsExtensions.cs:91,121` — Blazor form validation
+
+---
+
+## Chosen Approach: Option C — `AsyncValidationAttribute` Deriving from `ValidationAttribute`
+
+Per review feedback, the chosen approach is a new `AsyncValidationAttribute` class that **derives from** `ValidationAttribute`. This provides:
+- Clear separation: async-only attributes are a distinct type
+- Clean failure: the sync `IsValid` override throws `NotSupportedException` in sync paths
```
Youssef1313, line 69: This will be a sealed override, I think?

Edit: probably not. We want to allow users to provide sync path when needed.

proposed/async-validation-plan.md
```diff
@@ -0,0 +1,383 @@
...
+- No silent skipping: since it derives from `ValidationAttribute`, sync `Validator` still discovers it
+- No sync-over-async: sync callers get a clear error telling them to use async APIs
+  - TODO: Add opt-in to sync-over-async (via ValidationContext and/or AppContext switch?)
```
Youssef1313, line 72: This can easily lead to issues like threadpool starvation right? Do we have a scenario/motivation to do this?

proposed/async-validation-plan.md
```diff
@@ -0,0 +1,383 @@
...
+- TODO: IAsyncEnumerable API to get validation results ASAP
+
+### Core Design
+
+```csharp
+namespace System.ComponentModel.DataAnnotations;
+
+public abstract class AsyncValidationAttribute : ValidationAttribute
+{
+    protected override ValidationResult? IsValid(object? value, ValidationContext validationContext)
```
Youssef1313, line 82: Similar to a previous comment of mine, is this expected to be a sealed override?

Edit: no, we want to allow users to provide sync path when needed.

proposed/async-validation-plan.md
```diff
@@ -0,0 +1,383 @@
...
+    {
+        throw new NotSupportedException(
```
Youssef1313, line 84: https://learn.microsoft.com/en-us/dotnet/api/system.componentmodel.dataannotations.validationattribute.isvalid?view=net-10.0#system-componentmodel-dataannotations-validationattribute-isvalid(system-object-system-componentmodel-dataannotations-validationcontext)

If I would read the API documentation (before async validation is shipped), I would assume that IsValid can only throw NotImplementedException and InvalidOperationException, and exactly for the reasons documented (and to be clear, I'm not following with the explanation of the InvalidOperationException in this case). So, isn't this technically unexpected break if existing code encounters the new attribute? i.e, consider an already-built library that scans some user code for attributes and the user tries to use AsyncValidationAttribute.

proposed/async-validation-plan.md
```diff
@@ -0,0 +1,383 @@
...
+| `ValidationAttribute` subclass overriding both `IsValid` AND `IsValidAsync` | ✅ Uses `IsValid` | ✅ Uses `IsValidAsync` |
+
+---
+
+## Other Options Considered
+
+### Option A: Virtual `IsValidAsync` Only on `ValidationAttribute` (No Separate Subclass)
+Same as above but without the `AsyncValidationAttribute` intermediate class. Users override `IsValidAsync` directly on `ValidationAttribute` and the sync `IsValid` base uses reflection to detect async-only attributes.
+
+**Not chosen because:** Reflection-based override detection is fragile. Option C's override is cleaner and more explicit.
```
Youssef1313, line 112: I'm not completely following with the downsides of this approach. I would personally hope to keep everything under `ValidationAttribute` (obviously if possible and we think it's good). So trying to understand more the problems with this approach.

Also for the case of calling the sync `IsValid` on an attribute that only can provide async implementation, I wonder if we should have a special `ValidationResult` for it, instead of throwing an exception. This suggestion/question applies to both Option A and Option C.

proposed/async-validation-plan.md
ploeh, line 69: Throwing `NotSupportedException` is a code smell that the design violates the Liskov Substitution Principle. This should strongly indicate consideration of an alternative design.

If the class isn't going to honour the behaviour of the base class, then why even derive from it?

ploeh (commented):
This is not a hill I'm going to die on, but as I've already outlined in https://github.com/dotnet/runtime/issues/121536#issuecomment-4211654031, I think this is a bad idea in general.

PR #366, opened 2026-04-07
Add in-proc debugger load proposal

AaronRobinsonMSFT:
## Summary
- add the new in-proc-debugger-load.md proposal under proposed
- index the proposal in INDEX.md
- capture the host-controlled in-proc debugger load hook design for non-desktop hosts

### DIFF
--- a/proposed/in-proc-debugger-load.md
+++ b/proposed/in-proc-debugger-load.md
@@ -1,0 +1,158 @@
+# In-proc debugger load hook
+
+**Owner** [Aaron Robinson](https://github.com/AaronRobinsonMSFT) | [Steve Pfister](https://github.com/steveisok)
+
+Mobile and other non-desktop .NET hosts need a supported way to load native debugger support after runtime initialization but before the first managed instruction executes. This proposal introduces a narrow, host-controlled in-proc debugger load hook for that purpose.
+
+The hook is intended for Android and iOS debugging scenarios and similar host-owned diagnostics integrations. It is deliberately not a general-purpose native extensibility point.
+
+## Scenarios and User Experience
+
+### Android and iOS debugging
+
+Tooling that launches a mobile debug session needs a supported way to load a native debugger support library after runtime initialization but before the first managed code executes.
+
+With the proposed design, tooling sets a debugger-specific, host-recognized setting. The host resolves the specified library and loads it during startup so the native component can perform its debugger-specific initialization before managed entrypoint execution.
+
+```text
+DOTNET_ANDROID_INPROC_DEBUGGER=libdebugger.so
+```
+
+Conceptually, the host flow becomes:
+
+```c
+coreclr_initialize(...);
+dlopen(resolve_debugger_hook(getenv("DOTNET_ANDROID_INPROC_DEBUGGER")), RTLD_NOW);
+coreclr_execute_assembly(...);
+```
+
+The exact environment variable name is still under discussion. This document uses `DOTNET_INPROC_DEBUGGER_HOOK` as a conceptual name for the feature and `DOTNET_ANDROID_INPROC_DEBUGGER` / `DOTNET_IOS_INPROC_DEBUGGER` as examples of host-specific variants.
+
+### Standard app launch
+
+If no debugger load hook is configured, application startup is unchanged. There is no additional runtime cost, API surface, or user-visible behavior.
+
+### WASM and other non-desktop targets
+
+Hosts that need early debugger loading can opt into this mechanism. Hosts that do not implement it remain unchanged. The goal is to provide a clear debugger hook where needed, not to require a uniform implementation across every runtime and host.
+
+## Requirements
+
+### Goals
+
+- Provide a supported, host-controlled mechanism for loading a native debugger or diagnostics component during startup.
+- Preserve existing Android and iOS debugger bootstrapping scenarios.
+- Keep the replacement mechanism narrowly scoped to in-process debugger and diagnostics scenarios.
+- Allow hosts to define platform-appropriate path semantics, such as bundle-relative resolution on iOS.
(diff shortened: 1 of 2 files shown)


rolfbjarne (approved):

proposed/in-proc-debugger-load.md
```diff
@@ -0,0 +1,158 @@
...
+
+The hook is intended to support debugger and diagnostics components that must be present before the app's managed entrypoint runs. Typical uses include:
+
+- loading the native side of an Android or iOS in-proc debugger,
+- performing debugger startup work that has to happen before managed code begins, and
+- keeping launch-time debugger behavior under host and tooling control rather than exposing a new runtime-wide extension point.
+
+### Naming and scope
+
+Several naming shapes were discussed:
```
jkotas: Do we want to agree on the name we are going with?

(I think I like `DOTNET_INPROC_DEBUGGER` the best.)

AaronRobinsonMSFT: > `DOTNET_INPROC_DEBUGGER`

+1

AaronRobinsonMSFT: Any other feedback? @elinor-fung @noahfalk ?

proposed/in-proc-debugger-load.md
```diff
@@ -0,0 +1,158 @@
+# In-proc debugger load hook
+
+**Owner** [Aaron Robinson](https://github.com/AaronRobinsonMSFT) | [Steve Pfister](https://github.com/steveisok)
+
+Mobile and other non-desktop .NET hosts need a supported way to load native debugger support after runtime initialization but before the first managed instruction executes. This proposal introduces a narrow, host-controlled in-proc debugger load hook for that purpose.
```
jkotas: This should describe the actual motivation: It is not possible to create a separate process to run the debugger (mscordbi/dac + the external debugger) on mobile. We are solving that for .NET 11 by loading and running the debugger inproc.

proposed/in-proc-debugger-load.md
```diff
@@ -0,0 +1,158 @@
...
+
+- `dotnet/runtime` maintainers for mobile, browser, and other non-desktop hosts
+- Debugger and diagnostics tooling teams that rely on early native component loading
+- IDE and launch tooling teams that need predictable, supported mobile debugging startup behavior
+
+## Design
+
+### Background
+
+Some current mobile debugging flows reuse profiler-related startup plumbing as a practical way to load a native binary between `coreclr_initialize` and `coreclr_execute_assembly`. That was a convenient implementation detail, not the intended contract. The real requirement is a supported way for participating hosts to load an in-proc debugger component during startup.
```
noahfalk: FWIW I don't know that the ordering with respect to coreclr_initialize matters, as long as it happens before coreclr_execute_assemly. I think of this as the replacement for a scenario where we would have launched a debugger process before the .NET process had started. Now that we are in-proc it can't be quite so early, but as long as it can run before managed code starts executing that should ensure breakpoints early in the code aren't missed.

proposed/in-proc-debugger-load.md
noahfalk: That name sounds good to me.

proposed/in-proc-debugger-load.md
```diff
@@ -0,0 +1,158 @@
...
+1. Introduce the host-controlled debugger hook on the mobile hosts that need it.
+2. Update debugger tooling to use the new mechanism.
+3. Retire the older profiler-based bootstrapping path once the supported debugging scenarios no longer depend on it.
+
+Temporary compatibility behavior during the transition can be host-specific, but the long-term direction should be the debugger hook described here rather than indefinite dual support.
+
+### Open issues
+
+- Final environment variable name and whether it should be shared or platform-specific.
+- Whether the initial design should support exactly one library or a delimited list.
```
noahfalk: As long as we can make it work with exactly one library I'd prefer that for simplicity.

proposed/in-proc-debugger-load.md
```diff
@@ -0,0 +1,158 @@
...
+- Whether Android should align with iOS on bundle-relative or host-relative resolution semantics.
+- How much of this behavior should be documented publicly versus maintained as host-specific implementation detail.
```
noahfalk: I think we need to treat the feature as public. However given the expected very limited set of users I don't think we need to make large efforts on documentation. For example as long as we are keeping this doc up-to-date with what we implement then maybe this is the documentation + a note in public learn.microsoft.com docs that points here for the particular environment variable.

noahfalk (approved):

proposed/in-proc-debugger-load.md
```diff
@@ -0,0 +1,158 @@
...
+- create a new extensibility surface that we are not prepared to support broadly, and
+- solve a much larger problem than the one at hand.
+
+If future evidence shows that multiple runtimes or hosts need the same capability, the design can be revisited. The current need is narrower and is better addressed in the host.
+
+### Security and supportability
+
+This proposal does not meaningfully change the trust model because environment variables are already treated as trusted process input. The key supportability concern is not security but accidentally creating a new public extension surface. For that reason, the hook should be explicitly described as intended for debugger or diagnostics tooling on participating hosts only.
+
+### Transition plan
```
jkotas: This is internal VS housekeeping, not particularly interesting for a persisted design doc on github.

jkotas:
@thaystg has been exploring alternative designs using the diagnostic server.

AaronRobinsonMSFT:
> @thaystg has been exploring alternative designs using the diagnostic server.

I assume we've moved passed the approach described here? @thaystg?

AaronRobinsonMSFT:
I think we've mvoed past this approach. Closing.

PR #369, opened 2026-05-28
Document thread safety issues and provide example

agocke:
Added a note on memory safety violations due to data races and provided an example of a NativeBuffer class that illustrates potential issues with concurrent access.

### DIFF
--- a/accepted/2025/memory-safety/caller-unsafe.md
+++ b/accepted/2025/memory-safety/caller-unsafe.md
@@ -156,6 +156,50 @@ According to the above definition, there are two valid constructors: the explici
 
 These data races are considered fundamental to the .NET type system. The safe subset of C# does not protect against them.
 
+Note that memory safety violations that occur _due to data races_ would be considered invalid and any code that can produce this result must be marked unsafe. Consider the following example:
+
+```C#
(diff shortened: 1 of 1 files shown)


accepted/2025/memory-safety/caller-unsafe.md
```diff
@@ -158,6 +156,50 @@ According to the above definition, there are two valid constructors: the explici
...
+        {
+            // SAFETY: _ptr is null or was returned by NativeMemory.Alloc; Free accepts both
+            NativeMemory.Free(_ptr);
+            _ptr = null;
+        }
+    }
+}
+```
+
+The above code can produce memory safety violations in the presence of data races. Consider two threads, one that invokes `Dispose` and the other that invokes `ReadAt` simultaneously. If the first thread executes right between the `_ptr[index]` derefence, the dereference will be invalid (out of bounds/use after free). Both the `Dispose` method and `ReadAt` method would need to be marked `unsafe` to correctly communicate the requirement that they must not be executed concurrently, or the implementation would have to be rewritten to use synchronization.
```
jkotas, line 201: Marking Dispose method as unsafe is not an option since it implements safe interface method.

PR #370, opened 2026-06-17
Add design for dotnetclitool v3

richlander:
This is a draft design for dotnetclitool v3, a successor to https://github.com/dotnet/designs/pull/333.

The basic challenge is that v2 is based on a runtime-dependency model. v2 is a major step forward, however, it is an evolution of v1. That's why the runtime-dependency is so strong. The intent of this proposal to start with NAOT (and self-contained friends) as the primary form factor and then make runtime-dependence the fallback. It also comes with the realization that v2 is very similar to what the container community did with the manifest list format. It does not however do what the community did with the OCI format to make it a simple format that welcoming to multiple readers and writers. v3 offer analogous improvements.

Containers are basically manifests and `.tar.gz` files. NuGet packages are manifests and zip-compressed `.nupkg` files. Very similar. This design takes advantage of NuGet where it provides value and removes it where it doesn't. There is an element of similarity to that process to the OCI one as well.

Design for:

- https://github.com/dotnet/sdk/issues/54333
- https://github.com/dotnet/sdk/issues/52609


accepted/2026/dotnet-cli-tools-v3.md
```diff
@@ -0,0 +1,383 @@
...
+
+The fundamental problem is *where compatibility is stated*. With v2, the file system is used to define the compatibility statement. With v3, a manifest file is used, but critically without the deep directory hierarchy. This means that the pointer package no longer acts as a compatibility gate, at least not to read the tools manifest.
+
+A new DotNetCli spec version enables us to make substantially different choices:
+
+- Use JSON as the format.
+- Rename the file as `manifest.json`.
+- Store the manifest at the top of the tools directory, at `tools/manifest.json`.
+
+Note: There is a desire to make it easier to write standalone tools to install .NET tools packages. It has always been unfortunate that tools installation required the SDK. With BYOR tools, that situation becomes untenable. A recent effort with [dotnet-install](https://github.com/richlander/dotnet-install) went down just this path. Tools like that are part of the reason to move to JSON. Many tools already have a System.Text.Json dependency and JSON serialization works with NAOT while XML serialization doesn't.
```
jkotas, line 106: This is a weak argument. Reading Xml with a simple schema it - like the one used for package - is trivial to handroll in fully AOT compatible way, without any Xml serializer.

xml vs. json is a cosmetic implementation detail. It should not be part of the leading design point. It should be in remarks towards at the end.

richlander: You are right. JSON vs XML is not a strong argument. My main point about format is that tools are more likely to have a JSON dependency already, however, that is speculative. The key point is that it will be easier to right standalone tools because the format is simpler.

I'll clean it up.

PR #371, opened 2026-06-29
Add proposal for apps resolving a deployed shared set of dependencies

elinor-fung:
cc @agocke @baronfel @dsplaisted @ericstj @marcpopMSFT @richlander

### DIFF
--- a/proposed/shared-dependencies-across-applications.md
+++ b/proposed/shared-dependencies-across-applications.md
@@ -42,7 +42,7 @@ The host reads a `sharedDependencies` array in `runtimeconfig.json`. Each entry 
 
 ### Developer experience
 
-An app can reference a shared set so it compiles against the libraries without copying them into its output, and the build records the deployed `.deps.json` in the app's `sharedDependencies`. The shared set can be a project that aggregates the libraries to be shared.
+An app references a shared set so it compiles against the libraries without deploying them, and the build records the deployed `.deps.json` in the app's `sharedDependencies`. The shared set can be a project or package that aggregates the libraries to be shared.
 
 `App.csproj` (app):
 
@@ -59,10 +59,15 @@ An app can reference a shared set so it compiles against the libraries without c
                       SharedDependencyLocation="/opt/AppSuite.Shared/" />
   </ItemGroup>
 
-  <!-- SharedDependency item (emits the entry) + ProjectReference with Private=false (no copy-local) -->
+  <!-- SharedDependency item (emits the entry)
+       ProjectReference with Private=false (no copy-local)
+       or PackageReference with HostProvided=true (compile-only asset) -->
   <ItemGroup>
     <ProjectReference Include="../AppSuite.Utilities/AppSuite.Utilities.csproj" Private="false" />
     <SharedDependency Include="/opt/AppSuite.Utilities/AppSuite.Utilities.deps.json" />
+
+    <PackageReference Include="AppSuite.Extras" Version="1.0.0" HostProvided="true" />
+    <SharedDependency Include="/opt/AppSuite.Extras/AppSuite.Extras.deps.json" />
   </ItemGroup>
 </Project>
 ```
@@ -86,9 +91,12 @@ An app can reference a shared set so it compiles against the libraries without c
 - `SharedDependency` item points at a deployed `.deps.json` and emits the entry.
 - Publishing the shared set produces a layout with the dependency assemblies and a `deps.json`. That layout is deployed to the shared location and the `deps.json` is pointed at by `sharedDependencies`.
 
+This relies on existing or future functionality for marking a reference as external — compiled against but not deployed with the app. `Private="false"` (copy-local off) on a project reference exists today. `HostProvided="true"` on a package reference would come from a future NuGet feature for compile-only assets and audit ownership transfer to whoever deploys the set. That NuGet design is still in progress and not finalized. Support for both need not land together — project references alone are already valuable.
+
 ## Related
 
 - [dotnet/runtime#53834](https://github.com/dotnet/runtime/issues/53834) — Support deploying multiple exes as a single self-contained set
+- [dotnet/runtime#71282](https://github.com/dotnet/runtime/issues/71282) — Locally shared deployment
 - [`--additional-deps`/`DOTNET_ADDITIONAL_DEPS`](https://github.com/dotnet/runtime/blob/main/docs/design/features/additional-deps.md) — existing support for additional deps files via command line or environment variable
 - Not directly related, but in the space of configuring how dependencies can be laid out and found
     - [Configure .NET install search behavior](https://learn.microsoft.com/dotnet/core/deploying#configure-net-install-search-behavior) — existing configuration for how the apphost finds the .NET install
--- a/proposed/shared-dependencies-across-applications.md
+++ b/proposed/shared-dependencies-across-applications.md
@@ -36,7 +36,7 @@ The host reads a `sharedDependencies` array in `runtimeconfig.json`. Each entry 
 }
 ```
 
-- Each `sharedDependencies` entry is an absolute path to the shared set's `.deps.json`. We can relax this later if there is a specific need (for example, paths relative to the app or environment variables).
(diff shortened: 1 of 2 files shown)


proposed/shared-dependencies-across-applications.md
```diff
@@ -0,0 +1,95 @@
...
+    ]
+  }
+}
+```
+
+- Each `sharedDependencies` entry is an absolute path to the shared set's `.deps.json`. We can relax this later if there is a specific need (for example, paths relative to the app or environment variables).
+- The host merges each shared set's `deps.json` into the app's dependencies at startup. The shared set is treated as part of the app, so it resolves with the same priority as assets directly in the app's `deps.json` (that is, before frameworks). This is similar to an [additional `deps.json` via `--additional-deps`/`DOTNET_ADDITIONAL_DEPS`](https://github.com/dotnet/runtime/blob/main/docs/design/features/additional-deps.md).
+- The assets in a shared set's `deps.json` are resolved relative to that `deps.json`'s folder (`/opt/AppSuite.Shared/` in the above example). The presence of `sharedDependencies` does not trigger file existence checks at startup. This differs from `--additional-deps`/`DOTNET_ADDITIONAL_DEPS`, which resolves an additional deps file's assets relative to the app directory and enables those file existence checks.
+
+### Developer experience
```
elinor-fung, line 43: I just kind of put something that somewhat made sense to me here. Happy to pull in whatever experience others folks think makes more sense.

marcpopMSFT:
CC @nkolev92 @aortiz-msft as the External feature from NuGet fits with this feature.

proposed/shared-dependencies-across-applications.md
```diff
@@ -0,0 +1,95 @@
...
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net11.0</TargetFramework>
+  </PropertyGroup>
+
+  <!-- SharedDependencyLocation metadata (no copy-local + emits the entry) -->
+  <ItemGroup>
+    <ProjectReference Include="../AppSuite.Shared/AppSuite.Shared.csproj"
+                      SharedDependencyLocation="/opt/AppSuite.Shared/" />
```
marcpopMSFT, line 59: If we had External=True from nuget, would I assume you wouldn't use SharedDependencyLocation on any packagereference?

elinor-fung: That seems like it would fit. What is the mechanism for run time resolution for External=True? Is there a doc around that feature we can link to?

marcpopMSFT: I asked @nkolev92 offline if he could add more details and share any specs they might have with you. We should update this design to include an example using that new feature once we have the details.

proposed/shared-dependencies-across-applications.md
```diff
@@ -0,0 +1,95 @@
...
+      "version": "11.0.0"
+    },
+    "sharedDependencies": [
+      "/opt/AppSuite.Shared/AppSuite.Shared.deps.json"
+    ]
+  }
+}
+```
+
+- Each `sharedDependencies` entry is an absolute path to the shared set's `.deps.json`. We can relax this later if there is a specific need (for example, paths relative to the app or environment variables).
```
marcpopMSFT: I know the customer requesting this said absolute paths were sufficient but I imagine having relative paths would make this much more useable by additional customers. Is there a reason not to support relative paths? Security could potentially be an issue but an absolute path could also point anywhere so not sure it matters.

@AArnott I know VS is looking into a shared assembly cache. Would ya'll be interested in a shared framework implementation for the parts of VS that run .NET code and would the current design work for you? Or are you limited by needing to support framework as well.

elinor-fung: I mostly just went with the smallest requirement as the starting point when writing the doc. I'd agree relative paths seems like it would expand the usability. And be simple to do - I have no issue allowing both absolute and relative (to the app's directory) paths.

akoeplinger: I also agree supporting relative paths would make this much more usable so you don't need to encode the deployment environment in your build.

elinor-fung:
```suggestion
- Each `sharedDependencies` entry is either an absolute path or relative path (based on the app's directory) to the shared set's `.deps.json`. Environment variables are not supported.
```

proposed/shared-dependencies-across-applications.md
```diff
@@ -0,0 +1,103 @@
...
+      "version": "11.0.0"
+    },
+    "sharedDependencies": [
+      "/opt/AppSuite.Shared/AppSuite.Shared.deps.json"
+    ]
+  }
+}
+```
+
+- Each `sharedDependencies` entry is either an absolute path or relative path (based on the app's directory) to the shared set's `.deps.json`. Environment variables are not supported.
```
AaronRobinsonMSFT: I would call out if we do/don't follow file system links.

proposed/shared-dependencies-across-applications.md
```diff
@@ -0,0 +1,103 @@
...
+**Non-goals**
+- Extension of the shared framework concept (global install, name-based framework references, framework versioning) to third parties
+- Machine-wide assembly store cache
+- Deployment of the applications and their shared set(s)
+
+## Proposed approach
+
+### Runtime resolution
+
+The host reads a `sharedDependencies` array in `runtimeconfig.json`. Each entry points at a shared set's `.deps.json`.
```
AaronRobinsonMSFT, line 20: If I recall `runtimeconfig.json` is considered trusted. We should also then transitively call out that all `sharedDependencies` files are considered trusted.

proposed/shared-dependencies-across-applications.md
AaronRobinsonMSFT: @marcpopMSFT Did we ever get these details?

proposed/shared-dependencies-across-applications.md
```diff
@@ -0,0 +1,103 @@
...
+- The shared set can be updated independently of the apps
+
+**Non-goals**
+- Extension of the shared framework concept (global install, name-based framework references, framework versioning) to third parties
+- Machine-wide assembly store cache
+- Deployment of the applications and their shared set(s)
+
+## Proposed approach
+
+### Runtime resolution
```
normj, line 18: Thinking of how we used the `dotnet store` mechanism with AWS Lambda and we didn't know the location of the shared dependencies at publish time. We relied on during deployment setting the `DOTNET_SHARED_STORE` environment variable so the runtime could found it's dependencies. 

Since I don't want to modify the `runtimeconfig.json` file would I use the `DOTNET_ADDITIONAL_DEPS` environment variable. If that is true would there be any issues if the deps file listed in the `sharedDependencies` section does not exist in the runtime environment?

proposed/shared-dependencies-across-applications.md
marcpopMSFT: I'll ping him offline to see if we can get those details though looks like he's oof at the moment.

elinor-fung: I did ping @nkolev92 offline and he pointed me at some (internal, non-finalized) doc. That is what the `HostProvided` syntax I added in the package reference example is based on.

proposed/shared-dependencies-across-applications.md
elinor-fung: `DOTNET_ADDITIONAL_DEPS` wouldn't fit exactly, since assets in those deps are still resolved based on the app directory and existing probing paths (not relative to the deps file specifying them).

In the current proposal, there isn't a replacement for locations that are not relative to the app and are known only at deployment time. My thought would be that in the future we could add support for enviroment variables in the `sharedDependencies` entries (so allowing something like `${MY_SHARED_SET}/AppSuite.Shared.deps.json`)

> would there be any issues if the deps file listed in the sharedDependencies section does not exist in the runtime environment?

I'd intend for this to be an error. Basically, the app has declared it needs something that doesn't exist (similar to if it declared it needs a framework that doesn't exist). If it doesn't actually know a path at build time, it shouldn't specific a shared dependency location.

PR #372, opened 2026-07-22
Add dotnetup "happy path" design proposal

dsplaisted:
Proposes how dotnetup makes its managed hive win for SDK and runtime resolution: a near-term change to `everywhere` mode that prepends the dotnetup dotnet directory before the Program Files entry on the system PATH, and a longer-term user-scoped, source-gated muxer redirect that relocates both SDK and runtime (including `dotnet foo.dll`).

### DIFF
--- a/accepted/2026/dotnetup/dotnetup-system-integration.md
+++ b/accepted/2026/dotnetup/dotnetup-system-integration.md
@@ -1,0 +1,328 @@
+# Making `dotnetup` win: system PATH and muxer integration for the "happy path"
+
+**Status:** Proposed
+**Owners:** (dotnetup / SDK team — fill in)
+**Area:** Acquisition / host / SDK resolution
+**Related:** [`accepted/2026/dotnetup/cli-acquisition-tool.md`](https://github.com/dotnet/designs/blob/main/accepted/2026/dotnetup/cli-acquisition-tool.md) · [dotnet/designs#350 (closed)](https://github.com/dotnet/designs/pull/350) · [install-location-per-architecture](https://github.com/dotnet/designs/blob/main/accepted/2021/install-location-per-architecture.md) · [local-sdk-global-json](https://github.com/dotnet/designs/blob/main/accepted/2025/local-sdk-global-json.md)
+
+## Summary
+
+We want `dotnetup` to be **opinionated** and to **win**: once a user adopts `dotnetup`, the
+CLI, build tooling, background agents (e.g. GitHub Copilot), and C# Dev Kit should all resolve
+the **same** SDK — the latest, from the `dotnetup`-managed hive — so a machine has one
+consistent .NET that every build uses. This is the "happy path": tooling, CLI, and agents all
+agree on one SDK, so a developer gets predictable behavior no matter how a build is launched,
+instead of different tools silently picking different SDKs.
+
+The hard part is *making* `dotnetup` win. `dotnetup` installs .NET into a user-local hive so
+users can manage SDKs and runtimes without elevation — but on Windows the .NET that actually
+*wins* is usually the machine-wide install under `C:\Program Files\dotnet`, because its
+`dotnet.exe` is on the **system** `PATH` and the muxer resolves SDKs relative to its own
+location. So after a user runs `dotnetup`, their CLI, C# Dev Kit, GitHub Copilot, and other
+tools may still pick a *different* SDK than the one `dotnetup` manages. This document is about
+closing that gap: the system `PATH` and muxer/host integration that lets the `dotnetup` hive
+win.
+
+It proposes a two-phase plan to make the `dotnetup`-managed hive reliably win:
+
+- **Near term (.NET 10 timeframe, no runtime changes):** when a machine-wide install is
+  present, insert the `dotnetup` dotnet directory into the **system** `PATH` *immediately
+  before* the `Program Files\dotnet` entry. This is a
+  one-time elevated change that survives later .NET updates because .NET's MSIs *append*
+  to `PATH`.
+- **Long term (.NET 11):** introduce a single new redirect mechanism in the machine-wide
+  muxer/host that relocates **both SDK and runtime resolution** — including `dotnet foo.dll` —
+  to a per-user `dotnetup` hive, driven by a **user-scoped** environment variable. Because the
+  variable is user-scoped, no elevation is required, each user opts in independently, a later
+  machine-wide SDK install cannot silently take ownership back, and there is no system `PATH`
+  manipulation.
+
+## Motivation
+
+### What "winning" needs to mean
+
+The end-user goal (the "happy path"): after adopting `dotnetup`, the SDK it manages is the
+one *everything* uses — `dotnetup` should be opinionated and "update paths everywhere to win."
+
+Why it matters:
+
+- **`dotnetup` only delivers its purpose if the installs it manages actually win.**
+  `dotnetup` lets users acquire, update, and manage .NET at the user level, without elevation.
+  But if a user installs or updates an SDK with `dotnetup` and their shell, IDE, and agents
+  keep resolving a *different* machine-wide .NET, nothing they run uses it. Making the managed
+  hive win is what turns `dotnetup` from a parallel, ignored install into the .NET the machine
+  actually uses.
+- **Predictability.** Different tools silently resolving different SDKs is confusing and hard
+  to diagnose; one SDK across CLI, tooling, and agents means a build behaves the same however
+  it is launched.
+- **Consistency of caches and results.** Up-to-date checks and build caches (e.g. LSCaches)
+  stay consistent when every build on the machine uses the same SDK.
+- **Uniform delivery of SDK improvements.** Value we are moving into the SDK reaches users
+  uniformly. One concrete driver is the decomposition of C# Dev Kit and pushing its
+  capabilities into the SDK (the up-to-date check / FUTDC, the single project system,
+  LSCaches), which light up consistently only when tooling, CLI, and agents all resolve the
+  same SDK — but the underlying consistency benefit is general.
+
+Success looks like:
+
+- New command-line sessions resolve the `dotnetup` dotnet.
+- New UI apps resolve it too — VS Code / C# Dev Kit, the GitHub Copilot app, and ideally
+  any app that shells out to `dotnet`.
+- It works when **no** other .NET is installed.
+- It works when a machine-wide ("system") .NET *is* installed.
+- It works even when that machine-wide .NET is **older** (.NET 6/7/8).
+
+## Goals
+
+- Make the `dotnetup`-managed hive the SDK that wins for CLI and tools, reliably, after a
+  simple `dotnetup` action.
+- Survive later machine-wide .NET updates (VS / Windows Update) without the user having to
+  re-run a fix.
+- Avoid a design that requires ongoing elevation.
+- Keep loose-zip and hand-extracted .NET usage predictable.
+- Make the `dotnetup`-managed hive win for **both** SDK/tooling and runtime resolution
+  (including `dotnet app.dll`), not just the SDK.
+
+## Non-goals
+
+- Changing how `dotnetup` acquires or lays down SDKs/runtimes
+- Forcing `global.json` adoption on users as the mechanism for winning (it helps some IDE
+  cases but is not the general answer; see [Relationship to existing mechanisms](#relationship-to-existing-mechanisms)).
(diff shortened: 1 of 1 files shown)


accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,328 @@
...
+- Be **durable**: a later machine-wide .NET install can't silently reclaim ownership.
+- Keep **loose/hand-extracted** .NET predictable: running a `dotnet.exe` you unzipped yourself
+  is unaffected.
+
+**Proposed approach (subject to change).**
+
+This seems to lead to the following solution:
+
+- The `Program Files\dotnet` muxer (`dotnet.exe`) can be the one resolved via the `PATH`
+- The muxer supports a **user-scoped** environment variable that redirects its resolution to a
```
elinor-fung: Checking the scope wouldn't (can't) be on the consumer/host. The install of `dotnetup` would set it in the desired scope.

accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,328 @@
...
+- New UI apps resolve it too — VS Code / C# Dev Kit, the GitHub Copilot app, and ideally
+  any app that shells out to `dotnet`.
+- It works when **no** other .NET is installed.
+- It works when a machine-wide ("system") .NET *is* installed.
+- It works even when that machine-wide .NET is **older** (.NET 6/7/8).
+
+## Goals
+
+- Make the `dotnetup`-managed hive the SDK that wins for CLI and tools, reliably, after a
+  simple `dotnetup` action.
```
elinor-fung, line 78: We should make sure this is:
- clearly messaged during `dotnetup` install and in any docs
- surfaced (with pointers to resolution if possible) in non-happy-path cases - for example, I have some repo that uses local sdk paths, I try dotnetup at some point, later I go back to my repo and it no longer finds the sdk required
- indicated in general diagnostic info like `dotnet --info`

Basically, the happy-path requires everything being on-boarded to `dotnetup`, so I want to make sure we are intentional about the experience in the non-happy-path / transition.

accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,328 @@
...
+  [#120472](https://github.com/dotnet/runtime/pull/120472) because historical details of how
+  the MSI components were authored could cause the value to be incorrectly removed from the PATH in some install/uninstall scenarios.
+
+These are the constraints this proposal is designed to respect: don't rely on removing the
+machine-wide `PATH` entry, don't break loose-zip users, and relocate the SDK **and** the
+runtime together rather than leaving the runtime behind.
+
+## Relationship to existing mechanisms
+
+- **`global.json` `sdk.paths` / `$host$`**
```
elinor-fung, line 244: Does this mean that the environment variable basically replaces `sdk.paths`, but everything else in the global.json is respected?

accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,328 @@
...
+the **default on Windows**; `none` and `shell` are unchanged. Both phases below describe what
+`everywhere` mode does.
+
+### Phase 1 (near term, .NET 10 timeframe): prepend before the Program Files entry
+
+In `everywhere` mode, when a machine-wide `Program Files\dotnet` install is on the system
+`PATH`, `dotnetup` today *removes* that entry. This proposal changes that to a durable prepend:
+
+1. Elevate once (the same elevation `dotnetup` already uses to edit the machine-wide `PATH`).
+2. **Insert the `dotnetup` dotnet directory into the system `PATH` immediately before the
```
elinor-fung, line 119: Does `dotnetup` do multi-arch today or only the OS architecture?

accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,328 @@
...
+This seems to lead to the following solution:
+
+- The `Program Files\dotnet` muxer (`dotnet.exe`) can be the one resolved via the `PATH`
+- The muxer supports a **user-scoped** environment variable that redirects its resolution to a
+  different dotnet hive.
+
+Concretely, the muxer would support the following environment variables:
+
+- `DOTNET_ROOT_REDIRECT_TARGET` - The path of the hive to redirect to
+- `DOTNET_ROOT_REDIRECT_SOURCES` - A list of sources to redirect from.  This allows the redirect to be restricted to the Program Files muxer, so that other muxers (such as from an unzipped runtime/SDK) are not affected
```
elinor-fung, line 168: Is this list of sources expected to be something that changes much? Would it be anything other than the global install?

General concern around the source path matching:
- I assume this would require some sort of canonicalization given the myriad of ways the same path can be specified.
- different behaviour based on where an install (that is otherwise xcopy-able) is located seems rather confusing.

(I don't have a good solution - just calling it out)

accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,328 @@
...
+  runtime team has objected to launch-path-dependent behavior (jkotas on #350). Settle the
+  exact matching rules — which source location(s) qualify, and how symlinked/copied hosts are
+  treated — and confirm the trade-off is acceptable.
+- **Runtime-load-behavior buy-in.** Relocating `dotnet foo.dll` requires the runtime/host
+  team to honor the redirect for app launches, which #350 avoided changing. Securing that
+  agreement is a prerequisite for Phase 2.
+- **Non-Windows.** The goals are the same on macOS/Linux, but this proposal focuses on the
+  Windows solution (likely the hardest). Phase 1 is Windows-specific and likely doesn't apply;
+  Phase 2's redirect may or may not be the right approach there. The cross-platform design
+  remains to be worked out.
```
elinor-fung: I'd rather consider this for all platforms (even if Windows is the motivating one).

accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,328 @@
...
+  precedent: tools that jam themselves at the front of `PATH` have shadowed same-named OS
+  utilities — e.g. Java breaking Windows Kerberos tools.)
+
+Why this isn't ideal:
+
+The system `PATH` is machine-wide but points at one user's hive. Other users generally can't
+access it, so they fall through to the machine-wide install — but an elevated process (or an account such as SYSTEM)
+may resolve `dotnet` from that hive, so the wrong user's install can win. This "multi-user oddness"
+(accepted for the near term) is why Phase 1 is a bridge; Phase 2 (per-user, no system `PATH`
+change) removes it.
```
elinor-fung, line 142: > removes it

Does that mean a one-time elevation as well?

elinor-fung:
cc @richlander

accepted/2026/dotnetup/dotnetup-system-integration.md
dsplaisted: We would want local repo paths to continue to work.  The `sdk.paths` feature should continue to work (and override or be applied after the environment-variable based redirect).  Putting a specific version of dotnet on the path shouldi also work as long as we include the "REDIRECT_SOURCE" type environment variable.

It would be good for `dotnet --info` to include information about the redirection if it applied.

accepted/2026/dotnetup/dotnetup-system-integration.md
dsplaisted: We'd like `sdk.paths` and everything else in global.json to continue to work.  The behavior we'd like is that everything would work the same as if `C:\Program Files\dotnet` was not on the PATH, and instead the folder specified by the redirect (such as `C:\Users\daplaist\AppData\Local\dotnet`) were on the PATH instead.

accepted/2026/dotnetup/dotnetup-system-integration.md
dsplaisted: We don't do multi-arch today but we should think about how the redirect support would interact with multi-arch.  I'm not sure if it would, you can't have two different architectures of dotnet on the PATH (or you can, but it doesn't matter since only the first one will apply).

accepted/2026/dotnetup/dotnetup-system-integration.md
dsplaisted: We would expect to only set it to the global install.  So theoretically we could have the muxer check its install location to determine whether to apply the redirect without the second environment variable.  But it seems probably better (and more testable) to use an environment variable.

accepted/2026/dotnetup/dotnetup-system-integration.md
dsplaisted: Phase 2 would have probably include a one-time elevation to install the updated muxer, if it wasn't already installed in Program Files.

accepted/2026/dotnetup/dotnetup-system-integration.md
dsplaisted: I updated the wording to clarify that the consumer doesn't need to check the scope.

accepted/2026/dotnetup/dotnetup-system-integration.md
dsplaisted: Updated this to try to make clear that the muxer feature should apply to all platforms.

accepted/2026/dotnetup/dotnetup-system-integration.md
elinor-fung: > as if C:\Program Files\dotnet was not on the PATH, and instead the folder specified by the redirect (such as C:\Users\daplaist\AppData\Local\dotnet) were on the PATH instead.

So as if it was actually `C:\Users\daplaist\AppData\Local\dotnet` that was launched? In my mind that means:
- global.json without `sdk.paths`: only look in the redirect folder, respecting everything in global.json
  - this means that if the global.json specifies an sdk version that is not in the redirect (but is in the global install), the resolution would fail
- global.json with `sdk.paths`: redirect folder replaces `$host$` in `sdk.paths`, if `$host$` is not listed in `sdk.paths` we won't look in the redirect
- no global.json: only look in redirect folder, use latest sdk found

Does that match your thinking?

SDK/runtime resolution can be confusing enough - I want to make sure we're very explicit about how anything we add interacts with existing features, where it sits in the resolution logic, and if/how much it adds to the complexity.

accepted/2026/dotnetup/dotnetup-system-integration.md
elinor-fung: I wonder if something like a marker file next to `dotnet` that indicates enabling redirection and is only included in the installation bundle (and not the archives) would make sense. It would avoid path matching. Different acquisition methods would still have different behaviour (when dotnetup has been in play), but it wouldn't change on xcopy to another location.

accepted/2026/dotnetup/dotnetup-system-integration.md
```diff
@@ -0,0 +1,329 @@
...
+Success looks like:
+
+- New command-line sessions resolve the `dotnetup` dotnet.
+- New UI apps resolve it too — VS Code / C# Dev Kit, the GitHub Copilot app, and ideally
+  any app that shells out to `dotnet`.
+- It works when **no** other .NET is installed.
+- It works when a machine-wide ("system") .NET *is* installed.
+- It works even when that machine-wide .NET is **older** (.NET 6/7/8).
+
+## Goals
```
jkotas, line 75: Number of existing scenarios break once a user opts into this feature in their environment, for example:

- Servicing using https://learn.microsoft.com/en-us/dotnet/core/install/windows#microsoft-update once a user opts into it. The central servicing via Microsoft Update is critical for customers with locked down environments.

- Framework dependent apps that install the runtime via MSI 
https://learn.microsoft.com/en-us/visualstudio/deployment/installer-projects-net-core?view=visualstudio#prerequisites 

Is that right? What are we going to do about that?

JakeRadMSFT: Correct this is a user level override and it takes preference for the user's tooling.

"Framework dependent apps that install the runtime via MSI" - how common is this?

accepted/2026/dotnetup/dotnetup-system-integration.md
JakeRadMSFT: > global.json with sdk.paths: redirect folder replaces $host$ in sdk.paths, if $host$ is not listed in sdk.paths we won't look in the redirect

I think something like that is fine. We should make sure dotnetup (and C#DK) complain if sdk.paths isn't setup properly for dotnetup to work. Similar to needing version to be 10/11 with roll forward latestFeature being needed to be using latest.

PR #373, opened 2026-07-30
Proposal for having a monthly minor releases of .NET SDK starting with net12

marcpopMSFT:
Aligning the SDK with the current tooling and VS release schedule and removing the confusion of feature bands while simplifying our infrastructure.

### DIFF
--- a/proposed/single-sdk-feature-band.md
+++ b/proposed/single-sdk-feature-band.md
@@ -194,6 +194,12 @@ For repositories with public-to-internal mirroring, branch topology must preserv
 
 This model needs validation with engineering systems owners before adoption.
 
+Concerns:
+
+- We would have two versions of the .1xx band shipping on the same day. One would be stable branded and one would be preview branded. That would then create 6 different feature combinations depending on the channel setting as well.
+  - Example: 11.0.103 would ship in VS stable and 11.0.1xx-preview.27053.4 would release in VS canary on Feb patch tuesday.
+  - Note: 11.0.1xx-preview.27053.4 versioning wouldn't work with workloads as is as we expect a preview version.
+
 ### Servicing safety rules
 
 In servicing trains:
--- a/proposed/single-sdk-feature-band.md
+++ b/proposed/single-sdk-feature-band.md
@@ -1,6 +1,17 @@
 # Monthly SDK Minor Releases
 
+Two-band transition in .NET 11 and monthly minor releases beginning in .NET 12
+
 **Owner** [marcpopMSFT](https://github.com/marcpopMSFT)
+
+## Decision Summary
+
+We will make this change in two stages:
+
+- **.NET 11:** Ship exactly two SDK feature bands. `11.0.1xx` is the long-lived servicing band and accepts only security fixes and fixes required to unblock servicing. `11.0.2xx` is the feature-delivery band.
+- **.NET 12:** Replace SDK feature-band versions with monthly minor SDK releases, such as `12.1.0`, `12.2.0`, and `12.3.0`.
+
+We will announce both stages together. During the early .NET 12 previews, we will address the infrastructure and ecosystem gaps required to support minor-version SDK releases.
 
 ## Motivation
 
@@ -22,7 +33,22 @@ Feature bands add another layer of complexity. Versions such as `12.0.100`, `12.
 
 ## Proposal
 
-We propose replacing SDK feature bands with monthly minor SDK releases. For example, the .NET 12 GA SDK would be `12.0.0`, followed by `12.1.0`, `12.2.0`, and so on. Each monthly minor is an explicit feature boundary. The latest minor receives new features, while customers that require maximum stability can remain on the original `12.0.x` line and receive only security and reliability fixes.
+We propose a phased rollout that reduces the number of release trains in .NET 11 while the ecosystem prepares for monthly minor SDK releases in .NET 12.
+
+### Phase 1: .NET 11 two-band transition
+
+.NET 11 ships exactly two SDK feature bands:
+
+| Version | Purpose | Allowed changes |
+| --- | --- | --- |
+| `11.0.1xx` | Long-lived servicing baseline | Security fixes and fixes required to unblock servicing |
+| `11.0.2xx` | Feature delivery | New SDK and tooling features, plus applicable servicing fixes |
+
+There will be no `11.0.3xx` or `11.0.4xx` feature bands. Existing feature-band-aware infrastructure remains in place for .NET 11, avoiding a versioning migration before the required host, workload, installer, acquisition, update, and release infrastructure is ready.
+
+### Phase 2: .NET 12 monthly minor releases
+
+Beginning with .NET 12, SDK feature bands are replaced with monthly minor SDK releases. The .NET 12 GA SDK is `12.0.0`, followed by `12.1.0`, `12.2.0`, and so on. Each monthly minor is an explicit feature boundary. The latest minor receives new features, while the original `12.0.x` line provides a stability-oriented servicing boundary.
 
 Release boundaries provide stronger isolation than a shared feature-flag system. In particular, analyzer, compiler, and API-shape changes can affect customers even when a tool attempts to gate user-visible behavior.
 
@@ -73,6 +99,8 @@ Outcome:
 
 ### Goals
 
+- Ship exactly two SDK feature bands in .NET 11, with distinct servicing and feature-delivery roles.
+- Preserve existing feature-band-aware infrastructure during the .NET 11 transition.
 - Replace quarterly SDK feature bands with monthly minor SDK releases.
 - Make the SDK version an explicit and reliable feature boundary.
 - Deliver stable tooling features monthly.
@@ -228,9 +256,27 @@ The release library and other compliance, inventory, and support tooling may als
 
 ## Transition
 
-Adopting this model at the beginning of a major release is simpler than changing versioning after customers and tooling have consumed feature-band versions. The proposed transition is to retain the existing feature-band model through .NET 11, including a final feature band if needed, and begin monthly minor releases with .NET 12.
+Adopting monthly minor versions at the beginning of a major release is simpler than changing versioning after customers and tooling have consumed feature-band versions. .NET 11 therefore uses a two-band transition model:
 
-The transition plan must identify and update automation that parses SDK feature bands, including workload tooling, `global.json` handling, installers, the .NET uninstall tool, the release library, Microsoft Update, release pipelines, compliance and inventory systems, support tooling, and documentation. Tools that encounter both the old feature-band scheme and the monthly-minor scheme need defined compatibility behavior.
+- `11.0.1xx` remains the rollback and servicing boundary for customers and source-build partners that require maximum stability. It receives only security fixes and fixes required to unblock servicing.
+- `11.0.2xx` receives new SDK and tooling features.
+- No additional .NET 11 feature bands are produced.
+
+During the early .NET 12 previews, we will identify, assign, and close the gaps required to use monthly minor versions. This work includes host and SDK resolution, `global.json`, workload tooling, installers, the .NET uninstall tool, the release library, Microsoft Update, WinGet, DNIM, getdotnet, release pipelines, VMR infrastructure, compliance and inventory systems, support tooling, third-party version parsing, and documentation.
+
+Tools that encounter both the feature-band scheme used through .NET 11 and the monthly-minor scheme beginning in .NET 12 need defined compatibility behavior. Changes that affect the ecosystem should land early enough in the .NET 12 preview cycle to allow validation and correction before stabilization.
+
+## External Communication
+
+We will publish a blog post that explains both stages of the plan:
+
+- Why .NET 11 retains feature-band numbering while reducing the release model to two bands.
+- The servicing-only role of `11.0.1xx` and the feature-delivery role of `11.0.2xx`.
+- Why monthly minor versions are the intended .NET 12 direction.
+- Which infrastructure and ecosystem gaps will be addressed during the .NET 12 previews.
+- What customers, partners, and tool authors should expect for SDK selection, workloads, installation, servicing, and version parsing.
+
+Migration guidance for affected tooling must be available before the first .NET 12 preview that uses the new versioning model.
 
 ## Benefits and Drawbacks
 
@@ -251,16 +297,27 @@ Drawbacks:
 - SDK and runtime versions can differ in ways that require clear diagnostics and documentation.
 - Monthly minors increase the number of SDK versions encountered by customers and support teams.
 
-## Risks and Open Questions
+## Implementation Work and Remaining Questions
+
+The following decisions are settled:
+
+- .NET 11 has exactly two SDK feature bands.
+- `11.0.1xx` receives only security fixes and fixes required to unblock servicing.
+- `11.0.2xx` is the .NET 11 feature-delivery band.
+- Monthly minor SDK versions are the intended .NET 12 direction.
+- Both stages will be communicated publicly.
+
+The .NET 12 preview work must:
 
 - Choose code flow with mirror conflict resolution or dual check-ins for tooling fixes needed by both the SDK and Visual Studio.
 - Define ownership and service expectations for the selected tooling propagation path.
-- Validate the monthly promotion and branding automation.
+- Validate monthly promotion and branding automation.
 - Define security-fix propagation for the stability line, current stable minor, next preview, Visual Studio, and source build.
 - Define workload compatibility and patch-number ordering.
 - Specify `global.json` roll-forward behavior across monthly minors.
-- Define installer, Microsoft Update, and side-by-side installation behavior.
-- Confirm the model and operational responsibilities with tooling teams and VMR owners.
+- Define installer, Microsoft Update, uninstall, acquisition, rollback, and side-by-side installation behavior.
+- Audit first-party and third-party systems that parse or compare SDK versions.
+- Confirm operational responsibilities with tooling teams and VMR owners.
 
 ## Alternatives Considered
 
--- a/proposed/single-sdk-feature-band.md
+++ b/proposed/single-sdk-feature-band.md
@@ -50,6 +50,20 @@ Outcome:
 - Easy to experiment with new features without waiting for broader SDK releases.
 - With each major release, VSCode will update the SDK installed to a new band resulting in a new baseline.
 
+## Alternatives
+
+### Monthly SDK minor releases
+
+Instead of using feature channels within a single SDK feature band, the SDK could adopt monthly minor releases. The .NET 11 GA SDK would be `11.0.0`, followed by `11.1.0`, `11.2.0`, and so on in subsequent months. Each monthly SDK would replace the previous monthly release and contain the latest available runtime, including applicable security and reliability fixes. I considered just having monthly feature bands as an alternative but the feeling was that having 11.0.100 and 11.0.1000 would lead to more confusion.
+
+Only the latest monthly minor and the original `11.0.x` release line would be serviced. The `11.0.x` line would receive security and reliability fixes but no new features, providing a stability-oriented option for source-build and enterprise partners. Existing `global.json` version selection and roll-forward behavior would allow customers to remain on `11.0.x`, select a particular monthly SDK, or move to newer monthly releases.
+
+Workload set versions would generally match the SDK version. For example, the workload set for `11.2.0` would use that version, while SDK hotfixes and mid-month workload updates would consume successive patch numbers such as `11.2.1` and `11.2.2`. A patch number could therefore represent either kind of update, avoiding the need for four-part workload set versions.
+
+This model would likely require three branches: the `11.0` stability branch, stable, and preview. DARC channels would flow monthly releases from preview to stable together with the required branding updates.
+
+The primary benefit of this approach is avoiding the governance and coordination risk of requiring every SDK-loaded component to implement feature channels correctly. Its primary drawbacks are that source-build customers remaining on `11.0.x` could not opt into new features through a feature-channel update, and the SDK would need to manage additional branching, flow, and branding complexity. This basically takes the feature band model and tries to find a reasonable way of collapsing it into monthly releases.
+
 ## Requirements
 
 ### Goals
--- a/proposed/single-sdk-feature-band.md
+++ b/proposed/single-sdk-feature-band.md
@@ -54,15 +54,103 @@ Outcome:
 
 ### Monthly SDK minor releases
 
-Instead of using feature channels within a single SDK feature band, the SDK could adopt monthly minor releases. The .NET 11 GA SDK would be `11.0.0`, followed by `11.1.0`, `11.2.0`, and so on in subsequent months. Each monthly SDK would replace the previous monthly release and contain the latest available runtime, including applicable security and reliability fixes. I considered just having monthly feature bands as an alternative but the feeling was that having 11.0.100 and 11.0.1000 would lead to more confusion.
+Instead of using feature channels within a single SDK feature band, the SDK could adopt monthly minor releases. For example, the .NET 12 GA SDK would be `12.0.0`, followed by `12.1.0`, `12.2.0`, and so on in subsequent months. Each monthly SDK would replace the previous monthly release as the actively supported feature release and contain the latest available runtime, including applicable security and reliability fixes.
 
-Only the latest monthly minor and the original `11.0.x` release line would be serviced. The `11.0.x` line would receive security and reliability fixes but no new features, providing a stability-oriented option for source-build and enterprise partners. Existing `global.json` version selection and roll-forward behavior would allow customers to remain on `11.0.x`, select a particular monthly SDK, or move to newer monthly releases.
+Monthly feature bands were also considered, but versions such as `12.0.100` followed by `12.0.200` or `12.0.1000` continue the feature-band concept and do not communicate the monthly release ordering as directly.
 
-Workload set versions would generally match the SDK version. For example, the workload set for `11.2.0` would use that version, while SDK hotfixes and mid-month workload updates would consume successive patch numbers such as `11.2.1` and `11.2.2`. A patch number could therefore represent either kind of update, avoiding the need for four-part workload set versions.
+#### Release and branch model
 
-This model would likely require three branches: the `11.0` stability branch, stable, and preview. DARC channels would flow monthly releases from preview to stable together with the required branding updates.
+Development would primarily occur in a long-lived preview branch. It could be as simple as a public release/12.x branch with all public codeflow to this branch. Once a month, the preview content intended for release would be promoted to a long-lived stable branch which couild just be internal/release/12.x which has internal codeflow of the runtime version. Each month we'd update the branding in each branch separately.
 
-The primary benefit of this approach is avoiding the governance and coordination risk of requiring every SDK-loaded component to implement feature channels correctly. Its primary drawbacks are that source-build customers remaining on `11.0.x` could not opt into new features through a feature-channel update, and the SDK would need to manage additional branching, flow, and branding complexity. This basically takes the feature band model and tries to find a reasonable way of collapsing it into monthly releases.
+The original `12.0.x` release line would remain available as the stability-oriented release for customers that want security and reliability fixes without new features (specifically, enterprise source build partners). This produces at least three logical release paths:
+
+- The original `12.0.x` stability line.
+- The latest stable monthly minor release.
+- The next monthly minor release in preview.
+
+Monthly promotion introduces a risk of drift between preview and stable. We could have codeflow from stable tooling to the stable branch but that would cause mirror conflicts or alternatively, we could require all stable tooling changes be made in the internal/release branch. Any security fixes would have to be ported on release day to the preview branch.
+
+#### Servicing and security fixes
+
+Only the latest monthly minor and the original `12.0.x` release line would be serviced. Superseded monthly minors would remain installable but would not receive further fixes. The `12.0.x` line would receive security and reliability fixes but no new features, providing a predictable option for enterprise source-build and other stability-focused customers.
+
+The servicing policy needs to distinguish among:
+
+- Runtime security and reliability updates included in an SDK release.
+- SDK or tooling hotfixes that do not require a runtime release.
+- Workload updates released between monthly SDK releases.
+- Fixes required by Visual Studio or source-build consumers that remain on a different SDK release path.
+
+Security and reliability fixes need an explicit propagation path to the original stability line, the current stable monthly release, the next preview, Visual Studio's supported SDK builds, and any dedicated source-build branch. Stable tooling fixes would need to flow through the VMR or an equivalent internal release process rather than flowing only into the preview branch.
+
+The policy must also define what happens when a runtime security release is delayed, skipped, or released out of band. A runtime release should not necessarily require a new SDK minor, but supported SDK lines still need a way to acquire that runtime update.
+
+#### SDK and runtime version relationship
+
+SDK and runtime versions would intentionally be allowed to differ. For example, an SDK hotfix `12.2.1` might contain runtime `12.0.7`, and a new monthly SDK minor preview would contain the most recently available runtime patch rather than a runtime released on the same day. Release notes would need to make this relationship clear.
+
+The SDK version components would have the following meaning:
+
+- Major: the annual .NET release.
+- Minor: a monthly SDK feature release.
+- Patch: an SDK/tooling hotfix or a release used to carry an updated workload set between monthly feature releases.
+- Prerelease label: builds of the next monthly minor before promotion to stable.
+
+Using the patch component for both SDK hotfixes and workload updates preserves a three-part version but means patch numbers identify release order rather than a particular type of change. Release tooling and documentation must not assume that every patch increment has the same cause.
+
+#### Visual Studio integration
+
+The preview SDK would continue to flow to Visual Studio Canary for validation. The stable monthly SDK would be available to supported Visual Studio channels according to their insertion and servicing policies. Visual Studio Insiders could continue to consume the preview flow through Canary rather than requiring a separate SDK build.
+
+
+#### Source-build contract
+
+Enterprise source-build partners that require maximum stability could remain on `12.0.x` and receive security and reliability fixes without taking monthly feature releases. Other source-build consumers could move to newer monthly minors through their normal distribution update process.
+
+Unlike the feature-channel proposal, a customer remaining on `12.0.x` could not opt into selected newer SDK features without moving to a newer monthly SDK. This loses a benefit of the feature-channel model for customers that must acquire the SDK from a distribution feed but want newer features. On the other hand, release boundaries provide stronger isolation than relying on every component to gate all behavior correctly. This is particularly important for changes such as analyzer fixes that may introduce new warnings even when the change is otherwise considered a correctness fix.
+
+#### Workloads, installers, and SDK selection
+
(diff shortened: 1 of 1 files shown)


proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
+# Single SDK Feature Band with Global Feature Channels
+
+**Owner** [marcpopMSFT](https://github.com/marcpopMSFT)
+
+Today, SDK feature bands, SDK tools, and Visual Studio are coupled in ways that is complex to manage and confusing for users.  The challenge with changing this is users can receive new features or behavior changes without intentionally opting in. This is especially problematic for managed environments and source build partners that prioritize stability. We propose a single SDK feature band strategy, combined with a global feature channel setting that all SDK-loaded components honor. This setting has three states: baseline, stable, and preview. New behavior is always disabled in baseline, and initially enabled only in preview.
```
Frulfump: Stable isn't explained in the same way that baseline and preview was.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
...
+
+A developer on Visual Studio stable channel receives SDK updates from the internal branch insertion path. They get behavior considered production-ready for the stable channel, while preview-only changes remain off. They get the latest runtime with all security fixes and any tooling security fixes as well.
+
+Outcome:
+
+- Stable channel receives deliberate feature lighting.
+- Behavior aligns with VS stable expectations.
+- Mid-cycle surprises are reduced.
+
+### Scenario 3: Canary/Insider users validate upcoming behavior
```
Frulfump:
```suggestion
### Scenario 3: Visual Studio Canary/Insider users validate upcoming behavior
```
This seems like it's just for VS

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
+# Single SDK Feature Band with Global Feature Channels
+
+**Owner** [marcpopMSFT](https://github.com/marcpopMSFT)
+
+Today, SDK feature bands, SDK tools, and Visual Studio are coupled in ways that is complex to manage and confusing for users.  The challenge with changing this is users can receive new features or behavior changes without intentionally opting in. This is especially problematic for managed environments and source build partners that prioritize stability. We propose a single SDK feature band strategy, combined with a global feature channel setting that all SDK-loaded components honor. This setting has three states: baseline, stable, and preview. New behavior is always disabled in baseline, and initially enabled only in preview.
+
+## Scenarios and User Experience
```
Frulfump, line 7: I'd like to see scenarios described for stable, baseline and preview for SDK users not using VS. Including if there's a difference for the different acquisition methods.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
...
+- baseline: off
+- stable: off (unless explicitly promoted)
+- preview: on (unless explicitly held back)
+
+Promotion policy:
+
+- A feature starts as preview-on, stable-off, baseline-off.
+- Promotion to stable requires explicit approval and release notes in a planned update.
+- Baseline remains off by default, with opt-in available for customers that explicitly choose it.
+- Minimum preview soak time before promotion is two months.
```
Frulfump, line 124: Two months seems a bit long for low-risk changes. UPDATE: I see this was mentioned as well in the Risks and Unknown section

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
...
+Customers can override with the environment variable if they choose a different risk posture.
+
+Precedence (highest to lowest):
+
+1. Command-line explicit override (if introduced)
+2. Environment variable explicit value
+3. Host-computed default written by SDK startup
+4. SDK build-time default
+
+Global.json integration may be added later, but is not required for the initial design.
```
Frulfump:
```suggestion
global.json integration may be added later, but is not required for the initial design.
```

I understand why you maybe don't want to add this initially but this was the method I was hoping to use so I could share the config with the other developers and know that CI will use the same config.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
...
+
+- Public branch 11.0.1xx:
+  - Preview branding
+  - Insert into VS canary
+  - Global channel default: preview
+
+- Internal branch build:
+  - Stable branding
+  - Includes MSRC fixes and latest runtimes
+  - Insert into VS stable and oobstable
```
Frulfump:
```suggestion
  - Insert into VS stable and unstable
```
Maybe a typo or an actual internal name :P

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
...
+### Enforcement approach
+
+Open questions:
+
+- Do we need an enfrocement bot enabled in every participating repo?
+- Do we need the ability to dump all new features into a machine and human readable format for review and comparison against intent/docs/PRs?
+
+Unknowns to validate during review:
+
+- Whether `baseline` is intuitive enough for external customers and partners.
```
Frulfump, line 256: Baseline is the most stable so I guess there could be some confusion, but I worry that some will misunderstand the guarantees given by baseline and stable and unnecessarily choose(mandate) baseline (irrespective of the name but naming helps alot with conveying intent). So by itself baseline is understandable but also having a channel named stable makes it maybe less obvious.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,237 @@
...
+- Whether existing per-tool feature flags can be harmonized without migration churn.
+- Which minimal cross-repo policy checks are required for v1 to avoid silent non-compliance.
+- Whether current public-to-internal mirror infrastructure can support a preview-branch intermediary with controlled internal stable promotion.
+- Whether stable/canary insertion branches should be structurally separated in all participating repos.
+
+### Does this replace existing feature flags?
+
+No. Component-level flags can remain. This proposal adds a single, global compatibility channel that defines default behavior posture.
+
+### Open questions to resolve
```
Frulfump, line 268: Should a change to N.0.1xx be up for consideration with this change, maybe it should now align with the runtime version? Support for the SDK and the different feature bands have been misunderstood sometimes and especially related to the runtime as that ships with the SDK but ia installed seperately where you deploy.

Frulfump:
@zivkan as we've touched on feature roll out/behavior changes for some of our discussions.

proposed/single-sdk-feature-band.md
tmat: I don't quite follow how introducing a new concept (channel setting) makes it less complex.
I'd be good to explain why staying on older SDK version that only receives security fixes isn't simpler and sufficient for customers who need stability. Is it because VS is tied to the latest SDK?

proposed/single-sdk-feature-band.md
marcpopMSFT: I defaulted to two months as that's roughly the timeline VS has for a change going from canary to stable. I think it's up to teams to decide how long they need depending on the feature as some features they may want longer and some things could be shorter. <1 month is probably not possible unless we change our release schedule.

proposed/single-sdk-feature-band.md
marcpopMSFT: We'll discuss with our team on what the design looks like for how customers enable this. I'm most concerned about ensuring tools don't accidentally make breaks in stable or baseline with how to enable this setting being second to that. For most customers though, I would hope this would be a noop where they wouldn't have to think about it and wouldn't want to change.

proposed/single-sdk-feature-band.md
marcpopMSFT: VS does a final release of the latest minor version on the same day as the .0 release of the new version. This is how they distinguish the two in internal discussions. I couldn't find any public labeling so clarified in the doc.

proposed/single-sdk-feature-band.md
marcpopMSFT: You're right that there is risk of us making this more complicated where there's a 1xx version that could have 3 different behaviors though we already have plenty of features you can opt into and out of. I think the confusion is customers not knowing which band they have or what the bands mean. It also simplifies some of our internal engineering though does add some separate branch complexity.

The main simplification is there's just be a 1xx build and so customers could just install that and not actually have to think about it. Most customers would never change the actual setting for this as source build partners would default to baseline, stable VS customers would default to stable, VSCode customers would default to stable, and insiders/canary customers would be preview. Most customers would probably just stick with the default.

jkotas:
cc @dotnet/distro-maintainers

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,251 @@
...
+- Cherry-pick debt risk: if internal stable and preview diverge, security and reliability fixes may require manual multi-branch propagation.
+- Validation matrix risk: every participating repo may need baseline/stable/preview validation, increasing CI cost and latency.
+- Diagnostics burden: support teams may need additional tooling to quickly determine the effective channel in customer environments.
+
+### Enforcement approach
+
+Open questions:
+
+- Do we need an enfrocement bot enabled in every participating repo?
+- Do we need the ability to dump all new features into a machine and human readable format for review and comparison against intent/docs/PRs?
```
MichaelSimons, line 251: Related, we should do a good job at documenting in release notes what channel new features are included in.  That includes when a release includes promotion of an existing feature to a different channel.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Does this replace existing feature flags?
+
+No. Component-level flags can remain. This proposal adds a single, global compatibility channel that defines default behavior posture.
+
+### Open questions to resolve
+
+1. Whether `DOTNET_SDK_FEATURE_CHANNEL` should be the only v1 control surface or whether a CLI switch should also ship in v1.
+2. Minimum schema for the machine-readable behavior manifest.
+3. Is there a change to the versioning schema we could do for the SDK? We could change to match runtime versioning but we sometimes have SDK hotfixes so it might be confusing to release an 11.0.108 that has 11.0.7 runtime in it.
```
MichaelSimons, line 272: In my experience the SDK versioning has been a source of confusion.  Eliminating the multiple feature bands will help significantly.  What is the need to retain 1xx vs xx?  Is the thought that if SDK needs a hotfix having misalignment between the SDK and runtime will cause confusion?  e.g. runtime 11.0.7 vs sdk 11.0.8 will be confusing but runtime 11.0.7 vs sdk 11.0.108 avoids this?

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,251 @@
...
+- Mirror topology risk for split public/internal versions: current one-way public to internal mirroring can accidentally blur stable versus preview intent.
+- Cherry-pick debt risk: if internal stable and preview diverge, security and reliability fixes may require manual multi-branch propagation.
+- Validation matrix risk: every participating repo may need baseline/stable/preview validation, increasing CI cost and latency.
+- Diagnostics burden: support teams may need additional tooling to quickly determine the effective channel in customer environments.
+
+### Enforcement approach
+
+Open questions:
+
+- Do we need an enfrocement bot enabled in every participating repo?
```
MichaelSimons: Yes, we need mechanisms in place up front to help prevent accidental/human error.  IMO it should be a goal to provide these enforcement mechanisms.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,251 @@
+# Single SDK Feature Band with Global Feature Channels
+
+**Owner** [marcpopMSFT](https://github.com/marcpopMSFT)
+
+Today, SDK feature bands, SDK tools, and Visual Studio are coupled in ways that is complex to manage and confusing for users.  The challenge with changing this is users can receive new features or behavior changes without intentionally opting in. This is especially problematic for managed environments and source build partners that prioritize stability. We propose a single SDK feature band strategy, combined with a global feature channel setting that all SDK-loaded components honor. This setting has three states: baseline, stable, and preview. New behavior or new features should never show up in baseline. New features should initially be enabled only in preview. Stable is for features that have gone through some preview validation and make include some behavior changes or new warnings.
+
+## Scenarios and User Experience
+
+### Scenario 1: Source build partner requires maximum stability
```
MichaelSimons:
```suggestion
### Scenario 1: Enterprise source build partner requires maximum stability
```

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,251 @@
...
+
+A source build partner consumes servicing builds and needs strict behavioral predictability. They configure the global channel to baseline. New features and behavior changes remain off by default across SDK, MSBuild, Roslyn, NuGet, MSTest, and other participating components. Today, these partners stay on the N.0.1xx feature band in order to prioritize stability.
+
+Outcome:
+
+- Servicing updates can include security and reliability fixes.
+- Existing behavior remains stable unless explicitly opted in.
+- Partner validation costs and risk are reduced.
+- Customers using source build SDKs can use the same SDK on multiple platforms with the same behavior with a simple environment variable
```
MichaelSimons, line 18: This is a big win.  Consider emphasizing this more.  There are a class of customers who are required to get their SDK through distro feeds but want to make use of newer features - this enables them.

joeloff:
@marcpopMSFT Can you add a scenario for Microsoft Update? Currently feature bands are serviced separately. Someone on 1xx will only be offer 1xx updates. But if they are on 2xx or 4xx, they'll get those too. SDKs from VS are serviced as from VS. The standalone SDK includes all the runtimes so those fixes are rolled up and a single offering can target multiple products.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,251 @@
+# Single SDK Feature Band with Global Feature Channels
```
rolfbjarne, line 1: If there's only ever a single feature band, can't just the entire "feature band" concept go away?

proposed/single-sdk-feature-band.md
Frulfump: Ah ok I see, maybe 1 month minimum is the sweet spot

proposed/single-sdk-feature-band.md
Frulfump: Slightly related if this would be available for Issues and PRs on GitHub as well that would help with planning for when we are dependant on a fix or feature being made available so maybe some new tags as I guess the branch would only be N.0.1xx or similar

proposed/single-sdk-feature-band.md
Frulfump: I raised something similar here https://github.com/dotnet/designs/pull/373/changes#r3688674762 but yes I guess you brought up the main concern with the hotfixing

proposed/single-sdk-feature-band.md
Frulfump: I look forward to hear what you come up with. I don't understand how it would affect accidentally making breaks could you elaborate? So you are hoping the default for most teams would be baseline right? Will baseline be the default?

proposed/single-sdk-feature-band.md
Frulfump: I think I don't understand the distinction

proposed/single-sdk-feature-band.md
Frulfump: I had to think again when responding to some comment about what baseline and stable represented so maybe an indication that it will take some time to get used too or that other words would be better but I haven't come up with any yet.

proposed/single-sdk-feature-band.md
MichaelSimons: I should have expanded on what I meant here. Not every Source Build partner requires the maximum level of stability. The requests we've received have specifically come from our enterprise Source Build partners, which have agreements around non-breaking changes. Those are the partners that require this level of stability.

I don't think it's accurate to say that all Source Build partners require maximum stability. However you'd like to word it is fine, but I think it's important to make that distinction.

marcpopMSFT:
> @marcpopMSFT Can you add a scenario for Microsoft Update? Currently feature bands are serviced separately. Someone on 1xx will only be offer 1xx updates. But if they are on 2xx or 4xx, they'll get those too. SDKs from VS are serviced as from VS. The standalone SDK includes all the runtimes so those fixes are rolled up and a single offering can target multiple products.

@joeloff can you provide more info on what you're looking for? Us collapsing into a single band means we'd only offer MU for that one band.

proposed/single-sdk-feature-band.md
marcpopMSFT: @MichaelSimons have any source build partners taken advantage of the fact that 10+ source build now supports feature bands beyond 1xx? My understanding is no so even if they don't require the least risk, they haven't changed their process to adopt the newer bands. Maybe with this making it much easier, they'd be more willing to though so it could just be a friction issue today.

proposed/single-sdk-feature-band.md
marcpopMSFT: Correct, hotfixing is the concern and them getting out of sync. I don't know if that would cause more confusion than having the 1 on front and then having them be out of sync. (ie is .108 having the .7 runtime less confusing than .8 with the .7 runtime)?

proposed/single-sdk-feature-band.md
marcpopMSFT: We'll have to brainstorm on how to track this and doc this. it's easy for an ai to look at a build release and see which features are in baseline, stable, and preview. Tying that back to specific PRs/issues is probably still pretty feasible for a computer to do. We'll have to explore tooling as we don't want the management of this to become a toil.

proposed/single-sdk-feature-band.md
marcpopMSFT: As covered above, there are SDK hotfixes that have no runtime release. Is a .108 SDK containing a .7 runtime more or less confusing than a .8 SDK with a .7 runtime? We typically do 1-3 SDK only hotfixes in a year and probably only the first year as our focused shifted to VNext.

proposed/single-sdk-feature-band.md
marcpopMSFT: To clarify, if we switch to a model with a single band, I have to ensure that all partner teams that contribute to the SDK follow this model. We've had a hard time in the past ensuring everyone followed the breaking change model for feature bands and this makes the consequences worse of having a breaking change if it slips into baseline.

How customers configure this is the second item on my list to figure out. This is a very early proposal but I wanted to get it out to get feedback before we got too far along.

proposed/single-sdk-feature-band.md
marcpopMSFT: Thanks, we'll have to keep brainstorming here to see if we can come up with something that doesn't require a lot of thinking. I picked what I picked initially as we already have previews and VS monthly releases are called the stable channel already. baseline was a best effort.

proposed/single-sdk-feature-band.md
marcpopMSFT: Thinking more on this, there's a lot of existing code that assumes feature bands like workloads code and global.json parsing. I think it may be more trouble than it's worth to get rid of the band versioning.

proposed/single-sdk-feature-band.md
jjonescz: Can we keep the 1xx versioning but stop talking about it as "feature band" concept? (Since we have only one "feature band", we don't need it as a concept, it seems to be making things unnecessarily complicated.)

rolfbjarne: Yes, that was exactly my point, we can keep versioning the same way, just say it's "how we version" and not use a "feature band" concept.

proposed/single-sdk-feature-band.md
Frulfump: I think for me when I help coworkers it will be easier for me to detect when they give me the runtime version instead of SDK version that I asked for if this follows the current versioning.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,261 @@
...
+
+### Enforcement approach
+
+Open questions:
+
+- How do we enforce this in every contributing repo?
+  - PR review guidance as a minimum
+  - What other enforcement can we create?
+- Do we need the ability to dump all new features into a machine and human readable format for review and comparison against intent/docs/PRs?
+  - It'd be great to be able to tag issues/PRs on which release they are in as well.
```
Frulfump, line 252: release here meaning baseline, stable or preview? When I hear release I think about as an example 10.0.302 or 10.0.10. GitHub natively currently show which GitHub release an issue is fixed in but this isn't extended to PRs yet but hoping that will come sooner rather than later so the largest benefit here would be to know ahead of time which release something will land in and what configuration baseline, stable or preview.

proposed/single-sdk-feature-band.md
marcpopMSFT: I think this is a nice middle ground that could work. We wouldn't have to figure out how to change our whole infrastructure which is what we'd need if we did a 11.0.0 or a 11.1.0 versioning of the SDK and we'd avoid the mismatched issue with runtime. We just call it the SDK version. global.json would still have rollforward feature but that wouldn't really apply going forward anymore.

proposed/single-sdk-feature-band.md
marcpopMSFT: features would never land in baseline presumably. But the idea would be to tie a new feature/setting to the patch version that it showed up in as stable. I think that's the most important piece to track as that's when most customers would end up getting the feature. Tracking when it released in preview is probably next most interesting. .NET has milestones we use for servicing and that's what I meant when extending back to PRs and issues.

proposed/single-sdk-feature-band.md
marcpopMSFT: @MichaelSimons an alternative proposal added to the doc would be to switch to minor releases. The drawback is that we'd lose the benefit this proposal has of enabling customers who use the source build packages to opt into newer features which is unfortunate. It does let us simplify the versioning somewhat but I don't know if minors are more intuitive (they'd be monthly and probably we'd stop doing minor after a year and switch to patches once there were no more feature changes coming in).

Thoughts on this tradeoff?

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,360 @@
...
+
+All participating components must query this value through a shared contract and use it to gate behavior changes.
+
+Initial policy for any new feature or behavior change:
+
+- baseline: off
+- stable: off (unless explicitly promoted)
+- preview: on (unless explicitly held back)
+
+Promotion policy:
```
mthalman: There may be features which cut across multiple components/repos. This will need to be carefully coordinated to ensure that all components are promoted together.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,360 @@
...
+
+Risk:
+
+This potentially limits the amount of feedback tooling teams would get on new features before promoting them to stable as they'd only be available in vNext or VS/msbuild.exe.
+
+### Servicing safety rules
+
+In servicing trains:
+
+- No new behavior should silently turn on for stable or baseline.
```
mthalman: It should be noted that "stable" doesn't mean you can't be broken by a servicing update. The simplest example of this is a new analyzer being added. When the user has "treat warnings as errors" turned on, they would be broken by taking a new update when they violating code. There are probably other examples of non-analyzer breaking changes as well.

One way to solve this is with a compatibility feature level setting.

```json
{
  "sdk": {
    "version": "11.0.106",
    "featureChannel": "stable",
    "featureLevel": "11.0.103"
  }
}
```

This means: I want the 11.0.106 SDK with stable behavior but only turn on features up to version 11.0.103. As long as gating is done correctly on our part, that guarantees they won't be broken by any new features that were added between 11.0.103 and 11.0.106 while also giving them the benefit of having security and bug fixes delivered.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,360 @@
...
+- preview: on (unless explicitly held back)
+
+Promotion policy:
+
+- A feature starts as preview-on, stable-off, baseline-off.
+- Promotion to stable requires explicit approval and release notes in a planned update.
+- Baseline remains off by default, with opt-in available for customers that explicitly choose it.
+- Minimum preview soak time before promotion is two months.
+
+### Configuration sources and precedence
```
mthalman: Since you can dynamically control which feature channel is used each time you execute `dotnet`, there needs to be some consideration of the impact of cached content. Some content that exists in a cache may be incompatible across feature channels. So having some kind of cache invalidation logic could be important. That would mean tying the state of the content to the feature channel to know whether it matches or doesn't.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,360 @@
...
+
+## Q and A
+
+### Why a three-state model instead of a binary preview toggle?
+
+A binary model cannot represent baseline needs independently from mainstream stable users. Three states separate strict stability from normal stable adoption and preview validation.
+
+## Risks and Unknowns
+
+- Cross-component adoption lag: not all components may adopt `SdkFeatureChannel` and `DOTNET_SDK_FEATURE_CHANNEL` at the same pace, leading to inconsistent behavior.
```
mmitche: Also worth calling out: Increased validation matrices. Likely SDK and tooling repo would need to increase validation in various ways as new features shift from branch-specific to feature flags.

mmitche (commented):
Worth calling out that source build validation will probably stay about the same, but to be more explicit, we're saying that N(flag=baseline) must be able to build N+1. We should make it an explicit goal to avoid taking dependencies on preview or stable features to be able to build N+1.

I don't necessarily think it's a **hard** requirement, but it's probably good guidnace.

mthalman:
> Worth calling out that source build validation will probably stay about the same, but to be more explicit, we're saying that N(flag=baseline) must be able to build N+1. We should make it an explicit goal to avoid taking dependencies on preview or stable features to be able to build N+1.
> 
> I don't necessarily think it's a **hard** requirement, but it's probably good guidnace.

Eesh, that's a really good point. And actually that's a regression from how it currently works in 10.0. In 10.0.4xx, you can depend on features that came from 10.0.3xx, for example. Relevant doc links:
- https://github.com/dotnet/source-build/blob/main/Documentation/feature-band-source-building.md#3xx4xx-band-initial-release-n0b00
- https://github.com/dotnet/source-build/blob/main/Documentation/feature-band-source-building.md#3xx4xx-band-servicing-n0b01

So saying you can only depend on baseline channel definitely limits things.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,360 @@
...
+
+Outcome:
+
+- Stable channel is a sensible default for most developers.
+- Power users have direct control via environment variable configuration.
+- Independent of IDE choice; behavior is consistent across any editor using the SDK.
+- Easy to experiment with new features without waiting for broader SDK releases.
+- With each major release, VSCode will update the SDK installed to a new band resulting in a new baseline.
+
+## Alternatives
```
jkotas: I think both the current system, this proposal and this alternative are complicated in different ways.

What would it take to switch to simpler plan?

- New .NET SDK features ship together with new major .NET version each November
- In rare cases a feature absolutely needs to be backported to servicing, it is handled as a one-off, with feature-specific opt-in to avoid breakages
- Users who cannot wait to use the new .NET SDK features can use next version previews at their own risk 
- This is the plan that the runtime itself is on. We do not have to explain the difference between the runtime and SDK versioning.

marcpopMSFT: There are a lot of pieces of the SDK that want to update more frequently than yearly. In particular, I know there are changes going in that will improve the experience in VSCode and CLI against particularly in the build and language experience. I don't think we can require agents be using vNext previews as we want these sort of experiences to light up as they are ready.

The feature flag proposal is simplest from a code construction and branching perspective but adds complication onto the tooling teams and testing. The minor version proposal is probably the simplest compromise here as it's similar enough to what others are doing like VS so it's less weird.

There is a further proposal to pull the SDK CLI out entirely and have it always be tip-only but that is a much larger investment with a lot of issues to sort out.

marcpopMSFT:
> > Worth calling out that source build validation will probably stay about the same, but to be more explicit, we're saying that N(flag=baseline) must be able to build N+1. We should make it an explicit goal to avoid taking dependencies on preview or stable features to be able to build N+1.
> > I don't necessarily think it's a **hard** requirement, but it's probably good guidnace.
> 
> Eesh, that's a really good point. And actually that's a regression from how it currently works in 10.0. In 10.0.4xx, you can depend on features that came from 10.0.3xx, for example. Relevant doc links:
> 
> * https://github.com/dotnet/source-build/blob/main/Documentation/feature-band-source-building.md#3xx4xx-band-initial-release-n0b00
> * https://github.com/dotnet/source-build/blob/main/Documentation/feature-band-source-building.md#3xx4xx-band-servicing-n0b01
> 
> So saying you can only depend on baseline channel definitely limits things.

With how it may be landing, we could potentially have it so that the next minor can be built with the previous minor. The minor versions are essentially feature bands but with simpler darc channel flows, dedicated branching infrastructure, and less confusing numbering (in theory)

proposed/single-sdk-feature-band.md
jkotas: > There are a lot of pieces of the SDK that want to update more frequently than yearly. 

Do we have a list? Do they *have to* update? What are we going to lose when they update yearly?

> I know there are changes going in that will improve the experience in VSCode and CLI against particularly in the build and language experience.

This can be handled as one-off that I have mentioned in my comment if it is really important to ship off cycle.

proposed/single-sdk-feature-band.md
jkotas: We should have a list of components/repos that we expect to be the source of the new potentially breaking features based on the historical data. It will help us understand the scale of the coordination problem.

(Related to my other comment https://github.com/dotnet/designs/pull/373#discussion_r3826093249 .)

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,272 @@
...
+- `12.2.0-preview.<buildnumber>`: a preview of the next monthly SDK feature release.
+- `12.2.1`: a hotfix or workload update after `12.2.0`.
+
+The SDK and runtime versions are intentionally independent. An SDK hotfix such as `12.2.1` might contain runtime `12.0.3`. A monthly SDK preview contains the most recently available runtime patch rather than requiring a runtime release on the same day. Release notes and diagnostics must make the relationship clear.
+
+Using the patch component for both SDK hotfixes and workload updates means patch numbers identify release order rather than a particular kind of change. Release tooling and documentation must not assume that every patch increment has the same cause.
+
+## Release and Branch Model
+
+Development occurs primarily in a long-lived public `release/12.x` branch. Public code flow from tooling repositories targets this branch. Preview builds of the next monthly minor are produced from it and flow to Visual Studio Canary.
```
jkotas, line 151: Is the idea that *all* SDK development occurs primarily in the `release/12.x` branch? How does it work between major releases - when is the primary development going to switch to `release/13.x` branch?

marcpopMSFT: This is pretty much the same as today. All tooling teams have been working out of their main that is flowing to both 10.0.4xx (and 10.0.3xx before that) and main (and later 11.0.1xx). Most feature work throughout the year has happened and shipped in the feature bands with a smaller set of changes that are riskier only going into 11 (like the AOT enablement).

proposed/single-sdk-feature-band.md
marcpopMSFT: All the tooling teams (and by that I mean msbuild, nuget, roslyn, and vstest as the major ones) have been shipping features monthly with the new VS model. Going back to a yearly release for them even with a path to backport would be challenging. Features from msbuild like multithreading and server that they only want on by default in 11 are still available in 10 branches as the code goes to both. Things like that would be hard to separate out and backport cleanly.

When tooling teams were given a choice to go quarterly with the SDK or monthly with VS, they all chose monthly. That's part of what's driving the minor version plans here as it aligns the SDK with everyone else.

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,284 @@
...
+
+A customer remaining on `12.0.x` cannot opt into selected features from later monthly minors without moving to a newer SDK. This tradeoff is intentional: release boundaries provide stronger isolation than relying on every component to gate behavior correctly. It is particularly important for analyzer and compiler fixes that can introduce warnings, change syntax trees or APIs, or otherwise affect downstream tooling.
+
+The next minor preview should be buildable with a prior stable release so 12.4.0 can be built with 12.3.0. The previews of the SDK will have to build with N-2 SDKs as the N-1 will still be unreleased and being worked on.
+
+## Workloads, Installers, and SDK Selection
+
+Existing workload and SDK infrastructure encodes feature-band identity in workload manifests, workload sets, package IDs, version validation, and selection logic. Monthly minors become the new compatibility boundary, and these systems must be updated accordingly.
```
jonathanpeppers: can we give an example of what a workload manifest name before & after would be?

marcpopMSFT: For workload manifests (this will match the GA SDK version):
`Microsoft.NET.Workload.Mono.ToolChain.Current.Manifest-11.0.100` -> `Microsoft.NET.Workload.Mono.ToolChain.Current.Manifest-12.0.0`

For workload sets (this will match the SDK versions):
`Microsoft.NET.Workloads.11.0.206` ->
`Microsoft.NET.Workloads.12.7.0`

proposed/single-sdk-feature-band.md
```diff
@@ -0,0 +1,284 @@
...
+- Minor: a monthly SDK feature release.
+- Patch: an SDK or tooling hotfix, or a release used to carry an updated workload set between monthly feature releases.
+- Prerelease label: builds of the next monthly minor before promotion to stable.
+
+For example:
+
+- `12.0.0-preview.1.<buildnumber>`: First preview of .NET 12 SDK
+- `12.0.0`: the annual GA and the beginning of the stability-oriented release line.
+- `12.0.1`: a security or reliability update to the original stability line.
+- `12.1.0`: the first monthly SDK feature release.
```
omajid, line 135: When a minor is bumped, the patch version is reset, right? In other words, 12.1.0 will be newer (or at least as recent) as 12.0.1 in both the SDK and Runtime contents, right?

proposed/single-sdk-feature-band.md
jonathanpeppers: I wondered if we could put in the doc, too, thanks.

We'll be able to do this, too? `Microsoft.NET.Workload.Mono.ToolChain.Current.Manifest-12.1.0`

Like today we could technically do `Microsoft.NET.Workload.Mono.ToolChain.Current.Manifest-11.0.200`, but haven't _needed_ to.

marcpopMSFT: Yes, if there were a future minor release where you wanted to make a risky change, you could re-baseline the workload set in that one. Let's say you did that in 12.7.0. That would mean all 12.0.N SDKs as well as all 12.1-12.6 SDKs would use the original 12.0.0 set of workloads. Starting in 12.7.0 and beyond, it would use that 12.7.0 band workloads. A potential point of confusion but mostly that should be an implementation detail customers don't have to think about as it's all handled for them through their workload commands (or VS installs).

proposed/single-sdk-feature-band.md
marcpopMSFT: Correct. Think of 12.0.x as it's own patch family though for source build enterprise customers and customers who want to stick with security and critical fixes only.

So 6 months in we could be shipping 12.0.6 (which is the security fix channel/family) and 12.6.0 (which is the feature channel/family). We can even do hotfixes of the feature releases from time to time if we don't want to wait the month (eg. 12.6.1). We will also likely stop doing minors at some point an only do patches from then on once we're no longer doing any feature work. So 12.11.0 is the last minor update and then patch updates from then on....

In all cases, the same runtime should ship in both releases that go out in the same month. Next month's preview will be on N-1.

PR #374, opened 2026-08-06
Update unsafe evolution opt in doc

jjonescz:


333fred (approved):
I like this with names as is. I'm fine with just `2` as the rules number.

